PR #6, opened 2023-12-12
Package and operation common interface

tadelesh:


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -0,0 +1,660 @@
...
+    const headers: SdkServiceResponseHeader[] = [];
+    let body: Type | undefined;
+    for (const data of response.responses) {
+      for (const header of Object.values(data.headers || [])) {
+        headers.push({
+          __raw: header,
+          description: getDocHelper(context, header).description,
+          details: getDocHelper(context, header).details,
+          serializedName: getHeaderFieldName(context.program, header),
+          type: getClientType(context, header.type),
```
jhendrixMSFT, line 350: Not sure if this is the correct location but all of the header responses have `encode` set to `rfc3339` which breaks some [tests](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/encode/datetime/main.tsp#L304-L342).

iscai-msft: oh interesting, that's a good point because our code is only doing `encode` for input headers

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -228,135 +272,256 @@ export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
...
+interface SdkServiceOperationBase {}
+
+export type SdkParameter = SdkEndpointParameter | SdkCredentialParameter | SdkMethodParameter;
+
+export interface SdkHttpOperation extends SdkServiceOperationBase {
+  __raw: HttpOperation;
+  kind: "http";
+  path: string;
+  verb: HttpVerb;
+  pathParams: SdkPathParameter[];
```
jhendrixMSFT, line 348: Historically, we've tried to keep the order of the generated params consistent with the order from OpenAPI. With the params split like this, we can no longer achieve this. Is this intentional and if so why is that?

iscai-msft: that's a good point, we wanted to separate them out specifically for http calls since some languages do different things for each parameter. Originally we had also a `parameters` property that had the same ordering as the spec, I removed in thinking it was a duplication, but you're right this is somethign we should have. Let me add it back

jhendrixMSFT: Adding `parameters` back sounds great.

Is there value in keeping the others? They duplicate the source. If other languages have a need to split them up then IMO they should do it.

jhendrixMSFT: It also complicates the API and might not be clear when to use `parameters` or the (duplicated) variants.

The API should try to push consumers into the "pit of success" as much as possible.

iscai-msft: that is fair, let me remove and then add back if needed later. thanks!

jhendrixMSFT: I see that header/path/query params have been consolidated into `parameters` which is great. Should `bodyParams` also be consolidated? Is it possible in tsp to model an operation with an ordered set of params that include body/header/path/query?
```
fooOp(id: string, w: Widget, count: int32)
```
Where `id` is a header param, `w` is a body param, and `count` is a query param. I presume we'd want the emitted param order to match the tsp defined order.

iscai-msft:
fyi: to use the latest artifact build, you need to have `@typespec/compiler: "0.52.0-dev.14"`

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -0,0 +1,734 @@
...
+        nameInClient: "body",
+        description: getDocHelper(context, tspBody.type).description,
+        details: getDocHelper(context, tspBody.type).details,
+        onClient: false,
+        contentTypes: contentTypes,
+        defaultContentType: defaultContentType,
+        isApiVersionParam: false,
+        apiVersions: getAvailableApiVersions<SdkHttpOperation>(context, tspBody.type),
+        type: bodyType,
+        optional: false,
```
jhendrixMSFT, line 107: I suspect this is the culprit that's making body params always required?

iscai-msft: yeah, i'm trying to figure out how to find out if a body param is optional though, bc there's not that property on the tsp body param

jhendrixMSFT: I added a test for this and I see that the method param is correctly marked as optional, it's just the operation param has it as required.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -0,0 +1,734 @@
...
+  throw new Error("blah");
+}
+
+function createContentTypeOrAcceptHeader(
+  context: SdkContext<SdkHttpOperation>,
+  bodyObject: SdkBodyParameter | SdkHttpResponse
+): Omit<SdkMethodParameter, "kind"> {
+  const nameInClient = bodyObject.kind === "body" ? "contentType" : "accept";
+  let type: SdkType = {
+    kind: "string",
```
jhendrixMSFT, line 153: Shouldn't this be `kind: "constant"`? I'm seeing the `accept` header show up as parameter due to its non-constant nature.

jhendrixMSFT: Ok maybe not here as we need to handle multiple content types. But it should be `constant` for the `accept` header.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -0,0 +1,734 @@
...
+    encode: "string",
+    nullable: false,
+  };
+  if (
+    bodyObject.contentTypes &&
+    bodyObject.contentTypes.length === 1 &&
+    /json/.test(bodyObject.contentTypes[0])
```
jhendrixMSFT, line 160: If `length === 1` shouldn't this always be constant? Unclear why it's limited to `json`.

packages/typespec-client-generator-core/src/package.ts
jhendrixMSFT: To clarify my first point, I see `accept` show up as non-constant for things like `application/octet-stream` and `image/png`. Removing the `json` constraint below resolves the issue. But is that correct?

iscai-msft: That's a good point, just because we don't default `contentType` bc of not having a clear default, doesn't mean we shouldn't default `accept`, bc adding an `accept` new mimetype would be breaking

packages/typespec-client-generator-core/src/package.ts
iscai-msft: For content types, we currently only want to have a constant when it's `json`, i believe this was the behavior as well with `m4`. The reason is bc we don't want to break users. Let's say a service starts off just accepting an `image/png` body, and we generate the `contentType` with constant value `image/png`. If they later add `image/jpeg`, we are now going to require users to input `contentType` automatically during generation, because it's no longer a constant, and between these two, the only clear indicator of what should be defaulted is what was added first, but I feel that's a brittle heuristic since we don't want to get into client default values

jhendrixMSFT: That makes sense, thanks for the explanation.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -224,137 +270,177 @@ export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
   visibility?: Visibility[];
```
jhendrixMSFT, line 270: Visibility is intended to be a bit-field so it doesn't need to be an array.

packages/typespec-client-generator-core/src/package.ts
jhendrixMSFT: Fixed in b8756c8.

jhendrixMSFT:
Fixes https://github.com/Azure/typespec-azure/pull/6

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -58,6 +59,24 @@ export interface SdkClient {
...
+export interface SdkClientType<TServiceOperation extends SdkServiceOperation> {
+  kind: "client";
+  name: string;
+  description?: string;
+  details?: string;
+  initialization?: SdkInitializationType;
+  methods: SdkMethod<TServiceOperation>[];
+  apiVersions: string[];
+  nameSpace: string; // fully qualified
+  endpoint: string;
```
jhendrixMSFT, line 75: Should this be `endpoint?: string`? I see cases where it's an empty string.

jhendrixMSFT: This was observed for non-instantiable clients which is expected at present.

jhendrixMSFT:
LGTM!

iscai-msft (approved):

PR #7, opened 2023-12-12
Tryit comment use bot token instead

timotheeguerin:
Posting a comment on a PR is a permission the azure pipeline doesn't get when building a fork regardless of who's fork it is unless the main repo is private. As we made this repo public, the tryit comment cannot be created by the ado pipeline anymore and we have to do the same thing as in `mcrosoft/typespec` where a github action will create the comment for ahead of time.
Core PR https://github.com/microsoft/typespec/pull/2749


.github/workflows/tryit-comment.yml
```diff
@@ -0,0 +1,36 @@
...
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const prNumber = ${{ github.event.pull_request.number }};
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: [
+                `Changes in this PR will be published to the following url to try(check status of TypeSpec Pull Request Try It pipeline for publish status):`,
+                `Playground: https://cadlplayground.z22.web.core.windows.net/cadl-azure/prs/${prNumber}/`,
```
weshaggard, line 31: Isn't this dependent on this getting the artifacts published by the devops pipeline? Does that mean the link doesn't work until after that?

timotheeguerin: indeed, this is a bit unfortunate but couldn't find a better way when we did that for `microsoft/typespec`. Don't think it really has really been causing problems since though. But open to suggestion if you have a better way.

weshaggard (approved):

PR #16, opened 2023-12-13
Add LRO Support and Location support to ARM

markcowl:


packages/typespec-azure-core/src/decorators.ts
```diff
@@ -603,9 +614,142 @@ export function isLroFailedState(program: Program, entity: EnumMember): boolean
...
+    }
+  }
 
-export function $pollingLocation(context: DecoratorContext, entity: ModelProperty) {
-  context.program.stateSet(pollingLocationsKey).add(entity);
+  program.stateSet(pollingLocationsKey).add(entity);
+}
+
+/**
+ * Gets pollign information stored with a field that contains a link to an Lro polling endpoint
```
tjprescott, line 669: typo: polling

docs/libraries/azure-core/reference/data-types.md
```diff
@@ -141,6 +141,14 @@ model Azure.Core.Page<TResource>
 | --------- | ------------------ |
 | TResource | The resource type. |
 
+### `PollingOptions` {#Azure.Core.PollingOptions}
+
+Generic polling options for LRO operations.
+
+```typespec
+model Azure.Core.PollingOptions
```
tjprescott, line 149: Does this not have any properties?

docs/libraries/azure-core/reference/data-types.md
```diff
@@ -231,6 +239,30 @@ Provides the most common query parameters for list operations.
...
+```typespec
+model Azure.Core.StatusMonitorOptions
+```
+
+### `StatusMonitorPollingOptions` {#Azure.Core.StatusMonitorPollingOptions}
+
+Options for overriding a polling endpoint that uses a StatusMonitor
+
+```typespec
+model Azure.Core.StatusMonitorPollingOptions<TPollingModel, TFinalResult, TFinalProperty>
```
tjprescott, line 255: Didn't Will just merge his PR that changes the settled on not using the T prefix for template args?

docs/libraries/azure-core/reference/data-types.md
```diff
@@ -231,6 +239,30 @@ Provides the most common query parameters for list operations.
...
+```
+
+#### Template Parameters
+
+| Name           | Description                                              |
+| -------------- | -------------------------------------------------------- |
+| TPollingModel  | The model that is returned when polling should continue. |
+| TFinalResult   |                                                          |
+| TFinalProperty |                                                          |
```
tjprescott, line 264: These need documentation

docs/libraries/azure-core/reference/data-types.md
```diff
@@ -247,6 +279,12 @@ Supported versions of Azure.Core TypeSpec building blocks.
 enum Azure.Core.Versions
 ```
 
+### `PollingOptionKind` {#Azure.Core.PollingOptionKind}
+
+```typespec
+union Azure.Core.PollingOptionKind
```
tjprescott, line 285: Could be a docs bug, but why doesn't it show the values?

docs/libraries/azure-core/reference/decorators.md
```diff
@@ -22,7 +22,9 @@ Identifies a ModelProperty as containing the final location for the operation re
 
 #### Parameters
 
-None
+| Name        | Type                  | Description                                                                                                                                                 |
+| ----------- | --------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| finalResult | `union Model \| void` | Sets the expected return value for the final result. Overrides<br />any value provided in the decorated property, if the property uses ResourceLocation<T>. |
```
tjprescott, line 27: Do we warn if there's a conflict?  Like, not a warning if there's an inferred value and they specify something that overrides it, but that they seemingly specific conflicting things (not sure that's possible)?

docs/libraries/azure-resource-manager/reference/data-types.md
```diff
@@ -61,9 +90,15 @@ model Azure.ResourceManager.ArmCreatedResponse<T>
...
 ```typespec
-model Azure.ResourceManager.ArmDeleteAcceptedLroResponse
+model Azure.ResourceManager.ArmDeleteAcceptedLroResponse<LroHeaders>
 ```
 
+#### Template Parameters
+
+| Name       | Description |
+| ---------- | ----------- |
+| LroHeaders |             |
```
tjprescott, line 100: Needs documentation

docs/libraries/azure-resource-manager/reference/data-types.md
```diff
@@ -98,19 +149,35 @@ model Azure.ResourceManager.ArmNoContentResponse<TMessage>
...
 ```
 
 #### Template Parameters
 
-| Name      | Description                |
-| --------- | -------------------------- |
-| TResource | The resource being updated |
+| Name       | Description                                                |
+| ---------- | ---------------------------------------------------------- |
+| TResource  | The resource being updated                                 |
```
tjprescott, line 179: Should be `Resource` I think

docs/libraries/azure-core/reference/decorators.md
markcowl: No, this works the same way as the override for pollingOperation (which also overrides the link values), since it is an intentional override, and is largely used with templates (in the case of pollingLink, as an augment decorator)

packages/samples/test/output/azure/resource-manager/arm-scenarios/tracked/@azure-tools/typespec-autorest/2021-10-01-preview/openapi.json
```diff
@@ -241,6 +241,14 @@
                 "type": "integer",
                 "format": "int32",
                 "description": "The Retry-After header can indicate how long the client should wait before polling the operation status."
+              },
+              "Azure-AsyncOperation": {
```
allenjzhang, line 245: Looks like by default we output both header. Should the samples be updated to use a single recommended pattern one or the other?

markcowl: We have options for all 3 patterns (Azure-AsyncOperations only,  Location only, and combined).  The template default uses the recommended async implementation for RPaaS, and matches the existing `final-state-via`  value.

docs/libraries/azure-core/reference/data-types.md
markcowl: This is similar to other models in this md file and elsewhere - template parameters are described, but not model properties.

docs/libraries/azure-core/reference/data-types.md
markcowl: Yeah, and since this is a new template, it should use the new paradigm, good catch

docs/libraries/azure-core/reference/data-types.md
markcowl: It doesn't go intodetail on model, enum, union types.

allenjzhang:
Could you add ARM sample for location based Resource?

allenjzhang (approved):

markcowl:
Validated against azure-rest-api-specs here: https://github.com/Azure/azure-rest-api-specs/pull/27140

Note that npm ls -a will fail, but spec validation was successful in all runs

Issue #17, opened 2023-12-13
TCGC types should support types for additional properties


tadelesh:
`additionalProperties` in `SdkModelType` is boolean, which lost the info of additional property type.
Will change the type to `SdkType`.

Original from: https://github.com/Azure/typespec-azure-pr/pull/3343#discussion_r1414485827

cc: @iscai-msft

PR #18, opened 2023-12-13
Add type info for additional properties in model type

tadelesh:
Fix https://github.com/Azure/typespec-azure/issues/17

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -197,7 +197,7 @@ export interface SdkModelType extends SdkTypeBase {
   details?: string;
   access: AccessFlags | undefined;
   usage: UsageFlags;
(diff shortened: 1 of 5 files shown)


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -197,7 +197,7 @@ export interface SdkModelType extends SdkTypeBase {
   details?: string;
   access: AccessFlags | undefined;
   usage: UsageFlags;
-  additionalProperties: boolean;
+  additionalProperties: SdkType | undefined;
```
jhendrixMSFT, line 200: The docs define this as `additionalProperties?: SdkType;` which is slightly different, but I assume is identical?

tadelesh:
@iscai-msft could you help to review?

iscai-msft (approved):

PR #34, opened 2023-12-18
Adding typespec-azure-portal-core package

yejee94:


packages/typespec-azure-portal-core/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "../../core/packages/tsconfig.json",
+  "composite": true,
+  "references": [
+    { "path": "../../core/packages/compiler/tsconfig.json" },
+    { "path": "../../core/packages/rest/tsconfig.json" },
+    { "path": "../../core/packages/openapi/tsconfig.json" },
+    { "path": "../../core/packages/lint/tsconfig.json" }
```
timotheeguerin, line 8: Shouldn't need that one

packages/typespec-azure-portal-core/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@azure-tools/typespec-azure-resource-manager": "workspace:~0.37.0",
+    "@azure-tools/typespec-client-generator-core": "workspace:~0.37.0",
+    "@azure-tools/typespec-autorest": "workspace:~0.37.1",
+    "@types/mocha": "~10.0.6",
+    "@types/node": "~18.11.9",
+    "@typespec/compiler": "workspace:~0.51.0",
+    "@typespec/openapi": "workspace:~0.51.0",
+    "@typespec/http": "workspace:~0.51.0",
+    "@typespec/rest": "workspace:~0.51.0",
+    "@typespec/versioning": "workspace:~0.51.0",
```
timotheeguerin, line 35: Do you need all those typespec dependencies?

common/changes/@azure-tools/typespec-azure-portal-core/yejeelee-typespecAzureCore_2023-12-18-15-44.json
timotheeguerin, line 1: Have we decided that we want this in our repo? Just wondering is there going to be any other use than the portal team? If no it feels maybe a bit unnecessary to have this package here where we have to maintain it and portal team is limited by our repo structure.

packages/typespec-azure-portal-core/package.json
yejee94: Yes I need all of them for my testing

packages/typespec-azure-portal-core/tsconfig.json
yejee94: deleted it :) thank you!

common/changes/@azure-tools/typespec-azure-portal-core/yejeelee-typespecAzureCore_2023-12-18-15-44.json
yejee94: Yes, talked to @witemple-msft and @allenjzhang, we want to place our library in typespec-azure because any ARM users can use our library decorators for Azure Portal UI customization :)

packages/typespec-azure-portal-core/tsconfig.json
timotheeguerin: sorry I just meant the last one(lint), you technically want this to be a list of all the local dependencies you have if you want to be able to do `rush watch` and work on thoses pacakges so it knows in which order to build. It probably doesn't matter for this one as its unlikely you'd be working on those core packages too but this is good to keep in sync with dependencies.

yejee94: got it! yes, reverted back and deleted only last one :) thank you!

packages/typespec-azure-portal-core/.mocharc.yaml
```diff
@@ -0,0 +1,4 @@
+timeout: 5000
```
timotheeguerin, line 1: Over the holidays we migrated from mocha to vitest, you'll have to do the same here. 

here is the PR for reference of what to do https://github.com/Azure/typespec-azure/pull/54

It should be:
- remove all the mocha files and reference to mocha
- add the vitest.config.ts file like the other packages 
- update the tsconfig.json to include `vitest.config.ts`
- update thje package.json to use vitest for test instead of mocha

packages/typespec-azure-portal-core/package.json
```diff
@@ -0,0 +1,51 @@
...
+  ],
+  "author": "",
+  "license": "MIT",
+  "type": "module",
+  "main": "dist/src/index.js",
+  "tspMain": "lib/main.tsp",
+  "peerDependencies": {
+    "@typespec/compiler": "~0.51.0"
+  },
+  "devDependencies": {
```
witemple-msft, line 23: We'll need to sort out which of these are bona fide dependencies and which are merely for testing. It seems like at a minimum typespec/compiler and azure-tools/typespec-azure-resource-manager need to be moved to the dependencies list.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+      const sourceLocation = getSourceLocation(target);
+      const dirPath =
+        sourceLocation && sourceLocation.file.path && getDirectoryPath(sourceLocation.file.path);
+      let argQueryPath = query.type && (query.type as Model).properties.get("filePath");
+      const argQueryPathValue =
+        argQueryPath && argQueryPath.type && (argQueryPath.type as StringLiteral).value;
+      let filePath = resolvePath(dirPath, argQueryPathValue); //if given path is fullpath, it will return the fullPath
+      if (filePath && argQueryPath && argQueryPathValue) {
+        filePath = normalizePath(filePath);
+        if (!fs.existsSync(filePath)) {
```
witemple-msft, line 41: Need to use `ctx.program.host.stat` instead of `fs.existsSync` here for correctness.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+) {
+  if (
+    getArmResourceKind(target) !== ("Tracked" as ArmResourceKind) &&
+    getArmResourceKind(target) !== ("Proxy" as ArmResourceKind)
+  ) {
+    reportDiagnostic(program, {
+      code: "invalidUsageDecorator",
+      messageId: decoratorName,
+      target,
+    });
```
witemple-msft, line 77: We probably want to target the decorator application itself here and not the target type. Unfortunately it's complicated to do so. Long-standing issue for some compiler assistance doing it here: https://github.com/microsoft/typespec/issues/1650

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+  }
+}
+/**
```
witemple-msft, line 80: Nitpick: Declarations want one line of space between them for readability.
```suggestion
}

/**
```

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+ * @param options AboutOptions of the property.
+ */
+export function $about(context: DecoratorContext, target: Model, options: AboutOptions) {
+  const { program } = context;
+  validateDecoratorUniqueOnNode(context, target, $about);
+  isARMResource(program, target, "about");
+  if (options && (options as Model).properties) {
+    const icon = (options as Model).properties.get("icon");
+    if (icon) {
+      if (icon.type.kind === "Model") {
```
witemple-msft, line 92: @timotheeguerin is there a helper for converting models in decorator args into object literals? This is something I run into frequently as well.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+        const sourceLocation = getSourceLocation(target);
+        const dirPath =
+          sourceLocation && sourceLocation.file.path && getDirectoryPath(sourceLocation.file.path);
+        let iconPath = icon.type && (icon.type as Model).properties.get("filePath");
+        const iconPathValue = iconPath && iconPath.type && (iconPath.type as StringLiteral).value;
+        let filePath = resolvePath(dirPath, iconPathValue); //if given path is fullpath, it will return the fullPath
```
witemple-msft, line 98: I would suggest creating a helper function to handle getting a path from a FilePath relative to a target.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+        if (filePath && iconPath && iconPathValue) {
+          filePath = normalizePath(filePath);
+          if (!fs.existsSync(filePath)) {
+            reportDiagnostic(program, {
+              code: "fileNotFound",
+              messageId: "aboutIcon",
+              target,
+            });
+          }
+          (iconPath.type as StringLiteral).value = filePath;
```
witemple-msft, line 108: I don't really understand what setting the value of the StringLiteral is doing.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+        filePath = normalizePath(filePath);
+        if (!fs.existsSync(filePath)) {
+          reportDiagnostic(program, {
+            code: "fileNotFound",
+            messageId: "browseargQuery",
+            target,
+          });
+        }
+        (argQueryPath.type as StringLiteral).value = filePath;
+      }
```
witemple-msft, line 49: Are you missing an else case for when argQuery is a string? I don't think I follow how that case is handled.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+  program.stateMap(PortalCoreKeys.about).set(target, options);
+}
+
+export function getAbout(program: Program, target: Type) {
+  return program.stateMap(PortalCoreKeys.about).get(target);
+}
+
+export function getAboutDisplayName(program: Program, target: Type) {
+  const about = getAbout(program, target);
+  return about.properties.get("displayName");
```
witemple-msft, line 122: What I would suggest is converting the AboutOptions into an object first before storing it in the stateMap so that you don't have to use the Model interface everywhere downstream of it. You can convert once when the decorator is applied.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+          }
+          (iconPath.type as StringLiteral).value = filePath;
+        }
+      }
+    }
+  }
+  program.stateMap(PortalCoreKeys.about).set(target, options);
+}
+
+export function getAbout(program: Program, target: Type) {
```
witemple-msft, line 116: Most of these functions that return things from the stateMap could use type annotations since our typing around stateMap is very weak in the compiler.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+export function getAboutKeywords(program: Program, target: Type) {
+  const about = getAbout(program, target);
+  return about.properties.get("keywords");
+}
+
+export function getAboutKeywordsItemsArray(program: Program, target: Type) {
+  const keywords = getAboutKeywords(program, target);
+  return keywords.type.values
+    .filter((value: Type) => value.kind === "String")
+    .map((value: Type) => (<StringLiteral>value).value);
```
witemple-msft, line 134: tip: I recommend using new-style `value as StringLiteral` rather than old-style `<StringLiteral>value`.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+}
+
+export function getDisplayName(program: Program, target: Type): string | undefined {
+  return program.stateMap(PortalCoreKeys.displayName).get(target);
+}
+
+export function getDisplayNameValue(program: Program, target: Type): string | undefined {
+  const result = program.stateMap(PortalCoreKeys.displayName).get(target);
+  return result && result.value;
+}
```
witemple-msft, line 205: Something a little weird is going on here. The first function doesn't actually return `string`, but a StringLiteral object from the compiler. I would suggest just making `getDisplayName` do what `getDisplayNameValue` does and only having one of them.

packages/typespec-azure-portal-core/src/lib.ts
```diff
@@ -0,0 +1,37 @@
+import { createTypeSpecLibrary } from "@typespec/compiler";
+
+export const $lib = createTypeSpecLibrary({
+  name: "@azure-tools/typespec-azure-portal-core",
+  diagnostics: {
+    fileNotFound: {
```
witemple-msft, line 6: diagnostic messages should be skewer-case by default e.g. `"file-not-found"` instead of `fileNotFound`.

packages/typespec-azure-portal-core/src/lib.ts
```diff
@@ -0,0 +1,37 @@
...
+      },
+    },
+    invalidUsageDecorator: {
+      severity: "error",
+      messages: {
+        browse: `@browse decorator can be only applied to trackedResource and proxyResource`,
+        about: `@about decorator can be only applied to trackedResource and proxyResource`,
+        marketplaceOffer: `@marketPlaceOffer decorator can be only applied to trackedResource and proxyResource`,
+      },
+    },
```
witemple-msft, line 20: I suggest making the diagnostic names very specific and parameterizing, like this:

```suggestion
   "not-a-resource": {
      severity: "error",
      messages: {
        default: paramMessage`@${"decoratorName"} can only be applied to TrackedResource and ProxyResource models`
      },
    },
```

This way you don't have to repeat the message, and the diagnostic name (which appears in error messages) alone communicates what's gone wrong.

packages/typespec-azure-portal-core/src/lib.ts
```diff
@@ -0,0 +1,37 @@
...
+    invalidType: {
```
witemple-msft, line 21: Suggest `"invalid-offer-id"` or similar for the diagnostic name.

packages/typespec-azure-portal-core/src/lib.ts
```diff
@@ -0,0 +1,37 @@
...
+      severity: "error",
+      messages: {
+        marketplaceOfferId: `@marketplaceOffer id cannot have a blank space.`,
+      },
+    },
+    "essentials-maximum-usage": {
```
witemple-msft, line 27: Suggest `"too-many-essentials"`

packages/typespec-azure-portal-core/src/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface FilePath {
```
witemple-msft, line 1: @bterlson Do we have any mechanism for generating these types from the declarations in TypeSpec?

packages/typespec-azure-portal-core/test/decorators.test.ts
```diff
@@ -0,0 +1,172 @@
...
+  //   const result = getBrowseArgQuery(runner.program, Foo);
+  //   //const actual = getDisplayName(runner.program, name as ModelProperty);
+  //   strictEqual(Foo.kind, "Model");
+  // //strictEqual(actual, "nickName");
+  // });
+
+  it("test @browse with string", async () => {
+    const browseString = createbrowseTestDecorator(`"helloThisISArgQuery"`);
+    const { Foo } = await runner.compile(createTestSpec(browseString));
+    const result = getBrowseArgQuery(runner.program, Foo);
```
witemple-msft, line 65: I think it'll be more useful if the getBrowseArgQuery function returns a string rather than the model property where the string was declared, that way you will be able to just call it and get the value back rather than having to further process the ModelProperty node.

packages/typespec-azure-portal-core/test/decorators.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    strictEqual(Foo.kind, "Model");
+    strictEqual(result.kind, "ModelProperty");
+    strictEqual(result.type.kind, "String");
+    strictEqual(result.type.value, "helloThisISArgQuery");
+  });
+
+  it("test @browse on non-ARM resource", async () => {
```
witemple-msft, line 72: tip: no need to put "test" in front of the descriptions, since it's implied.

packages/typespec-azure-portal-core/test/decorators.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        ${browseString}
+        model Bar {}
+      `);
+    expectDiagnostics(diagnostics, {
+      code: "@azure-tools/typespec-azure-portal-core/invalidUsageDecorator",
+      message: "@browse decorator can be only applied to trackedResource and proxyResource",
+    });
+  });
+
+  it("test @about", async () => {
```
witemple-msft, line 84: tip: group the related tests for the same decorator into a `describe` block

witemple-msft (commented):
This is very nice progress, and I think it's going to be awesome when RPs can get default UX by just annotating their specs.

common/changes/@azure-tools/typespec-azure-portal-core/yejeelee-typespecAzureCore_2023-12-18-15-44.json
witemple-msft: Yes, just to chime in, we want this to be a first class component of the azure-tools system. While integrating it with our repo and release cycle has costs it will help make sure we don't break anyone using it, especially while it's new. The other component, which is the default UX emitter counterpart to this library, will be off-repo I think.

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+  normalizePath,
+  resolvePath,
+  validateDecoratorUniqueOnNode,
+  type DecoratorContext,
+  type Model,
+  type ModelProperty,
+  type Program,
+  type Type,
+} from "@typespec/compiler";
+import fs from "fs";
```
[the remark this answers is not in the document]
timotheeguerin: yeah that's a good idea, yeah definitely do not use any node apis (not just `fs`)

packages/typespec-azure-portal-core/lib/decorators.tsp
```diff
@@ -0,0 +1,31 @@
...
+model marketplaceOfferOptions {
+  id?: string;
+}
+
+model AboutOptions {
+  icon?: FilePath;
+  displayName?: string;
+  keywords?: string[];
+  learnMoreDocs?: string[];
+}
```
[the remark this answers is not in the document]
yejee94: added description :)

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: Question on this

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: fixed :) thank you!

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: question on this

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: question on this

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: fixed :) thank you!

packages/typespec-azure-portal-core/package.json
yejee94: question on this!

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -0,0 +1,253 @@
...
+  const { program } = context;
+  validateDecoratorUniqueOnNode(context, target, $browse);
+  isARMResource(program, target, "browse");
+  if (options && (options as Model).properties) {
+    const query = (options as Model).properties.get("argQuery");
+    if (query && query.type.kind === "Model") {
+      const sourceLocation = getSourceLocation(target);
+      const dirPath =
+        sourceLocation && sourceLocation.file.path && getDirectoryPath(sourceLocation.file.path);
+      let argQueryPath = query.type && (query.type as Model).properties.get("filePath");
```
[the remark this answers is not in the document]
yejee94: question on this!

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: will do! :) thank you so much!

packages/typespec-azure-portal-core/.mocharc.yaml
yejee94: converted to vitest :) thank you so much Tim!

packages/typespec-azure-portal-core/src/decorators.ts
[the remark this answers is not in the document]
yejee94: chat with Will : how to get the sourceLocation of the decorator instead of the target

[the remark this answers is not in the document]
timotheeguerin: Do you mean you want to emit a diagnostic on the decorator itself instead of the type it decorates?
context.decoratorTarget I think would be that or context.getArgumentAt()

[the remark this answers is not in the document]
witemple-msft: @timotheeguerin theres two different things that kind of got mixed up in this thread. What I suggested when chatting with @yejee94 is that we should be doing everything relative to the decorator application instead of the target, both the diagnostics, which is simple through ctx.decoratorTarget, but also computing the relative file path relative to the decorator rather than relative to the target, which could be important in the case of augmentations.

packages/typespec-azure-portal-core/src/lib.ts
yejee94: fixed :)

packages/typespec-azure-portal-core/src/lib.ts
yejee94: fixed :) thank you!

packages/typespec-azure-portal-core/src/lib.ts
yejee94: fixed :) thank you!!

packages/typespec-azure-portal-core/src/lib.ts
yejee94: fixed :) thank you!

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: talked to Will offline about this :) dismissed

packages/typespec-azure-portal-core/src/decorators.ts
yejee94: talked to Will offline about this :) dismissed

PR #68, opened 2023-12-28
fix(tcgc): temporary work-around of `@format("date")`

### LINKED ISSUE #64: [Bug][getAllOperations] Exception when defining a `utcDateime` parameter
**Describe the bug**

In `autorest.csharp`, we have a test which defines a `utcDatetime` parameter: https://github.com/Azure/autorest.csharp/blob/74938fddeede57bd4758720b05ea355eabd29645/test/TestProjects/sdk/newprojecttypespec/Azure.NewProject.TypeSpec/Azure.NewProject.TypeSpec.tsp#L292

```typescript
op addTimeHeader(
    @header("Repeatability-First-Sent") repeatabilityFirstSent?: utcDateTime 
): void;
```

Adopting tcgc on `getAllOperations` branch, and there will be an exception when compiling the test.
```
Error: Unknown format date
    at addFormatInfo (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/types.ts:146:15)
    at getSdkModelPropertyType (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/types.ts:850:3)
    at file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/package.ts:451:17
    at Array.map (<anonymous>)
    at getSdkBasicServiceMethod (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/package.ts:451:6)
    at getSdkServiceMethod (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/package.ts:500:10)
    at getSdkMethods (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/package.ts:613:17)
    at createSdkClientType (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/package.ts:653:14)
    at getSdkPackage (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/package.ts:673:5)
    at createSdkContext (file:///C:/Users/mingzhehuang/workspaces/archerzz/typespec-azure/packages/typespec-client-generator-core/src/decorators.ts:487:27)

```

**To Reproduce**
Build the test project in `autorest.csharp`, using a generator built with TCGC on `getAllOperations` branch.

**Expected behavior**
There should be no exception.

**Additional context**

archerzz:
- add new `SDKBuiltIntType`: `utcDatetime`
  - we cannot use `datetime` because it's occupied by SDKDateTimeType
- change `addFormatInfo()` to add suuport of format `date` and `date-time`
- add test case

resolve #64

### DIFF
--- a/packages/typespec-client-generator-core/test/types.test.ts
+++ b/packages/typespec-client-generator-core/test/types.test.ts
@@ -184,6 +184,8 @@ describe("typespec-client-generator-core: types", () => {
           azureLocationProperty: string;
           @format("etag")
           etagProperty: string;
+          @format("date")
+          dateProperty: string;
         }
       `
       );
@@ -340,6 +342,19 @@ describe("typespec-client-generator-core: types", () => {
       strictEqual(sdkType.kind, "datetime");
       strictEqual(sdkType.wireType.kind, "string");
       strictEqual(sdkType.encode, "rfc3339");
+    });
+    it("used as parameter", async function () {
+      await runner.compileWithBuiltInService(
+        `
+        op addTimeHeader(
+          @header("Repeatability-First-Sent") repeatabilityFirstSent?: utcDateTime 
+        ): void;
+      `
+      );
+      const m = runner.context.sdkPackage.clients[0].methods[0].parameters[0];
+      strictEqual(m.type.kind, "datetime");
+      strictEqual(m.type.wireType.kind, "string");
+      strictEqual(m.type.encode, "rfc7231");
     });
     it("rfc3339", async function () {
       await runner.compileWithBuiltInService(
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -116,6 +116,7 @@ export type SdkBuiltInKinds =
   | "boolean"
   | "date"
   | "time"
+  | "utcDatetime"
   | "any"
   | "int32"
   | "int64"
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -141,6 +141,10 @@ export function addFormatInfo<TServiceOperation extends SdkServiceOperation>(
       case "azurelocation":
         propertyType.kind = "azureLocation";
         break;
+      // TODO: we need to handle how to process @format("date-time") and utcDateTime
+      // We have a separate SdkDateTimeType, but for @format("date-time") the SDKType is SDKBuiltInType
+      // I cannot use `datetime` here, because it will conflict with SdkDateTimeType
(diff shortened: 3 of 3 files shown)


tadelesh:
Since TSP has `utcDateTime` type and `offsetDateTime` type, I think `@format("date-time")` is not a good way to express string with datetime format. Need to decide whether we should support it or add lint to core to avoid it.

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -184,6 +184,8 @@ describe("typespec-client-generator-core: types", () => {
           azureLocationProperty: string;
           @format("etag")
           etagProperty: string;
+          @format("date")
```
timotheeguerin, line 187: this shouldn't not be a thing, this means a string that is with a known pattern. Please don't make anything to have this being a way to specify dates

timotheeguerin (changes requested):

packages/typespec-client-generator-core/test/types.test.ts
timotheeguerin: if you want a date you use the `utcDateTime` or `offsetDateTime` type and specify the encoding.

archerzz: Yeah, @tadelesh and I talked about this. Right now, it's feasible. See test case in `autorest.csharp`: https://github.com/Azure/autorest.csharp/blob/00c74a505f8140f7789934c18a95b3749e0ecf80/test/TestProjects/sdk/newprojecttypespec/Azure.NewProject.TypeSpec/Azure.NewProject.TypeSpec.tsp#L265

So, I think we need to discourage people from using it:
- add a linter rule to report error on `@format("date")` and `@format("date-time")`
- What do we do when the TypeSpec spec is used with an OpenAPI emitter? see https://github.com/microsoft/typespec/blob/da99aa955b497c34cd6deb8f873fecd0d583c85b/packages/website/versioned_docs/version-latest/standard-library/reference/js-api/functions/%24format.md?plain=1#L18-L21

timotheeguerin: it is feasible to write but you shouldn't do anything about it. If the user says the type is a string it is a string not a date regardless of what the format says. `@format` is not a type indication is it just a known string pattern.
This is NOT the same as openapi where erverything neeeded to be a primitive type.

PR #91, opened 2024-01-05
Add support for `@encodedName` in typespec-autorest

timotheeguerin:
Core PR adding the decorator https://github.com/microsoft/typespec/pull/2779

### DIFF
--- a/packages/typespec-autorest/test/models.test.ts
+++ b/packages/typespec-autorest/test/models.test.ts
@@ -65,7 +65,7 @@ describe("typespec-autorest: model definitions", () => {
 
     expect(res.defs.Foo).toMatchObject({
       properties: {
-        xJson: { type: "integer", format: "int32" },
+        xJson: { type: "integer", format: "int32", "x-ms-client-name": "x" },
       },
     });
   });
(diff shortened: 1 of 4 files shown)


packages/typespec-autorest/test/models.test.ts
```diff
@@ -50,13 +47,43 @@ describe("typespec-autorest: model definitions", () => {
...
+      "Foo",
+      `model Foo {
+        @encodedName("application/json", "xJson")
+        x: int32;
+      };`
+    );
+
+    expect(res.defs.Foo).toMatchObject({
+      properties: {
+        xJson: { type: "integer", format: "int32" },
```
markcowl, line 68: shouldn't we be setting x-ms-client-name to match the original name here?

packages/typespec-autorest/test/models.test.ts
```diff
@@ -50,13 +47,43 @@ describe("typespec-autorest: model definitions", () => {
...
+      `model Foo {
+        @encodedName("application/json", "xJson")
+        @projectedName("json", "projectedJson")
+        x: int32;
+      };`
+    );
+
+    expect(res.defs.Foo).toMatchObject({
+      properties: {
+        xJson: { type: "integer", format: "int32" },
```
markcowl, line 85: same question

markcowl (approved):
A couple of questions on the tests, otherwise, lgtm

packages/typespec-autorest/test/models.test.ts
timotheeguerin: let me add it I think this might just be the `toMatchObject` which doesn't care if there is extra properties

timotheeguerin: Add to the test

PR #93, opened 2024-01-05
various enhancements

mario-guerra:
Added support for various features that are useful for data gathering during mid-semester reviews. Also corrected what I think is an oversight in that the local repo was not using the same branch as the upstream query.


mario-guerra:
@allenjzhang could you please review/approve these changes?

timotheeguerin (approved):

PR #95, opened 2024-01-08
[tcgc] add @encodedName support

### LINKED ISSUE #92: Add support for @encodedName in tcgc
Example simple implementation for autorest https://github.com/Azure/typespec-azure/pull/91 with json support only

iscai-msft:
fixes #92

### DIFF
--- a/packages/typespec-client-generator-core/src/public-utils.ts
+++ b/packages/typespec-client-generator-core/src/public-utils.ts
@@ -6,7 +6,6 @@ import {
   getEffectiveModelType,
   getFriendlyName,
   getNamespaceFullName,
-  getProjectedName,
   getSummary,
   ignoreDiagnostics,
   listServices,
@@ -14,6 +13,7 @@ import {
   ModelProperty,
   Namespace,
   Operation,
+  resolveEncodedName,
   Scalar,
   Type,
   Union,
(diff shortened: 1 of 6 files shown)


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -237,7 +231,12 @@ export function getDocHelper(context: SdkContext, type: Type): DocWrapper {
 }
 
 export function getWireName(context: SdkContext, type: Type & { name: string }) {
-  return getProjectedName(context.program, type, "json") ?? type.name;
+  // 1. Check if there's an encoded name
+  const encodedName = resolveEncodedName(context.program, type, "application/json");
+  if (encodedName !== type.name) return encodedName;
+  // 2. Check if there's deprecated language projection
+  const jsonProjectedProgram = createProjectedNameProgram(context.program, "json");
+  return jsonProjectedProgram.getProjectedName(type);
```
timotheeguerin, line 239:
```suggestion
  return getProjectedName(context.program, type, "json") ?? type.name;
```
still use what you had before otherwise you'll have the projection issue we encountered a while back

iscai-msft: ok removed references to `createProjectedNameProgram` throughout. Thanks!

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -956,7 +959,13 @@ describe("typespec-client-generator-core: types", () => {
       strictEqual(clientProjectedProp.kind, "property");
       strictEqual(clientProjectedProp.serializedName, "clientProjectedName");
 
-      // wire name test
+      // wire name test with encoded and projected
+      const jsonEncodedProp = sdkModel.properties.find(
+        (x) => x.kind === "property" && x.serializedName === "encodedWireName"
+      )!;
```
joheredi, line 965: find could return undefined and cause this to crash. It may be worth removing the `!` and handling undefined and throw an error that can help when debugging

joheredi (approved):
Left a minor comment

markcowl (approved):

PR #127, opened 2024-01-13
add @clientName

### LINKED ISSUE #114: Add `@clientName` or whatever else name to tcgc
To replace the 2nd part of `@projectedName` we need this new decorator to override the name for certain client

```
extern dec clientName(scope, name: valueof string)
```

iscai-msft:
fixes #114 

Bigger code change because I found out that our scoped decorator can't handle multiple different values per scope


tadelesh:
also fixes: https://github.com/Azure/typespec-azure-pr/issues/3496

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -62,24 +62,30 @@ function setScopedDecoratorData(
...
-  // for transitivity situation, we could allow scope extension
-  if (existing.scopes && !scope) {
-    existing.scopes = undefined;
-    return true;
+  if (!Object.keys(targetEntry).includes("") && !scope) {
+    for (const key of Object.keys(targetEntry)) {
+      if (key !== "") {
+        delete targetEntry[key];
+      }
+    }
```
tadelesh, line 88:
```suggestion
    context.program.stateMap(key).set(target, { [""]: value });
```

tadelesh: only nit, bc this logic is only for deprecated decorator.

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -2,6 +2,19 @@ using TypeSpec.Reflection;
...
+ * @param rename The rename you want applied to the object
+ * @param scope The language scope you want this decorator to apply to. If not specified, will apply to all language emitters
+ *
+ * @example
+ * ```typespec
+ * @clientName("nameInClient")
+ * op nameInService: void;
+ * ```
+ */
+extern dec clientName(target: unknown, rename: valueof string, scope?: valueof string);
```
timotheeguerin, line 16: Have you validated that's the name you want to use?

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -2,6 +2,19 @@ using TypeSpec.Reflection;
 
 namespace Azure.ClientGenerator.Core;
 
+/**
+ * Changes the name of a method, parameter, property, or model generated in the client SDK
+ * @param rename The rename you want applied to the object
+ * @param scope The language scope you want this decorator to apply to. If not specified, will apply to all language emitters
+ *
+ * @example
```
timotheeguerin, line 10: Might be good to have a scoped example as well

packages/typespec-client-generator-core/lib/decorators.tsp
iscai-msft: yes we're leaning more towards `clientName`, but will follow up and make sure we're all on the same page

srnagar (approved):
LGTM

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -38,19 +38,19 @@ import { getAllModels, getSdkEnum, getSdkModel } from "./types.js";
...
 
 function getScopedDecoratorData(context: SdkContext, key: symbol, target: Type): any {
-  const retval = context.program.stateMap(key).get(target);
+  const retval: Record<string, any> = context.program.stateMap(key).get(target);
   if (retval === undefined) return retval;
-  if (!retval.scopes) return retval.value; // this means it applies to all languages
-  if (retval.scopes.includes(context.emitterName)) return retval.value;
-  return undefined;
+  if (Object.keys(retval).includes(context.emitterName)) return retval[context.emitterName];
+  return retval[""]; // in this case it applies to all languages
```
joheredi, line 44: nit: any chance we could use a value to determine all instead of empty string? `*` maybe?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -795,3 +800,18 @@ export function $flattenProperty(context: DecoratorContext, target: ModelPropert
...
   return getScopedDecoratorData(context, flattenPropertyKey, target) ?? false;
 }
+
+const clientNameKey = createStateSymbol("clientName");
+
+export function $clientName(
+  context: DecoratorContext,
+  entity: Type,
+  value: string,
+  scope?: string
```
joheredi, line 810: I forgot if we can use string literal unions for TSP decorator parameters. I was thinking it may be useful to provide hints about possible values.

```ts
export function $clientName(
  context: DecoratorContext,
  entity: Type,
  value: string,
  scope?: "dotnet" | "java" | "python" | "javascript" | "go" | (string & {}))
```

That, in typescript will allow you to input any string but intellisense will suggest the strings in the union. Not sure if this would work with the TSP language server

joheredi (approved):

packages/typespec-client-generator-core/src/decorators.ts
timotheeguerin: or a `symbol` to prevent any conflicts

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: yep will do

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: I'm not sure it would, but might as well try

iscai-msft: tests still pass, not sure how much it will help the UI, but I think it's good to have even just for us. Thanks!

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -38,19 +38,19 @@ import { getAllModels, getSdkEnum, getSdkModel } from "./types.js";
 export const namespace = "Azure.ClientGenerator.Core";
 
 function getScopedDecoratorData(context: SdkContext, key: symbol, target: Type): any {
-  const retval = context.program.stateMap(key).get(target);
+  const retval: Record<string, any> = context.program.stateMap(key).get(target);
   if (retval === undefined) return retval;
-  if (!retval.scopes) return retval.value; // this means it applies to all languages
-  if (retval.scopes.includes(context.emitterName)) return retval.value;
-  return undefined;
+  if (Object.keys(retval).includes(context.emitterName)) return retval[context.emitterName];
```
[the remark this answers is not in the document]
iscai-msft: `context.emitterName` will always be available. If someone hasn't passed it in, we call our own function to extract the emitter name. We already have this as a semi-requirement, but tcgc doesn't work if there's no `emitterName`. In the future, we're thinking of making `emitterName` a required argument to `createSdkContext`

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -36,21 +37,22 @@ import { createStateSymbol, reportDiagnostic } from "./lib.js";
 import { getAllModels, getSdkEnum, getSdkModel } from "./types.js";
 
 export const namespace = "Azure.ClientGenerator.Core";
+const AllScopes = Symbol.for("@azure-core/typespec-client-generator-core/all-scopes");
 
 function getScopedDecoratorData(context: SdkContext, key: symbol, target: Type): any {
-  const retval: Record<string, any> = context.program.stateMap(key).get(target);
+  const retval: Record<string | symbol, any> = context.program.stateMap(key).get(target);
```
timotheeguerin, line 43:
```suggestion
  const retval: Record<string | typeof AllScopes, any> = context.program.stateMap(key).get(target);
```

that way only that symbol is allowed

PR #156, opened 2024-01-22
Union enum

iscai-msft:
fixes https://github.com/Azure/typespec-azure-pr/issues/3813


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -531,6 +536,54 @@ export function getSdkEnum(context: SdkContext, type: Enum, operation?: Operatio
...
+      nullable: false,
+    });
+  }
+  return values;
+}
+
+function getSdkUnionEnum(context: SdkContext, type: UnionEnum, operation?: Operation) {
+  let sdkType = context.modelsMap?.get(type.union) as SdkEnumType | undefined;
+  if (!sdkType) {
+    if (!type.union.name) throw new Error("Your union must be named in order to generate an enum");
```
srnagar, line 564:
```suggestion
    if (!type.union.name) throw new Error("Union must be named in order to generate an enum");
```

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -531,6 +536,54 @@ export function getSdkEnum(context: SdkContext, type: Enum, operation?: Operatio
...
+      ...getSdkTypeBaseHelper(context, type.union, "enum"),
+      name: getLibraryName(context, type.union),
+      description: docWrapper.description,
+      details: docWrapper.details,
+      valueType: { ...getSdkTypeBaseHelper(context, type.kind, "string"), encode: "string" },
+      values: [],
+      nullable: false,
+      isFixed: !type.open,
+      isFlags: false,
+      usage: UsageFlags.None, // We will add usage as we loop through the operations
```
srnagar, line 577: Can we leave this as undefined until it's set later instead of setting it to `None`?

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -1370,6 +1370,43 @@ describe("typespec-client-generator-core: types", () => {
...
+    it("union to extensible enum values", async () => {
+      await runner.compileWithBuiltInService(`
+      union PetKind {
+        Cat: "cat",
+        Dog: "dog",
+        string,
+      }
+
+      @route("/extensible-enum")
+      @get
```
srnagar, line 1382: Should this be a `put`?

packages/typespec-client-generator-core/src/types.ts
srnagar: Should we include a couple of values to inform the user which union we are referring to in the error? For e.g. `Union with values a, b, c.... must be named in order to generate an enum`.

srnagar (approved):

markcowl (changes requested):
Wait for release

packages/typespec-client-generator-core/src/types.ts
iscai-msft: `usage` is a required value so we'll have a compile error if we set it to `undefined`

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -235,22 +230,26 @@ export function getWireName(context: SdkContext, type: Type & { name: string })
...
   context: SdkContext,
-  type: Type,
+  type: Type | string,
   kind: TKind
-): DefaultSdkTypeBase<TKind> {
+): {
+  __raw?: Type;
+  nullable: boolean;
+  deprecation?: string;
+  kind: TKind;
```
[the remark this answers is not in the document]
iscai-msft: ok we can, I thought it would be cleaner to not create a type, but you are the expert so I've learned something 🖖

markcowl (approved):

packages/typespec-client-generator-core/src/types.ts
iscai-msft: We can, also there's actually no way we'll enter this code if the union isn't named, bc we then generate it as a union of literals bc we only call this function if the union is named. I honestly might just use `!` to avoid confusion

packages/typespec-client-generator-core/test/types.test.ts
iscai-msft: good catch, thanks!

PR #166, opened 2024-01-25
create multipart file sdktype for bytes in a multipart formdata model

### LINKED ISSUE #165: [tcgc] add support for a `MultiPartFile` type
Currently, our code at large assumes models to be JSON objects. Since we need to support `multipart/form-data`, this is not the case. For example, in the following model
```typespec
model JsonModel {
  prop: bytes
}
```
`prop` will default to a b64 encoded string, which is the correct behavior. However, in a model used for multipart, we would instead want to recognize the `bytes` as being for a file, and also add information about `content-type` and `filename`. 

```typespec
model MultipartInputModel {
  picture: bytes;
}
```

So proposal is to add a new primitive type like the following, which would be the above property `picture`'s type

```typescript
interface MultipartFormDataFile:
  kind: "multipartformdatafile";
  fileStream: SdkBytesType (with default encoding of `binary`, still has access to @encodedType);
  filename: string | undefined;
  contentType: string | undefined;
```

iscai-msft:
fixes #165


common/changes/@azure-tools/typespec-client-generator-core/multipartType_2024-01-25-00-03.json
timotheeguerin, line 1: Delete this file we just migrated to Pnpm

lmazuel (approved):
LGTM for what I can read TypeScript :)

joheredi (approved):

PR #184, opened 2024-01-29
HotFix: Autorest emitter should generated format: decimal

timotheeguerin:
`decimal128` might be the correct format as established in the registry but autorest doesn't support it today and treats `decimal` as a 128bit decimal


packages/typespec-autorest/src/openapi.ts
```diff
@@ -2111,7 +2111,7 @@ function createOAPIEmitter(program: Program, options: ResolvedAutorestEmitterOpt
       case "decimal":
         return { type: "number", format: "decimal" };
       case "decimal128":
-        return { type: "number", format: "decimal128" };
+        return { type: "number", format: "decimal" };
```
markcowl, line 2114: Need to update primitive type test in autorest, otherwise LGTM

markcowl (approved):

PR #189, opened 2024-01-30
[emitter-python] update DPG how-to guide

msyyc:
update how-to guide for python part


docs/howtos/DataPlane Generation - DPG/03convenient.mdx
```diff
@@ -168,9 +167,9 @@ The two possible value for the `Access` enum are `internal` and `public`.
 <TabItem value="python" label="Python">
 
 ```python
-class _GetModel:
-  ...
-# FIXME
+# can't import form models directly
+from petstorenamespace.models import GetModel # will report error
+from petstorenamespace.models._models import GetModel # OK but not recommended in Python
```
iscai-msft, line 172: we shouldn't document private imports

msyyc:
already done https://github.com/Azure/typespec-azure/blob/main/docs/howtos/Client%20Generation/03convenient.mdx

docs/howtos/DataPlane Generation - DPG/03convenient.mdx
iscai-msft: I think we should just not even include the private imports

msyyc: Updated @iscai-msft

docs/howtos/Client Generation/05union.mdx
```diff
@@ -37,11 +37,11 @@ op completions(@body input: CompletionInput): CompletionResult;
 <TabItem value="python" label="Python">
 
 ```python
-def analyze(*, mode: Literal["strict", "lenient"], image: bytes) -> AnalyzeResult:
+def analyze(image: bytes, *, mode: Literal["strict", "lenient"]) -> AnalyzeResult:
   ...
 
 class CompletionInput:
-  input: str | List[str] = rest_field(readonly=True)
+  input: Union[str, List[str]] = rest_field(readonly=True)
```
msyyc, line 44: Introduced in 3.10: https://docs.python.org/3/library/stdtypes.html#types-union and 3.8/3.9 doesn't support yet.

iscai-msft (approved):

PR #200, opened 2024-02-01
[tcgc] Fix list of bytes

iscai-msft:


timotheeguerin (approved):

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -800,14 +800,18 @@ function getSdkBodyModelPropertyType(
...
+    (base.type.kind === "array" && base.type.valueType.kind === "bytes");
   return {
     ...base,
     kind: "property",
     optional: type.optional,
     visibility: getSdkVisibility(context, type),
     discriminator: false,
     serializedName: getPropertyNames(context, type)[1],
-    isMultipartFileInput: base.type.kind === "bytes" && operationIsMultipart,
+    isMultipartFileInput: isBytesInput && operationIsMultipart,
```
MaryGao, line 814: I was wondering the case 

```
@post op uploadFile(
              @body body: {
                name: string;
                foo: {
                      name: string;
                      content: bytes;
                };
              }
            ): void;
```

iscai-msft: that's a good point @MaryGao, I do think it's fair that if there's an explicit encoding set on the `bytes`, we treat it not as a multipart file object, but instead as encoded bytes. Let me post this question in the dpg redmond sync team and see what other people think

lmazuel: The whole point of MFD is that you can pass the bytes directly without doing a base64 of them (perf impact, lower bandwidth). If, as a service, you're ok with base64, then just design your service to take a JSON as input, it's much more simple.
IOW, it's a corner case of MFD as an HTTP design in the first place. I don't want to go there at all. 
I would say let's even refuse `@encode` for clarity for now, so we don't shoot ourselves later. I'd rather people complain early, than they ship undertested SDK that reach people and it doesn't work.

MaryGao: @lmazuel I may not get the point completely and do you mean we could ignore them today?

BTW I created an issue to explain myself and for any further discussion.
https://github.com/Azure/typespec-azure/issues/308

PR #202, opened 2024-02-01
update-js-dpg-docs

qiaozha:


joheredi (approved):
Thanks for documenting this. It may be worth mentioning instead of just NOT SUPPORTED that RLC output is not impacted by client.tsp

docs/howtos/DataPlane Generation - DPG/05union.mdx
```diff
@@ -249,7 +299,7 @@ public Response<long> Foo(string x, Cancellation token);
 <TabItem value="typescript" label="Typescript" >
 
 ```typescript
-FIXME;
+NOT_SUPPORTED;
```
qiaozha, line 302:
```suggestion
JS RLC is not in the business of customization with client.tsp
```

PR #203, opened 2024-02-01
Fix Location resource, add mechanism for additional non-resource path segments

markcowl:


packages/typespec-azure-resource-manager/lib/decorators.tsp
```diff
@@ -186,3 +186,21 @@ extern dec armCommonTypesVersion(
...
   version: valueof string | EnumMember
 );
+
+/**
+ * This decorator is used on Azure Resource Manager resources that are not based on
+ * Azure.ResourceManager common types.
+ *
+ * @param propertiesType: The type of the resource properties.
+ */
+extern dec armBuiltInResource(target: Model, propertiesType: Model);
```
timotheeguerin, line 196: Are those decorators supposed to be private?

packages/typespec-azure-resource-manager/package.json
```diff
@@ -52,33 +52,33 @@
...
-    "@typespec/versioning": "workspace:~0.52.0",
-    "@types/node": "~18.11.9",
-    "@typespec/eslint-config-typespec": "workspace:~0.52.0",
-    "@typespec/library-linter": "workspace:~0.52.0",
-    "@typespec/eslint-plugin": "workspace:~0.52.0",
-    "@typespec/tspd": "workspace:~0.46.0",
-    "eslint": "^8.55.0",
-    "vitest": "^1.2.0",
-    "@vitest/coverage-v8": "^1.1.0",
+    "@azure-tools/typespec-autorest": "workspace:~",
```
timotheeguerin, line 64: Did pnpm change this? Wondering if this might break our dev bumping script

daviwil (approved):
Overall design looks good, thanks for fixing this! Agree with Tim's comments, too

packages/typespec-azure-resource-manager/package.json
allenjzhang: I noticed core is also listed as changed. I wonder if there is a rebase of the latest missing here.

packages/typespec-azure-resource-manager/src/operations.ts
```diff
@@ -205,7 +211,8 @@ export function $armRenameListByOperation(
     [parentTypeName, parentFriendlyTypeName] = getArmParentName(context.program, resourceType);
   }
   const parentType = getParentResource(program, resourceType);
-  if (parentType) {
+
+  if (parentType && !isArmBuiltInResource(program, parentType)) {
```
allenjzhang, line 215: We originally talked about allowing non-resource model in @parentResource to participate in path resolution for this fix. Are we to track and fix that separately now?

packages/typespec-azure-resource-manager/lib/decorators.tsp
markcowl: No,  A spec author could use these to more easily create additional path segments in the resource URI, so they should be public

packages/typespec-azure-resource-manager/package.json
markcowl: hmm.  I ran pnpm update and pnpm change, I certainly did not change this intentionally

packages/typespec-azure-resource-manager/src/operations.ts
markcowl: That is actually what this does.  If you apply the `@armBuiltInResource` decorator to a non-resource type with a path key and segment property, it can be made the parent of any resource.  It is basically how you mark a virtual resource that nas no operations in the path

packages/typespec-azure-resource-manager/package.json
timotheeguerin: I’ll need to keep an eye on that, if this keeps happening either we need a setting to prevent it or we need to make sure our system work with it

PR #210, opened 2024-02-02
Don't build packages from core twice

timotheeguerin:
This was to simplify some  install dependencies but this actually cause things in core from being built twice


package.json
```diff
@@ -4,28 +4,29 @@
   "private": true,
   "packageManager": "pnpm@8.13.1",
   "scripts": {
-    "build": "pnpm -r --workspace-concurrency=Infinity --aggregate-output --reporter=append-only build ",
+    "run-all": "pnpm -r --filter='!./core'",
+    "build": "pnpm run-all --workspace-concurrency=Infinity --aggregate-output --reporter=append-only build ",
```
markcowl, line 8: This makes it slightly less convenient to do a cross-cutting change, and even to start up from a new fork, where I will need to build both core and this repo.

At the very least, we should update the CONTRIBUTING.md for cross-cutting changes to match the new workflow

markcowl (approved):

PR #212, opened 2024-02-03
Fix: Prepare publish 

timotheeguerin:


eng/scripts/prepare-publish.js
```diff
@@ -19,45 +37,50 @@ const Major = 3;
...
 typespecRun("git", "fetch", "https://github.com/microsoft/typespec", "main");
 if (production) {
   typespecRun("git", "merge", "--ff-only", "FETCH_HEAD");
 }
 // Stage the typespec core publish
 typespecRun("pnpm", "change", "version");
-typespecRun("pnpm", "update-latest-docs");
+if (!args.values.onlyBumpVersions) {
+  typespecRun("pnpm", "update-latest-docs");
+}
```
markcowl, line 66: Shouldn't the 'git add -A' below only occur if there are changes (i.e. with the 'git commit -m' on line 51)

markcowl (approved):

Issue #215, opened 2024-02-05
[Bug][TCGC] TCGC report `conflicting-multipart-model-usage` wrongly


msyyc:
TCGC reports the following error when generating Python SDK for https://github.com/Azure/azure-rest-api-specs/pull/26702
```
D:/dev/autorest.python/packages/typespec-python/alpha/Azure.Analytics.Purview.DataMap/models.tsp:479:7 - error @azure-tools/typespec-client-generator-core/conflicting-multipart-model-usage: Invalid encoding
> 479 | model AtlasErrorResponse {
      |       ^^^^^^^^^^^^^^^^^^

Found 1 error.
```

msyyc:
I think it is related with https://github.com/Azure/typespec-azure/blob/ebfe63960277356c611f15b2404a0ae6f2d9e6ed/packages/typespec-client-generator-core/src/types.ts#L443-L454

According to code annotation, the error raised when multipart request is used in regular input while the logic doesn't distinguish input and output, so it may be wrongly reported for output model.

PR #223, opened 2024-02-06
[TCGC] fix multipart lint

msyyc:
fixes https://github.com/Azure/typespec-azure/issues/215

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -441,9 +441,7 @@ export function getSdkModel(context: SdkContext, type: Model, operation?: Operat
   let isFormDataType = false;
   if (httpBody) {
     const isMultipartOperation = httpBody.contentTypes.includes("multipart/form-data");
-    if (isMultipartOperation && httpBody.type === type) {
-      isFormDataType = true;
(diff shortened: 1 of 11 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -437,9 +437,12 @@ export function getSdkModel(context: SdkContext, type: Model, operation?: Operat
   const httpOperation = operation
     ? ignoreDiagnostics(getHttpOperation(context.program, operation))
     : undefined;
-  const isFormDataType = httpOperation
-    ? Boolean(httpOperation.parameters.body?.contentTypes.includes("multipart/form-data"))
-    : false;
+  const httpBody = httpOperation?.parameters.body;
+  let isFormDataType = false;
+  if (httpBody) {
+    const isMultipartOperation = httpBody.contentTypes.includes("multipart/form-data");
```
timotheeguerin, line 443: shouldn't you just check for startwith`multipart/` there is `multtipart/mixed` too

iscai-msft: ooh good point

timotheeguerin (approved):

PR #228, opened 2024-02-06
Bump version for February release

timotheeguerin:


markcowl (commented):
Looks like we need to exclude .changesets from formatting (or not have that added changeset)

markcowl (approved):
Just need to fix the changeset issue

PR #231, opened 2024-02-07
Azure `Page` type `nextLink` properties should be `readOnly: true`

daviwil:
This PR fixes https://github.com/Azure/typespec-azure-pr/issues/3924 by marking `nextLink` properties with `@visibility("read")` so that the OpenAPI output for these properties contains `readOnly: true`.


.changeset/next-link-read-only-2024-1-7-14-13-43.md
```diff
@@ -0,0 +1,5 @@
+---
+"@azure-tools/typespec-azure-core": patch
```
timotheeguerin, line 2:
```suggestion
"@azure-tools/typespec-azure-core": minor
```

timotheeguerin: As I think this is not really a fix but a change

timotheeguerin (approved):
Looks good, is there any breaking change issue with this? (apart from the swagger spec will need to be regenerated in spec repos)

.changeset/next-link-read-only-2024-1-7-14-13-43.md
daviwil: Good point, thanks!

daviwil:
@timotheeguerin We probably should mention it.  This is something that should have been set to begin with but got missed during Azure Core development. @allenjzhang and I discussed that it's worth introducing a break.  Anyone else have opinions here?  @markcowl @johanste

markcowl (approved):

PR #242, opened 2024-02-08
Add flatten to prop

iscai-msft:


.changeset/twelve-dingos-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@azure-tools/typespec-client-generator-core": minor
+---
+
+add .flatten onto the sdk type for a model's property, corresponds to @flatten
```
timotheeguerin, line 5:
```suggestion
add `.flatten` onto the sdk type for a model's property, corresponds to @flatten
```

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -421,6 +422,7 @@ function addDiscriminatorToModelType(
       nameInClient: discriminator.propertyName,
       apiVersions: getAvailableApiVersions(context, model.__raw!),
       isMultipartFileInput: false, // discriminator property cannot be a file
+      flatten: false,
```
joheredi, line 425: Can we add a comment why this is always false?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -238,6 +238,7 @@ export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
   serializedName: string;
   isMultipartFileInput: boolean;
   visibility?: Visibility[];
+  flatten: boolean;
```
joheredi, line 241: Adding this could potentially break emitters if they are building the SdkBodyModelPropertyType. Can we mark this as optional?

joheredi (approved):

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: Emitters shouldn't be building body model property types, the structure of the tcgc types has been that tcgc owns the structure of the interfaces, and emitters should either only access the properties on the sdk type

PR #248, opened 2024-02-09
Hotfix: Various issues with using union instead of enum

timotheeguerin:
Issues fixed
- [x] `reaonly` not applied on status enums if using `use-read-only-status-enum`
- [x] `union` declaration not allowed for lro status
- [x] `enum` inside of a union doesn't resolve to anything
- [x] `UnionType | null` would endup creating an inline enum instead of `$ref`


packages/typespec-azure-resource-manager/src/rules/arm-resource-provisioning-state-rule.ts
```diff
@@ -9,7 +10,7 @@ export const armResourceProvisioningStateRule = createRule({
   description: "Check for properly configured provisioningState property.",
   messages: {
     default:
-      "The RP-specific property model in the 'properties' property of this resource must contain a 'provisioningState property.  The property type should be an enum, and it must specify known state values 'Succeeded', 'Failed', and 'Canceled'.",
+      "The RP-specific property model in the 'properties' property of this resource must contain a 'provisioningState property.  The property type should be an enum or a union of string values, and it must specify known state values 'Succeeded', 'Failed', and 'Canceled'.",
     missingValues: paramMessage`The "@knownValues" decorator for provisioningState, must reference an enum with 'Succeeded', 'Failed', 'Canceled' values. The enum is missing the values: [${"missingValues"}].`,
```
allenjzhang, line 14: nit: Can you add recommend spreading `ResoourceProvisioningState` rather than declare themselves?

packages/typespec-azure-resource-manager/src/rules/arm-resource-provisioning-state-rule.ts
```diff
@@ -9,7 +10,7 @@ export const armResourceProvisioningStateRule = createRule({
   description: "Check for properly configured provisioningState property.",
   messages: {
     default:
-      "The RP-specific property model in the 'properties' property of this resource must contain a 'provisioningState property.  The property type should be an enum, and it must specify known state values 'Succeeded', 'Failed', and 'Canceled'.",
+      "The RP-specific property model in the 'properties' property of this resource must contain a 'provisioningState property.  The property type should be an enum or a union of string values, and it must specify known state values 'Succeeded', 'Failed', and 'Canceled'.",
```
allenjzhang, line 13: Also please have an issue for updating all examples to union if not already. Thx

packages/typespec-azure-resource-manager/test/rules/arm-resource-provisioning-state.test.ts
```diff
@@ -60,7 +83,7 @@ describe("typespec-azure-resource-manager: arm resource provisioning state rule"
       .toEmitDiagnostics({
         code: "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state",
         message:
-          "The RP-specific property model in the 'properties' property of this resource must contain a 'provisioningState property.  The property type should be an enum, and it must specify known state values 'Succeeded', 'Failed', and 'Canceled'.",
+          "The RP-specific property model in the 'properties' property of this resource must contain a 'provisioningState property.  The property type should be an enum or a union of string values, and it must specify known state values 'Succeeded', 'Failed', and 'Canceled'.",
```
allenjzhang, line 86: Do we still need to mention `enum` here?

allenjzhang (approved):
See comments.

packages/typespec-azure-resource-manager/src/rules/arm-resource-provisioning-state-rule.ts
timotheeguerin: well we can't spread with union we just include the enum as a variant. That message is already quite long I think we should have the markdown doc written for it though to show all the details.

We might also want to chagne this rule when we enabvle the `no-enum` rule to just say use union

packages/typespec-azure-resource-manager/test/rules/arm-resource-provisioning-state.test.ts
timotheeguerin: for now we do as this is a hotfix where we just allow both state to live side by side. We have an issue to do the cleanup for this sprint that will be in next release.

PR #255, opened 2024-02-12
[tcgc] fix string kinds to just use what is coming from tsp

iscai-msft:


.chronus/changes/fix_string_formats-2024-1-12-16-54-4.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+allows us to generate arm packages
```
timotheeguerin, line 7: not sure this will make much sense to what is fixed here to someone looking at the changelog

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -109,21 +109,15 @@ function addFormatInfo(
       case "uuid":
       case "password":
       case "etag":
+      case "arm-id":
```
timotheeguerin, line 112: I thought we agreed that `@format` was not going to be used

iscai-msft: oh you're right, these are all actual scalar types right?

timotheeguerin: there should be, `password` is just `@secret`, if there is some missing that are needed right now then we should look into adding them as soon as possible.

iscai-msft: ok made them actual scalar kinds. to clarify, we should never look at the `@format` decorator later down the line right?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -195,12 +189,20 @@ function getScalarKind(scalar: Scalar): SdkBuiltInKinds {
     case "decimal":
     case "plainDate":
     case "plainTime":
+    case "arm-id":
```
timotheeguerin, line 192: Sorry a lot of those are not built-in the compiler, they are in azure-core so you have to check the right namespace

tadelesh:
close as it has been refined in scalar refactor in tcgc.

PR #277, opened 2024-02-20
Support new added `@flattenProperty` decorator in autorest emitter

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/276

### DIFF
--- a/.chronus/changes/flatten-2024-1-20-11-30-53.md
+++ b/.chronus/changes/flatten-2024-1-20-11-30-53.md
@@ -4,4 +4,4 @@ packages:
   - "@azure-tools/typespec-autorest"
 ---
 
-Support new added @flattenProperty decorator.
+Support `@flattenProperty` decorator.
--- a/.chronus/changes/flatten-2024-1-20-11-30-53.md
+++ b/.chronus/changes/flatten-2024-1-20-11-30-53.md
@@ -4,4 +4,4 @@ packages:
   - "@azure-tools/typespec-autorest"
 ---
 
-Support new added @flattenProperty decorator.
+Support new added @flattenProperty decorator.
--- a/.chronus/changes/flatten-2024-1-20-11-30-53.md
+++ b/.chronus/changes/flatten-2024-1-20-11-30-53.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+Support new added @flattenProperty decorator.
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -12,6 +12,7 @@ import {
   SdkContext,
   createSdkContext,
   getClientNameOverride,
+  shouldFlattenProperty,
 } from "@azure-tools/typespec-client-generator-core";
 import {
   ArrayModelType,
@@ -1943,6 +1944,10 @@ function createOAPIEmitter(
         enumSchema["x-ms-enum"]!.name = (getPropertyType(typespecType) as Model).name;
         return enumSchema;
       }
+    }
+
+    if (typespecType.kind === "ModelProperty" && shouldFlattenProperty(tcgcSdkContext, typespecType)) {
+      newTarget["x-ms-client-flatten"] = true;
     }
 
     attachExtensions(typespecType, newTarget);
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -1946,7 +1946,10 @@ function createOAPIEmitter(
       }
     }
 
-    if (typespecType.kind === "ModelProperty" && shouldFlattenProperty(tcgcSdkContext, typespecType)) {
+    if (
+      typespecType.kind === "ModelProperty" &&
+      shouldFlattenProperty(tcgcSdkContext, typespecType)
+    ) {
       newTarget["x-ms-client-flatten"] = true;
     }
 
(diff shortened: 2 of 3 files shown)


packages/typespec-autorest/src/openapi.ts
timotheeguerin, line 1: Do we expect this decorator to be added in the main files or client.tsp? As if it is in the client.tsp it won't get picked up by autorest today.

tadelesh: It should be in `client.tsp`. Is there any blocking that we could not change autorest emitter's compile enter point to `client.tsp`?

timotheeguerin: I think that defeats the point of the separate client.tsp. However i have always been of the opinion that client.tsp should be imported from main

tadelesh: Since swagger contains not only the API definition but also some client related information, I think autorest emitter compile from client.tsp is reasonable.

timotheeguerin: It is but then why do we have a different entry point, this separation is just for the spec repo, we should just import client.tsp from main.

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
+import { deepStrictEqual } from "assert";
+import { describe, it } from "vitest";
+import { openApiFor } from "./test-host.js";
+
+describe("typespec-autorest: flatten", () => {
```
timotheeguerin, line 5: now that we moved to vitest we don't actually need this top level describe, it just adds another level of nesting in the UI too

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  it("tcgc flatten decorator", async () => {
+    const res = await openApiFor(
+      `
+      @service
+      namespace Test;
+
```
timotheeguerin, line 11:
```suggestion
```
I don't think this needed, good to keep the test as small as possible

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        #suppress "deprecated" "@flattenProperty decorator is not recommended to use."
+        @flattenProperty
+        properties?: WidgetProperties;
+      }
+
+      model WidgetProperties {
+        prop: string;
+      }
+
+      op get(): void;
```
timotheeguerin, line 22: remove this too

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      },
+      type: "object",
+    });
+  });
+
+  it("openapi extension decorator", async () => {
+    const res = await openApiFor(
+      `
+      @service
+      namespace Test;
```
timotheeguerin, line 41: same here

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      model Widget {
+        @extension("x-ms-client-flatten", true)
+        properties?: WidgetProperties;
+      }
+
+      model WidgetProperties {
+        prop: string;
+      }
+
+      op get(): void;
```
timotheeguerin, line 52: and here

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
+import { deepStrictEqual } from "assert";
+import { describe, it } from "vitest";
+import { openApiFor } from "./test-host.js";
+
+describe("typespec-autorest: flatten", () => {
+  it("tcgc flatten decorator", async () => {
```
timotheeguerin, line 6:
```suggestion
  it("applies x-ms-client-flatten for property marked with @flattenProperty", async () => {
```

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        properties: {
+          $ref: "#/definitions/WidgetProperties",
+          "x-ms-client-flatten": true,
+        },
+      },
+      type: "object",
+    });
+  });
+
+  it("openapi extension decorator", async () => {
```
timotheeguerin, line 37: do we need this test?

packages/typespec-autorest/src/openapi.ts
tadelesh: Oh, if that is the only reason for import main from client, then it is non-sense. I stand for import client from main.

.chronus/changes/flatten-2024-1-20-11-30-53.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+Support new added @flattenProperty decorator.
```
timotheeguerin, line 7:
```suggestion
Support `@flattenProperty` decorator.
```

packages/typespec-autorest/test/flatten.test.ts
```diff
@@ -0,0 +1,28 @@
+import { deepStrictEqual } from "assert";
+import { it } from "vitest";
+import { openApiFor } from "./test-host.js";
+
+it("applies x-ms-client-flatten for property marked with @flattenProperty", async () => {
+  const res = await openApiFor(
+    `
+    model Widget {
+      #suppress "deprecated" "@flattenProperty decorator is not recommended to use."
```
timotheeguerin, line 9:
```suggestion
      #suppress "deprecated" "for test"
```

timotheeguerin (approved):

PR #283, opened 2024-02-20
Add ArmResourceHead operation template

allenjzhang:


timotheeguerin (approved):
If thats a template we agreed on adding looks good to me

packages/typespec-azure-resource-manager/lib/operations.tsp
```diff
@@ -108,9 +108,32 @@ op ArmResourceRead<
...
 
+/**
+ * A resource HEAD operation
+ * @template Resource the resource being checked
+ * @template BaseParameters Optional. Allows overriding the operation parameters
+ * @template Parameters Optional. Additional parameters after the path parameters
+ * @template Response Optional. The success response for the read operation
+ * @template Error Optional. The error response, if non-standard.
+ */
+@autoRoute
```
johanste, line 119: Is there any way to take an existing GET operation? To avoid any potential mismatches of parameters/response etc?

packages/typespec-azure-resource-manager/lib/operations.tsp
```diff
@@ -108,9 +108,32 @@ op ArmResourceRead<
...
+@doc("Check a {name} via HEAD operation", Resource)
+@head
+op ArmResourceHead<
```
markcowl, line 122: Should this be `ArmResourceExists`?

packages/typespec-azure-resource-manager/lib/operations.tsp
```diff
@@ -108,9 +108,32 @@ op ArmResourceRead<
...
+/**
+ * A resource HEAD operation
+ * @template Resource the resource being checked
+ * @template BaseParameters Optional. Allows overriding the operation parameters
+ * @template Parameters Optional. Additional parameters after the path parameters
+ * @template Response Optional. The success response for the read operation
+ * @template Error Optional. The error response, if non-standard.
+ */
+@autoRoute
+@doc("Check a {name} via HEAD operation", Resource)
```
markcowl, line 120:
```suggestion
@doc("Check for the existence of a {name}.", Resource)
```

Since this will be user-facing docs, customers likely don't need to see the operation verb in the description

packages/typespec-azure-resource-manager/lib/operations.tsp
```diff
@@ -108,9 +108,32 @@ op ArmResourceRead<
...
+@head
+op ArmResourceHead<
+  Resource extends ArmResourceBase,
+  BaseParameters = DefaultBaseParameters<Resource>,
+  Parameters extends {} = {},
+  Response extends {} = OkResponse,
```
markcowl, line 126: This should return one of the ARM responses, either with an empty body, or no content.  (i.e. ArmResponse).  Also, don't we specifically want to describe the 404 in this case, since this should not be treated as a standard error?

packages/typespec-azure-resource-manager/test/resource.test.ts
```diff
@@ -1021,4 +1021,77 @@ describe("typespec-azure-resource-manager: ARM resource model", () => {
...
+          description: "The name of the widget",
+          in: "path",
+          name: "widgetName",
+          required: true,
+          type: "string",
+        },
+      ],
+      responses: {
+        "200": {
+          description: "The request has succeeded.",
```
markcowl, line 1086: Here I think we want 200 with description "The resource exists."  and 404 with "The resource does not exist."

packages/typespec-azure-resource-manager/test/resource.test.ts
```diff
@@ -1021,4 +1021,77 @@ describe("typespec-azure-resource-manager: ARM resource model", () => {
       });
     });
   });
+
+  it("emits correct paths for ArmResourceHead operation", async () => {
+    const openApi = await openApiFor(`
+      @armProviderNamespace
+      @useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
```
markcowl, line 1028: Add v5 types

markcowl (changes requested):
I also think we should add this to the `arm-scenarios` sample, and to the samplke ARM spec in the playground

packages/typespec-azure-resource-manager/lib/operations.tsp
allenjzhang: I was thinking about the same thing but was concerned with CheckNameExist operations. But hearing second on it, I will update it and bring to ARM for final review.

packages/typespec-azure-resource-manager/lib/operations.tsp
markcowl: Looking through existing specs, especially this: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/resources/resource-manager/Microsoft.Resources/stable/2023-07-01/resources.json#L2995-L3069

I think the two default responses should be NoContent and NotFound (204, 404)

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/operations.tsp
johanste: Returning a 204 would be wrong/violating RFC9110 (unless the corresponding GET request would have returned 204). So I really hope that is not what ARM does...

johanste: Well, following the link @markcowl provided above. And now I'm sad.

packages/typespec-azure-resource-manager/lib/operations.tsp
johanste: I think existence check is one of the use cases for the endpoint. It is not the only one. We have chosen to provide a client library convention that uses it for existence checks. But that is more of a client library concern, no?

packages/typespec-azure-resource-manager/lib/operations.tsp
allenjzhang: Can you elaborate? Are you saying using operation overload (yet to be design/implemented) on ArmResourceRead? Or are you suggesting to support arbitrary HEAD operation?

packages/typespec-azure-resource-manager/lib/operations.tsp
allenjzhang: ARM indeed picked `204` over `200` because of empty body. As this is only limited in ARM lib, Azure.Core or TCGC could still add support as add-on to existing GET as you have suggested. As for impl, while it is easy to slap on a `@EnableHead()` on any GET operation there, it would tricky for  customizations such as ResponseCode and potentially additional headers.

packages/typespec-azure-resource-manager/lib/operations.tsp
johanste: Something like [this](https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CgpAcm91dGVPZihUT3ApCkBUeXBlU3BlYy5IdHRwLmhlYWQgb3AgSGVhZDxUT3AgZXh0ZW5kcyDJJ1JlZmxlY3Rpb24uT3BlcmHECj7ETDo6cGFyYW1ldGVycyk6IMUScmV0dmFsxHXOZ%2BUAgygiL3NvbWV0aGluZy97aWR9IikKb3AgZ2V0U8gXKM84cGF0aCBpZDogc3RyaW5nKcgJOwrEOslORXhpc3RzIGlz5gDIzFI%2BOwoK&e=%40typespec%2Fopenapi3&options=%7B%7D)

packages/typespec-azure-resource-manager/lib/operations.tsp
allenjzhang: This is scoped to ARM which has its owe standards such as name and response codes. Client Lib may provide a template that is more generic and would not conflict with this one similar to other Life cycle operations.

packages/typespec-azure-resource-manager/lib/operations.tsp
allenjzhang: What you suggested is just an implementation detail which is how we implemented in ARM now. GET and HEAD share a base operation template.

markcowl (approved):

PR #287, opened 2024-02-21
Adding @promotion decorator and learnMoreDocs validation

yejee94:
adding a decorator called @promotion which helps us know which apiVersion for RP should be deploy to Portal.

also adding validation rule for @about.learnMoreDocs checking if the links are starting with https://


packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -15,9 +15,9 @@ import {
   resolvePath,
   validateDecoratorUniqueOnNode,
 } from "@typespec/compiler";
+import * as versioning from "@typespec/versioning";
```
timotheeguerin, line 18: try to not do `import *` but import only what you need

yejee94: fixed thank you!!

.chronus/changes/yejeelee-promotions-2024-1-21-1-4-17.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-portal-core"
+---
+
+adding new decorator called @promotion and adding validation rule for @about.LearnMoreDocs
```
timotheeguerin, line 7:
```suggestion
adding new decorator called `@promotion` and adding validation rule for `@about.LearnMoreDocs`
```

packages/typespec-azure-portal-core/src/types.ts
```diff
@@ -17,3 +19,8 @@ export interface BrowseOptions {
 export interface marketplaceOfferOptions {
   id?: string;
 }
+
+export interface PromotionOptions {
+  apiVersion: string | EnumMember;
+  autoUpdate?: boolean;
+}
```
timotheeguerin, line 26:
```suggestion
export interface PromotionOptions {
  readonly apiVersion: string | EnumMember;
  readonly autoUpdate?: boolean;
}
```

timotheeguerin: good practice so that other libraries don't modify your state

.chronus/changes/yejeelee-promotions-2024-1-21-1-4-17.md
yejee94: I think this get auto-generated?

timotheeguerin (approved):

packages/typespec-azure-portal-core/lib/decorators.tsp
```diff
@@ -63,3 +70,9 @@ model AboutOptions {
   /** Set of links which can help learn more about the resource */
   learnMoreDocs?: string[];
 }
+
+/** Options for promotion of ARM resource. */
+model PromotionOptions {
```
allenjzhang, line 75: Since it is a single data entry, it should be singular name.

packages/typespec-azure-portal-core/lib/decorators.tsp
```diff
@@ -15,6 +15,13 @@ extern dec about(target: Model, options: AboutOptions);
  */
 extern dec browse(target: Model, options: BrowseOptions);
 
+/**
+ * Provides a Model customizing deployment promotion apiVersion for ARM resource.
+ * The apiVersion will be used as a version to deploy to Portal.
+ * @param options Property options provides promotion information of the resourceType.
+ */
+extern dec promotion(target: Model, options: PromotionOptions);
```
allenjzhang, line 23: Do you plan to always take a single version or an array?

yejee94: only single version always :)

packages/typespec-azure-portal-core/src/types.ts
```diff
@@ -17,3 +19,8 @@ export interface BrowseOptions {
 export interface marketplaceOfferOptions {
   id?: string;
 }
+
+export interface PromotionOptions {
```
allenjzhang, line 23: Same comments as above. Singular name

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -52,6 +55,72 @@ export function $browse(context: DecoratorContext, target: Model, options: Model
   }
 }
 
+export function $promotion(context: DecoratorContext, target: Model, options: Model) {
+  const { program } = context;
+  validateDecoratorUniqueOnNode(context, target, $promotion);
```
allenjzhang, line 60: nit: Since we are retiring `@service{version}` and will only rely on `@versioned`, it only rely on enum versions, the validation of each value provided can be greatly simplified. Only need to validate the enum specified is same as the one specified in `@versioned` on the namespace. Compiler will ensure the values are correct.

yejee94: talked to Allen offline, it is okay to keep this logic since customers are still using the service logic!

PR #295, opened 2024-02-22
have separate datetime types

### LINKED ISSUE #75: tcgc needs to expose if a time is UTC or not
tsp exposes utcDateTime and offsetDateTime, see [reference](https://microsoft.github.io/typespec/language-basics/built-in-types#date-and-time-types). tcgc doesn't expose the distinction between time types that are in UTC or contain a time-zone offset.

iscai-msft:
fixes #75


timotheeguerin (approved):
So you will just go with the breaking change?

PR #298, opened 2024-02-22
ARM: Add rule to enable disabling `DeleteResponseCodes` LintDiff rule

tjprescott:
Closes https://github.com/Azure/typespec-azure-pr/issues/3836

**Impact** 
- 9 violations in REST API specs (https://github.com/Azure/azure-rest-api-specs/pull/27948)

**azure-openapi-validator**
https://github.com/Azure/azure-openapi-validator/pull/668


packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
```diff
@@ -0,0 +1,44 @@
...
+import { getHttpOperation } from "@typespec/http";
+import { getArmResources } from "../resource.js";
+
+/**
+ * Verify that a delete operation only
+ */
+export const armDeleteResponseCodesRule = createRule({
+  name: "arm-delete-operation-response-codes",
+  severity: "warning",
+  description: "Ensure delete operations have the appropriate status codes.",
```
timotheeguerin, line 13: Should add the markdown doc link too and write the doc (example `no-enum` rule)

packages/typespec-azure-resource-manager/test/rules/arm-delete-response-codes.test.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  BasicTestRunner,
+  LinterRuleTester,
+  createLinterRuleTester,
+} from "@typespec/compiler/testing";
+import { beforeEach, describe, it } from "vitest";
+import { armDeleteResponseCodesRule } from "../../src/rules/arm-delete-response-codes.js";
+import { createAzureResourceManagerTestRunner } from "../test-host.js";
+
+describe("typespec-azure-resource-manager: arm delete response codes rule", () => {
```
timotheeguerin, line 10: we don't need a top level describe anymore with vitest as everything is already grouped per file

.vscode/extensions.json
```diff
@@ -2,7 +2,6 @@
   "recommendations": [
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
-    "streetsidesoftware.code-spell-checker",
-    "ZixuanChen.vitest-explorer"
+    "streetsidesoftware.code-spell-checker"
```
timotheeguerin, line 5: should we recommend the new extension instead?

packages/typespec-azure-resource-manager/test/rules/arm-delete-response-codes.test.ts
tjprescott: I'd prefer to keep it consistent with other files.

timotheeguerin: Changing all the existing test will cause massive diff so it’s not really worth it I think unless refactoring significantly a test file. But here for new files this just adds extra layer that make the cli and test explorer a bit worse

tjprescott: I filed this issue to restore support for mocha https://github.com/Azure/typespec-azure/issues/302.

The describes grouping remains useful when using that framework.

timotheeguerin: mocha is vastly inferior in the capacity it gives us in test, we also were using completely outdated packages that had vulnerability, reverting is most likely not going to happen. apart from the extension issue vitest is better in every way. vitest is a modern alternative to `jest` which was the most popular test framework but we switch right at the time they got to v1 and the extension wasn't really ready but they have been pushing many updates in the last weeks so this feels uncessary to revert just because they have a bug they will most likely fix.

packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const armDeleteResponseCodesRule = createRule({
+  name: "arm-delete-operation-response-codes",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/delete-operation-response-codes",
+  description: "Ensure delete operations have the appropriate status codes.",
+  messages: {
+    sync: `Synchronous delete operations must have 200, 204 and default responses. They must not have any other responses.`,
+    async: `Long-running delete operations must have 202, 204 and default responses. They must not have any other responses.`,
+  },
```
tjprescott, line 18: @markcowl would you prefer that I refactor this as a "response-codes" rule and have separate messages for delete/post/etc or make each it's own rule? I'm fine either way.

docs/libraries/azure-resource-manager/rules/delete-operation-response-codes.md
```diff
@@ -0,0 +1,62 @@
...
+  };
+}
+```
+
+#### ✅ Correct
+
+```tsp
+@armResourceOperations
+interface Employees {
+  @armResourceDelete(Employee)
```
markcowl, line 31: Here we should show usage of the ArmResourceDeleteSync template

docs/libraries/azure-resource-manager/rules/delete-operation-response-codes.md
```diff
@@ -0,0 +1,62 @@
...
+  } | {
+    @statusCode _: 204;
+    result: boolean;
+  } | ErrorResponse;
+}
+```
+
+## Asynchronous
+
+Long-running (LRO) delete operations must have 202, 204, and default responses. They must not have any other responses.
```
markcowl, line 44: In the text, we should recommend the ArmResourceDeleteWithoutOkAsync template

packages/typespec-azure-resource-manager/test/rules/arm-delete-response-codes.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      @path
+      @segment("employees")
+      name: string;
+    }
+    
+    model EmployeeProperties {}
+    
+    @armResourceOperations
+    interface Employees {
+      @armResourceDelete(Employee)
```
markcowl, line 79: We should add a test that validates ArmResourceDeleteSync does not throw the error, as regression protection

markcowl (changes requested):
Looks good.  A coupl eof comments.

Also, we should deprecate `ArmResourceDeleteAsync` unfortunately, this will likely mean adding some deprecation suppressions in samples.

packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
tjprescott: Going to leave this as-is for now since code complete it approaching.

packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Verify that a delete operation only
+ */
+export const armDeleteResponseCodesRule = createRule({
+  name: "arm-delete-operation-response-codes",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/delete-operation-response-codes",
+  description: "Ensure delete operations have the appropriate status codes.",
+  messages: {
+    sync: `Synchronous delete operations must have 200, 204 and default responses. They must not have any other responses.`,
+    async: `Long-running delete operations must have 202, 204 and default responses. They must not have any other responses.`,
```
markcowl, line 17: Here we should recommend the ArmResourceDeleteWithoutOKAsync template

packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
markcowl: I think we can add this in a separate PR

packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Verify that a delete operation only
+ */
+export const armDeleteResponseCodesRule = createRule({
+  name: "arm-delete-operation-response-codes",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/delete-operation-response-codes",
+  description: "Ensure delete operations have the appropriate status codes.",
+  messages: {
+    sync: `Synchronous delete operations must have 200, 204 and default responses. They must not have any other responses.`,
```
markcowl, line 16: Here we should recommend the ArmResourceDeleteSync operation template.

markcowl (changes requested):

tjprescott:
@markcowl updated. Please take a look at the splash radius associated with deprecating ArmResourceDeleteAsync.

docs/libraries/azure-resource-manager/rules/delete-operation-response-codes.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+#### ✅ Correct
+
+```tsp
+@armResourceOperations
+interface Employees {
+  @armResourceDelete(Employee)
+  delete(...ApiVersionParameter): {
+    @statusCode _: 200;
```
markcowl, line 33: Looks like this wasn't changed, this should use ArmResourceDeleteSync instead of a custom operation.

packages/typespec-azure-resource-manager/test/rules/arm-delete-response-codes.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      delete(...ApiVersionParameter): {
+        @statusCode _: 200;
+        result: boolean;
```
markcowl, line 82: We should use ArmResourceDeleteSync for one of these valid sync test instances - we could either do this here, or add another test that uses that template

packages/typespec-azure-resource-manager/test/rules/core-operations.test.ts
```diff
@@ -179,6 +179,7 @@ describe("typespec-azure-resource-manager: core operations rule", () => {
           }
     
           @armResourceOperations
+          #suppress "deprecated" "test"
```
markcowl, line 182: It doesn't look like there are any deprecations in the operation template for ArmResourceDeleteAsync.

markcowl (changes requested):
It looks like some changes might be missing from the PR.  Comments point these out

docs/libraries/azure-resource-manager/rules/delete-operation-response-codes.md
tjprescott: It looks like rebasing threw away my changes. Fun times.

packages/typespec-azure-resource-manager/test/rules/core-operations.test.ts
timotheeguerin: just click the update PR button

packages/typespec-azure-resource-manager/test/rules/arm-delete-response-codes.test.ts
tjprescott: There's another test that uses the template.

packages/samples/specs/resource-manager/tenantResource/main.tsp
```diff
@@ -180,6 +180,7 @@ model FlangeProperties {
   ...DefaultProvisioningStateProperty;
 }
 
+#suppress "deprecated" "Should replace TenantResourceDelete in the future"
```
tjprescott, line 183: @markcowl I was able to update the other samples to simply use `ArmResourceDeleteWithoutOkAsync` but this is an interface template of a template of a template, etc. so I just suppressed the deprecation warning...

packages/typespec-azure-resource-manager/lib/operations.tsp
```diff
@@ -412,6 +412,7 @@ op ArmResourceDeleteAsyncBase<
  * @template Response Optional. The success response(s) for the delete operation
  * @template Error Optional. The error response, if non-standard.
  */
+#deprecated "Use 'ArmResourceDeleteWithoutOkAsync' instead"
```
tjprescott, line 415: You can see the cascading effects of formally deprecating this.

tjprescott:
@markcowl all comments should be resolved. The REST API specs PR has been merged

allenjzhang (approved):

PR #299, opened 2024-02-23
change to singular for data types

yejee94:
changed

all options => option

### DIFF
--- a/packages/typespec-azure-portal-core/lib/decorators.tsp
+++ b/packages/typespec-azure-portal-core/lib/decorators.tsp
@@ -5,28 +5,28 @@ namespace Azure.Portal;
 
 /**
  * Provides a Model describing about of ARM resource.
- * @param options Property options allows more detailed infomation about the resourceType.
+ * @param option Property option allows more detailed infomation about the resourceType.
  */
-extern dec about(target: Model, options: AboutOptions);
+extern dec about(target: Model, option: AboutOption);
 
 /**
  * Provides a Model customizing browse view of ARM resource.
- * @param options Property options provides browsing information of the resourceType.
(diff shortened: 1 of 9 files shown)


timotheeguerin (commented):
This doesn’t feel right to me there is multiple options

yejee94:
@timotheeguerin this was suggested by Allen from last PR! I'm actually changing from options to option though??

allenjzhang (approved):

packages/typespec-azure-portal-core/lib/decorators.tsp
```diff
@@ -71,8 +71,8 @@ model AboutOptions {
   learnMoreDocs?: string[];
 }
 
-/** Options for promotion of ARM resource. */
-model PromotionOptions {
+/** Option for promotion of ARM resource. */
+model PromotionOption {
```
timotheeguerin, line 75: I don't really get why this is a singlular options, there is 2? And the point of an option bag even if there is a single option is you can add more later which would be breaking if you had to change the name.

@allenjzhang can you expand on why you wanted this?

allenjzhang: My original thinking was on `xxxOptions` like `PromotionOptions` as it has a single one apiVersion settings. And plural of options is addressed as an [], so a `plural[]` sounded weird.

Also I had mentally thought of `xxOption` as a single option model for `xx` with bunch of properties (ie `@service`) rather a bunch of discret/independent option property bag grouped under `xx`. **But I do see your point on expandability and the names, particularly they have an `option` suffix. xxOptions sounds ok and reflect the multiple items under it. I am ok to take back my original comment.**

PR #304, opened 2024-02-23
ARM: Add rule to allow disabling LintDiff `AvoidAdditionalProperties`

tjprescott:
Closes https://github.com/Azure/typespec-azure-pr/issues/4083

**Impact**
14 violations in REST API Specs (https://github.com/Azure/azure-rest-api-specs/pull/27926)

**azure-openapi-validator**
https://github.com/Azure/azure-openapi-validator/pull/667


docs/libraries/azure-resource-manager/rules/no-record.md
```diff
@@ -0,0 +1,67 @@
...
+```text title=- Full name-
+@azure-tools/typespec-azure-resource-manager/no-record
+```
+
+ARM requires Resource provider teams to define types explicitly. This is to ensure good customer experience in terms of the discoverability of concrete type definitions.
+
+#### ❌ Incorrect
+
+```tsp
+model Address extends Record<string> {}
```
timotheeguerin, line 14: Nit, would put the `is Record<T>` example first as this is recommended over `extends`

packages/typespec-azure-resource-manager/src/rules/arm-no-record.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const armNoRecordRule = createRule({
+  name: "arm-no-record",
+  severity: "warning",
+  description: "Don't use Record types for ARM resources.",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-record",
+  messages: {
+    default:
+      "Definitions must not be of type Record. ARM requires Resource provider teams to define types explicitly.",
```
timotheeguerin, line 14:
```suggestion
      "Properties must not be of type Record. ARM requires Resource provider teams to define types explicitly.",
```
as this is currently where its used for(though maybe changing the key to be explicit)

packages/typespec-azure-resource-manager/test/rules/arm-no-record.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      namespace Microsoft.Contoso;
+
+      @Azure.ResourceManager.tenantResource
+      model Widget is ProxyResource<WidgetProperties> {
+        @key("widgetName")
+        @segment("widgets")
+        @path
+        @visibility("read")
+        name: string;
+      }
```
timotheeguerin, line 90: Do you actually need all this boilerplate for those tests?

packages/typespec-azure-resource-manager/src/rules/arm-no-record.ts
markcowl: I would make this 'Model Properties and Parameters', as I think this is actually what will be flagged by the rule.

packages/typespec-azure-resource-manager/src/rules/arm-no-record.ts
```diff
@@ -0,0 +1,47 @@
...
+    extends:
+      "Definitions must not extend type Record. ARM requires Resource provider teams to define types explicitly.",
+    is: "Definitions must not equate to type Record. ARM requires Resource provider teams to define types explicitly.",
+  },
+  create(context): SemanticNodeListener {
+    return {
+      model: (model: Model) => {
```
markcowl, line 21: Do we have a mechanism for omitting types in Azure.ResourceManager?  We want to make sure that valid usages of `Record<unknown>`, like 'tags' for example, are exempted.

markcowl (changes requested):
Minor nits, otherwise lgtm

packages/typespec-azure-resource-manager/test/rules/arm-no-record.test.ts
tjprescott: Yes because I updated the logic to examine only ARM resources. I refactored it to be a little cleaner.

packages/typespec-azure-resource-manager/src/rules/arm-no-record.ts
tjprescott: I'm still working on that bit.

packages/typespec-azure-resource-manager/test/rules/arm-no-record.test.ts
timotheeguerin: so if it only check arm resources is there not a gap that `bad-record` would have reported but now this doesn't?

packages/typespec-azure-resource-manager/src/rules/arm-no-record.ts
tjprescott: I added a test for this... and it passes... but I don't understand _why_ it passes....

tjprescott: Is passes because of the design of the linter rule framework. Using `ArmTagsProperty` triggers the rule to emit a diagnostic, but because it originated outside of user code, it omits the diagnostic.

markcowl (approved):

packages/typespec-azure-resource-manager/test/rules/arm-no-record.test.ts
timotheeguerin: I don't think the azure core rule should be aware of arm at all. I think we should reapply the same logic manually?

Wouild it hurt to just complain about any reference of `Record` instead of limiting to arm resources?

tjprescott: Mark's response, per Teams:

> [12:51 PM] Mark Cowlishaw
> Yeah, I don't think here is any particular admonition against Record<unknown> per se, in non-resource types.  As long as we are including the entire resource definition (properties of properties (of properties...)) then I think this is acceptable.  The main outliers will be Action request/response in that case

timotheeguerin (approved):

PR #305, opened 2024-02-23
Add operation status

### LINKED ISSUE #204: [Bug][getAllModels] `OperatoinStatus` not found even `filter-out-core-models : false`
**Describe the bug**
I bumped to typespec `0.52.0` (along with tcgc `0.38.9`).
See `cadl-ranch` LRO test case: https://github.com/Azure/cadl-ranch/blob/c611311c3d36a3048a3077137cbd4e6be762d034/packages/cadl-ranch-specs/http/azure/core/lro/standard/main.tsp#L217

`getLroMetadata` return the `finalEnvelopResult` as `OperatoinStatus` which is a foundation type. But you cannot get it through `getAllModels()` even I set `filter-out-core-models` to false.
image

**To Reproduce**
Try to parse the `cadl-ranch` test case with  `filter-out-core-models: false`.

**Expected behavior**
I understand `OperationStatus` should be found.

iscai-msft:
fixes #204

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -1077,16 +1077,18 @@ function updateTypesFromOperation(
     }
   }
   const lroMetaData = getLroMetadata(program, operation);
-  if (lroMetaData) {
+  if (lroMetaData && generateConvenient) {
     const logicalResult = diagnostics.pipe(
       checkAndGetClientType(context, lroMetaData.logicalResult, operation)
     );
-    if (
-      logicalResult &&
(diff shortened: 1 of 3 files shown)


timotheeguerin (approved):

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1070,17 +1070,19 @@ function updateTypesFromOperation(
     }
   }
   const lroMetaData = getLroMetadata(program, operation);
-  if (lroMetaData) {
+  if (lroMetaData && generateConvenient) {
```
tadelesh, line 1073: Should `generateConvenient` impact only usage or the model detection? In other logics, it only impact usage.

Issue #309, opened 2024-02-26
Support @clientName for union variant name


tadelesh:
related code: https://github.com/Azure/typespec-azure/blob/00f500c4ba7cf3fcae628bfbad651c87695fc0a5/packages/typespec-client-generator-core/src/types.ts#L599

weidongxu-microsoft:
cadl-ranch https://github.com/Azure/cadl-ranch/pull/525

PR #314, opened 2024-02-26
support @clientName for extensible enum variant

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/309

add a fix for operation group path does not adopt `@clientName`

### DIFF
--- a/packages/typespec-client-generator-core/test/types.test.ts
+++ b/packages/typespec-client-generator-core/test/types.test.ts
@@ -12,6 +12,7 @@ import {
   SdkType,
   SdkUnionType,
 } from "../src/interfaces.js";
+import { isErrorOrChildOfError } from "../src/public-utils.js";
 import {
   getAllModels,
   getAllModelsWithDiagnostics,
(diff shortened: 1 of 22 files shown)


packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -930,7 +930,37 @@ describe("typespec-client-generator-core: types", () => {
...
+            string
+          }
+          op x(body: TestUnion): void;
+        }
+      `,
+        `
+        namespace Customizations;
+
+        @@clientName(N.TestUnion, "TestUnionRename");
+        @@clientName(N.TestUnion.B, "BRename");
```
timotheeguerin, line 953: There is one issue right in that you cannot augment `A`?

This doesn't make this change wrong but I think it greatly limit its purpose

tadelesh: Yeah. I suppose it is by design since union variant `A` does not has an explicit name.

iscai-msft (approved):

PR #317, opened 2024-02-26
Fix missing canonical-versioning ruleset in arm lib

timotheeguerin:


packages/typespec-azure-resource-manager/src/linter.ts
```diff
@@ -64,7 +64,10 @@ export const $linter = defineLinter({
         // TODO: Enable this rule once azure-rest-api-specs repo is ready (issue #3839)
         [`@azure-tools/typespec-azure-resource-manager/${armCommonTypesVersionRule.name}`]: false,
       },
-      extends: ["@azure-tools/typespec-azure-core/all"],
+      extends: [
+        "@azure-tools/typespec-azure-core/all",
+        "@azure-tools/typespec-azure-core/canonical-versioning",
```
tjprescott, line 69: If this already extends the `azure-core/all` ruleset, then why does this rule need to be extended here?  Is "all" not really all?

timotheeguerin: yes all doesn't include those, this was a mistake to name it `all` we should have had
- `required` (What is required for azure core)
- `all` (Actually everything)
- `<anyoptional set>`

timotheeguerin: i'll file an issue to consider changing the names

timotheeguerin: https://github.com/Azure/typespec-azure/issues/330

markcowl (approved):

tjprescott (approved):

PR #323, opened 2024-02-27
Add visibility and optionality for provisioning state & linting rules

allenjzhang:
https://github.com/Azure/typespec-azure-pr/issues/2638


packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -169,7 +170,8 @@ enum ResourceProvisioningState {
 @doc("Contains a default provisioningState property to be spread into resource property types")
 model DefaultProvisioningStateProperty {
   @doc("The provisioning state of the resource.")
-  provisioningState: ResourceProvisioningState;
+  @visibility("read")
+  provisioningState?: ResourceProvisioningState;
```
timotheeguerin, line 174: that's going to cause a diff in the swagger and new linting validation right?

allenjzhang: Yes. Though this is an older model that are not showing up in our recent samples, so usage is minimal. I have searched public main and it is not used by any spec. Looking at RPSaaSxxx now.

timotheeguerin (approved):

tjprescott:
Here is the azure-openapi-validator PR that will disable this rule once this is released:
https://github.com/Azure/azure-openapi-validator/pull/666

packages/samples/test/output/azure/resource-manager/liftr.confluent/@azure-tools/typespec-autorest/openapi.json
```diff
@@ -516,12 +516,10 @@
       "properties": {
         "provisioningState": {
           "$ref": "#/definitions/Azure.ResourceManager.ResourceProvisioningState",
-          "description": "The provisioning state of the resource."
+          "description": "The provisioning state of the resource.",
+          "readOnly": true
```
markcowl, line 520: Note: We should make sure these samples are using the suggested read-only-status-schema: true setting.  We can do this in a separate PR

markcowl (approved):

Issue #326, opened 2024-02-28
fix issues of union as enum in TCGC


tadelesh:
1. anonymous union to extensible enum naming ([example](https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgVGVzdCB7DQogIHR5cGU6IHN0cmluZyB8ICJBIsQGQiINCn0%3D&e=%40typespec%2Fopenapi3&options=%7B%7D))
2. union as enum value as a discriminator ([example](https://cadlplayground.z22.web.core.windows.net/?c=dW5pb24gRG9nS2luZCB7DQogIHN0cmluZyzEC0dvbGRlbjogImfFCSIsDQp9DQoNCkBkaXNjcmltaW5hdG9yKCJraW5kIikNCm1vZGVsxE7GSsQXOshdO8QSd2VpZ2h0OiBpbnQzMjvHTsY2xmwgZXh0ZW5kc9dFLsYmxjo%3D&e=%40typespec%2Fopenapi3&options=%7B%7D))

PR #328, opened 2024-02-28
Fix union as enum issues

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/326

### DIFF
--- a/packages/typespec-client-generator-core/test/types.test.ts
+++ b/packages/typespec-client-generator-core/test/types.test.ts
@@ -1427,7 +1427,9 @@ describe("typespec-client-generator-core: types", () => {
     it("union to extensible enum values", async () => {
       await runner.compileWithBuiltInService(`
       union PetKind {
+        @doc("Cat")
         Cat: "cat",
+        @doc("Dog")
         Dog: "dog",
         string,
       }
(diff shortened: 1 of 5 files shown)


packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -1450,17 +1452,40 @@ describe("typespec-client-generator-core: types", () => {
...
       strictEqual(dogValue.value, "dog");
+      strictEqual(dogValue.description, "Dog");
       strictEqual(dogValue.enumType, petKind);
       strictEqual(dogValue.valueType, petKind.valueType);
       strictEqual(dogValue.kind, "enumvalue");
     });
 
+    it("property of anonymous union as enum", async () => {
+      await runner.compileWithBuiltInService(`
+      model Pet {
```
iscai-msft, line 1470: we don't want to create an enum from an anonymous union, we only do it if the union is named. let me roll back this part of the pr, rest looks good as far as i can see

timotheeguerin: is this not true anymore?

iscai-msft: i didn't know we changed this, i know @lmazuel before really didn't want to generate a named enum from anonymous unions. can you approve it as is and @tadelesh depending on what @lmazuel says can you make a separate PR if we do want to name anonymous unions?

timotheeguerin (approved):

Issue #333, opened 2024-02-29
Support anonymous union enum


tadelesh:
[example](https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgVGVzdCB7DQogIHR5cGU6IHN0cmluZyB8ICJBIsQGQiINCn0%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)
need to take care of `TestEnum | null` situation.

PR #335, opened 2024-02-29
Anonymous union enum

tadelesh:
fix: #333

@iscai-msft one problem for nullable, i don't think it should be in sdk types for now, it should be an attribute of model property, what do you think?


iscai-msft:
Yes it shouldn't be a separate type, just mark it as `nullable`

PR #343, opened 2024-02-29
Fix azure scalars

iscai-msft:


packages/typespec-client-generator-core/doc/types.tsp
```diff
@@ -77,7 +77,7 @@ model SdkBuiltInType extends SdkType {
     | "armId"
     | "ipAddress"
     | "azureLocation"
-    | "etag";
+    | "eTag";
```
timotheeguerin, line 80: is that going to be breaking?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -190,33 +185,10 @@ function addEncodeInfo(
...
-    case "boolean":
-    case "string":
-    case "url":
-    case "decimal":
-    case "plainDate":
-    case "plainTime":
-      return scalar.name;
-    default:
-      throw Error(`Unknown scalar kind ${scalar.name}`);
+  if (isSdkBuiltInKind(scalar.name)) {
```
timotheeguerin, line 188: Here this is still going to find wong scalars. It should also be checking the namespace

timotheeguerin: If you want what you could do is for every scalar get the fully qualified name and then you create a map of known fully qualified name

```
const knownScalars = {
  "TypeSpec.int32": "...",
 "Azure.Core.eTag": "...",
}
```

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -114,30 +119,20 @@ function addFormatInfo(
...
-        break;
-      case "ipaddress":
-        propertyType.kind = "ipAddress";
-        break;
-      case "azurelocation":
-        propertyType.kind = "azureLocation";
-        break;
-      default:
-        break;
+  const format = getFormat(context.program, type);
```
timotheeguerin, line 122: Can we keep the map on the left exactly as it is with a `legacyFormat` function. Otherwise here as we keep adding new known scalar `@format` with the same name will work even though it shouldn't

packages/typespec-client-generator-core/doc/types.tsp
iscai-msft: kind of, but no one's using it rn. I'll make a note of it though

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -79,41 +79,87 @@ export interface SdkBuiltInType extends SdkTypeBase {
...
+  for (const kind in SdkAzureBuiltInStringKindsEnum) {
+    if (!isSdkBuiltInKind(kind)) continue; // it will always be true
+    retval[`Azure.Core.${kind}`] = kind;
+  }
+  return retval;
+}
+
+export function isSdkBuiltInKind(kind: string): kind is SdkBuiltInKinds {
+  return (
+    Object.keys(SdkBuiltInKindsMiscellaneousEnum).includes(kind) ||
```
timotheeguerin, line 156: Why not do `kind in SdkBuiltInKindsMiscellaneousEnum` that will be faster than doing `.includes`

iscai-msft: oh ok nice thanks!

PR #359, opened 2024-03-04
ARM: add `arm-resource-name-pattern` rule to allow disabling LintDiff `ResourceNamePattern` rule

tjprescott:
Closes https://github.com/Azure/typespec-azure-pr/issues/3903

**REST API Specs** (6 violations)
https://github.com/Azure/azure-rest-api-specs/pull/28085

**OpenAPI Validator PR**
https://github.com/Azure/azure-openapi-validator/pull/669


packages/typespec-azure-resource-manager/src/rules/arm-delete-response-codes.ts
```diff
@@ -18,25 +17,27 @@ export const armDeleteResponseCodesRule = createRule({
...
-        const resources = getArmResources(context.program);
-        const armResource = resources.find((re) => re.typespecType === model);
-        if (armResource && armResource.operations.lifecycle.delete) {
-          const deleteOperation = armResource.operations.lifecycle.delete;
-          const isAsync = getLroMetadata(context.program, deleteOperation.operation) !== undefined;
-          const [httpOp, _] = getHttpOperation(context.program, deleteOperation.operation);
-          const statusCodes = new Set([...httpOp.responses.map((r) => r.statusCodes.toString())]);
-          const expected = new Set(["204", "*"]);
-          expected.add(isAsync ? "202" : "200");
+      root: (program: Program) => {
```
tjprescott, line 20: This is just a refactoring of this existing rule to be more efficient. It doesn't change the behavior.

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
```diff
@@ -0,0 +1,37 @@
...
+          const nameProperty = [...resource.typespecType.properties.values()].find(
+            (p) => p.name === "name"
+          );
+          if (nameProperty !== undefined) {
+            const pattern = getPattern(program, nameProperty);
+            if (pattern === undefined) {
+              context.reportDiagnostic({
+                target: nameProperty,
+              });
+            }
```
tjprescott, line 31: So, per the LintDiff rule it says it's also fine to apply the default ARM pattern here if one is not provided. We have a couple options:
1) don't do that and just always throw this warning
2) always apply the default if not specified, in which case this linter rule actually becomes a warning diagnostic and the logic would change
3) provide an easy mechanism to provide the default pattern and have the error message tell people to use that (like we did for the error codes rule pointing to other templates).

@markcowl @timotheeguerin @johanste any thoughts on what is preferrable?

johanste: It would be preferable (to me) if there was an implicit default. But I'm not sure how many resources adhere to the same constraints.

.chronus/changes/arm-resourceNamePatternRule-2024-2-4-21-52-56.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
timotheeguerin, line 3: Should be feature

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
timotheeguerin: That rule could also be an easy code fix candidate so user can auto fix

.chronus/changes/arm-resourceNamePatternRule-2024-2-4-21-52-56.md
tjprescott:
```suggestion
changeKind: feature
```

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
```diff
@@ -0,0 +1,37 @@
...
+  messages: {
+    default: `The resource name parameter should be defined with a 'pattern' restriction.`,
+  },
+  create(context) {
+    return {
+      root: (program: Program) => {
+        const resources = getArmResources(program);
+        for (const resource of resources) {
+          // find the name property
+          const nameProperty = [...resource.typespecType.properties.values()].find(
```
timotheeguerin, line 22: why not do `resource.typespecType.properties.get("name")`?

tjprescott: Thanks for the suggestion! Done!

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
tjprescott: @timotheeguerin do we have documentation on how to do that?

timotheeguerin: https://typespec.io/docs/extending-typespec/linters#provide-a-codefix

For linter in particular

timotheeguerin: The auto suppress code fix in the compiler is a simple one that is similar in the logic

timotheeguerin: Also feels like adding a decorator on a node feels like a common code fix we should maybe build a common helper piece in the same way we have the update identfier codefix

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Verify that a delete operation only
+ */
+export const armResourceNamePatternRule = createRule({
+  name: "arm-resource-name-pattern",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/resource-name-pattern",
+  description: "The resource name parameter should be defined with a 'pattern' restriction.",
+  messages: {
+    default: `The resource name parameter should be defined with a 'pattern' restriction.`,
```
markcowl, line 14:
```suggestion
    default: `The resource name parameter should be defined with a 'pattern' restriction.  Decorate the "name" property in the resource definition using the @pattern decorator, with a regular expression indicating the allowed characters in the resource name.`,
```

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
markcowl: Agreed, a codefix with the default value is the best way to document this.  In general, I think we should make a codefix an acceptance criterion for new linting rules, where that is possible.

markcowl: But, please, file an issue for adding the codefix for this, it is important we get this rule in.

markcowl (changes requested):

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
```diff
@@ -0,0 +1,75 @@
...
+  DiagnosticTarget,
+  Program,
+  SourceLocation,
+  createRule,
+  defineCodeFix,
+  getPattern,
+  getSourceLocation,
+} from "@typespec/compiler";
+
+import { isWhiteSpace } from "../../../../core/packages/compiler/src/core/charcode.js";
```
timotheeguerin, line 11: hhm this is going to fail in the packaged package

timotheeguerin: I would  just do `=== " "`  think it will work just as well. Or don't reindent and let the formatter do it when they trigger it.

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
```diff
@@ -0,0 +1,75 @@
...
+// Issue: https://github.com/microsoft/typespec/issues/3044
+function createPatternCodeFix(diagnosticTarget: DiagnosticTarget) {
+  return defineCodeFix({
+    id: "add-pattern-decorator",
+    label: "Add `@pattern` decorator to the resource name property with the default ARM pattern.",
+    fix: (context) => {
+      const location = getSourceLocation(diagnosticTarget);
+      const { lineStart, indent } = findLineStartAndIndent(location);
+      const updatedLocation = { ...location, pos: lineStart };
+      return context.prependText(updatedLocation, `${indent}@pattern(/^[a-zA-Z0-9-]+$/)\n`);
```
tjprescott, line 24: @markcowl is this actually the default ARM pattern? If not, what is?

packages/typespec-azure-resource-manager/src/rules/arm-resource-name-pattern.ts
```diff
@@ -0,0 +1,74 @@
...
+  messages: {
+    default: `The resource name parameter should be defined with a 'pattern' restriction.  Decorate the "name" property in the resource definition using the @pattern decorator, with a regular expression indicating the allowed characters in the resource name.`,
+  },
+  create(context) {
+    return {
+      root: (program: Program) => {
+        const resources = getArmResources(program);
+        for (const resource of resources) {
+          // find the name property
+          const nameProperty = resource.typespecType.properties.get("name");
```
markcowl, line 60: This is likely to change soon, but is accurate for now

markcowl (approved):

Issue #353, opened 2024-03-04
[Bug] Open union of float value type becomes int value type


pshao25:
**Describe the bug**
```
union FloatExtensibleEnum {
    float32,
    One: 1.0,
    Two: 2.0,
    Four: 4.0,
}
```
I defined an extensible enum with type `float32`, and several values `1.0`. 

**To Reproduce**
Call https://github.com/Azure/typespec-azure/blob/53cc4dee48de3e8b774643d5946bed1f768f8609/packages/typespec-client-generator-core/src/types.ts#L800

**Expected behavior**
Value type `float32`

**Actual behavior**
Value type `int32`

**Additional context**
See this line: https://github.com/Azure/typespec-azure/blob/53cc4dee48de3e8b774643d5946bed1f768f8609/packages/typespec-client-generator-core/src/types.ts#L658

The value type comes from actual value type which becomes `1` after it reads into memory, which is originally `1.0`. Since we now have the ability to identify the real type of extensible enum as open union, we should respect the type defined in the spec, i.e. `float32`.

lirenhe:
cc @tadelesh and @iscai-msft

iscai-msft:
i'll look into this.
@timotheeguerin should tsp know that the type of `1.0` is a float and not an int?

timotheeguerin:
typespec doesn't today see this isssue https://github.com/microsoft/typespec/issues/2240 you have to figure out yourself if you care

PR #362, opened 2024-03-05
Enhance logic for value type of enum and union as enum for TCGC

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/353

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -22,6 +22,7 @@ import {
   Tuple,
   Type,
   Union,
+  UnionVariant,
   UsageFlags,
   createDiagnosticCollector,
   getDiscriminator,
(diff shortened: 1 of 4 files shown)


iscai-msft (approved):

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -563,18 +565,51 @@ export function getSdkModelWithDiagnostics(
 
 function getSdkEnumValueType(
   context: TCGCContext,
-  type: EnumMember | StringLiteral | NumericLiteral
+  values:
+    | IterableIterator<EnumMember>
+    | IterableIterator<UnionEnumVariant<string>>
+    | IterableIterator<UnionEnumVariant<number>>
 ): SdkBuiltInType {
   let kind: "string" | "int32" | "float32" = "string";
```
weidongxu-microsoft, line 573: what about int64 and float64?

tadelesh: This is only for enum value. Since we could not get exact type for enum value, we just use int32 or float32 to be the enum value type. And I believe it is the most widely used case.

weidongxu-microsoft: Got it. emitter can enlarge the type if found it to be "fixed enum".

PR #364, opened 2024-03-05
Fix is form data

tadelesh:
rollback https://github.com/Azure/typespec-azure/issues/215
when we use the spread way to define the body parameter, each time we call getHttpOperation, we will get a new model type for body.

example: https://github.com/bterlson/typespec-todo/blob/ca8a3aab5f3ef52e52cf2b48d2980f808165fada/main.tsp#L293


iscai-msft:
@tadelesh this is an http issue right? Shouldn't we wait for the http library to fix it?

timotheeguerin:
> @tadelesh this is an http issue right? Shouldn't we wait for the http library to fix it?

there is no issue in the http library, this is by design today we don't deduplicate types. You should not be calling it twice for the same operation anyway https://github.com/microsoft/typespec/issues/2961

tadelesh:
> @tadelesh this is an http issue right? Shouldn't we wait for the http library to fix it?

as timothee said, it is by design, so i prefer to remove it temporally. i'm doing a refactor for `getAllOperations` and do one call for `getHttpOperation`. after all emitter switch to it, many issues (alias, spread, etc.) could be resolved.

iscai-msft:
@tadelesh can this be closed bc I'm assuming your caching `getHttpOperation` fixes this issue as well?

https://github.com/Azure/typespec-azure/pull/395

PR #376, opened 2024-03-06
Moving RetryAfter header to LROHeaders to enabled easier override if needed.

allenjzhang:
During migrations, there are many expressed concerns on the swagger diff from our standard/best practices operation templates and wanted to stay the same without the header. To do that, we will have to jump thru hoops and may involve multiple level of redefinition. Having these moved to the default LroHeaders on these Async operation template would allow easier override.

Please note this is a "breaking change" because of Response changes. However polling the rest-api-spec repo, there are about 5 instance of usage which all indicated not best usages as they bypassed the ARM template and used Ops directly. It's best to fix those to standard patterns anyway.

https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20ArmAcceptedLroResponse&type=code


markcowl (approved):
I feel like we should add in some tests here, but this change looks good

allenjzhang:
> I feel like we should add in some tests here, but this change looks good

I feel unit tests will be very limited, but our sample regen have sufficient scenarios to cover this.

allenjzhang:
@tadelesh , FYI.

PR #386, opened 2024-03-07
block @browse.argQuery usage for ProxyResource and adding more validations

yejee94:
* @browse.argQuery should only be used in TrackedResource
* @browse.argQuery should not take file name
* @browse.argQuery.filePath should only allow .kql or .kml files
* @about.icon.filePath should only allow .svg

### DIFF
--- a/.chronus/changes/yejeelee-proxyFix-2024-2-7-15-0-25.md
+++ b/.chronus/changes/yejeelee-proxyFix-2024-2-7-15-0-25.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@azure-tools/typespec-azure-portal-core"
+---
+
+put more validation and remove browse.argQuery from proxyResource usage
--- a/packages/typespec-azure-portal-core/src/decorators.ts
+++ b/packages/typespec-azure-portal-core/src/decorators.ts
@@ -158,18 +158,29 @@ export function checkIsArmResource(
   target: Model,
   decoratorName: "browse" | "about" | "marketplaceOffer" | "promotion"
 ) {
-  if (
-    getArmResourceKind(target) !== ("Tracked" as ArmResourceKind) &&
-    getArmResourceKind(target) !== ("Proxy" as ArmResourceKind)
-  ) {
-    reportDiagnostic(program, {
-      code: "not-a-resource",
-      format: {
-        decoratorName: decoratorName,
-      },
-      target,
-    });
-    return false;
+  if (decoratorName === "browse") {
+    if (getArmResourceKind(target) !== ("Tracked" as ArmResourceKind)) {
+      reportDiagnostic(program, {
+        code: "not-a-resource",
+        messageId: "browse",
+        target,
+      });
+      return false;
+    }
+  } else {
+    if (
+      getArmResourceKind(target) !== ("Tracked" as ArmResourceKind) &&
+      getArmResourceKind(target) !== ("Proxy" as ArmResourceKind)
+    ) {
+      reportDiagnostic(program, {
+        code: "not-a-resource",
+        format: {
+          decoratorName: decoratorName,
+        },
+        target,
+      });
+      return false;
+    }
   }
   return true;
 }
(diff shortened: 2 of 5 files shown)


packages/typespec-azure-portal-core/src/lib.ts
```diff
@@ -9,9 +9,18 @@ export const $lib = createTypeSpecLibrary({
         default: paramMessage`cannot find @${"decoratorName"} file ${"propertyName"} from path ${"filePath"}`,
       },
     },
+    "invalid-type": {
+      severity: "error",
+      messages: {
+        iconSvg: paramMessage`@about.icon.filePath only allows svg file, current file: ${"filePath"}`,
```
Olgaian, line 15: don't we need an icon?

packages/typespec-azure-portal-core/test/decorators.test.ts
```diff
@@ -150,6 +192,32 @@ describe("TypeSpec-Azure-Portal-Core decorators test", () => {
...
 
+  it("@about with icon with wrong file extension", async () => {
+    const aboutTest = `
+      @test @about({
+        icon: {
+          filePath: "./icon.txt"
+        }
+      })`;
+    const diagnostics = await runner.diagnose(createTestSpec(undefined, aboutTest));
+    strictEqual(diagnostics.length, 1);
```
timotheeguerin, line 203: use expect diagnostic api, doing that manually is very hard to diagnose after

timotheeguerin: the message can be a regex if you don't want an exact match

.chronus/changes/yejeelee-proxyFix-2024-2-7-15-0-25.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: internal
```
timotheeguerin, line 2:
```suggestion
changeKind: feature
```

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -153,10 +153,26 @@ export function getBrowseArgQuery(program: Program, target: Type) {
   return getBrowse(program, target)?.argQuery;
 }
 
+export function checkIsArmTrackedResource(
+  program: Program,
+  target: Model,
+  decoratorName: "browse"
+) {
+  if (getArmResourceKind(target) !== ("Tracked" as ArmResourceKind)) {
```
timotheeguerin, line 161: Why do you need to do this `as ArmResourceKind`

packages/typespec-azure-portal-core/test/decorators.test.ts
```diff
@@ -150,6 +188,34 @@ describe("TypeSpec-Azure-Portal-Core decorators test", () => {
...
+    const aboutTest = `
+      @test @about({
+        icon: {
+          filePath: "./icon.txt"
+        }
+      })`;
+    const diagnostics = await runner.diagnose(createTestSpec(undefined, aboutTest));
+    expectDiagnostics(diagnostics, {
+      code: "@azure-tools/typespec-azure-portal-core/invalid-type",
+      message: new RegExp("^@about.icon.filePath only allows svg file, current file:"),
```
timotheeguerin, line 201:
```suggestion
      message: /^@about.icon.filePath only allows svg file, current file:/,
```
you can do that for regex, makes it much more obvious and will be better for dealing with escapes

packages/typespec-azure-portal-core/src/decorators.ts
```diff
@@ -34,7 +34,7 @@ import {
 export function $browse(context: DecoratorContext, target: Model, options: Model) {
   const { program } = context;
   validateDecoratorUniqueOnNode(context, target, $browse);
-  checkIsArmResource(program, target, "browse");
+  checkIsArmTrackedResource(program, target, "browse");
```
timotheeguerin, line 37: Do you need to return if it is invalid here?

timotheeguerin (approved):

Issue #385, opened 2024-03-07
[tcgc] roll back change we made that allowed any union of values to be an extensible enum


iscai-msft:
In the last release, we decided to always generate a union of constants as an enum. However, after talking more with @lmazuel and seeing the confusion here, we've decided to only generate an enum from a union if the union is named

tadelesh:
The root cause is not the naming, It is from hierarchy. My fix will rollback the change for hierarchy.

Issue #379, opened 2024-03-07
[Bug][tcgc] getSdkUnion would merge underlying enum members to top level


MaryGao:
**Describe the bug**
When we adopt to the latest tcgc our [two cases failed](https://github.com/Azure/autorest.typescript/blob/d78e698f87ffb97d16688852908292383aa3519f/packages/typespec-ts/test/modularUnit/type.spec.ts#L119-L174) and we found that 
- Sometimes the getSdkUnion could return correct isFixed value like Case 2 but sometimes it can't like Case 1
- Also for union of EnumA | EnumB tcgc would flatten all underlying enums into top level 

TCGC are supposed to provide consistant isFixed information, otherwise it would be a big issue for JS because RLC didn't adopt tcgc but Modular does. So it would introduce type inconsistancy between them and finally fails the compile process.

Also I think current behavior for prompting members to top level is debetable. It would lose the name info which may be useful for emitters.

### Case 1: Nullable enum
```
enum Color {
  Color1: 1,
  Color2: 2
}
model Test {
  color: Color | null;
}
op read(@body body: Test): void;
``` 

![image](https://github.com/Azure/typespec-azure/assets/9943211/4099ea27-591d-4781-86ba-234c3d83cefe)

### Case 2: EnumeA | EnumB

```
 enum LR {
  left,
  right,
}
enum UD {
  up,
  down,
}

model Test {
  color: LR | UD;
}
op read(@body body: Test): void;
```

![image](https://github.com/Azure/typespec-azure/assets/9943211/489ecfd2-488d-4c22-a015-d5d0adebaef2)

tadelesh:
TCGC get this info from two places, first is `isFixed` ([code](https://github.com/Azure/typespec-azure/blob/a15638510b0f8059b69b78e05cdc08a3ec0ef7a2/packages/typespec-azure-core/src/decorators.ts#L58)) from `@fixed` decorator, second is the `open` property in `UnionEnum` type ([code](https://github.com/Azure/typespec-azure/blob/a15638510b0f8059b69b78e05cdc08a3ec0ef7a2/packages/typespec-azure-core/src/helpers/union-enums.ts#L23)). I think current problem should mainly from losing information for hierarchy union.

lirenhe:
@MaryGao, in JS, do we have a workaround for this issue?

MaryGao:
> isFixed
I don't think we have a quick workaround here, currently we just fallback to tcgc dev.17 version.

[...]

PR #389, opened 2024-03-08
Add union and enum doc and refine logic

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/379
resolve: https://github.com/Azure/typespec-azure/issues/385


packages/typespec-client-generator-core/src/types.ts
timotheeguerin, line 1: So this mean you don't flatten those cases(union composed of other unions or enum?) how are language that cannot represent those reusable union going to treat it? Are they going to have to reflatten it? Wouldn't it be better to provide both view to the emitter?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -353,9 +353,17 @@ export function getSdkUnionWithDiagnostics(
     return diagnostics.wrap(clientType);
   }
 
-  const unionAsEnum = diagnostics.pipe(getUnionAsEnum(type));
-  if (unionAsEnum) {
-    return diagnostics.wrap(getSdkUnionEnum(context, unionAsEnum, operation));
+  // judge if the union can be converted to enum
```
iscai-msft, line 356: I would say we don't need to judge if it's specifically a union of enum, just that if it's named

joheredi (approved):

tadelesh:
@iscai-msft I saw you rollbacked my change for union as enum with hierarchy. I'm wondering do we have the agreement of erasing the hierarchy types and flattened into one layer of enum? It will break JS RLC, since the type has been changed. For the following example, I don't think we could omit `BaseEnum` and just generate `ExtendedEnum` as enum. I prefer to generate `ExtendedEnum` as union and `BaseEnum` as enum. Also, do we have the agreement of not generating anonymous union as enum? I don't like the inconsistency. 
```
union BaseEnum {
  string,
  EnumValue1: "value1",
}

union ExtendedEnum {
  BaseEnum,
  EnumValue2: "value2",
}
```

iscai-msft:
@tadelesh this was after talking with @lmazuel , but I'm going to start a discussion in dpg sync about this

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,505 @@ import TabItem from "@theme/TabItem";
...
+<TabItem value="" label="Python">
+
+Python generates inline unions as a union of literals, not as enums. We also don't generate a closed set of literals
+
+```python
+from typing import Literal
+
+type WidgetHorizontal = Literal["left"] | Literal["right"] | str
+
+model Widget:
```
lmazuel, line 85: Not Python :)

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,505 @@ import TabItem from "@theme/TabItem";
...
+
+Python generates a union of literals again. Since we always generate our typing as open, this will
+look the same as our previous example where we generate an open union of strings.
+
+```python
+from typing import Literal
+
+type WidgetColor = Literal["red"] | Literal["blue"] | str
+
+model Widget:
```
lmazuel, line 190: Same, need Pythonization

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,505 @@ import TabItem from "@theme/TabItem";
...
+union Orientation {
+  LR,
+  UD,
+}
+```
+
+</TabItem>
+<TabItem value="" label="Python">
+
+Since
```
lmazuel, line 302: @iscai-msft guess you wanted to write something here?

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,537 @@ import TabItem from "@theme/TabItem";
...
+</TabItem>
+<TabItem value="java" label="Java" >
+
+```java
+public enum Orientation {
+   LEFT("left"),
+   RIGHT("right"),
+   UP("up"),
+   DOWN("down");
+}
```
weidongxu-microsoft, line 354: Behavior of union of union depends on `getUnionAsEnum`.

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,537 @@ import TabItem from "@theme/TabItem";
...
+</TabItem>
+<TabItem value="java" label="Java" >
+
+```java
+public final class ProvisioningState extends ExpandableStringEnum<ProvisioningState> {
+    public static final ProvisioningState IN_PROGRESS = fromString("inProgress");
+    public static final ProvisioningState SUCCEEDED = fromString("Succeeded");
+    public static final ProvisioningState FAILED = fromString("Failed");
+    public static final ProvisioningState CANCELED = fromString("Canceled");
+}
```
weidongxu-microsoft, line 286: Behavior of union of union depends on `getUnionAsEnum` (I assume `ResourceProvisioningState` would be eventually a `union`).

However, in this case, if we do not flatten this, we still need a way to say `ProvisioningState` is something `inProgress` + those in `Azure.ResourceManager.ResourceProvisioningState`

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,537 @@ import TabItem from "@theme/TabItem";
...
+```
+
+</TabItem>
+<TabItem value="java" label="Java" >
+
+```java
+public enum Horizontal {
+   LEFT("left"),
+   RIGHT("right");
+}
```
weidongxu-microsoft, line 113: Currently, Java does not distinguish the type on whether it is named or not.

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -5,9 +5,537 @@ import TabItem from "@theme/TabItem";
...
+
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+union LR {
+  left: "left",
+  right: "right",
+}
+```
```
pshao25, line 22: I would suggest something like
```
union LR {
  left: "leftWire",
  right: "rightWire",
}
``

File: docs/howtos/DataPlane Generation - DPG/06types.mdx [hunk 9]
```diff
@@ -5,9 +5,537 @@ import TabItem from "@theme/TabItem";
...
+```typespec
+union LR {
+  left: "left"
+  right: "right"
+}
+
+union UD {
+  up: "up"
+  down: "down"
+}
```
pshao25, line 308:
```suggestion
union LR {
  left: "left",
  right: "right"
}

union UD {
  up: "up",
  down: "down"
}
```

PR #391, opened 2024-03-08
[tcgc] Use own usage flags

iscai-msft:
I've verified that this code still works, so language emitters currently relying on `typespec/compiler`'s `UsageFlags` won't be broken

```typescript
import { UsageFlags } from "@typespec/compiler";
import { getAllModels } from "@azure-tools/typespec-client-generator-core";


const models = getAllModels(sdkContext).filter(x => x.usage === UsageFlags.Input);
```

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -505,6 +505,9 @@ export type SdkHttpPackage = SdkPackage<SdkHttpOperation>;
 
 export type LanguageScopes = "dotnet" | "java" | "python" | "javascript" | "go" | string;
 
+/**
+ * This enum represents the different ways a model can be used in a method.
+ */
 export enum UsageFlags {
   None = 0,
   Input = 1 << 1,
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -9,7 +9,6 @@ import {
   Namespace,
   Operation,
   Type,
-  UsageFlags,
 } from "@typespec/compiler";
 import {
   HttpAuth,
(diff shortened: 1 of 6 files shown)


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -505,3 +504,9 @@ export interface SdkPackage<TServiceOperation extends SdkServiceOperation> {
 export type SdkHttpPackage = SdkPackage<SdkHttpOperation>;
 
 export type LanguageScopes = "dotnet" | "java" | "python" | "javascript" | "go" | string;
+
+export enum UsageFlags {
```
m-nash, line 508: Can we name this ModelUsageFlags so it doesn't get confused with other kinds of usage?

m-nash (approved):

timotheeguerin (approved):

packages/typespec-client-generator-core/src/interfaces.ts
timotheeguerin: or `ClientUsage`

iscai-msft: after verification, this will break emitter authors who are currently filtering using `@typespec/compiler`'s `UsageFlags`. Going to improve docs instead

timotheeguerin: it doesn't break if it is only used with bitwise operations https://www.typescriptlang.org/play?#code/KYOwrgtgBAYg9nKBvAUFdUCWIAOYAuUAvFAIxQA8FZANGhnAXoSeVVAEx0C+KKokKADlgAd3iJUGLLgLEylaqTrTG+ZvLbUuKXigBmYEAGN8mOCCgBzYPgkAKAJQAuEeITJ66AE62w3yzcJADpsZgBuXT5ogHoYmSgAG0wAIz5jCwBnQn0PEhs7BCdIlEx9e1zEADJYBFDZfEdPXSA

if however you used the type and pass it around it will fail

PR #392, opened 2024-03-08
Fix @lroStatus not detecting correctly when using union

timotheeguerin:


markcowl (approved):

packages/typespec-azure-core/src/decorators.ts
```diff
@@ -350,7 +350,7 @@ export function extractLroStates(
 
         return diagnostics.wrap(undefined);
       } else {
-        storeLroState(program, result, option.value, variant);
+        storeLroState(program, result, variant.name.toString(), variant);
```
markcowl, line 353: Do we have to account for both enum and union types?

timotheeguerin: yes that was the problem this broke the anonymous variant case. So in case of named variant change the variant name, in the case of unnamed check the value

Issue #334, opened 2024-02-29
Give a good name for templated model without `@friendlyName`


tadelesh:
original issue: https://github.com/Azure/typespec-azure-pr/issues/3418
1. let's get the consensus giving a non-duplicated name for templated model if there is no `@friendlyName` is a must and shall apply to all languages
2. add test case in cadl-ranch
3. get agreement for naming logic and implement in tcgc `getLibraryName`, `getAllModels` uses `getLibraryName`, so it is transitive

cc: @iscai-msft

tadelesh:
@pshao25 could you help to explain what .NET did and link the related test here? thanks

weidongxu-microsoft:
Java emitter would be fine, if TCGC goes with `Template<Model1, Model2>` to `TemplateModel1Model2`. (it should be fine as long as it is not just `Template` which could cause lots of duplicated names.

pshao25:
.Net has same requirement to Java.

BTW, is `getLibraryName` expected to give the name of anonymous model?

tadelesh:
> .Net has same requirement to Java.
> 
> BTW, is `getLibraryName` expected to give the name of anonymous model?

Nope. Anonymous model name is in the `generatedName` property of tcgc model type.

pshao25:
I think we could have a temporary solution: if we cannot find any other overridden name, give me a flag instead of just returning its original name.

PR #393, opened 2024-03-11
give a nonredundant name for templated instance model

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/334

### DIFF
--- a/packages/typespec-client-generator-core/test/public-utils.test.ts
+++ b/packages/typespec-client-generator-core/test/public-utils.test.ts
@@ -777,6 +777,28 @@ describe("typespec-client-generator-core: public-utils", () => {
       await helper("@azure-tools/typespec-ts");
       await helper("@azure-tools/typespec-python");
     });
+    it("template without @friendlyName renaming", async () => {
+      await runner.compileWithBuiltInService(`
+      op GetResourceOperationStatus<
+        Resource extends TypeSpec.Reflection.Model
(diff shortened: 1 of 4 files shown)


packages/typespec-client-generator-core/test/public-utils.test.ts
```diff
@@ -1404,13 +1426,13 @@ describe("typespec-client-generator-core: public-utils", () => {
...
           [
-            "ResourceOperationStatus",
+            "ResourceOperationStatusUserExportedUserError",
             "OperationState",
             "Error",
             "InnerError",
             "ExportedUser",
             "ErrorResponse",
-            "OperationStatus",
+            "OperationStatusExportedUserError",
```
iscai-msft, line 1435: i believe these are both core models, correct?
nit: I don't think we should modify the names of core models based off of templating. Not the biggest issue though, bc ideally no one should be generating with core models. If it makes the code much harder, no need to change your existing code

tadelesh: AFAIK, `OperationStatus` is exported in Java and .NET. I don't think this will only happen for core model. This will happen when implicitly using templated model without `@friendlyName` in templated operation. And it is a possible usage for customer. Since .NET and Java need this nonredundant naming insurance, I prefer to add it in TCGC.

iscai-msft (approved):

PR #399, opened 2024-03-11
Fix operation trait not applied correctly in Foundations.Operation

timotheeguerin:
fix https://github.com/microsoft/typespec/issues/2938

Problem seems to be that we tried to applied the request trait using the ApiVersion location so it wouldn't include trait that are meant to be included as request parameters


packages/typespec-azure-core/lib/foundations.tsp
```diff
@@ -285,10 +285,8 @@ op Operation<
   Traits extends TypeSpec.Reflection.Model = {},
   ErrorResponse = Azure.Core.Foundations.ErrorResponse
 >(
-  ...TraitProperties<
-    Traits & VersionParameterTrait<ApiVersionParameter>,
-    TraitLocation.ApiVersionParameter
-  >,
+  ...TraitProperties<VersionParameterTrait<ApiVersionParameter>, TraitLocation.ApiVersionParameter>,
+  ...TraitProperties<Traits, TraitLocation.Parameters>,
```
timotheeguerin, line 289: Ok so this was done on purpose because the addition of the trait parameters is done earlier in most cases.

PR #400, opened 2024-03-11
Fixing correct name for managed identity and customer managed key, updated enum to union

### LINKED ISSUE #278: [Bug] `ManagedServiceIdentityType` enum name is different between TSP and Swagger
This enum is a ModelProperty type in common model `ManagedServiceIdentity`: https://github.com/Azure/typespec-azure/blob/ebfe63960277356c611f15b2404a0ae6f2d9e6ed/packages/typespec-azure-resource-manager/lib/arm.foundations.tsp#L387

In swagger common types, it's named [ManagedServiceIdentityType](https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/common-types/resource-management/v4/managedidentity.json#L35-L48).
Though in tsp, it's named [ManagedIdentityType](https://github.com/Azure/typespec-azure/blob/ebfe63960277356c611f15b2404a0ae6f2d9e6ed/packages/typespec-azure-resource-manager/lib/arm.foundations.tsp#L439).

allenjzhang:
closes #278


packages/typespec-azure-resource-manager/lib/arm.foundations.tsp
```diff
@@ -417,9 +417,11 @@ model ManagedSystemIdentityProperties {
   principalId?: string;
 
   @doc("The type of managed identity assigned to this resource.")
-  type: ManagedSystemIdentityType;
+  type: SystemAssignedServiceIdentityType;
 }
 
+/** Alias of ManagedServiceIdentityType for back compatability. Please change to ManagedServiceIdentityType. */
+alias ManagedIdentityType = ManagedServiceIdentityType;
```
timotheeguerin, line 424:
```suggestion
#deprecated "Please change to ManagedServiceIdentityType." 
alias ManagedIdentityType = ManagedServiceIdentityType;
```

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -72,10 +72,14 @@ export function $resourceParameterBaseFor(
 ) {
   const resolvedValues: string[] = [];
   for (const value of values.values) {
-    if (value.kind !== "EnumMember") {
+    // must be enum or string Union. TODO: Once Union extend string support has been added, the condition below can be further constrained.
+    if (value.kind === "EnumMember") {
+      resolvedValues.push(value.name);
+    } else if (value.kind === "UnionVariant" && value.type.kind === "String") {
+      resolvedValues.push(<string>value.name);
```
timotheeguerin, line 79:
```suggestion
      resolvedValues.push(String(value.name));
```

packages/typespec-azure-resource-manager/lib/arm.foundations.tsp
```diff
@@ -308,20 +308,20 @@ model ArmTagsProperty {
 /**
  * The kind of entity that created the resource.
  */
-// NOTE: This is how the enum is named in types.json
+// NOTE: This is how the extensible enum is named in types.json
 @doc("The kind of entity that created the resource.")
-enum createdByType {
+union createdByType {
```
timotheeguerin, line 313: Those should have `string` as a variant

timotheeguerin: one issue with doing this right now is that if someone was spreading those in their enum it will fail vs the other way around works fine.

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -72,10 +72,14 @@ export function $resourceParameterBaseFor(
 ) {
   const resolvedValues: string[] = [];
   for (const value of values.values) {
-    if (value.kind !== "EnumMember") {
+    // must be enum or string Union. TODO: Once Union extend string support has been added, the condition below can be further constrained.
```
timotheeguerin, line 75: I think can revert that change too

packages/typespec-azure-resource-manager/lib/arm.foundations.tsp
```diff
@@ -99,26 +99,32 @@ enum ResourceHome {
...
-Enum. Indicates the action type. "Internal" refers to actions that are for internal only APIs.
+Extensible enum. Indicates the action type. "Internal" refers to actions that are for internal only APIs.
 """)
-enum ActionType {
+union ActionType {
   @doc("Actions are for internal-only APIs.")
-  Internal,
+  Internal: "Internal",
+
+  @doc("modelAsString=true")
```
markcowl, line 126: Not sure we need this

markcowl (commented):
Are we still thinking of merging this one, or are the conversion to union parts taken up by the ProvisioingState one?

allenjzhang:
> Are we still thinking of merging this one, or are the conversion to union parts taken up by the ProvisioingState one?

They are separate. I will get this one in first and then ProvisioningState.

packages/typespec-azure-resource-manager/lib/arm.foundations.tsp
timotheeguerin: I don't know, this is a doc talking about some autorest property which we should try to detatch arm from autorest emitter so this feels kinda useless, people should get used to extensible unions anyway

timotheeguerin: This could show up in the docs depending on who's emitting it, typescript for example might show a doc here

.chronus/changes/azhang_fixManagedIdentity-2024-2-11-13-47-14.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Updated ManagedIndentity and CustomerManagedKey TypeSpec model definition to be consistent with Swagger commont-types
```
timotheeguerin, line 7:
```suggestion
Updated `ManagedIndentity` and `CustomerManagedKey` TypeSpec model definition to be consistent with Swagger commont-types
```

timotheeguerin (approved):

PR #410, opened 2024-03-12
fixing missing Retry-After in DeleteLroResponse

allenjzhang:


timotheeguerin (approved):

packages/typespec-azure-resource-manager/lib/responses.tsp
```diff
@@ -65,8 +74,8 @@ model ArmAcceptedResponse<Message extends valueof string = "Resource operation a
 @doc(Description)
 model ArmAcceptedLroResponse<
   Description extends valueof string = "Resource operation accepted.",
-  LroHeaders extends TypeSpec.Reflection.Model = ArmLroLocationHeader &
-    Azure.Core.Foundations.RetryAfterHeader
+  LroHeaders extends TypeSpec.Reflection.Model = Azure.Core.Foundations.RetryAfterHeader &
```
markcowl, line 77: I don't think we want to change the order

markcowl (changes requested):

markcowl (approved):

Issue #233, opened 2024-02-07
`@clientName` shouldn't carry over 


timotheeguerin:
Similar to discussion we just had for `@friendlyName`, `@clietnName` shouldn't carry over `model is`

[Playground](https://azure.github.io/typespec-azure/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstY2xpZW50LWdlbmVyYXRvci1jb3JlIjsKCnVzaW5nIFR5cGVTcGVjLkh0dHA70BVSZXN00RVWyX3IG0HEeS5DxWpHyGkuQ29yZTsKCgpAc2VydmljZQpuYW1lc3BhY2UgQ29udG9zby5XaWRnZXRNYW5hZ2VyOwoKQOYArk5hbWUoIkN1c3RvbUZvbyIpCkByb3V0ZSgiZsQOIG9wIGZvbygpOiB2b2lkxWTHIGJhciIpIMQhYmFyIGlzxCg7&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)

tadelesh:
From TypeSpec doc: https://typespec.io/docs/language-basics/models#is, `is` will carry over the decorators, Does it mean it is done by compile?

timotheeguerin:
Yeah by design decorator get carried over by the compiler and we don't have a way to opt out of that behavior but there is other way to make it behave the same.

If you check the target of the decorator node and compare to the decorator node parent and make sure they match then it should only run in the case where client name was defined there

tadelesh:
> Yeah by design decorator get carried over by the compiler and we don't have a way to opt out of that behavior but there is other way to make it behave the same.
> 
> If you check the target of the decorator node and compare to the decorator node parent and make sure they match then it should only run in the case where client name was defined there

Will compiler provide an API to distinguish if a decorator is a direct decorator? Have `@friendlyName` in core resolve such problem? I suppose it should be done in the decorator set function.

timotheeguerin:
yes `@friendlyName` has the same problem but we can't change it as it will be breaking, `@clientName` has the opportunity before its too late. 

And no there will not be a compiler api to do the workaround. There will be an api at some point to make decorator not carry over. The workaround shouldn't be too hard to run.

tadelesh:
[...]

PR #412, opened 2024-03-13
prevent carry over for `@clientName`

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/233

### DIFF
--- a/.chronus/changes/prevent-carry-over-2024-2-13-14-34-30.md
+++ b/.chronus/changes/prevent-carry-over-2024-2-13-14-34-30.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+prevent carry over for @clientName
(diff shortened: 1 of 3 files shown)


.chronus/changes/prevent-carry-over-2024-2-13-14-34-30.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+prevent carry over for @clientName
```
timotheeguerin, line 7:
```suggestion
prevent carry over for `@clientName`
```
make sure you wrap decorator as this is markdown and in github it will tag the user of that name otherwise

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -818,6 +821,25 @@ export function $clientName(
   value: string,
   scope?: LanguageScopes
 ) {
+  if (entity.kind === "Model" || entity.kind === "Operation") {
```
timotheeguerin, line 824: Can you add a reference to this issue https://github.com/microsoft/typespec/issues/2717

that this is a workaround for the lack of functionality in the compiler.

timotheeguerin (approved):
Awesome thanks @[tadelesh](https://github.com/tadelesh) just left some small comments should be good to merge after

Issue #403, opened 2024-03-12
Named union should support `@usage` and `@access` in TCGC


tadelesh:
Since we will use union to represent enum, this should be a must.

PR #413, opened 2024-03-13
Add `@access` and `@usage` support for named union

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/403
Also add deprecation doc.

### DIFF
--- a/packages/typespec-client-generator-core/lib/decorators.tsp
+++ b/packages/typespec-client-generator-core/lib/decorators.tsp
@@ -92,6 +92,8 @@ extern dec client(target: Namespace | Interface, value?: {}, scope?: valueof str
 extern dec operationGroup(target: Namespace | Interface, scope?: valueof string);
 
 /**
+ * DEPRECATED: Use `@usage` and `@access` decorator instead.
+ *
  * Whether to exclude a model from generation for specific languages. By default we generate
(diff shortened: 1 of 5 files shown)


packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -92,6 +92,8 @@ extern dec client(target: Namespace | Interface, value?: {}, scope?: valueof str
 extern dec operationGroup(target: Namespace | Interface, scope?: valueof string);
 
 /**
+ * DEPRECATED: Use `@usage` and `@access` decorator instead.
```
timotheeguerin, line 95: Do you want to mark them explicitly as deprecated instead of just a comment that most people don't read?

iscai-msft: @timotheeguerin they've already been previously marked as deprecated. They've been deprecated for a while, is it fair for us to remove support for it now?

timotheeguerin: we have still never removed any deprecated element from typespec but for tcgc probably ok if you do as it shouldn't be used in RPSaaSmaster yet and otherwise it would have already failed

iscai-msft (approved):

PR #414, opened 2024-03-13
updating autorest to emit response headers in lexicographic order

allenjzhang:


packages/typespec-autorest/src/openapi.ts
```diff
@@ -848,7 +848,7 @@ function createOAPIEmitter(
     for (const data of response.responses) {
       if (data.headers && Object.keys(data.headers).length > 0) {
         openapiResponse.headers ??= {};
-        for (const [key, value] of Object.entries(data.headers)) {
+        for (const [key, value] of Object.entries(data.headers).sort()) {
```
timotheeguerin, line 851: So this was actually done explicitly in the schema sorter to keep the order, I don't really remmber why https://github.com/Azure/typespec-azure/blob/dbe7d3afebba18d1ae9f49b2861a7873e033389e/packages/typespec-autorest/schema/autorest-openapi-schema.tsp#L306-L307


But if we wnat to fix this its the schema we should update not sort here.

timotheeguerin (changes requested):

packages/typespec-autorest/test/response-headers-order.test.ts
timotheeguerin, line 1: I would move this test to `sorting.test.ts`

packages/typespec-autorest/test/sorting.test.ts
```diff
@@ -72,4 +79,34 @@ describe("typespec-autorest: OpenAPI output should be determinstic", () => {
...
+      ${headerDefinitions}
+      model Headers { ...A_Header, ...B_Header, ...C_Header };
+
+      op read(): {@statusCode _: 200, content: string, headers: Headers};
+      `
+    );
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers).length, 3);
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers)[0], "a");
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers)[1], "b");
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers)[2], "c");
```
timotheeguerin, line 95:
```suggestion
    deepStrictEqual(Object.keys(res.paths["/"].get.responses["200"].headers), [
      "a",
      "b",
      "c"
    ]);
```

packages/typespec-autorest/test/sorting.test.ts
```diff
@@ -72,4 +79,34 @@ describe("typespec-autorest: OpenAPI output should be determinstic", () => {
...
+      ${headerDefinitions}
+      model Headers { ...C_Header, ...A_Header, ...B_Header };
+
+      op read(): {@statusCode _: 200, content: string, headers: Headers};
+      `
+    );
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers).length, 3);
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers)[0], "a");
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers)[1], "b");
+    strictEqual(Object.keys(res.paths["/"].get.responses["200"].headers)[2], "c");
```
timotheeguerin, line 110:
```suggestion
    deepStrictEqual(Object.keys(res.paths["/"].get.responses["200"].headers), [
      "a",
      "b",
      "c"
    ]);
```

timotheeguerin (approved):
Just nit to make the test cleaner and more in line with the other sorting test. Wondering if we should not wait till the last miniute to merge that one so we don';t need to keep regenerating specs in teh spec repo

PR #420, opened 2024-03-13
Update references to projected names

timotheeguerin:


packages/typespec-azure-core/lib/foundations.tsp
```diff
@@ -168,7 +168,6 @@ model InnerError {
   code?: string;
 
   @doc("Inner error.")
-  @projectedName("csharp", "InnerErrorObject")
```
markcowl, line 171: This feels like a breaking change for c-sharp, can we add this decoration in tcgc?

markcowl (approved):

PR #425, opened 2024-03-14
[tcgc] make artifical content type parameter optional

iscai-msft:
Since it wasn't defined in the tsp, we can't assume that it's required on the service. Should default to optional. If it's required, it should be explicitly defined in the tsp

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -175,13 +175,15 @@ function createContentTypeOrAcceptHeader(
     };
   }
   // No need for clientDefaultValue because it's a constant, it only has one value
+  // Since these are the created accept and content type header parameters, we don't make them required.
+  // If they are defined as required, we won't enter this code and they will be generated as required
   return {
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -175,13 +175,15 @@ function createContentTypeOrAcceptHeader(
     };
   }
   // No need for clientDefaultValue because it's a constant, it only has one value
+  // Since these are the created accept and content type header parameters, we don't make them required.
```
jhendrixMSFT, line 178: What's the motivation for this change? Since these are synthesized, constant, params, they won't show up in public surface area so there's no way for callers to set them. Maybe I'm missing something.

tadelesh:
@iscai-msft is it a still valid requirement? i think the content type should be required if there is a body.

iscai-msft:
hmm yes I agree. let me close this pr and issue. @jhendrixMSFT if you don't agree, please reopen. thanks!

PR #430, opened 2024-03-14
Updated .NET to 8 in PR pipeline

abatishchev:


timotheeguerin (approved):

timotheeguerin (approved):

eng/pipelines/pr.yml
```diff
@@ -14,32 +14,42 @@ jobs:
...
-
-        "Windows - Node 20.x":
+        "Windows - Node 20.x + .NET 6":
           pool: azsdk-pool-mms-win-2022-general
           imageName: windows-2022
-          nodeVersion: "20.x"
+          nodeVersion: 20.x
           dotnetVersion: 6.0.x
 
+        # Node 20 + .NET 8
```
timotheeguerin, line 41: Yeah I think here we don't need either to make a new matrix of test, we can just upgrade to dotnet 8 to build the VS extension on linux.

eng/pipelines/pr.yml
```diff
@@ -14,31 +14,29 @@ jobs:
 
     strategy:
       matrix:
-        # Node 18
-        "Linux - Node 18.x - Dotnet 6":
+        # Node 18 + .NET 8
+        "Linux - Node 18.x + .NET 8":
```
timotheeguerin, line 18: I would maybe just omit the fact that we have a version of `.net` in the matrix name as it is not super relevant.

timotheeguerin: this was just probably not cleaned since the split

abatishchev: Agreed, removed

timotheeguerin (approved):

PR #434, opened 2024-03-18
Fix wrong client cache for package clients calculation

tadelesh:


iscai-msft (approved):

jhendrixMSFT:
This has introduced a regression when generating https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/client/structure/default

PR #445, opened 2024-03-19
[tcgc] `.name` -> `.nameInClient` on `SdkModelPropertyType`s

iscai-msft:


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -288,7 +288,11 @@ export interface SdkCredentialType extends SdkTypeBase {
 export interface SdkModelPropertyTypeBase {
   __raw?: ModelProperty;
   type: SdkType;
+  /**
```
m-nash, line 291: Can we create a follow up to remove the `nameInClient` method and link the issue here so we don't forget.

Given tcgc is all internal we can be more aggressive with breaking changes if we all agree.

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -2656,14 +2654,14 @@ describe("typespec-client-generator-core: types", () => {
       const models = runner.context.experimental_sdkPackage.models;
       strictEqual(models.length, 1);
       const scopes = (models[0] as SdkModelType).properties.find(
-        (x) => x.nameInClient === "scopes"
```
m-nash, line 2659: if we are going to deprecate in a transition we should be asserting both are the same in all these cases until / unless we actually remove it completely.

Otherwise we aren't validating that the old property still works.

timotheeguerin (approved):

PR #455, opened 2024-03-20
[tcgc] add UsageFlags.JsonMergePatch

iscai-msft:


tadelesh:
I'd like to hold this PR before we could meet all requirements from Java and .NET from this [issue](https://github.com/Azure/typespec-azure/issues/460).

iscai-msft:
@tadelesh confirmed with the rest of the dpg crew that we want a `JsonMergePatch` usage. A good amount of languages generate extra info for the model in that case. We will still mark it as `Input`, so they're is no rule that both can't be set

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1264,14 +1264,18 @@ function updateTypesFromOperation(
...
-      checkAndGetClientType(context, httpOperation.parameters.body.type, operation)
-    );
+  const httpBody = httpOperation.parameters.body;
+  if (httpBody) {
+    const bodies = diagnostics.pipe(checkAndGetClientType(context, httpBody.type, operation));
     if (generateConvenient) {
       bodies.forEach((body) => {
         updateUsageOfModel(context, UsageFlags.Input, body);
       });
+      if (httpBody.contentTypes.includes("application/json-merge-patch")) {
```
srnagar, line 1274: The content type should be `application/merge-patch+json` - https://datatracker.ietf.org/doc/html/rfc7386

```suggestion
      if (httpBody.contentTypes.includes("application/merge-patch+json")) {
```

srnagar (approved):

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -529,4 +529,5 @@ export enum UsageFlags {
   Input = 1 << 1,
   Output = 1 << 2,
   Versioning = 1 << 3,
```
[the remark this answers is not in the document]
iscai-msft: yep, that's a separate pr, let me send!

packages/typespec-client-generator-core/src/types.ts
iscai-msft: oh oops, thank you so much for hte catch!

joheredi (approved):

PR #459, opened 2024-03-21
Enums are always fixed after we switch to use union to represent extensible enum

tadelesh:
Change TCGC behavior.


timotheeguerin (changes requested):
Before we merge that I think we need to also have everything else ready to merge(Not need to be part of your PR), 
- linter rule already enabled
- autorest also changing behavior

timotheeguerin (approved):
It was decided today during the meeting that we'll get that in for this sprint
Made the autorest change https://github.com/Azure/typespec-azure/pull/473

PR #467, opened 2024-03-21
Azure Core: Enable `no-enum` rule by default in `all` ruleset

timotheeguerin:


.chronus/changes/enable-no-enum-rule-2024-2-21-16-42-36.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
timotheeguerin, line 3:
```suggestion
changeKind: feature
```

markcowl (approved):

packages/typespec-azure-core/test/rules/no-enum.test.ts
```diff
@@ -46,6 +47,27 @@ describe("typespec-azure-core: no-enum rule", () => {
       .toBeValid();
   });
 
+  it("emit warning about other enums in versioned service", async () => {
```
MaryGao, line 50: I am wondering why we have this warning in azure scope?

Looking at the swagger in mgmt, there are a lot of cases where `"modelAsString": false`.
https://github.com/Azure/azure-rest-api-specs/blob/feature/health-lro/specification/compute/resource-manager/Microsoft.Compute/CloudserviceRP/preview/2020-10-01-preview/cloudService.json#L1483-L1492

timotheeguerin: Being an azure scope is the entire point of this linting rule. Actual enums cannot deal with new values being added so azure policy is you shouldn’t use them. There is of course always exceptions.

MaryGao: I see this guideline: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#enums--sdks-client-libraries, is it from there?

PR #500, opened 2024-03-22
Migrate enum to union in samples

timotheeguerin:


packages/samples/specs/data-plane/language/authoring/authoring.tsp
```diff
@@ -80,15 +80,17 @@ model Project {
 }
 
 @doc("Represents the project kind.")
-enum ProjectKind {
+union ProjectKind {
+  string,
```
allenjzhang, line 84: nit, @doc?

allenjzhang (approved):

packages/samples/specs/data-plane/language/authoring/authoring.tsp
timotheeguerin: Not sure there is much value of describing this one. I don't think it is required by the required-documentation linting rule

PR #502, opened 2024-03-22
[tcgc] add UsageFlags.MultipartFormData and deprecate `.isFormDataType`

iscai-msft:


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -302,3 +303,20 @@ export function isNullable(type: Type | SdkServiceOperation): boolean {
...
 }
+
+function isOperationBodyType(context: TCGCContext, type: Type, operation?: Operation): boolean {
+  if (type.kind !== "Model") return false;
+  if (!isHttpOperation(context, operation)) return false;
+  const httpBody = operation
+    ? getHttpOperationWithCache(context, operation).parameters.body
+    : undefined;
+  return (
+    !!httpBody &&
```
joheredi, line 314: nit: Boolean(httpBody) can be used here, it has the same effect but reads a bit nicer

joheredi (approved):

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -302,3 +303,20 @@ export function isNullable(type: Type | SdkServiceOperation): boolean {
...
+    httpBody.type.kind === "Model" &&
+    getEffectivePayloadType(context, httpBody.type) === getEffectivePayloadType(context, type)
+  );
+}
+
+export function isFormDataType(context: TCGCContext, type: Type, operation?: Operation): boolean {
```
srnagar, line 320: Should we call this method `isMultipartFormDataType`?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -529,4 +532,5 @@ export enum UsageFlags {
   Input = 1 << 1,
   Output = 1 << 2,
   Versioning = 1 << 3,
+  MultipartFormData = 1 << 5,
```
srnagar, line 535: We should document that when this is set, `Input` is also set?

srnagar (approved):

PR #520, opened 2024-03-26
Migrate issue templates to forms

timotheeguerin:
Convert to form so its not just a template that can be easily ignored. Actively recommend repro steps with the playground.
image


.github/ISSUE_TEMPLATE/bug-report.md
timotheeguerin, line 1: I only updated bug-report and feature-request but we also have 
- general
- onboarding request
- breaking change 

do we still need them?

the breaking change one in particular I think leads poeple to think this is meant for them to file and I don't think we actually use it direcly either

markcowl: We need a way for us to file bugs (which is what general is for).  The breaking change template is useful for the information we need in breaking change reviews.

markcowl (approved):

PR #521, opened 2024-03-26
Updating ProvisioningState from enum to union

allenjzhang:
1. Updated ARM lib default `ResourceProvisioningState` to union
2. Updated samples `enum ProvisioningState` to union
3. Updated @lroStatus to support nested union ProvisioningState


packages/samples/specs/resource-manager/arm-scenarios/nobody-action/main.tsp
```diff
@@ -55,20 +55,23 @@ model EmployeeProperties {
...
 
   /** The resource is being deleted */
-  Deleting,
+  Deleting: "Deleting",
 
   /** The resource create request has been accepted */
-  Accepted,
+  Accepted: "Accepted",
+
+  /** modelAsString=true */
```
timotheeguerin, line 73: not sure we should show that in the doc

XiaofeiCao:
FYI, getstarted docs may need update as well:
https://github.com/Azure/typespec-azure/blob/8b8d7c05f168d9305a09691c4fedcb88f4a57652/docs/getstarted/azure-resource-manager/step05.md?plain=1#L42-L48

packages/samples/specs/resource-manager/arm-scenarios/nobody-action/main.tsp
markcowl: I don't think this is needed

markcowl (approved):

Issue #418, opened 2024-03-13
nextLink property should not be marked `readOnly: true`


mikekistler:
A recent change in TypeSpec Azure adds `readOnly: true` to the `nextLink` property in a pageable list response. 

`readOnly: true` is meaningless on a property of a response-only schema, and pageable list responses are always response-only -- they are never part of a request body.

We should not be including meaningless attributes in the OpenAPI that is generated from TypeSpec.

markcowl:
@mario-guerra  There is a current discussion about this with differing opinions, can you coordinate coming to consensus?

mikekistler:
Capturing a workaround shared in email:
> There is a one-line change to a spec that will update the visibility of these types across the whole spec:
> ```
> @@visibility(Azure.Core.Page.nextLink)​
> ```
[...]

PR #549, opened 2024-03-29
Remove readOnly from nextLink (#418)

markcowl:
Fix for #418


package.json
timotheeguerin, line 1: are those changes on purpose, I think it will complain that it doesn't match the version in core repo?

pnpm-lock.yaml
timotheeguerin, line 1: probably revert this file too

package.json
markcowl: this was just from running pnpm update.  Should I update core at the same time?

timotheeguerin: I would just revert here, don't think it brings much value to update right now

timotheeguerin (approved):

PR #559, opened 2024-04-01
[tcgc] take lro finalResult into consideration

### LINKED ISSUE #124: [Bug] LROs that don't return a model should not set `sdkMethod.response.type`
For reference: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/lro/standard/main.tsp#L167

This is a `DELETE` which does not return a model. We can see that `sdkMethod.operation.responses` contains an `OperationStatus` response type which makes sense at the protocol level. However, we _also_ see that `sdkMethod.response.type` is an `OperationStatus` model type which I don't think makes sense at the method level.

iscai-msft:
fixes #124

### DIFF
--- a/.chronus/changes/versioning_test-2024-3-22-18-14-21.md
+++ b/.chronus/changes/versioning_test-2024-3-22-18-14-21.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+enhance versioning and add tests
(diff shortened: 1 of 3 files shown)


jhendrixMSFT (changes requested):
This appears to have introduced a regression. For https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/lro/rpc-legacy/main.tsp#L148 the operation no longer has a return type.

tadelesh:
> This appears to have introduced a regression. For https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/lro/rpc-legacy/main.tsp#L148 the operation no longer has a return type.

I do think it is a problem from `getLroMetadata`. @markcowl , could you help to identify why for this cadl-ranch test, we get `void` for `finalResult`?

jhendrixMSFT (approved):

PR #563, opened 2024-04-01
[tcgc] filter out resource manager models as well

iscai-msft:


jhendrixMSFT (approved):

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -205,7 +205,9 @@ export function isAzureCoreModel(t: Type): boolean {
   return (
     t.kind === "Model" &&
     t.namespace !== undefined &&
-    ["Azure.Core", "Azure.Core.Foundations"].includes(getNamespaceFullName(t.namespace))
+    ["Azure.Core", "Azure.Core.Foundations", "Azure.ResourceManager"].includes(
```
tadelesh, line 208: Why we filter out the arm model? Some basic models like `TrackedResource` and `ProxyResource` is in this namespace. They are needed for emitter to generate.

PR #565, opened 2024-04-01
[tcgc] body spread formdata with multiple properties

iscai-msft:


packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2987,6 +2987,60 @@ describe("typespec-client-generator-core: package", () => {
       deepStrictEqual(bodyParam.type, shelfModel);
       deepStrictEqual(bodyParam.correspondingMethodParams, createShelfRequest.type.properties);
     });
+    it("formdata model without body decorator in spread model", async () => {
```
iscai-msft, line 2990: just look at this method

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2987,6 +2987,60 @@ describe("typespec-client-generator-core: package", () => {
...
+        document: bytes;
+      
+        /** Glossary / translation memory will be used during translation in the form. */
+        glossary?: bytes[];
+      }
+      
+      #suppress "@azure-tools/typespec-azure-core/byos"
+      @post
+      @route("document:translate")
+      op documentTranslate is RpcOperation<
```
timotheeguerin, line 3016: can you reduce this to `op foo(...DocumentTranslatecontent): void` and then don't even need the azure core runner

iscai-msft: it didn't repro without the azure core runner

timotheeguerin (approved):

timotheeguerin (approved):

PR #572, opened 2024-04-02
Refine tcgc logic for spread

tadelesh:
1. Set spread model with none usage
2. Workaround for arm provider method parameter


iscai-msft (approved):

packages/typespec-client-generator-core/package.json
```diff
@@ -75,6 +79,7 @@
     "@typespec/rest": "workspace:~",
     "@typespec/tspd": "workspace:~",
     "@typespec/versioning": "workspace:~",
+    "@typespec/openapi": "workspace:~",
```
timotheeguerin, line 82: remove this too, you shouldn't need to look at openapi

packages/typespec-client-generator-core/package.json
```diff
@@ -57,13 +57,17 @@
...
+    "@azure-tools/typespec-autorest": "workspace:~",
+    "@azure-tools/typespec-azure-core": "workspace:~",
     "@typespec/compiler": "workspace:~",
     "@typespec/http": "workspace:~",
     "@typespec/rest": "workspace:~",
     "@typespec/versioning": "workspace:~"
   },
   "devDependencies": {
     "@azure-tools/typespec-azure-core": "workspace:~",
+    "@azure-tools/typespec-azure-resource-manager": "workspace:~",
```
timotheeguerin, line 69: Arm library might also be a problem right now because it depends on autorest

iscai-msft, line 69: we can't test with arm or autorest bc of circular dependencies. cc @timotheeguerin, going to remove for this release

tadelesh: how should we do test with arm? we have such flag in tcgc.

timotheeguerin: Today we can’t because of that cicurlar dependencies issue. If this is a priority we need to speed up removing autorest from arm library

PR #582, opened 2024-04-02
[tcgc] Method Input Signatures Doc

iscai-msft:


docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,509 @@
...
+</TabItem>
+</Tabs>
+
+### Anonymous Model
+
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+op upload(firstName: string, lastName: string): void;
```
tadelesh, line 180: same with `op upload({firstName: string, lastName: string}): void;`
but not same with ` op upload(@path id: string, firstName: string, lastName: string): void;`

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,509 @@
...
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+alias User = {
+  firstName: string,
+  lastName: string,
+};
+
+op upload(...User): void;
```
tadelesh, line 232: for me, it is same as anonymous model spread. i wonder can we arrange the cases with more essential way. i mean:
1. `op update(param1, param2): void` raw parameter without decorator
2. `op update(@path path1, param1): void` raw parameters with decorator
3. `op update(...ModelWithoutDecoratedProperties): void` spread model without decorated properties
4. `op update(...ModelWithDecoratedProperties): void` spread model with decorated properties
5. `op update(...Alias/AnonymouModelWith/WithoutDecoratedProperties): void` spread anonymous model with/without decorated properties

iscai-msft: i'll switch the raw parameters to alias bc you are right, they are the same

tadelesh:
how should we deal with model mixed with http decorator? for example:
```
model Test {
  @path
  prop1: string;
  @path
  prop2: string;
  prop3: string;
}
op test(...Test): void;
```
if we don't want to spread `Test` model, then we need to keep `prop1` and `prop2` in the `Test` model and do the mapping when handling request. then we could separate method and protocol clean.
if we want to put path parameter as method parameter, and generate model without path property, then we need to use protocol information when exporting method signature which make our initial design broken.

weidongxu-microsoft:
Java would like to add a comment "use with caution". @srnagar 

Condition
1. the anonymous model should only contain very small number of properties (4 maybe?)
2. the anonymous model should likely stay same across versions (each optional property added in api-version would result in a new overload of the API in Java SDK)
3. the anonymous model cannot be used in json-merge-patch
4. the model (that now be spread to op) cannot be polymorphic (<-- this one probably be cross-language)

For 4, I mean this
```ts
@discriminator("kind")
model ModelBase {
  kind: string;
}

model Model extends ModelBase {
  kind: "abc";
  prop1: string;
}

@post
op post(...ModelBase, prop2: string): OkResponse;
```

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,456 @@
...
+
+## Basic
+
+### HTTP Get
+
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+op @get get(): User;
```
weidongxu-microsoft, line 16: All HTTP method decorator such as `@get` should be before `op`.

weidongxu-microsoft, line 16: All HTTP method decorator such as `@get` should be before `op`.

```suggestion
@get op get(): User;
```

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,456 @@
...
+</TabItem>
+</Tabs>
+
+### HTTP Post
+
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+op @post post(User): void;
```
weidongxu-microsoft, line 61:
```suggestion
@post op post(@body user: User): void;
```

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,456 @@
...
+```python
+class User:
+  first_name: str
+  last_name: str
+
+class UserRequest:
+  user: User
+
+def upload(userRequest: UserRequest) -> None:
+  ...
```
weidongxu-microsoft, line 308: Not sure if Python code is correct.

If `@body user: User`, the request body is the `User` model.

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,456 @@
...
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+model BlobProperties {
+  @key
+  @visibility("read")
+  @path name: string;
+
+  @header contentLength: string;
+  @header contentType: string;
```
weidongxu-microsoft, line 350: tsp error

```
contentType parameter must be a string literal or union of string literalsTypeSpec(@typespec/http/content-type-string)
```

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
msyyc: @iscai-msft The code seems not correct and here is the generated code with latest typespec-emitter:
image

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,456 @@
...
+}
+
+op setBlobProperties(...BlobProperties): void;
```
weidongxu-microsoft, line 353: route need to define the `name` as it is `@path`.

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
weidongxu-microsoft: I would recommend not using multiple-content-type as example here.

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
chunyu3: with/without @body decorator should have different behavior.  Does this part want to show the scenaro without @body decorator which mean anonymous body?

weidongxu-microsoft: I just mean `post(User)` is not valid in typespec... I just guess Isabella's intended code, not sure.

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,513 @@
...
+
+### Model with Decorated Properties
+
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+model BlobProperties {
+  @key
+  @visibility("read")
```
chunyu3, line 388: Can `@visibility("read")` decorate a path parameter together with `@key` decorator? 
```suggestion
```

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -0,0 +1,567 @@
...
+</TabItem>
+
+<TabItem value="python" label="Python">
+
+```python
+class User:
+  first_name: str
+  last_name: str
+
+def upload(user: User) -> None:
```
msyyc, line 280: if no explicit `@body user: User`, the generated python code will be
```
    def upload(  # pylint: disable=inconsistent-return-statements
        self, body: Union[_models.User, JSON, IO[bytes]], **kwargs: Any
    ) -> None:
```

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
msyyc: It is permitted: 
image

msyyc: nit: it seems `@path` is not necessary if it is already decorated with `@key`

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
tadelesh: this case aims to show the basic usage, so weidong is right, i changed the doc.

qiaozha:
js pending #712

lmazuel (approved):
All languages confirmed they have a design, by pushing code to it. Approving.

PR #619, opened 2024-04-05
Add `no-generic-numeric` rule to disable LintDiff `IntegerTypeMustHaveFormat`

### LINKED ISSUE #669: {Linting]: Discourage use of imprecise numeric types in Azure.Core and Azure.ResourceManager
At least the following types:
- float
- integer
- decimal
- numeric

tjprescott:
Closes #669.

**REST API Specs**
https://github.com/Azure/azure-rest-api-specs/pull/28731


packages/typespec-azure-core/src/rules/use-standard-integer.ts
```diff
@@ -0,0 +1,35 @@
...
+  name: "use-standard-integer",
+  description: "Use only types that map to int32 or int64.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-core/rules/use-standard-integer",
+  messages: {
+    default: "Recommended integer types are 'int32', 'int64' or 'safeint'.",
+  },
+  create(context) {
+    return {
+      model: (model: Model) => {
```
timotheeguerin, line 13: Probably should check for `scalar` extends as well

packages/typespec-azure-core/src/rules/use-standard-integer.ts
```diff
@@ -0,0 +1,35 @@
...
+        const nonRecommendedInts = [
```
timotheeguerin, line 14: Nit: would move out of the rule and declare as a `Set`

docs/libraries/azure-core/rules/use-standard-integer.md
```diff
@@ -0,0 +1,25 @@
...
+
+Azure services favor the standard types int32 and int64 (or safeint) when generating code using Autorest.
+
+#### ❌ Incorrect
+
+```tsp
+model Widget {
+  id: int8;
+}
+```
```
markcowl, line 17: I think we should have an example that uses int32 and an example that uses safeInt

packages/typespec-autorest/test/formats.test.ts
```diff
@@ -71,4 +71,48 @@ describe("typespec-autorest: format", () => {
...
+        },
+        intB: {
+          format: "int64",
+          type: "integer",
+        },
+        intC: {
+          format: "int64",
+          type: "integer",
+        },
+        intD: {
```
markcowl, line 105: This emits an integer?

packages/typespec-azure-core/README.md
```diff
@@ -63,6 +63,7 @@ Available ruleSets:
...
 | `@azure-tools/typespec-azure-core/rpc-operation-request-body`                                                                                                | Warning for RPC body problems.                                                                                                                       |
 | `@azure-tools/typespec-azure-core/spread-discriminated-model`                                                                                                | Check a model with a discriminator has not been used in composition.                                                                                 |
+| `@azure-tools/typespec-azure-core/use-standard-integer`                                                                                                      | Use only types that map to int32 or int64.                                                                                                           |
```
markcowl, line 66: This should name the specific types `safeInt` and `int32` and `int64`

packages/typespec-azure-core/src/rules/use-standard-integer.ts
```diff
@@ -0,0 +1,35 @@
+import { Model, createRule } from "@typespec/compiler";
+
+export const useStandardInteger = createRule({
+  name: "use-standard-integer",
+  description: "Use only types that map to int32 or int64.",
```
markcowl, line 5: use `safeInt` or `int32` or `int64`

packages/typespec-azure-core/test/rules/use-standard-integer.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        prop1: int32;
+        prop2: int64;
+        prop3: safeint;
+      }
+      `
+    )
+    .toBeValid();
+});
+
+it("does not emit a warning diagnostic for non-standard integer types that map to supported integers in Autorest", async () => {
```
markcowl, line 60: Do we want this test?  I think both numeric and integer should produce a separate warning about imprecise numeric types

markcowl (changes requested):

docs/libraries/azure-core/reference/linter.md
```diff
@@ -59,5 +59,6 @@ Available ruleSets:
 | `@azure-tools/typespec-azure-core/response-schema-problem`                                                           | Warn about operations having multiple non-error response schemas.                                                                                    |
 | `@azure-tools/typespec-azure-core/rpc-operation-request-body`                                                        | Warning for RPC body problems.                                                                                                                       |
 | `@azure-tools/typespec-azure-core/spread-discriminated-model`                                                        | Check a model with a discriminator has not been used in composition.                                                                                 |
+| `@azure-tools/typespec-azure-core/use-standard-integer`                                                              | Use only types that map to int32 or int64.                                                                                                           |
```
johanste, line 62: I think this is actively harmful. It will push service teams to be less precise than they can/should be (e.g. the service exposes an 8 bit unsigned integer for an embedding vector - should we now say that it is an int32? And do that at the expense of DPG emitters that actually *do* support unsigned byte sized types?). It is especially bad since reviewers have no reliable way to understand what the intent of the service spec author was. And, nitpick, we also tell spec authors to not use int64 because ECMA script - that would be `safeint` in TypeSpec.

If we insist on this rule, we need to better understand what tools it protects. E.g. if we are not generating code from the generated openapi/swagger, then it doesn't matter what auto rest does. If we *do* have to protect downstream tools, if is probably better that the autorest-openapi emitter emits the formats.

packages/typespec-autorest/test/formats.test.ts
tjprescott: It does. https://github.com/Azure/typespec-azure/blob/main/packages/typespec-autorest/src/openapi.ts#L2149

packages/typespec-azure-core/test/rules/use-standard-integer.test.ts
tjprescott: I think so, but I could just move these into the above test. I'm not testing for the non-precise integer diagnostic (there's already a test for that) but just to make sure that the "use-standard-int" is not thrown for these since they make to "int64" in autorest.

packages/typespec-autorest/test/formats.test.ts
```diff
@@ -71,4 +71,48 @@ describe("typespec-autorest: format", () => {
     );
     expectDiagnosticEmpty(diagnostics);
   });
+
+  it("ensures certain scalars emit int32 and int64 formats", async () => {
```
timotheeguerin, line 75: we already have tests for that in `primitive-types.test.ts`

packages/typespec-azure-core/README.md
timotheeguerin: actually should we recommend `int64`? wouldn't we always want people to use `safeint`

packages/typespec-autorest/test/formats.test.ts
tjprescott: Perfect. Just wanted to make sure this behavior was codified since it underlies the logic of the rule, which doesn't live here.

packages/typespec-azure-core/src/rules/use-standard-integer.ts
tjprescott: Done

packages/typespec-autorest/test/formats.test.ts
tjprescott: That being said, this test has been removed. There is already a test in autorest.

docs/libraries/azure-core/rules/no-generic-types.md
```diff
@@ -0,0 +1,25 @@
+---
+title: "no-generic-types"
```
timotheeguerin, line 2:
```suggestion
title: "no-generic-numerics"
```

what do you think of this rename, `no-generic-types` makes me thinkg we don't allow templates specially for someone coming from `TS` or `C#`

packages/typespec-azure-core/src/rules/no-generic-types.ts
```diff
@@ -0,0 +1,38 @@
+import { Model, createRule, paramMessage } from "@typespec/compiler";
+
+const disallowList = new Set(["integer", "numeric", "float", "decimal"]);
+const alternatives = new Map([
```
timotheeguerin, line 4: nice

packages/typespec-autorest/test/formats.test.ts
tjprescott: Test removed.

packages/typespec-azure-core/src/rules/no-generic-types.ts
```diff
@@ -0,0 +1,38 @@
...
+    return {
+      model: (model: Model) => {
+        for (const [_, prop] of model.properties) {
+          if (prop.type.kind === "Scalar") {
+            if (disallowList.has(prop.type.name)) {
+              context.reportDiagnostic({
+                target: prop.type,
+                format: {
+                  name: prop.type.name,
+                  alternative: alternatives.get(prop.type.name)!,
```
timotheeguerin, line 29: I think with the alternative we could provide the codefix. shopuld be easy we already provide the codefix from changing an identifier `createChangeIdentifierCodeFix` (used today to change `number` to `float64`)

```ts
return [createChangeIdentifierCodeFix(node, "float64")];
```

docs/libraries/azure-core/rules/no-generic-types.md
tjprescott: Updated.

packages/typespec-azure-core/test/rules/use-standard-integer.test.ts
tjprescott: These tests were refactored

packages/typespec-azure-core/README.md
tjprescott: The new message just recommends using bit-specific sizes.

markcowl:
Fixes: https://github.com/Azure/typespec-azure/issues/669

docs/libraries/azure-core/reference/linter.md
tjprescott: This has been replaced with a rule that simply checks for generic numeric types (numeric, decimal, float, integer). @johanste  please take a look at the updated logic and let me know your thoughts.

packages/typespec-azure-core/src/rules/no-generic-types.ts
tjprescott: I can certainly do that, but do we really want to put _something_ there just to resolve the rule if it's potentially wrong? Would it be better to just force the service team to do the research and put the size-appropriate value?

timotheeguerin (approved):
Approving as this looks good just to figure out if we want a codefix or not

packages/typespec-azure-core/src/rules/no-generic-numeric.ts
```diff
@@ -0,0 +1,59 @@
+import { Model, Scalar, createRule, paramMessage } from "@typespec/compiler";
+
+const disallowList = new Set(["integer", "numeric", "float", "decimal"]);
+const alternatives = new Map([
+  ["integer", "int32"],
```
markcowl, line 5: What are your thoughts about writing a codefix for this?  I am thinking that the 'integer' and 'float' fixes are helpful, but we shoyuld not attempt an automated fix for'numeric' or 'decimal'

markcowl (approved):

packages/typespec-azure-core/src/rules/no-generic-numeric.ts
tjprescott: https://github.com/Azure/typespec-azure/pull/619#discussion_r1566359749

packages/typespec-azure-core/src/rules/no-generic-types.ts
johanste: I think we need the service team to specify the actual behavior. How many instances of existing specs are running into this issue?

PR #620, opened 2024-04-05
Verify samples against linters

timotheeguerin:
Needs https://github.com/microsoft/typespec/pull/3124

### DIFF
--- a/packages/typespec-azure-playground-website/samples/arm.tsp
+++ b/packages/typespec-azure-playground-website/samples/arm.tsp
@@ -55,20 +55,21 @@ model EmployeeProperties {
 
 /** The provisioning state of a resource. */
 @lroStatus
-enum ProvisioningState {
-  ...ResourceProvisioningState,
+union ProvisioningState {
+  string,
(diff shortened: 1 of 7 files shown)


packages/typespec-azure-playground-website/samples/arm.tsp
```diff
@@ -55,20 +55,21 @@ model EmployeeProperties {
 
 /** The provisioning state of a resource. */
 @lroStatus
-enum ProvisioningState {
-  ...ResourceProvisioningState,
+union ProvisioningState {
+  string,
+  ResourceProvisioningState,
```
markcowl, line 60: I think here we should just add in the values from the standard provisioning state, since this is probably not what we want
```suggestion
  /** Resource has been created.*/
  Succeeded: "Succeeded",

  /** Resource creation failed.*/
  Failed: "Failed",

   /** Resource creation was canceled.*/
  Canceled: "Canceled",
```

markcowl (changes requested):

markcowl (approved):

PR #624, opened 2024-04-07
fix orphan union as enum has no usage problem

tadelesh:
cc: @haolingdong-msft

### DIFF
--- a/packages/typespec-client-generator-core/test/types.test.ts
+++ b/packages/typespec-client-generator-core/test/types.test.ts
@@ -823,7 +823,7 @@ describe("typespec-client-generator-core: types", () => {
         string,
       }
       `);
-      
+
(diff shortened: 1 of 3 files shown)


haolingdong-msft (approved):
LGTM.

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -812,6 +812,25 @@ describe("typespec-client-generator-core: types", () => {
       strictEqual(unionAsEnum.usage, UsageFlags.Input | UsageFlags.Output);
       strictEqual(unionAsEnum.access, "public");
     });
+
+    it("usage override for orphan union as enum", async function () {
```
haolingdong-msft, line 816: how about add another case for `@access(Access.Internal)`?

Issue #633, opened 2024-04-08
Add flattenProperty doc


lmazuel:
Should be documented here: https://azure.github.io/typespec-azure/docs/howtos/DataPlane%20Generation%20-%20DPG/06types

lmazuel:
https://github.com/Azure/typespec-azure/pull/635

PR #635, opened 2024-04-08
Add flattenProperty in the doc

lmazuel:
Fix https://github.com/Azure/typespec-azure/issues/633

### DIFF
--- a/docs/howtos/DataPlane Generation - DPG/06types.mdx
+++ b/docs/howtos/DataPlane Generation - DPG/06types.mdx
@@ -95,7 +95,9 @@ TODO
 <TabItem value="typescript" label="Typescript" >
 
 ```typescript
-TODO;
+interface Foo {
+  name: string;
+}
 ```
 
 </TabItem>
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft (commented):
We still have a common problem of this not working when it matters most.
https://github.com/weidongxu-microsoft/autorest.java/blob/tsp_mgmt_test-flatten/typespec-tests/tsp/arm.tsp#L147-L148

The 1st one does not work.

I think @archerzz had an issue to typespec team.

weidongxu-microsoft:
Also, Java disabled the flatten for DPG, regardless of whether there is such decorator.

tadelesh:
> We still have a common problem of this not working when it matters most. https://github.com/weidongxu-microsoft/autorest.java/blob/tsp_mgmt_test-flatten/typespec-tests/tsp/arm.tsp#L147-L148
> 
> The 1st one does not work.
> 
> I think @archerzz had an issue to typespec team.

It seems we could add decorator to template, not the instance model.

iscai-msft (approved):

PR #642, opened 2024-04-09
[tcgc] support nullable additional properties (#625)

tadelesh:


.chronus/changes/fix_nullable-2024-3-8-15-5-44.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
```
timotheeguerin, line 2: This cannot be a feature if you do a hotfix

timotheeguerin (changes requested):
you cannot do a feature as a hotfix this will bump the minor version and on next release skip a version for everyone

PR #656, opened 2024-04-11
change service method parameter impl

tadelesh:
Use http operation when generating service method parameters.


iscai-msft (approved):
thanks @tadelesh !

PR #661, opened 2024-04-12
Adding name property to ARMResource base model and introduce new ResourceNameParameter to simplify name

allenjzhang:
This PR contains:
- Adding `name` to `ArmResource` base model to match ARM definition
- Adding a spread model to be used by top level ARM resources to simplify `name` parameter declaration.

No tests added at this point.


packages/samples/test/output/azure/resource-manager/arm-scenarios/extension-resource/@azure-tools/typespec-autorest/2021-10-01-preview/openapi.json
```diff
@@ -98,7 +98,7 @@
           {
             "name": "roleAssignmentName",
             "in": "path",
-            "description": "Name of roleAssignment",
+            "description": "The name of the resource",
```
markcowl, line 101: We should be able to use interpolation to figure this out

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -7,6 +7,30 @@ using Azure.ResourceManager.Private;
 
 namespace Azure.ResourceManager;
 
+model DefaultNameParameter<
```
markcowl, line 10: Per discussion, if we rejigger the decorator slightly, so that they take a keyname and segment name, and the name being set to null or never means that they calculate the default (and otherwise, they use the supplied name).  Should allow us to consolidate these templates.

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -7,6 +7,30 @@ using Azure.ResourceManager.Private;
...
+  Resource extends ArmResource,
+  NamePattern extends valueof string = "^[a-zA-Z0-9-]{3,24}$"
+> {
+  @doc("The name of the resource")
```
markcowl, line 14:
```suggestion
  @doc("The name of the {name}", Resource)
```

packages/typespec-azure-resource-manager/lib/models.tsp
markcowl: Note that this also makes it easier to change one thing (say the pattern) without having to change others

packages/typespec-azure-resource-manager/lib/private.decorators.tsp
```diff
@@ -82,3 +82,9 @@ extern dec resourceParameterBaseFor(target: ModelProperty, values: unknown[]);
  * Marks an enum as representing the valid `common-types` versions.
  */
 extern dec armCommonTypesVersions(target: Enum);
+
+/**
+ * Provides default name decoration on resource name property with
+ * camelcased and pluralized key and segment name
+ */
+extern dec defaultResourceKeySegmentName(target: ModelProperty, armResource: Model);
```
markcowl, line 90: Per discussion, update to take a segmentName parameter and specially process null or empty (or never) values

markcowl (changes requested):

docs/getstarted/azure-resource-manager/step02.md
```diff
@@ -52,27 +50,27 @@ model UserResource is TrackedResource<UserResourceProperties> {
...
 @armResourceOperations
 interface Users {
-  get is ArmResourceRead<UserResource>;
-  create is ArmResourceCreateOrUpdateAsync<UserResource>;
-  update is ArmResourcePatchSync<UserResource, UserResourceProperties>;
-  delete is ArmResourceDeleteSync<UserResource>;
-  listByResourceGroup is ArmResourceListByParent<UserResource>;
-  listBySubscription is ArmListBySubscription<UserResource>;
+  get is ArmResourceRead<User>;
+  create is ArmResourceCreateOrUpdateAsync<User>;
```
markcowl, line 54:
```suggestion
  create is ArmResourceCreateOrReplaceAsync<User>;
```

docs/getstarted/azure-resource-manager/step05.md
```diff
@@ -89,25 +89,21 @@ model NotificationDetails {
...
 @armResourceOperations
 interface Users {
-  get is ArmResourceRead<UserResource>;
-  create is ArmResourceCreateOrUpdateAsync<UserResource>;
-  update is ArmResourcePatchSync<UserResource, UserResourceProperties>;
-  delete is ArmResourceDeleteSync<UserResource>;
-  listByResourceGroup is ArmResourceListByParent<UserResource>;
-  listBySubscription is ArmListBySubscription<UserResource>;
+  get is ArmResourceRead<User>;
+  create is ArmResourceCreateOrUpdateAsync<User>;
```
markcowl, line 93:
```suggestion
  create is ArmResourceCreateOrReplaceAsync<User>;
```

markcowl (changes requested):
A couple of nhits, also looks like we have a resource property duplication linting rule that will need to be modified to not detect the overridden name as a duplicate property.

I *think*t this is just supposed to detect properties from the envelope duplicated in the properties bag

markcowl (approved):
Looks great!

PR #670, opened 2024-04-15
Rename try it pipeline to pr tools and update chronus github commenter to 0.4.0

timotheeguerin:


eng/pipelines/pr-tools.yml
```diff
@@ -70,7 +71,7 @@ extends:
             steps:
               - checkout: self
 
-              - script: npx -p @chronus/github-pr-commenter@0.3.0 chronus-github-pr-commenter verify
+              - script: npx -p @chronus/github-pr-commenter@0.4.0 chronus-github-pr-commenter verify
```
markcowl, line 74: It's unfortunate we have to hard code the version here

markcowl (approved):

PR #671, opened 2024-04-15
[Autorest-canonical Emitter] Update resolveRef for the common type dir

dinwa-ms:
We tested the autorest-canonical emitter in this [PR](https://github.com/Azure/azure-rest-api-specs-pr/pull/17643/checks?check_run_id=23587064163) and it failed due to setting the common types. Update the emitter with logic to getSchemaOrRef and resolveRef to correctly set the common type dir.


packages/typespec-autorest-canonical/README.md
```diff
@@ -70,3 +76,29 @@ Omit unreachable types. By default all types declared under the service namespac
 
 If the generated openapi types should have the `x-typespec-name` extension set with the name of the TypeSpec type that created it.
 This extension is meant for debugging and should not be depended on.
+
+## Decorators
+
+### Autorest
+
+- [`@useRef`](#@useref)
```
timotheeguerin, line 84: I don't think this is a good way of dealing this. This will require adding this decorator to every common element in the library which we don't want to do. We are also going to be removing that from arm as we cannot have arm depend on autorest.

Instead you should build a set of known types and when resolving the type replace the reference

timotheeguerin (changes requested):
This is not a good solution as it will involve adding a dependency to that library from arm to add the new decorator which is something we are trying to move away from with autorest

packages/typespec-autorest-canonical/README.md
dinwa-ms: removed the decorator

.chronus/changes/update-resolveRef-2024-2024-4-17-11-10-36.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
timotheeguerin, line 3:
```suggestion
changeKind: feature
```

packages/typespec-autorest-canonical/src/decorators.ts
```diff
@@ -0,0 +1,37 @@
+import { Model, ModelProperty, Program, Service, Type } from "@typespec/compiler";
```
timotheeguerin, line 1: remove this file too

packages/typespec-autorest-canonical/src/openapi.ts
```diff
@@ -145,6 +148,7 @@ import {
   Refable,
 } from "./types.js";
 import { AutorestCanonicalEmitterContext, resolveOperationId } from "./utils.js";
+import { getRef } from "./decorators.js";
```
timotheeguerin, line 151: Have you tested this? This will not work either no? its not going to get the value from anywhere

timotheeguerin: you should call the `resolveRef` from autorest emitter

packages/typespec-autorest-canonical/src/decorators.ts
dinwa-ms: removed

.chronus/changes/update-resolveRef-2024-2024-4-17-11-10-36.md
dinwa-ms: fixed

packages/typespec-autorest-canonical/src/openapi.ts
dinwa-ms: import it from autorest

timotheeguerin (approved):

PR #673, opened 2024-04-15
[tcgc] add subscriptionId to client parameters

iscai-msft:


tadelesh (approved):
i'm ok with current change. but i believe we should introduce new decorator to promote operation level parameter to client level, just like swagger extension `x-ms-parameter-location`.

PR #678, opened 2024-04-16
Remove typespec-autorest from arm

lmazuel:
I don't have the skills to finish this PR, but we'd really like to remove autorest as dependency of the ARM package, and as far as I see, the only thing remaining is `private.decorators.ts` with:

```typescript
import { RefProducerParams, setRefProducer } from "@azure-tools/typespec-autorest";
```

And some code that uses it, so I thought I could push what I did so far :p


markcowl:
This dependency really only exists to communicate emitter configuration settings to the ARM library.  If we revived the old library options proposal, that would be the cleanest mechanism for resolving, I think.

This should mainly be about the dependency itself, though, there is nothing important to API shapes that comes from this dependency.

timotheeguerin:
done in #811

PR #689, opened 2024-04-18
dpg doc, add caution section to spread doc

weidongxu-microsoft:
Adding a caution section about the spread / anonymous model.
It is not always best to do this spread, as it affect SDK signature. For some languages, having lots of method arguments is not a good design.

### DIFF
--- a/docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
+++ b/docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
@@ -133,6 +133,11 @@ public void post(User user);
 
 ## Spread
 
+Please use the *spread* feature with caution.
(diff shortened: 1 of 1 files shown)


tadelesh (approved):

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -133,6 +133,11 @@ public void post(User user);
 
 ## Spread
 
+Please use the *spread* feature with caution.
+- The anonymous model to be spread into operation should have less than 6 settable properties. See [simple methods](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-parameters).
```
weidongxu-microsoft, line 137: .NET and Java have guideline on "simple methods" that should not have more than 6 arguments.

Python and TypeScript/JavaScript does not have such guideline (AFAIK). Python has keyword argument. TypeScript does not put optional arguments to method.

Issue #97, opened 2024-01-08
Add Azure.Core linter rule to encourage open union for discriminator


markcowl:
Implementation Issue for https://github.com/microsoft/typespec/issues/2589

- Add a linting rule that emits a warning if the discriminator property is not defined in the base type as an open string union

markcowl:
@markcowl talk to laurent about timing

PR #693, opened 2024-04-18
Add no-string-discriminator linter rule

timotheeguerin:
fix [#97](https://github.com/Azure/typespec-azure/issues/97) add linter rule to recommend the discriminator property to be explicitly defined as an extensible union


packages/typespec-azure-core/src/rules/no-string-discriminator.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+  create(context) {
+    return {
+      model: (model: Model) => {
+        const discriminator = getDiscriminator(context.program, model);
+        if (discriminator === undefined) {
+          return;
+        }
+
+        const prop = model.properties.get(discriminator.propertyName);
```
markcowl, line 21: Does `properties.get` resolve inherited properties?

packages/typespec-azure-core/test/rules/no-string-discriminator.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    .toEmitDiagnostics([
+      {
+        code: "@azure-tools/typespec-azure-core/no-string-discriminator",
+        message:
+          'Use an extensible union instead of a plain string (ex: `union PetKind { cat: "cat", dog: "dog", string };`)',
+      },
+    ]);
+});
+
+it("emits a warning @discriminator points to a property that is not a union", async () => {
```
markcowl, line 40: Should probably have a variant for enums, just in case.

markcowl (changes requested):

markcowl (approved):

Issue #577, opened 2024-04-02
`ResponseHeadersTrait` with invalid reference cause compiler crash


timotheeguerin:
```
alias ServiceTraits = ResponseHeadersTrait<invalid>;
```
```
TypeError: paramModel.properties is not iterable
```
[Playground Link](https://azure.github.io/typespec-azure/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3JlIjsKCnVzaW5nIFR5cGVTcGVjLkh0dHA70BVSZXN00RVWyXLIG0HEWS5Db3Jl0hIuVHJhaXRzOwoKQHNlcnZpY2UKICBAdXNlRGVwZW5kZW5jeSjLLsddcy52MV8wX1ByZXZpZXdfMikKCm5hbWVzcGFjZSBDb250b3NvLldpZGdldE1hbmFnZXI7CgovKiogVGjJICDGICDHISDHfyDnATIuICovCmVudW0gyHQgewogIMRFxxEgMjAyMi0wOC0zMcQsICBgyREwYCwKfQoKYWxpYXMgU8ZZ5gDpID0gUmVzcG9uc2VIZWFkZXJzxRg8Rm91bmRhdMRoLlJlcXVlc3RJZMYiPjsK&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)

Issue #332, opened 2024-02-28
Using `ResourceOperations` interface with an invalid identifier crash


timotheeguerin:
```
TypeError: traitModel.properties is not iterable
```
[Playground Link](https://azure.github.io/typespec-azure/playground?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtxhVjb3JlIjsKCnVzaW5nIEHEFC5Db3JlLlRyYWl0czsKCmFsaWFzIE9wZXJhdGlvbnMgPcwnUmVzb3VyY2XKIDxTZXJ2aWNlxkE%2BOwpz&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)

PR #694, opened 2024-04-18
Fix crash when `Traits` builders gets passed non model

timotheeguerin:
fix [#577](https://github.com/Azure/typespec-azure/issues/577)
fix [#332](https://github.com/Azure/typespec-azure/issues/332)


markcowl (approved):
I don't see any way this could exist in any specs that currently compile

PR #698, opened 2024-04-19
[tcgc] return projection of single api version

iscai-msft:


tadelesh:
close in favor of; https://github.com/Azure/typespec-azure/pull/700

PR #707, opened 2024-04-19
Remove non relevant linter rules

timotheeguerin:
All rules about the previous state of `@fixed` and open enums
- no `@fixed` - we should deprecate it as its a noop
- no fixed enum discriminator - new rule taking over recommending unions https://github.com/Azure/typespec-azure/pull/693


packages/typespec-azure-resource-manager/src/rules/missing-x-ms-identifiers.ts
```diff
@@ -12,7 +12,8 @@ import { isArmCommonType } from "../common-types.js";
 
 export const missingXmsIdentifiersRule = createRule({
   name: "missing-x-ms-identifiers",
-  description: "Azure services should not use enums.",
+  description:
+    "Array properties should describe their identifying properties with x-ms-identifiers.",
```
markcowl, line 16:
```suggestion
    `Array properties should describe their identifying properties with x-ms-identifiers. Decorate the property with @OpenAPI.extension("x-ms-identifiers", [id-prop])  where "id-prop" is a list of the names of identifying properties in the item type.`,
```

markcowl (approved):

PR #719, opened 2024-04-22
[tcgc] add versioning doc

### LINKED ISSUE #716: Projected version support TCGC Author doc

iscai-msft:
fixes #716


lmazuel:
We should add a section to override to a specific api-version, but otherwise LGTM

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,207 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+# Versioning
+
+This doc details what emitters will generate for versioned specs
+
+## Single API Version
+
+If there is just one API version in the spec, we will generate
```
lmazuel, line 10: Somehow it feels like you didn't finish your sentence?

catalinaperalta: I had the same impression when reading this sentence.

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+## Multiple API version
+
+The configuration flag `api-version` allows you to toggle the behavior that our emitters will generate.
+
+We will get the versioning information from the `Versions` enum that you define and link as part of the `@typespec/versioning` library.
```
catalinaperalta, line 16: What is "link as part of the `@typespec/verisoning` library" referring to?

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+> NOTE: The ordering of the values in the `Versions` enum is very important. We use this information to determine the order of versions. Our default value will be the **last entry** in the `Versions` list
+
+### Default
+
+By default our emitters will only generate the surface used by the latest api version if there are multiple defined.
+
+By default, only models used in api surface of the latest api version are included in generation. Documentation and enums showing the available API versions will still include all of the known api versions, meaning there will be documentation for both the preview and stable releases.
```
catalinaperalta, line 24: Maybe better to put these statements together into one block, like:
```suggestion
By default our emitters will only generate the surface used by the latest api version if there are multiple defined. This includes generating only the models used in the surface area of the latest api version. 

Documentation and enums showing the available API versions will still include all of the known api versions, meaning there will be documentation for both the preview and stable releases.
```

catalinaperalta: Also, this is a style nit but should we be consistent about the casing of API?

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,207 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+### Override to return all
+
+You can also override the signature to return the combined api surface area of all of the separate api versions. Different languages have different support for versioning validation
+
+**To override the behavior, pass in `api-version: all` in your `tspconfig.yaml`.**
```
catalinaperalta, line 117: Would it make sense to also include an example of how this looks in tspconfig.yaml?

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
iscai-msft: lmao yeah my b

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
iscai-msft: it means the enum you pass to `@versioned`. Updating doc, thanks!

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
iscai-msft: fair, let me always lowercase it

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,406 @@
...
+
+```typescript
+
+```
+
+</TabItem>
+<TabItem value="java" label="Java" >
+
+```java
+
```
weidongxu-microsoft, line 395: Discussion:

Even it is `"all"`, I don't think Java would ever want to generate a API not in stable api-version to a SDK in stable api-version.
Java likely won't generate a removed API (or property) in any case, especially for GA version.

weidongxu-microsoft: This discussion may be a bit on the real case, instead of a sample

Let's assume the versions in tsp be

```ts
enum Versions {
  v2023_11_01_preview: "2023-11-01-preview",
  v2023_11_01: "2023-11-01",
  v2023_12_01_preview: "2023-12-01-preview",
  v2023_12_01: "2023-12-01",
}
```

When release SDK for "v2023_12_01", typically it only want to generate for version
```
v2023_11_01
v2023_12_01
```

If any API / property only exists in preview version, SDK won't want to include them.

---

In data-plane, this may not be an issue from typespec source, as reviewer would suggest service to remove any preview prior to stable, so their typespec would only contain
```ts
enum Versions {
  v2023_11_01: "2023-11-01",
  v2023_12_01: "2023-12-01",
}
```
at this stage.

If there is future preview, it would be append to the list. Then the next stable would clear these previews.

However, this may not be the case for management.

tadelesh: For `all`, I don't think we have a clear decision for how to deal with version resiliency. I suggest removing this part for now.

lmazuel: `all` just means "no projection". Meaning you get access to the full list of things, and you have access to `api_versions` in each object to decide your own logic, if you want to.
This is literally what we have today, we all receive "all", and we made it work.
I'm ok to add a new conf option like "stable_only", but can someone explains to me how it would be different from generating the latest?

tadelesh: specifying `2023-12-01` will not get any removal or say breaking changes caused by previous stable version because of projection. use `all` will not do projection, so all info will be there.

weidongxu-microsoft: I am OK if this is just a case or demonstration.

But personally I would avoid advice service set this "all". At present, I don't see a convincing case that service really need to set "all" to handle their SDK.

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,539 @@
...
+    /**
+     * Gets the latest service version supported by this client library.
+     * 
+     * @return The latest {@link ServiceServiceVersion}.
+     */
+    public static ServiceServiceVersion getLatest() {
+        return V2023_11_01;
+    }
+}
```
weidongxu-microsoft, line 101: Could we put the public APIs (without implementation code) in doc?

The only code that matters would be the `getLatest()` API.

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,539 @@
...
+
+ServiceClientClient serviceClientClient = new ServiceClientClientBuilder()
+  // override the api version, even if only one version is defined in the spec
+  .serviceVersion(ServiceServiceVersion.V2023_11_01)
+  // other configurations
+  .buildClient();
```
weidongxu-microsoft, line 107: 4 space

XiaofeiCao: Fixed.

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
XiaofeiCao: updated

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,462 @@
...
+
+enum Versions {
+  v2023_11_01: "2023-11-01",
+}
+
+model StableModel {
+  stableFeature: string;
+}
+
+op stableFunctionality(...StableModel): void;
```
chunyu3, line 32: It better not to mix spread feature here. This doc only focus on versioning.

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,536 @@
...
+
+model StableModel {
+  stableFeature: string;
+}
+
+@added(Versions.v2023_11_01_preview)
+@removed(Versions.v2023_11_01)
+op previewFunctionality(...PreviewModel): void;
+
+op stableFunctionality(...StableModel): void;
```
chunyu3, line 290: we may need to add an operation which is added in `2023_11_01` version.

docs/howtos/DataPlane Generation - DPG/09versioning.mdx
```diff
@@ -0,0 +1,536 @@
...
+ServiceClientOptions options = new ServiceClientOptions(ServiceVersion.V2023_11_01);
+ServiceClient client = new ServiceClient(endpoint, options);
+//You can specify the service api-version when create client instance. Now client's api_version will be "2023-11-01"
+```
+
+</TabItem>
+<TabItem value="typescript" label="Typescript" >
+
+```typescript
+// there's no apiVersion defined in the all the operations, TypeScript emitter will ignore it.
```
qiaozha, line 82: All the examples in this document, don't really have an apiVersion parameter in the operation level, typescript emitter will just ignore it. 
@lmazuel @iscai-msft I should if we should add any test cases that apiVersion is actually being defined in the operation level?

iscai-msft: I'm going to start off without having it in the docs, since people will mainly use templates, and that takes care of it

lmazuel (approved):

PR #736, opened 2024-04-24
Adding arm extended location and re-organize common type definitions

allenjzhang:


packages/typespec-azure-resource-manager/lib/arm.tsp
```diff
@@ -7,16 +7,19 @@ import "@azure-tools/typespec-autorest";
 import "@azure-tools/typespec-azure-core";
 
 import "./arm.foundations.tsp";
-import "./common-types.tsp";
+import "./common-types/common-types.tsp";
+import "./common-types/types.tsp";
+import "./common-types/managed-identity.tsp";
```
markcowl, line 12: I really like having the swagger common-types directory structure mirrored here, it will make it easier for us to keep all these types in sync as the swagger common-types evolve

packages/typespec-azure-resource-manager/lib/common-types/extended-location.tsp
```diff
@@ -0,0 +1,21 @@
+namespace Azure.ResourceManager.Foundations {
+  /** The complex type of the extended location. */
+  model ExtendedLocation {
```
markcowl, line 3: We should ask for a definition of this in swagger common-types

markcowl (approved):
I think we should do a test run against typespec-next before merging this.  Otherwise, this looks great

PR #745, opened 2024-04-26
prevent tcgc versioning projection

tadelesh:
since autorest emitter needs to generate multiple version swagger, need to pass `all` to tcgc to prevent tcgc do the versioning projection to `latest` version.


iscai-msft (approved):
looks good to me, @timotheeguerin are you ok with this change? Are there any other places we should add this code?

Background is: we decided for tcgc we are going to by default project to the last version, you can override this with flag `api-version`

packages/typespec-autorest/src/openapi.ts
```diff
@@ -163,6 +163,8 @@ const defaultOptions = {
 
 export async function $onEmit(context: EmitContext<AutorestEmitterOptions>) {
   const resolvedOptions = { ...defaultOptions, ...context.options };
+  // set api-version to all to prevent tcgc versioning projection
+  (context.options as any)["api-version"] = "all";
```
timotheeguerin, line 167: I do feel like this is quite hacky, if another emitter wanted that they'd have to do the same thing. Could we maybe instead have some options in `createSdkContext` to use or not to use the projection

iscai-msft: yeah, my only concern with that approach is now there are two places and ways to input this information. Would you be ok for now if we just use this code, open an issue to keep track of this and put it on the backlog, and if another library runs into this issue, we can brainstorm more?

timotheeguerin: well I don't see it as 2 place to input the info. One is what the user ask for the other is what does your emitter support. With autorest it handles the versioning itself so that param is pointless. 

I think it would be better to have 
```ts
createSDKContext(context, {
   versionStrategy: "latest-or-api-version" | "ignore",
});
```

timotheeguerin: if you think this is really not good then sure but doing what we have now is bypassing the type system so we could easily break it

iscai-msft: ok that seems fair, let me update the code here. Thanks!

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -571,16 +571,27 @@ export function listOperationsInOperationGroup(
...
+  options?: CreateSdkContextOptions
+): SdkContext<TOptions, TServiceOperation> {
   const diagnostics = createDiagnosticCollector();
   const protocolOptions = true; // context.program.getLibraryOptions("generate-protocol-methods");
   const convenienceOptions = true; // context.program.getLibraryOptions("generate-convenience-methods");
   const generateProtocolMethods = context.options["generate-protocol-methods"] ?? protocolOptions;
   const generateConvenienceMethods =
     context.options["generate-convenience-methods"] ?? convenienceOptions;
+  if (options?.ignoreVersioning === false) {
+    context.options["api-version"] ?? "all";
```
timotheeguerin, line 593: that's not going to set the value here

timotheeguerin: where is it used? also does `all` makes any sense for the others?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -571,16 +571,27 @@ export function listOperationsInOperationGroup(
   return operations;
 }
 
+interface CreateSdkContextOptions {
+  ignoreVersioning?: boolean;
```
timotheeguerin, line 575:
```suggestion
  readonly ignoreVersioning?: boolean;
```

could consider making this a union of string values in case there is different strategies in the future.

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: `all` means we don't do any version projection, we just return all versions. Thanks for pointing out that i'm not actually setting `api-version`. I think it's fair to just set `api-version` here, so we don't need multiple boolean statements down the pipeline

timotheeguerin: juste note that setting this exact object is a bad pattern you are updating an object passed by the consumer that they wouldn't expect to be modified. This could lead to some bug down the line, unlikely but it could.

PR #762, opened 2024-04-30
Initial commit of ARM bug fix and rename/reorg

allenjzhang:
This change contains following changes:

- Updated `extendedLocation` with optional and READ, CREATE
- Renamed base common types to be consistent with ARM common type, included backcompat aliases with #deprecate
  - ArmResource -> Resource
  - TrackResourceBase -> TrackResource
  - ProxyResourceBase -> ProxyResource
  - ExtensionResourceBase -> ExtensionResource
  - ResourceSku -> Sku
  - ResourcePlan -> Plan

Note: both the `#deprecated alias` and in some cases `TrackResource<T>` and `Foundations.TrackResource` may cause spec break. `rest-api-specs/typespec-next` will be checked and updated if this is approved.


packages/samples/specs/resource-manager/liftr.confluent/confluent.tsp
```diff
@@ -8,7 +8,7 @@ using TypeSpec.Http;
 using TypeSpec.Rest;
 using TypeSpec.Versioning;
 using Azure.ResourceManager;
-using Azure.ResourceManager.Foundations;
+// using Azure.ResourceManager.Foundations;
```
markcowl, line 11: I would just remove this

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -11,24 +12,34 @@ model ArmResourceBase {}
  * Base model that defines common properties for all Azure Resource Manager resources.
  */
 @doc("Common properties for all Azure Resource Manager resources.")
-model ArmResource extends ArmResourceBase {
+model Resource extends ArmResourceBase {
```
markcowl, line 15: Do we still need ArmResourceBase?  I think we may have this as a template parameter constraint, but I wonder if we can't just use Resource instead.

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -11,24 +12,34 @@ model ArmResourceBase {}
...
   @doc("Fully qualified resource ID for the resource. Ex - /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/{resourceProviderNamespace}/{resourceType}/{resourceName}")
   @visibility("read")
-  id: string;
+  id?: armResourceIdentifier;
```
markcowl, line 18: Hmm, yeah, I agree that we should describe this correctly, but I wonder if it won't cause some breaking changes for some sdks based on the v3 definitions

markcowl (approved):
Just a few questions and comments, but the changes look good.

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
allenjzhang: Agreed.

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
allenjzhang: You are right, This was changed to `arm-id` in v4. While it doesn't impact current specs, I will use versioning to correctly model it. Fixing.

packages/typespec-azure-resource-manager/lib/private.models.tsp
```diff
@@ -0,0 +1,10 @@
+namespace Azure.ResourceManager.Private;
+
+/**
+ * Describes the shape of strict constraint model to type check against `Resource` template
+ *  parameter of ARM operation templates & base models.
+ */
+model ArmResourceTypeConstraint {
```
markcowl, line 7: Per the issue this morning, this behavior will not work after we fix this issue:https://github.com/microsoft/typespec/issues/3267

So I think we need to go with one of the other options (a decorator, for example, would allow us to make this an error)

markcowl (changes requested):

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -48,6 +48,31 @@ export function $omitIfEmpty(context: DecoratorContext, entity: Model, propertyN
...
+  context: DecoratorContext,
+  entity: Operation | Model,
+  sourceType: Model,
+  constraintType: Model
+) {
+  if (sourceType !== undefined && constraintType !== undefined) {
+    // walk the baseModel chain until find a match or fail
+    let baseType: Model | undefined = sourceType;
+    do {
+      if (baseType === constraintType) return;
```
markcowl, line 61: Do we want to do this, though?  It feels like this will be too restrictive for brownfield APIs.    There are brownfield APIs that just mark their own internal resource as x-ms-azure-resource, and these will have to depend on Resource now.

packages/typespec-azure-resource-manager/test/enforce-constraint.test.ts
```diff
@@ -0,0 +1,77 @@
+import { expectDiagnosticEmpty, expectDiagnostics } from "@typespec/compiler/testing";
+import { describe, it } from "vitest";
+import { checkFor } from "./test-host.js";
+
+describe("typespec-azure-resource-manager: @enforceConstraint", () => {
+  it("emits no error when template param extends from Resource", async () => {
```
markcowl, line 6: I think we need some additional tests
- direct extension of Resource (succeed)
- model the *is* Resource (fail)

packages/typespec-azure-resource-manager/src/lib.ts
```diff
@@ -94,6 +94,12 @@ export const $lib = createTypeSpecLibrary({
         default: "Resource types must have a property with '@path` and '@segment' decorators.",
       },
     },
+    "template-type-constraint-no-met": {
+      severity: "error",
+      messages: {
+        default: paramMessage`The template parameter "${"sourceType"}" for "${"entity"}" does not satisfy the constraint type "${"constraintType"}".`,
```
markcowl, line 100:
```suggestion
        default: paramMessage`The template parameter "${"sourceType"}" for "${"entity"}" does not extend the type "${"constraintType"}". Use the "TrackedResource", "ProxyResource", or "ExtensionResource" template to define the resource.`
```

The important thing is that we give them some actionable thing to do in the message.  If we want this decorator to be generally useful, we might add a parameter to the decorator arguments that includes an optional extra action message, like this added sentence.

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -48,6 +48,31 @@ export function $omitIfEmpty(context: DecoratorContext, entity: Model, propertyN
   }
 }
 
+export function $enforceConstraint(
+  context: DecoratorContext,
+  entity: Operation | Model,
+  sourceType: Model,
```
markcowl, line 54: See above, might want to add an 'action message'  argument telling  the user what action to take to correct the error.

markcowl (changes requested):

packages/typespec-azure-resource-manager/src/private.decorators.ts
allenjzhang: I thought about it. But given that Foundations.Resource is really the lowest basic model, this gives us an opportunity not letting them continue on the forked path. I don't see there is any reason not to extend from this.

markcowl (approved):

.chronus/changes/azhang_armReorg-2024-3-29-17-13-46.md
```diff
@@ -0,0 +1,20 @@
+---
+changeKind: internal
```
timotheeguerin, line 2:
```suggestion
changeKind: deprecation
```

allenjzhang:
Updated and added `actionMessage`. Added to test scenarios.

PR #766, opened 2024-04-30
[tcgc] add generated names for constants

### LINKED ISSUE #553: [TCGC][Bug] getAllModels, Optional literal do not have correct generated name
**TSP definition:**
```tsp
model Model {
  literal: "literal";
  optionalLiteral?: "optionalLiteral";
}
```

**Java code:**
In Java's generated code, we will generate an enum for the `optionalLiteral`, the enum is called `ModelOptionalLiteral`, like below. I guess this optional literal is like anonymous model, and will need TCGC to generate a correct name for it.
```java
public enum ModelOptionalLiteral {
    /**
     * Enum value optionalLiteral.
     */
    OPTIONAL_LITERAL("optionalLiteral");

    /**
     * The actual serialized value for a ModelOptionalLiteral instance.
     */
    private final String value;

    ModelOptionalLiteral(String value) {
        this.value = value;
    }

    /**
     * Parses a serialized value to a ModelOptionalLiteral instance.
     * 
     * @param value the serialized value to parse.
     * @return the parsed ModelOptionalLiteral object, or null if unable to parse.
     */
    public static ModelOptionalLiteral fromString(String value) {
        if (value == null) {
            return null;
        }
        ModelOptionalLiteral[] items = ModelOptionalLiteral.values();
        for (ModelOptionalLiteral item : items) {
            if (item.toString().equalsIgnoreCase(value)) {
                return item;
            }
        }
        return null;
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public String toString() {
        return this.value;
    }
}

```

Discussed with @tadelesh offline, TCGC does not handle the name for it currently. It is better if TCGC can handle it.
@archerzz  @pshao25 Does `.Net` have this behavior? What will .Net generate for optional literal?

iscai-msft:
fixes #553

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -178,7 +178,7 @@ function getSdkHttpParameters(
   ) {
     // if we have a body param and no content type header, we add one
     const contentTypeBase = {
-      ...createContentTypeOrAcceptHeader(retval.bodyParam),
+      ...createContentTypeOrAcceptHeader(httpOperation, retval.bodyParam),
       description: `Body parameter's content type. Known values are ${retval.bodyParam.contentTypes}`,
     };
     if (!methodParameters.some((m) => m.__raw && isContentTypeHeader(context.program, m.__raw))) {
@@ -197,7 +197,7 @@ function getSdkHttpParameters(
   if (responseBody && !headerParams.some((h) => isAcceptHeader(h))) {
     // If our operation returns a body, we add an accept header if none exist
     const acceptBase = {
-      ...createContentTypeOrAcceptHeader(responseBody),
+      ...createContentTypeOrAcceptHeader(httpOperation, responseBody),
     };
     if (!methodParameters.some((m) => m.name === "accept")) {
       methodParameters.push({
@@ -221,6 +221,7 @@ function getSdkHttpParameters(
 }
 
 function createContentTypeOrAcceptHeader(
+  httpOperation: HttpOperation,
   bodyObject: SdkBodyParameter | SdkHttpResponse
 ): Omit<SdkMethodParameter, "kind"> {
   const name = bodyObject.kind === "body" ? "contentType" : "accept";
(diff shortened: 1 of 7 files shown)


tadelesh (approved):

tadelesh:
i had the same idea but not finished the impl. thanks for the work.

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -520,12 +531,19 @@ function buildNameFromContextPaths(
...
   let createName: string = "";
   for (let j = lastNonAnonymousModelNodeIndex; j < contextPath.length; j++) {
-    if (!contextPath[j]?.type?.name) {
+    const currContextPathType = contextPath[j]?.type;
+    if (
+      currContextPathType?.kind === "String" ||
+      currContextPathType?.kind === "Number" ||
+      currContextPathType?.kind === "Boolean"
+    ) {
+      createName = `${createName}${contextPath[j].displayName}${capitalCase(String(currContextPathType.value))}`;
```
weidongxu-microsoft, line 540: A minor concern on naming, whether it be `${createName}${contextPath[j].displayName}}` or `${createName}${contextPath[j].displayName}${capitalCase(String(currContextPathType.value))}`

A real case in storage blob
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/Microsoft.BlobStorage/stable/2021-12-02/blob.json#L6237-L6248
to 
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/SyncCopyStatusType.java

However, I guess if we really need a specified name in TypeSpec, we can do e.g.
```
enum SyncCopyStatusType {
  Success: "success"
}

model .. {
  `x-ms-copy-status`?: SyncCopyStatusType.Success
}
```

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -247,6 +248,8 @@ function createContentTypeOrAcceptHeader(
       kind: "constant",
       value: bodyObject.contentTypes[0],
       valueType: type,
+      name: `${httpOperation.operation.name}ContentType`,
```
pshao25, line 251: What if we keep it "ContentType"? I'm not sure if we could recognize it if it is not called "ContentType".

iscai-msft: this is the name for the type of the `Content-Type` parameter, not the `Content-Type` parameter itself. Since each content type parameter can have its own types, if all of the types are called `ContentType` then there are duplicate names for these constant types

PR #776, opened 2024-05-03
Move deprecated operations to separate file and flatten decorators file

timotheeguerin:


allenjzhang (approved):
add #deprecate in legacy

PR #793, opened 2024-05-07
[tcgc] release apiversion clientDefaultValue for template params fix

iscai-msft:


timotheeguerin (changes requested):
we are making a release right now

PR #806, opened 2024-05-09
Add tsp-client docs section

catalinaperalta:
Add a section with information about `tsp-client`.


docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
```diff
@@ -0,0 +1,182 @@
+# Getting started with `tsp-client`
+
+`tsp-client` is a simple command line tool to facilitate generating client libraries from TypeSpec.
+
+## Installation
+
+```
+npm install -g @azure-tools/typespec-client-generator-cli
```
timotheeguerin, line 8: Should that tool just be added to the spec repo `package.json` if the main goal is to use it there?

catalinaperalta: I dont think we require tsp-client for any part of the REST API spec development process currently. We only use it to do initial conversion for some brownfield services. Aside from that it is mainly used in the language repositories and their corresponding automation scripts (this last part is in progress).

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
```diff
@@ -0,0 +1,182 @@
...
+```
+
+## Commands
+
+Use one of the supported commands to get started generating clients from a TypeSpec project.
+This tool will default to using your current working directory to generate clients in and will
+use it to look for relevant configuration files. To specify a different output directory, use
+the `-o` or `--output-dir` option.
+
+### init
```
iscai-msft, line 24: nit: it might be more clear if you do `tsp-client init` instead of just `init` in the header. That way you can immediately see what the whole command is

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
```diff
@@ -0,0 +1,182 @@
...
+
+Initialize the client library directory using a tspconfig.yaml. When running this command pass in a path to a local or remote tspconfig.yaml with the `-c` or `--tsp-config` flag.
+
+The `init` command generates a directory structure following the standard pattern used across Azure SDK language repositories, creates a [tsp-location.yaml](#tsp-locationyaml) file to control generation, and performs an initial generation of the client library. If you want to skip client library generation, then pass the `--skip-sync-and-generate` flag.
+
+> IMPORTANT: This command should be run from the root of the repository.
```
iscai-msft, line 30: can you give an example of the root of the repository? it can get confusing for monorepos, is it `azure-sdk-for-python` or is it `azure-sdk-for-python/sdk/containerregistry`

iscai-msft: oh i see you include it in the below example, if you feel like it you can expound on it more here, but def just a small itty bitty nit

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
```diff
@@ -0,0 +1,182 @@
...
+
+Example:
+
+```
+azure-sdk-for-python/sdk/contosowidgetmanager/azure-contoso-widgetmanager> tsp-client update
+```
+
+### sync
+
+Sync a TypeSpec project with the parameters specified in tsp-location.yaml.
```
iscai-msft, line 50: maybe you can say that this is the first step of `update`, so if you want to split `update` up you can call `sync` and or `generate`

iscai-msft (approved):

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
timotheeguerin: yeah but isn't the folow that you'd want ot call `tsp-client` from the spec repo when you are ready to generate sdks?

catalinaperalta: Through the release planner we link folks to this page when they're ready to generate sdks: https://eng.ms/docs/products/azure-developer-experience/develop/sdk-develop?tabs=dataplane That page links folks to language specific generation instructions, which work within the language repos. Plus some languages, like C# and Python wrap tsp-client under language specific tooling. I believe C# requires their sdks to be generated with the dotnet build command since it does some other processing under the hood for them so that the generated sdks are compatible with the configurations used in the repo

catalinaperalta: We might still need to add tsp-client in the package.json once we integrate it more with the specs repo, since we want to have the automation scripts use tsp-client. Once we do that I can revisit and update this page

catalinaperalta: I could also remove this from the page and recommend folks run it with npx

timotheeguerin: no that's fine if the goal is do have it there in the future but you are not ready yet docs looks good then

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
catalinaperalta: done, thanks for the suggestion!

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
catalinaperalta: Done :)

docs/howtos/DataPlane Generation - DPG/09tsp_client.mdx
catalinaperalta: I was wondering if that could be confusing? Some of these commands have required flags, so I wouldnt want someone to think that they can run the command without them.

PR #807, opened 2024-05-09
feat(doc): update csharp support for @flattenProperty

archerzz:
A follow-up of #633 and #635


lmazuel (approved):
I don't know C#, but since it's documentation I trust that @archerzz knows what should be there :)

Issue #289, opened 2024-02-22
How to express example and title node from swagger in tsp


tadelesh:
Title example: https://github.com/Azure/azure-rest-api-specs-pr/blob/69142cc803bd9e4bbe0737cb0af1e25db2cc2f63/specification/hdinsight/resource-manager/Microsoft.HDInsight/stable/2024-05-01/hdinsight.json#L1574
Example example: https://github.com/Azure/azure-rest-api-specs-pr/blob/69142cc803bd9e4bbe0737cb0af1e25db2cc2f63/specification/hdinsight/resource-manager/Microsoft.HDInsight/stable/2024-05-01/hdinsight.json#L1675

timotheeguerin:
`@summary` is the way to do title but looks like we are not respecting that in autorest emitter. 

https://azure.github.io/typespec-azure/playground?c=QHN1bW1hcnkoIlRpdGxlIGlzIGhlcmUiKQptb2RlbCBGT28ge30%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

timotheeguerin:
`@example` is not supported today we have an issue for it

PR #813, opened 2024-05-09
`@summary` sets the title of definitions

timotheeguerin:
fix [#289](https://github.com/Azure/typespec-azure/issues/289)


packages/typespec-autorest/src/openapi.ts
```diff
@@ -1690,9 +1690,9 @@ export async function getOpenAPIForService(
 
   function resolveProperty(prop: ModelProperty, context: SchemaContext): OpenAPI2SchemaProperty {
     let propSchema;
-    if (prop.type.kind === "Enum" && prop.default) {
+    if (prop.type.kind === "Enum" && prop.defaultValue) {
```
markcowl, line 1693: hmm, why all the value type fixes here?

markcowl (approved):

packages/typespec-autorest/src/openapi.ts
timotheeguerin: Technically not related to this PR, just I think the eslint pipeline not reporting the deprecated in the ci anymore, I need to investigate. YOu just see them in the IDE as warning.

Issue #131, opened 2024-01-16
Customization for arm build-in model's visibility and optional setting for resource `properties` property


tadelesh:
For current ARM resource model the `properties` property is optional and with visibility read and create, but existed RP's swagger may have different setting for them. Other than redefine a new `ProxyResource` model, is there other way to do customization for them?
e.g.: https://github.com/Azure/azure-rest-api-specs/blob/05c4049bc22f0ec65acc18f9835132397049cb9e/specification/elasticsan/resource-manager/Microsoft.ElasticSan/stable/2023-01-01/elasticsan.json#L1652

markcowl:
Because the typing of templates is structural and not nominal, you can use your own resource types if you want a different 'properties' optionality/visibility.

The optionality and visibility in the templates is based on current LintDiff rules and required behaviors for RPaaS clients, so there is a disincentive to update the templates.

Need to understand how this impacts management client generation before we think about adding an extra degree of freedom to templates, but because a custom resource type should work just fine, this should not be blocking.

tadelesh:
It will cause breaking change for lots of languages as required property will become a param in default model ctor. Confirm that custom resource type could resolve it. But it seems this is a common issue for lots of existed swaggers. To prevent too much manual work, I'm going to change conversion logic not to use predefined resource template, instead extends predefined resource base and add properties property separately.

markcowl:
I think this is related to the inheritance hierarchy for resources as well - given how many resources don't conform to the represented TrackedResource/ProxyResource/ExtensionResource hierarchy, we may not want to use the existing model templates anyway

markcowl:
@markcowl ensure that swagger converter has a clear path to avoid breaking client changes

MaryGao:
@tadelesh @markcowl could you help me understand on how this issue is going to resolve? 

This would cause a big breaking for JS([here](https://github.com/Azure/autorest.typescript/issues/2335#issuecomment-1988422027)) and I was wondering how we could mitigate this.

tadelesh:
If service team insist on setting visibility and optionality which is different from current arm template, we will use customized definition to define the resource which will make the visibility and optionality correct.

MaryGao:
Thanks and it works for me!

allenjzhang:
@visibility issue has been addressed.  Still need to find the best way for optionality for brownfields

- Have parallel `legacy.XXResource` in ARM with required properties
- Manually code the above in each spec
[...]

PR #833, opened 2024-05-14
Adding a legacy Required properties resource mix models

allenjzhang:
Putting in a separate namespace avoid name conflict.

#131


packages/typespec-azure-resource-manager/lib/legacy/models.tsp
```diff
@@ -0,0 +1,59 @@
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using TypeSpec.OpenAPI;
+using Azure.ResourceManager.Foundations;
+using Azure.ResourceManager.Private;
+
+namespace Azure.ResourceManager.Legacy;
+
+//#region Standard Resource Operation Interfaces
```
timotheeguerin, line 9: do we supprot region in tsp?

packages/typespec-azure-resource-manager/lib/legacy/models.tsp
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Concrete tracked resource types can be created by aliasing this type using a specific property type.
+ *
+ * See more details on [different Azure Resource Manager resource type here.](https://azure.github.io/typespec-azure/docs/howtos/ARM/resource-type)
+ * @template Properties A model containing the provider-specific properties for this resource
+ */
+@doc("Concrete tracked resource types can be created by aliasing this type using a specific property type.")
```
markcowl, line 16: I think these changes should be made to the standard templates, and the older ones should be marked as Legacy and deprecated.

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/legacy/models.tsp
```diff
@@ -0,0 +1,59 @@
+using TypeSpec.Http;
```
markcowl, line 1: For legacy, I think we are still leaving out scenarios where folks extend Resource or EntityResource.  I would prefer it if we just let them extend a common-types resource type, and added a new decorator they would need to decorate the type with that is the equivalent of `@armResourceInternal` and `@includeInapplicableMetadataInPayload`, and relax the rules around extension to include this case.

packages/typespec-azure-resource-manager/lib/legacy/models.tsp
markcowl: remove

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -40,7 +40,7 @@ model ResourceNameParameter<
 @armResourceInternal(Properties)
 @includeInapplicableMetadataInPayload(false)
 model TrackedResource<Properties extends {}> extends Foundations.TrackedResource {
-  @doc("The resource-specific properties for this resource.")
+  @doc("The resource-specific optional properties for this resource.")
```
markcowl, line 43: I don't think these doc updates are necessary.  The typespec and the swagger already say it is optional, adding the documentation makes swagger changes without adding any information.

markcowl (changes requested):
I think we should consider a different approach, since this leaves out many important brownfield scenarios.  Also, a couple of inline nits

packages/typespec-azure-resource-manager/lib/legacy/models.tsp
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Concrete tracked resource types can be created by aliasing this type using a specific property type.
+ *
+ * See more details on [different Azure Resource Manager resource type here.](https://azure.github.io/typespec-azure/docs/howtos/ARM/resource-type)
+ * @template Properties A model containing the provider-specific properties for this resource
+ */
+@doc("Concrete tracked resource types can be created by aliasing this type using a specific property type.")
+@armResourceInternal(Properties)
+@includeInapplicableMetadataInPayload(false)
+model TrackedResourceWithRequiredProperties<Properties extends {}>
```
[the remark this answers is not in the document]
allenjzhang: I feel the namespace being legacy which has good separation. It is targeted for brownfield. Unless ARM has disallowed required properties (currently they don't) we probably don't want to put deprecate.

allenjzhang:
This PR is ON HOLD until the discussion of @visibility design issues..

Issue #832, opened 2024-05-14
[Bug]: Inconsistent type of `OperationStatusResult.percentComplete`


Alancere:
### Describe the bug

go sdk two different results:
![image](https://github.com/Azure/typespec-azure/assets/46921893/a971694c-f882-4373-9418-179cb20c8516)

### Reproduction

tsp: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/common-types/types.tsp#L220

spec common-type: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v3/types.json#L439

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

tadelesh:
cc: @allenjzhang

allenjzhang:
Per ARM RPC guide line, it would be float64 (double). https://github.com/Azure/azure-resource-manager-rpc/blob/8231d7df51aec87e6ccb5bcc04478136758b0a4c/v1.0/async-api-reference.md?plain=1#L164

PR #860, opened 2024-05-17
Fix percentComplete should be a float not an int

timotheeguerin:
fix [#832](https://github.com/Azure/typespec-azure/issues/832)


packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -217,7 +217,7 @@ model OperationStatusResult {
   @doc("Percent of the operation that is complete.")
   @minValue(0)
   @maxValue(100)
-  percentComplete?: int32;
+  percentComplete?: float64;
```
timotheeguerin, line 220: In the original issue it was mentioning `float32` but we also reference `float64` here https://github.com/Azure/typespec-azure/blob/a72d7efeacb6e6b8d64b9c65853ecc42558d77d5/packages/typespec-azure-resource-manager/lib/responses.tsp#L124

So probably good those 2 are consistent, so do we need to change the other to float32 or this one is good.

allenjzhang: https://github.com/Azure/azure-resource-manager-rpc/blob/8231d7df51aec87e6ccb5bcc04478136758b0a4c/v1.0/async-api-reference.md?plain=1#L164

timotheeguerin: so sound like this fix is good then

markcowl (approved):

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
markcowl: Yes, double and float64 are synonymous

PR #868, opened 2024-05-20
[Hotfix] Release/may 2024  Update `ManagedServiceIdenitity.UserAssignedIdentities` property back to inline Record

allenjzhang:
- Update `ManagedServiceIdenitity.UserAssignedIdentities` property back to Record
- Added some missing arm comment type versioning selector

### DIFF
--- a/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
+++ b/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
@@ -36,22 +36,6 @@ model ManagedServiceIdentity {
   userAssignedIdentities?: Record<UserAssignedIdentity>;
 }
 
-/** The set of user assigned identities associated with the resource. The userAssignedIdentities dictionary keys will be ARM resource ids in the form: '/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}. The dictionary values can be empty objects ({}) in requests.", */
-@armCommonDefinition(
-  "UserAssignedIdentities",
-  {
-    version: Azure.ResourceManager.CommonTypes.Versions.v4,
-    isDefault: true,
-  },
-  "managedidentity.json"
-)
-@armCommonDefinition(
-  "UserAssignedIdentities",
-  Azure.ResourceManager.CommonTypes.Versions.v5,
-  "managedidentity.json"
-)
-model UserAssignedIdentities is Record<UserAssignedIdentity>;
-
 #deprecated "Please change ManagedSystemIdentityProperties to SystemAssignedServiceIdentity."
 alias ManagedSystemIdentityProperties = SystemAssignedServiceIdentity;
 /**
--- a/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
+++ b/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
@@ -36,6 +36,23 @@ model ManagedServiceIdentity {
   userAssignedIdentities?: Record<UserAssignedIdentity>;
 }
 
+/** The set of user assigned identities associated with the resource. The userAssignedIdentities dictionary keys will be ARM resource ids in the form: '/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}. The dictionary values can be empty objects ({}) in requests.", */
+#deprecated "Please use `Record<UserAssignedIdentity>` directly for better SDK API."
+@armCommonDefinition(
+  "UserAssignedIdentities",
+  {
+    version: Azure.ResourceManager.CommonTypes.Versions.v4,
+    isDefault: true,
+  },
+  "managedidentity.json"
+)
+@armCommonDefinition(
+  "UserAssignedIdentities",
+  Azure.ResourceManager.CommonTypes.Versions.v5,
+  "managedidentity.json"
+)
+model UserAssignedIdentities is Record<UserAssignedIdentity>;
+
 #deprecated "Please change ManagedSystemIdentityProperties to SystemAssignedServiceIdentity."
 alias ManagedSystemIdentityProperties = SystemAssignedServiceIdentity;
 /**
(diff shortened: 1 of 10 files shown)


packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
```diff
@@ -33,10 +33,23 @@ model ManagedServiceIdentity {
   type: ManagedServiceIdentityType;
 
   @doc("The identities assigned to this resource by the user.")
-  userAssignedIdentities?: UserAssignedIdentities;
+  userAssignedIdentities?: Record<UserAssignedIdentity>;
 }
 
 /** The set of user assigned identities associated with the resource. The userAssignedIdentities dictionary keys will be ARM resource ids in the form: '/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}. The dictionary values can be empty objects ({}) in requests.", */
+@armCommonDefinition(
```
timotheeguerin, line 40: What was that doing before

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
```diff
@@ -76,6 +89,19 @@ alias ManagedIdentityType = ManagedServiceIdentityType;
 /**
  * A managed identity assigned by the user.
  */
+@armCommonDefinition(
```
timotheeguerin, line 92: same here, was that just missed and it wasn't replacing with $ref?

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
```diff
@@ -33,10 +33,23 @@ model ManagedServiceIdentity {
...
+    isDefault: true,
+  },
+  "managedidentity.json"
+)
+@armCommonDefinition(
+  "UserAssignedIdentities",
+  Azure.ResourceManager.CommonTypes.Versions.v5,
+  "managedidentity.json"
+)
 model UserAssignedIdentities is Record<UserAssignedIdentity>;
```
timotheeguerin, line 53: feel like this needs to be removed or it won't have the same meaning

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
markcowl: If people use the ManagedServiceIdentityProperty mix-in, then the common-types are referenced before you get here.  This is only to pave the way for brownfield APIs that reference this type directly (there are around 10, I believe)

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
markcowl: There are existing APIs that just use this common type, and we want to allow them not to break.  We would also like to generate the common types.

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
markcowl: Same reason.  Although here, I don't know if anyone references this property directly, but we do want to generate this into common-types

markcowl (approved):

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
timotheeguerin: If they use this type they will break when converting to typespec generation

timotheeguerin: This should either be removed and added just for the common types generation or it should be an alias and some magic done in the common type generation

allenjzhang: I added as alias in a different file.

allenjzhang: Another option is to leave it as model but add a deprecation on it. This way the common type swagger code gen is not as hacky. I actually like it slightly more than alias. Just slightly. :-)

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
```diff
@@ -33,10 +33,24 @@ model ManagedServiceIdentity {
   type: ManagedServiceIdentityType;
 
   @doc("The identities assigned to this resource by the user.")
-  userAssignedIdentities?: UserAssignedIdentities;
+  userAssignedIdentities?: Record<UserAssignedIdentity>;
 }
 
 /** The set of user assigned identities associated with the resource. The userAssignedIdentities dictionary keys will be ARM resource ids in the form: '/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}. The dictionary values can be empty objects ({}) in requests.", */
+#deprecated "Please use `Record<UserAssignedIdentity>` directly for better SDK API."
```
allenjzhang, line 40: Added deprecation notice.

timotheeguerin:
```suggestion
#deprecated "DO NOT USE, use `Record<UserAssignedIdentity>` directly, using this model will result in a different client SDK when generated from TypeSpec vs the swagger output."
```
can we make the message clearer that this should not be used at all? It nost just that it will give better SDK, using it will result in a breaking change when generation goes from swwgger to typespec

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
markcowl: @timotheeguerin  Yes, we want to have a conversation with anyone who is converting one of these specs that uses this type, so that they mave to managedserviceidentity, or use Record<UserAssignedIdentity> instead.  Meanwhile, we can use this type to generate common-types.

markcowl:
seems like the chronus check is broken,  typespec-autorest isn't impacted, is it?

I would guess you might need to regenerate samples, though

timotheeguerin:
> seems like the chronus check is broken, typespec-autorest isn't impacted, is it?
> 
> I would guess you might need to regenerate samples, though

I think it is comparing against `main`

timotheeguerin:
> seems like the chronus check is broken, typespec-autorest isn't impacted, is it?
> 
> I would guess you might need to regenerate samples, though

ok fixed for the check, the comment though still show the outdated comparaison but I think that's a missing functionality in chronus

PR #870, opened 2024-05-20
[tcgc] change nullable behavior

iscai-msft:
Instead of having `.nullable` as a property on an `SdkType` or an `SdkPropertyType`, we're going to deprecate all `nullable` properties on our sdk types, and have users call helper function `isNullable` instead and pass in an `SdkType` to figure out if it's nullable. Types will have complete fidelity with TypeScript, as we will return them as a union of an `SdkType` with an `SdkNoneType`


iscai-msft:
Code snippet for using the helper methods:

```ts
let property: SdkModelPropertyType;
const propertyTypeNullable = isNullable(property.type);
const propertyType = getUnderlyingNullableType(context, property.type) if propertyTypeNullable else property.type;
```

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -604,3 +605,40 @@ export function getHttpOperationWithCache(
...
+  return nullCount > 0 && nonNullCount > 0;
+}
+
+/**
+ * Since we don't remove null types from the values of a union type, this helper function helps return the type without null unioned in.
+ *
+ * Only call after first making sure the type isNullable with our helper function
+ * @param type
+ */
+export function getUnderlyingNullableType(type: SdkType): SdkType {
```
m-nash, line 634: I am wondering if we follow the same pattern we did for SdkEnumType.  SdkEnumType is a special case for SdkUnionType but it has a few additional values that we want to extract automatically as helpers from the union.  SdkNullableType feels like the same thing.  Its a union type of exactly 2 items and due to its type we know its nullable and it can have a helper property which tells me the underlying type.

```js
export interface SdkNullableType extends SdkTypeBase {
  kind: "nullable";
  type: SdkType;
}
```

Then just like we do when we come across the special union that matches enum we return an instance of SdkNullabletype insteda of SdkUnion so everyone has these two conveniences implicit at hand.  I think this will simplify the code quite a bit in tcgc and for emitters.

iscai-msft: i think i like this idea, the only thing i want to double-click on  your definition of a nullable item being a union of exactly 2 items. I've always thought that it has to be a union of at least 2 items, in which case it would ultimately be a nullable union.

This way you don't have to deal with the helper functions, and languages that don't care about nullable can easily ignore any `SdkNullableType`

m-nash: ah you are correct, I could have a nullable union or nullable enum.  I think my snippet interface above still works though it would just change the logic of when we return an SdkNullableType vs when we return an SdkUnionType.

Maybe we should consider having SdkNullableType extend SdkUnionType and just provide the type property which is what we call in this pr the underlying type?

srnagar: Just want to confirm - `foo: ModelA | ModelB | null` is also `SdkNullableType` right? If so, what does `getUnderlyingNullableType()` return? I think @m-nash's proposal should still work even if we have > 2 types.

srnagar, line 634: Should we call this method `getNonNullType()`? `getUnderlyingType` seems like we are returning a base type or some internal type.

m-nash: @srnagar getUnderlyingNullableType goes away and SdkNullableType.type is where you get the non-nullable version so it would contain an SdkUnionType of `ModelA | ModelB`

Btw all the naming of what I am proposing is up for discussion I was just throwing out something to start the discussion.

iscai-msft: this method goes away with @m-nash's proposal, since the `SdkNullableType.valueType` will get you the same value

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -604,3 +605,40 @@ export function getHttpOperationWithCache(
   context.httpOperationCache.set(operation, httpOperation);
   return httpOperation;
 }
+
+/**
+ * Determines if a type is nullable.
+ * @param type
+ * @returns
```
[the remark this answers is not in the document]
iscai-msft: this function will also get removed since we will look at whether `type.kind === "nullable"` to determine whether it's nullable

packages/typespec-client-generator-core/src/public-utils.ts
iscai-msft: updated!

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -97,6 +97,7 @@ export type SdkType =
   | SdkArrayType
   | SdkTupleType
   | SdkDictionaryType
+  | SdkNullableType
```
haolingdong-msft, line 100: `SdkNullableType`: I'm a bit confused on the nullable type, I thought it is type whose kind is `null` to represent `null` in tsp. As the example in chat shows.

image

haolingdong-msft:
Thanks for the change. Got two questions:

1. Wonder the backward compatibility as this will break emitters. Currently the `nullable` property on model and parameter/model property are marked as deprecated, but will still be available, right? When will TCGC remove the deprecated `nullable` property? 
2. Will TCGC provide `getNonNullType` helper, seems I did not find in the pr.
image

docs/howtos/DataPlane Generation - DPG/06types.mdx
```diff
@@ -806,6 +793,110 @@ public final class Animal implements JsonSerializable<Animal> {
...
+about nullability by inspecting the type of a property.
+
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+model Foo {
+  basicNullableProperty: string | null;
+  modelNullableProperty: Bar | null;
+}
```
haolingdong-msft, line 808: How about the case `nullableProperty: A|B|null`? Will the returned type be 'union' or 'nullable'?

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: The final implementation changed to introduce a new `SdkNullableType` in TCGC instead of reusing the `SdkUnionType` to reduce complexity for emitters.

docs/howtos/DataPlane Generation - DPG/06types.mdx
tadelesh: `nullable` with value type of `union`.

tadelesh:
> Thanks for the change. Got two questions:
> 
> 1. Wonder the backward compatibility as this will break emitters.  Currently the `nullable` property on model and parameter/model property are marked as deprecated, but will still be available, right? When will TCGC remove the deprecated `nullable` property?  Also I think after this version of TCGC released, emitters will need to change the handling logic to handle the newly added `nullable` type, otherwise emitter could break. **May be better to release minor version instead of patch veresion**.
> 2. Will TCGC provide `getNonNullType` helper, seems I did not find in the pr.
>    image
>    update: ok, seems we don't need this helper with the current design in your pr. Wonder the need for helper `isNullable()`, as we already add a `nullable` type, `isNullable()` is just the type whose kind is `nullable`?

1. it is a breaking change. and since every emitter need to switch to use new `nullable` type, i suggest to remove all deprecated `nullable` attribute at the same time. @iscai-msft what do you think?
2. as the impl changed, you just need to get `valueType` attribute from `nullable` type, no extra helper needed.

weidongxu-microsoft:
Seems the main advantage of a `SdkNullableType` over `SdkType.nullable` is that it could do a proper wrap of the underlying typespec union in `__raw` (and potential for more properties other than current `valueType`)?

There could also be diff on `description`/`details` between `SdkNullableType` and its type in `valueType`, though it probably not common and likely not used by emitter if they don't output a dedicated class for nullable type.

haolingdong-msft:
> > Thanks for the change. Got two questions:
> > 
> > 1. Wonder the backward compatibility as this will break emitters.  Currently the `nullable` property on model and parameter/model property are marked as deprecated, but will still be available, right? When will TCGC remove the deprecated `nullable` property?  Also I think after this version of TCGC released, emitters will need to change the handling logic to handle the newly added `nullable` type, otherwise emitter could break. **May be better to release minor version instead of patch veresion**.
> > 2. Will TCGC provide `getNonNullType` helper, seems I did not find in the pr.
> >    image
> >    update: ok, seems we don't need this helper with the current design in your pr. Wonder the need for helper `isNullable()`, as we already add a `nullable` type, `isNullable()` is just the type whose kind is `nullable`?
> 
> 1. it is a breaking change. and since every emitter need to switch to use new `nullable` type, i suggest to remove all deprecated `nullable` attribute at the same time. @iscai-msft what do you think?
> 2. as the impl changed, you just need to get `valueType` attribute from `nullable` type, no extra helper needed.

Make sense to me.
1. please release a minor version, instead of patch version, so that emitters will not automatically get upgraded and break. Thanks.

tadelesh:
> Seems the main advantage of a `SdkNullableType` over `SdkType.nullable` is that it could do a proper wrap of the underlying typespec union in `__raw` (and potential for more properties other than current `valueType`)?
> 
> I assume there could also be diff on `description`/`details` between `SdkNullableType` and its type in `valueType`, though it probably not common and likely not used by emitter if they don't output a dedicated class for nullable type.

it's kind of easy way to create two types.
for description, i believe only for the following case will get the doc.
```
@doc("Test")
union Test {
  string,
  int32,
  null
}
```

packages/typespec-client-generator-core/src/interfaces.ts
haolingdong-msft: Yeah, I realized the design changed from we have talked from chat. Thanks for the update.

I kind of feel that the current design looks similar as previous ‘sdktype.nullable’, as they both try to represent a nullable type, but current one introduce ‘SdkNullableType’ and brings another level of hierarchy/complexity. And the current design also breaks downstream.

Would like to know the reason why the design changed and the benefit of the current design (comparing to sdktype.nullable).

iscai-msft:
We'll wait to release this with the first of the month release that typespec does. We can't do a separate minor version release by ourselves, we can only do patch versions between releases, because we are tied to the typespec release schedule. This should be out within a week if we want to wait for the patch version to make these changes

iscai-msft:
@tadelesh I'm good with breaking the `nullable` properties at the same time, there are too many i think right now and it's confusing

m-nash (approved):
This looks great thanks for all the work here.

I would love to see at least one regen preview in some language successfully adopt this change, before merge would be awesome, but at least post merge and before release.

srnagar (approved):

joheredi (approved):

tadelesh:
@chunyu3 could you help to try this change and regen the sdk code to see if there is any hidden problem?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -242,7 +237,11 @@ export interface SdkDictionaryType extends SdkTypeBase {
   kind: "dict";
   keyType: SdkType;
   valueType: SdkType;
-  nullableValues: boolean;
+}
+
+export interface SdkNullableType extends SdkTypeBase {
+  kind: "nullable";
+  valueType: SdkType;
```
ArcturusZhang, line 244: could we have another name for this?
Because in this context, this is never a value, not like in the SdkArrayType or SdkDictionaryType, that thing is called "the value of this array/dictionary" therefore it is `valueType`, but here it makes no sense that we say "we have a value inside a null type".

How about just `type` or `innerType`?

Similar comment also applies to `SdkUnionType`, we have `values` but those are not the "values". `valueTypes` makes more sense but not so much, similar with above reason - in an union of `string | int`, the `string` or `int` are not called as `value` of the union, in this case, `variantTypes` might be a better name.

iscai-msft: i'm ok to switch it back to `type`, and I get what you're saying about `union` as well. Can you open a separate issue for what you'd like to see in unions, and we can do the correct deprecation?

iscai-msft:
@m-nash I've confirmed the behavior with Python

Issue #772, opened 2024-05-02
[tcgc] switch `...` to always mean spreading, regardless of model or alias


iscai-msft:
This issue is blocked by #771 

### Clear and concise description of the problem

Right now, tcgc has `...MyModel` maintain the logical grouping, but `...MyAlias` spread out the parameters. This is because there is a lack of support for `op(MyModel)` language definitions, and because we want to regroup models spread in azure core templates.

However, this behavior is often confusing and has led to a lot of back and forth. After talking with the TypeSpec team, we want to change the default, because eventually with more language support, `...MyModel` should mean spread. The TypeSpec team will also modify the spreading currently done in `Azure.Core` templates (#771). If we implement this change concurrently with the `Azure.Core` changes, emitters shouldn't really see any code changes

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

iscai-msft:
closing because we're documenting this change in #1043

tadelesh:
reopen since this is only for tcgc change.

PR #886, opened 2024-05-21
[tcgc] new spread behavior

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/772

several cases related with spread:

1.
```
op test(a: string, b: string, c: string): void;
```
2.
```
model Test {
  prop: string
}
op test(...Test): void;
```
3.
```
op test(@body body: {prop: string}): void;
```

4.
```
model Test {
  prop: string
}
op test(@body body: {...Test}): void;
```

tcgc will do http body spread for 1 and 2, not for 3 and 4. and for http body model, 1, 2 will have a model `TestRequest` with spread usage and internal access.

**Adoption guideline for emitters:**

For emitters that have already adopted `getAllOperations`, if an operation has spread parameters:
1. The parameters of `SdkServiceMethod` are spread.
2. The body parameter of `SdkServiceOperation` is an anonymous model with `Usage.Spread` usage and `Access.Internal` access. And the `correspondingMethodParams` of the body will be list of the parameters of `SdkServiceMethod`.
3. Emitters could rely on the above info to generate clients' code.

For emitters that have not adopted `getAllOperations`, emitters need to do:
1. Use `getHttpOperationWithCache` to prevent different body type resolving for spread cases.
2. Refer this code: https://github.com/Azure/typespec-azure/blob/5d77ba3e93521735a84051b4037c6f55e0a8a00a/packages/typespec-client-generator-core/src/types.ts#L1311-L1325 to change emitter's logic of handling operations with spread parameters.
3. When call `getClientType` from the body type of operation with spread, the return model will be an anonymous model with `Usage.Spread` usage and `Access.Internal` access. DO NOT CALL `getEffectivePayloadType` for body type if it is from spread.


packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2786,63 +2802,6 @@ describe("typespec-client-generator-core: package", () => {
       strictEqual(bodyParameter.type.properties[0].name, correspondingMethodParams[0].name);
     });
 
-    it("spread with discriminate type with implicit property", async () => {
```
tadelesh, line 2789: this test has been removed because if we treat spread as spread, then spread a discriminated type is invalid.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1260,8 +1259,16 @@ function updateTypesFromOperation(
...
     if (generateConvenient) {
-      // spread body model should be none usage
-      if (sdkType.kind !== "model" || !sdkType.isGeneratedName) {
+      // special logic for spread body model
+      if (
+        httpBody.type.kind === "Model" &&
+        httpBody.type.name === "" &&
+        (httpBody.type.sourceModels.some((x) => x.usage === "spread") ||
+          httpBody.type === operation.parameters) &&
+        !context.spreadModels?.has(httpBody.type)
```
tadelesh, line 1268: logic to judge whether a http body type is from spread.

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -569,6 +534,50 @@ export function getCorrespondingMethodParams(
...
+    ) {
+      return methodParam;
+    }
+    if (
+      serviceParam.kind === "header" &&
+      serviceParam.serializedName === "Accept" &&
+      methodParam.name === "accept"
+    ) {
+      return methodParam;
+    }
```
weidongxu-microsoft, line 566: This appears to be somewhat hack? Is it possible for us to know they are all?

PS: for header, should we always do case-insensitive? That would cover the "accept".

weidongxu-microsoft, line 566: This appears to be somewhat hack? Is it possible for us to know they are all?

PS: for header, should we always do case-insensitive on serializedName (path/query on serializedName  as well)? That would cover the "accept".

weidongxu-microsoft: Fine. I am not sure about the logic. Just from the name, if serviceOperation does not have such param, why there is `serviceParam` of such. They are also added?

haolingdong-msft: I would also agree with always do case-insensitive for serializedName, I don't think we need to keep both lowercase name and first-letter capital case serializedName.

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -569,6 +534,50 @@ export function getCorrespondingMethodParams(
...
+): SdkModelPropertyType | undefined {
+  const queue: SdkModelPropertyType[] = [...methodParameters];
+  const visited: Set<SdkModelType> = new Set();
+  while (queue.length > 0) {
+    const methodParam = queue.shift()!;
+    if (
+      methodParam.__raw &&
+      serviceParam.__raw &&
+      (methodParam.__raw === serviceParam.__raw ||
+        methodParam.__raw === serviceParam.__raw.sourceProperty)
```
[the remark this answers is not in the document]
tadelesh: add a comment

packages/typespec-client-generator-core/src/http.ts
weidongxu-microsoft: OK. May worth a comment as well. Next dev may not familiar with this.

packages/typespec-client-generator-core/src/types.ts
tadelesh: refined the logic. current judgement should be perfect.

iscai-msft (approved):
thanks @tadelesh

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -3091,17 +3055,429 @@ describe("typespec-client-generator-core: package", () => {
...
       strictEqual(documentMethodParam.kind, "method");
       const op = method.operation;
       ok(op.bodyParam);
       strictEqual(op.bodyParam.kind, "body");
-      strictEqual(op.bodyParam.name, "documentTranslateContent");
+      strictEqual(op.bodyParam.name, "testRequest");
       deepStrictEqual(op.bodyParam.correspondingMethodParams, [documentMethodParam]);
     });
+
+    it("anonymous model with @body should not be spread", async () => {
```
qiaozha, line 3068: Looks like we have to take the generated names for anonymous model, because it has to be used in the operation signature.

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -3091,17 +3055,429 @@ describe("typespec-client-generator-core: package", () => {
...
+
+    it("anonymous model from spread with @bodyRoot should not be spread", async () => {
+      await runner.compileWithBuiltInService(`
+        model Test {
+          prop: string;
+        }
+        op test(@bodyRoot body: {...Test}): void;
+        `);
+      const method = getServiceMethodOfClient(runner.context.experimental_sdkPackage);
+      const models = runner.context.experimental_sdkPackage.models;
```
qiaozha, line 3118: we haven't adopted to getAllOperations for modular yet, I guess we can't leveage the tcgc feature here?

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: yes.

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: yes. emitters that has not adopted to getAllOperations need to deal with the operation by themselves. you could refer to the judgement i used in types.ts.

tadelesh:
> Could we add some guidance on how should emitter adopt this?

added in description.

haolingdong-msft:
Overall looks good to me. Just to confirm: so emitter will use body param's `access` and `usage` of `SdkServiceOperation`
to determine whether it's spread body or not, if it is spread body, then read from `SdkServiceOperation.body.correspondingMethodParams` to get the spread parameter list? So seems emitters don't need to interact with `SdkServiceMethod`?

haolingdong-msft (approved):

weidongxu-microsoft:
Tested on Java emitter, should be good now.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -64,7 +64,6 @@ import {
   getCrossLanguageDefinitionId,
   getCrossLanguagePackageId,
   getDefaultApiVersion,
-  getEffectivePayloadType,
```
MaryGao, line 67: Deco getEffectivePayloadType?

weidongxu-microsoft: I think java won't use this. So, agree on deprecation.

packages/typespec-client-generator-core/test/package.test.ts
MaryGao, line 3068: Do we have a case for alias? I believe it is the same as spread model?

```
alias Test = {
  prop: string
};
op test(...Test): void;
```

MaryGao: How about this one? Also spread case, right?

```
model Test {
  @header
  foo: string;
  prop: string
};
op test(...Test): void;
```

packages/typespec-client-generator-core/src/package.ts
tadelesh: it is still used in response.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -598,4 +598,6 @@ export enum UsageFlags {
   JsonMergePatch = 1 << 4,
   // Input will also be set when MultipartFormData is set.
   MultipartFormData = 1 << 5,
+  // Used in spread.
+  Spread = 1 << 6,
```
MaryGao, line 602: Maybe not relevant to this pr, just want to clarify UsageFlags, if a model is used in 
- patch operation for input
- another operation for spread input
- another operation for output

How many medels we will get from TCGC?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1403,6 +1422,18 @@ function updateAccessOfModel(context: TCGCContext): void {
...
 }
 
+function updateSpreadModelUsageAndAccess(context: TCGCContext): void {
+  for (const sdkType of context.spreadModels?.values() ?? []) {
+    updateUsageOfModel(context, UsageFlags.Spread, sdkType);
+  }
+  for (const sdkType of context.modelsMap?.values() ?? []) {
+    // if a type only has spread usage, then it could be internal
+    if (sdkType.usage === UsageFlags.Spread) {
+      sdkType.access = "internal";
```
MaryGao, line 1432: If a model is only used for spread, does that mean the TCGC would not return that model in getAllModels by default?

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: two. one original model with path|outpu, one anonymous model with spread.

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: same as spread model, i think it has been covered by another case.

packages/typespec-client-generator-core/src/types.ts
tadelesh: it will return it but with access internal

packages/typespec-client-generator-core/test/package.test.ts
weidongxu-microsoft: What I take is 
1. if there is `@body` or `@bodyRoot` anywhere, it is explicit body, and it is not spread
2. otherwise, spread

Issue #864, opened 2024-05-20
[Bug]: support mix types for enum types in tcgc


ArcturusZhang:
### Describe the bug

I have this enum:
```typespec
enum Fixed {
	One: 1,
	Two: "2"
}
```
which is totally fine in typespec, but TCGC gives me `string` as its valueType, which is just incorrect.

### Suggestion

I think it is reasonable that we return a union as its value type which is more intuitive.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

tadelesh:
emm, i doubt if it should be fixed in tcgc enum type or add a lint rule to avoid it.

tadelesh:
@timotheeguerin is it a valid case for typespec?

timotheeguerin:
it is today a valid use case, we don't have a way to limit enums types. But either way I would agree that we probably would want a linting rule against that for azure.

tadelesh:
since we have lint in core that disallow azure service to have enums: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/src/rules/no-enum.ts, so i believe it is an only ask for unbranded. i finished the support pr, we could decide whether we need to have this breaking change in the sync up.

PR #888, opened 2024-05-21
[tcgc] support enum value with mix types

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/864


markcowl:
Do we still need this PR?  Will close at end of month if no response

PR #902, opened 2024-05-22
Add multipart explicit part support in autorest

timotheeguerin:
Add support for https://github.com/microsoft/typespec/pull/3342 in autorest emitter


packages/typespec-azure-resource-manager/src/rules/arm-post-response-codes.ts
```diff
@@ -25,7 +25,10 @@ export const armPostResponseCodesRule = createRule({
       if (response.responses.length > 1) {
         throw new Error("Multiple responses are not supported.");
       }
-      if (response.responses[0].body !== undefined) {
+      if (
+        response.responses[0].body !== undefined &&
```
markcowl, line 29: I feel like this is leaving out lro posts that might return multipart data.  This should be a rare case, but it would be nice to support it (and add a test to verify support).

markcowl (changes requested):

markcowl (approved):

PR #922, opened 2024-05-24
Create typespec feed for only operations RT

gaurimadhok:
This creates a template that will only generate the operations resource type for a resource provider.


eng/feeds/azure-operations-scaffolding.json
```diff
@@ -0,0 +1,47 @@
+{
+  "azure-arm-operations": {
+    "title": "(rest-api-spec repo) Azure Resource Manager Service Project with Operations Resource Type",
+    "description": "Create a project with operations resource type in rest-api-spec repo, representing an Azure service ARM API",
+    "compilerVersion": "0.45.2",
+    "libraries": [],
+    "skipCompilerPackage": true,
+    "config": {
+      "emit": ["@azure-tools/typespec-autorest", "@azure-tools/typespec-autorest-canonical"]
```
dinwa-ms, line 9: Do you need the canonical emitter for this

eng/feeds/azure-operations-scaffolding.json
timotheeguerin, line 1: is that meant to be a separate file not part of the main azure scaffolding?

eng/feeds/azure-operations-scaffolding.json
gaurimadhok: removed

eng/feeds/azure-operations-scaffolding.json
gaurimadhok: removed separate file and added to main scaffolding file

eng/feeds/azure-scaffolding.json
timotheeguerin, line 1: Can you help me understand when one would choose this template? Also I think the docs should be updated to include this new template

gaurimadhok: Someone would use this template if the only resource type they want is the operations resource type. This is the use case for some new resource providers who only need that functionality. I can update the docs

gaurimadhok: @timotheeguerin updated the docs

docs/getstarted/createproject.md
```diff
@@ -22,6 +22,7 @@ When initializing a new project, you'll encounter four templates:
 | **(rest-api-spec repo) Azure Resource Manager Service Project** | Similar to the above, but tailored for Azure Resource Manager services within the `azure-rest-api-specs` repository.                                                                 |
 | **(standalone) Azure Data Plane Service Project**               | This template is configured for local use or in a personal repo, allowing for API development without the specific directory structure of the `azure-rest-api-specs` repo.           |
 | **(standalone) Azure Resource Manager Service Project**         | Similar to the above but designed for Azure Resource Manager services, configured for local use or in a personal repo.                                                               |
+| **(standalone) Azure Resource Manager Service Project with only Operations Resource Type**         | Similar to the above but designed to only have the operations resource type.                                                              |
```
timotheeguerin, line 25: Doc is not aligned with the template, it says here standalone but the template says it for use in the azure spec repo.

markcowl, line 25: I also think we should specifically call out the specific folks we expect to use this template (3rd party? a special subset of RPs?) so that we avoid confusion for the rest of the users

eng/feeds/azure-scaffolding.json
```diff
@@ -240,5 +240,50 @@
         "initialValue": "Microsoft.Contoso"
       }
     }
+  },
+  "azure-arm-operations": {
+    "title": "(rest-api-spec repo) Azure Resource Manager Service Project with only Operations Resource Type",
```
markcowl, line 245: If there is a set of intended users, we should mention them here

markcowl (changes requested):

PR #923, opened 2024-05-25
Fix when rendering version enum only include current member and mark it modelAsString: true

timotheeguerin:
This ensure that when the version enum is still referenced that 
1. we still mark it as modelAsString true to not be flagged as a breaking change in spec repo(even though it won't matter as it gets special treatment)
2. only include the emitted version member so adding new versions in the future don't update the previous one

### DIFF
--- a/.chronus/changes/fix-emit-api-version-2024-4-25-9-53-10.md
+++ b/.chronus/changes/fix-emit-api-version-2024-4-25-9-53-10.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+When emitting version enum only include current version and mark with `modelAsString: true`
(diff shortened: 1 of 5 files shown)


allenjzhang:
#1 is good
#2 I first thought if it would be better emit all versions up to current one instead a single current version. But I guess as previous version like -preview could be taken out, it would impact the swaggers. As our current autorest SDK emitters don't need the full set of versions, this sounds good to me. This would still subject to the config to completely bypass version enum generation, right?

allenjzhang:
> #1 is good
> #2 Would be better emit all versions up to current one instead a single current version? I guess as previous version like -preview could be taken out, it would then impact the swaggers. Depending on the client codegen and breaking change or not, you effectively can only do single version client instead of offer version selectors providing no breaking changes.

Further thinking about it, swaggers/openapi, it only represent one version in our multi-version to single-version emission, it is accurate to render single version there. So I am good with this change.

allenjzhang (approved):

timotheeguerin:
So this only applies if you reference the version enum in other cases we already omitted it.

In those case the only real case is when it's used in the paramerrized host.

I don't think it matters at all what that value is as it's not used in swagger generators but only selecting the current version does feel right and the simplest so we don't create extra diff for new versions

PR #925, opened 2024-05-28
[tcgc] default `.access` to `"public"`

iscai-msft:
fixes https://github.com/Azure/typespec-azure/issues/926

### DIFF
--- a/docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
+++ b/docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
@@ -488,7 +488,7 @@ They each extend from the shared `SdkMethodBase`
 interface SdkMethodBase<TServiceOperation extends SdkServiceOperation> {
   __raw?: Operation;
   name: string;
-  access: AccessFlags | undefined;
+  access: AccessFlags;
   parameters: SdkParameter[];
   apiVersions: string[];
   description?: string;
@@ -1360,7 +1360,7 @@ export interface SdkModelType extends SdkTypeBase {
   generatedName: string;
   description?: string;
   details?: string;
-  access?: AccessFlags;
+  access: AccessFlags;
   usage: UsageFlags;
   additionalProperties?: SdkType;
   discriminatorValue?: string;
@@ -1387,7 +1387,9 @@ export enum UsageFlags {
 
 ### AccessFlags
 
-We have the `.access` property on a model etc. as optional, so emitters can choose what to default to.
+We default the value of `.access` property on model, enum, and method types to be `"public"`. So if the `@access` decorator isn't explicitly applied to one of these definitions, its value will be `"public"`.
+
+If you want to know if a tsp author explicitly set the value with an `@access` decorator, you can call `getAccessOverride`
 
 ```ts
 export type AccessFlags = "internal" | "public";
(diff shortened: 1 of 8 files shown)


docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -1360,7 +1360,7 @@ export interface SdkModelType extends SdkTypeBase {
   generatedName: string;
   description?: string;
   details?: string;
-  access?: AccessFlags;
+  access: AccessFlags;
```
m-nash, line 1363: @archerzz Our goal is to take the types from tcgc and directly serialize them into codemodle.json file.  Right now the decorator representation work is not done so you won't be able to serialize this as is and get the information on the other side when you deserialize unless you synthetically add a property `wasAccessOverriden`.  What are your thoughts on doing that in the short term until the decorator work is complete vs having this pr wait for the decorator work to happen first and then we do this?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -557,7 +557,7 @@ export function getSdkModelWithDiagnostics(
       details: docWrapper.details,
       properties: [],
       additionalProperties: undefined, // going to set additional properties in the next few lines when we look at base model
-      access: undefined, // dummy value since we need to update models map before we can set this
+      access: "public", // dummy value since we need to update models map before we can set this
```
srnagar, line 560: Since this is the default value now, we can change the comment

srnagar (approved):

haolingdong-msft:
Could we consider the anonymous model used in spread parameter in operation case? In this case, the model should be internal as users cannot see it. e.g.
```
  @post
  sendLong(
    ...LongRequest
  ): OkResponse;
```
Here `SendLongRequest` will be anonymous model and the access should be internal as it is not visible to user.

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
archerzz: > @archerzz Our goal is to take the types from tcgc and directly serialize them into codemodle.json file. Right now the decorator representation work is not done so you won't be able to serialize this as is and get the information on the other side when you deserialize unless you synthetically add a property `wasAccessOverriden`. What are your thoughts on doing that in the short term until the decorator work is complete vs having this pr wait for the decorator work to happen first and then we do this?

@m-nash I think we can go with a workaround. We can use `__raw` to check if the `@access` is explicitly set.

iscai-msft:
@haolingdong-msft @tadelesh is handling that in [this](https://github.com/Azure/typespec-azure/pull/886) PR

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
iscai-msft: I think it's fair to use `.__raw` for now, will wait on chat discussion to see what happens

m-nash (approved):

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
m-nash: @archerzz can you link the follow up issue to stop using __raw and start using .decorators once that is in place so we don't lose track of that?

joheredi (approved):

Issue #330, opened 2024-02-28
Update ruleset naming


timotheeguerin:
Problem is we have use `/all` as the only ruleset we had in azure-core and arm. 

However this is confusing as with the addition of cannonical ruleset `all` isn't actually all its more `required`.

Proposing we change to the following:
- `all` (Actually every single rule)
- `required` or `recommended` (Only required rules)
- `<any-optional-ruleset>`

This of course would need some change in the tspconfig validation of the spec repos as well as replacing.

timotheeguerin:
Should be done after this has been rollout #854

timotheeguerin:
https://github.com/Azure/typespec-azure/issues/654 is done, made a pr to automatically generate the `all` ruleset [microsoft/typespec#3462](https://github.com/microsoft/typespec/pull/3462) after this is merged we can remove all the explicit `all` rulesets and that will respect the established rules

PR #927, opened 2024-05-28
Remove `all` rulesets

timotheeguerin:
fix [#330](https://github.com/Azure/typespec-azure/issues/330)

`all` rulesets are automaticlaly provided by the compiler now https://github.com/microsoft/typespec/pull/3462


packages/typespec-azure-resource-manager/src/linter.ts
```diff
@@ -63,25 +63,6 @@ const rules = [
   unsupportedTypeRule,
 ];
 
-const allRulesEnabled = Object.fromEntries(
-  rules.map((rule) => [`@azure-tools/typespec-azure-resource-manager/${rule.name}`, true])
-);
-
 export const $linter = defineLinter({
   rules,
-  ruleSets: {
```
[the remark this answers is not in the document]
timotheeguerin: I was not sure anymore as we have the azure rulesets package. I think this can wait

packages/typespec-azure-core/src/linter.ts
```diff
@@ -4,7 +4,6 @@ import { authRequiredRule } from "./rules/auth-required.js";
 import { byosRule } from "./rules/byos.js";
 import { casingRule } from "./rules/casing.js";
 import { compositionOverInheritanceRule } from "./rules/composition-over-inheritance.js";
-import { friendlyNameRule } from "./rules/friendly-name.js";
```
markcowl, line 7: why remove this one?

markcowl: If we do want to remove it, we should remove the associated diagnostic and detection code as well, shouldn't we?

timotheeguerin: Oh weird It must mean it wasn’t added to the rules array

timotheeguerin: yeah that rule was never enabled. But I thought that we used to fail if you try to enable a rule thats not registered, will have to check that

markcowl (approved):

PR #929, opened 2024-05-28
Adding an overload parameter on ResourceNameParameter to allow `name` to be set to string union type

allenjzhang:
```
      /** Widget resource */
      model Widget is ProxyResource<WidgetProperties> {
         ...ResourceNameParameter<Widget, NameType=WidgetNameType>;
      }

      /** different type of widget used on resource path */
      union WidgetNameType {
        string,
        /** small widget */
        Small: "Small",
        /** large widget */        
        Large: "Large",
```

which will produce:

```
          {
            "name": "widgetName",
            "in": "path",
            "description": "The name of the Widget",
            "required": true,
            "type": "string",
            "enum": [
              "Small",
              "Large"
            ],
            "x-ms-enum": {
              "name": "WidgetNameType",
              "modelAsString": true,
              "values": [
                {
                  "name": "Small",
                  "value": "Small",
                  "description": "small widget"
                },
                {
                  "name": "Large",
                  "value": "Large",
                  "description": "large widget"
                }
              ]
            }
          }
        ],
```


allenjzhang:
The `ResourceNameParameter` is aimed at simplifying the name definition. The original explicit declaration still works. Unfortunately at this point any additional decorator (`minLength`, `maxLength`, `@doc` etc) could not be added to this template. User would either use @@ augment decorator OR explicitly define them.

allenjzhang:
The `ResourceNameParameter` is aimed at simplifying the name definition. The original explicit declaration still works. Unfortunately at this point any additional decorator (`minLength`, `maxLength`, `@doc` etc) could not be added to this template. User would either use @@ augement decorator OR explicitly define them.

packages/typespec-azure-resource-manager/CHANGELOG.md
timotheeguerin, line 1: was this supposed to change?

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -15,18 +15,20 @@ namespace Azure.ResourceManager;
...
  * @template NamePattern The RegEx pattern of the name. Default is `^[a-zA-Z0-9-]{3,24}$`.
+ * @template NameType The type of the name property. Default type is string. However you can pass an union with string values.
  */
 model ResourceNameParameter<
   Resource extends Foundations.Resource,
   KeyName extends valueof string = "",
   SegmentName extends valueof string = "",
-  NamePattern extends valueof string = "^[a-zA-Z0-9-]{3,24}$"
+  NamePattern extends valueof string = "^[a-zA-Z0-9-]{3,24}$",
+  NameType extends string = string
```
timotheeguerin, line 25: should it just be `Type`

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/models.tsp
allenjzhang: ARM mandates the name need to be string.

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -15,18 +15,20 @@ namespace Azure.ResourceManager;
  * @template KeyName Override default key name of the resource.
  * @template SegmentName Override default segment name of the resource.
  * @template NamePattern The RegEx pattern of the name. Default is `^[a-zA-Z0-9-]{3,24}$`.
+ * @template NameType The type of the name property. Default type is string. However you can pass an union with string values. For additional decorators such as @minLength, you can use either augment decorator on `[Resource].name` or passing in a scalar string type with decorators.
```
markcowl, line 18: This gets cust off, consider shortening the description or terminating the line earlier

packages/typespec-azure-resource-manager/test/resource.test.ts
```diff
@@ -652,4 +652,68 @@ describe("typespec-azure-resource-manager: ARM resource model", () => {
...
+      model Widget is ProxyResource<WidgetProperties> {
+         ...ResourceNameParameter<Widget, NameType=WidgetNameType>;
+      }
+
+      @doc("The properties of a widget")
+      model WidgetProperties {
+         size: int32;
+      }
+
+      /** different type of widget used on resource path */
```
markcowl, line 672: I feel like we should add an expected output test in typespec-autorest for each of these scenarios

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/models.tsp
timotheeguerin: I mean `NameType` param should be called `Type`?

timotheeguerin (approved):

markcowl (approved):

PR #930, opened 2024-05-28
[tcgc] expose enums in response headers

iscai-msft:


packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -2462,6 +2462,25 @@ describe("typespec-client-generator-core: types", () => {
...
+        @visibility("read")
+        @header("Repeatability-Result")
+        repeatabilityResult?: "accepted" | "rejected";
+      }
+      op foo(): RepeatableResponse;
+      `);
+      const sdkPackage = runner.context.experimental_sdkPackage;
+      strictEqual(sdkPackage.models.length, 0);
+      strictEqual(sdkPackage.enums.length, 1);
+      strictEqual(sdkPackage.enums[0].name, "FooResponse");
```
iscai-msft, line 2477: @tadelesh what are your thoughts on this generated name?

tadelesh: i think we need to add header name.

tadelesh: committed the code.

jhendrixMSFT (approved):

Issue #849, opened 2024-05-16
[Bug]: TCGC generates incorrect cross-language definition IDs


tjprescott:
### Describe the bug

See: https://apiview.dev/Assemblies/Review/e3195e5ca7954296b4e68393eb084773#ModelClient.generateImages

The cross-language definition ID for this operation should be `ModelClient.getChatCompletions`. However, TCGC is transforming the TypeSpec through it's client decorators and so forth and generating an ID of ` Customizations.Client1.create`. The method to generate the cross-language definition ID should not apply any such customizations.

The whole point of the cross-language definition ID is to be the same no matter the language. If a client.tsp applies different customizations for different languages then there's no way those IDs will match.

### Reproduction

1. Create a simple TypeSpec.
2. Apply TCGC customizations.
3. Generate the mapping file. 

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

tadelesh:
two problems here:
1. i wonder how we could identify whether an operation is from customization and find its source operation? if there a way that we could know a definition is from `client.tsp` file?
2. idw whether if we will add more customization which could make find source impossible. i want to know what the rule of cross-language definition id?

iscai-msft:
1. I'm not sure about that, that's a good point though, @tjprescott any ideas?
2. We want cross language definition id to be the same for the tsp definition, and for all of the language emitters, so we shouldn't have any client customizations included in it

tjprescott:
@iscai-msft is correct, in order for the cross-language ID to be "cross-language" you can't apply language specific customizations. The only source of truth is the TypeSpec, so the cross-langauge ID should always correspond to the TypeSpec ID.

If you can't trace something back to pure TypeSpec then there would be no cross-language ID. So if you had, for example, a decorator that created some kind of helper operation, there's be no cross-language ID for that helper, but it would only exist in Python.

mikeharder:
[...]

PR #936, opened 2024-05-29
[tcgc] update crossLanguageDefinitionId to correspond to apiview

iscai-msft:
fixes https://github.com/Azure/typespec-azure/issues/849


packages/typespec-client-generator-core/test/public-utils.test.ts
```diff
@@ -921,7 +923,7 @@ describe("typespec-client-generator-core: public-utils", () => {
         const models = runner.context.experimental_sdkPackage.models;
         strictEqual(models.length, 1);
         strictEqual(models[0].name, "TestRequest");
-        strictEqual(models[0].crossLanguageDefinitionId, "TestService.TestRequest");
+        strictEqual(models[0].crossLanguageDefinitionId, "TestService.test.Request.anonymous");
```
tjprescott, line 926: I see a lot of `Service.blah.Foo.anonymous` tests in here, what about one for:
```
model Foo {
  props: {
    name: string;
  }
}
```
The cross language id for that name would be `Foo.props.anonymous.name`.

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -636,7 +636,7 @@ describe("typespec-client-generator-core: package", () => {
 
       const fooAccessor = mainClient.methods[0];
       strictEqual(fooAccessor.kind, "clientaccessor");
-      strictEqual(fooAccessor.crossLanguageDefintionId, "TestService.getFoo");
+      strictEqual(fooAccessor.crossLanguageDefintionId, "TestService.Foo");
```
tjprescott, line 639: Annotating what the X-lang IDs from APIView would be in this test case. Do they jive with your changes?
```
namespace Foo {     // TestService.Foo
  interface Bar {        // TestService.Foo.Bar
    @route("/one")   
    one(): void;          // TestService.Foo.Bar.one
  }
}
interface Bar {        // TestService.Bar
  @route("/two")
  two(): void;          // TestService.Bar.two
}
```

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -646,7 +646,7 @@ describe("typespec-client-generator-core: package", () => {
       strictEqual(barAccessor.kind, "clientaccessor");
       strictEqual(barAccessor.access, "internal");
       strictEqual(barAccessor.name, "getBar");
-      strictEqual(barAccessor.crossLanguageDefintionId, "TestService.getBar");
+      strictEqual(barAccessor.crossLanguageDefintionId, "TestService.Bar");
```
tjprescott, line 649: Also, should there be an anonymous model test in this file as well?

tjprescott (commented):
A few comments.

packages/typespec-client-generator-core/test/public-utils.test.ts
iscai-msft: we do have quite a few of those, one example would be line 1541 in the same test file

packages/typespec-client-generator-core/test/package.test.ts
iscai-msft: yep they correspond

packages/typespec-client-generator-core/test/package.test.ts
iscai-msft: the anonymous tests are all in `public-utils.test.ts` and `types.test.ts`, the `package` test file is more to do with namesapce and interfaces / clients

tjprescott (approved):

Issue #931, opened 2024-05-29
[Bug]: visibility discrepancy between Swagger and TypeSpec defined `UserAssignedIdentity` 


XiaofeiCao:
### Describe the bug

TypeSpec:
https://github.com/Azure/typespec-azure/blob/a633ee158fe4da70ebe8f1e0a5a27230701244af/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp#L95-L101

Swagger: 
```json
"UserAssignedIdentity": {
  "type": "object",
  "description": "User assigned identity properties",
  "properties": {
    "principalId": {
      "description": "The principal ID of the assigned identity.",
      "format": "uuid",
      "type": "string",
      "readOnly": true
    },
    "clientId": {
      "description": "The client ID of the assigned identity.",
      "format": "uuid",
      "type": "string",
      "readOnly": true
    }
  }
}
```

The model is used in both input and output. 
For input, user sends something like:
```json
{
  "userAssignedIdentities": {
    "00000000-0000-0000-0000-000000000000": {}
  }
}
```
For output, service will reply:
```json
{
  "userAssignedIdentities": {
    "00000000-0000-0000-0000-000000000000": {
      "principalId": <some-principal-id>,
      "tenantId": <some-tenant-id>
    }
  }
}

```

If not defined with  `@visibility("read")`, unnecessary setter methods will be generated for Java. 

### Reproduction

See above

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

allenjzhang:
fixed in pending PR #945

PR #945, opened 2024-06-01
ARM Library reorganization

allenjzhang:
- Created fully versioned Azure.ResourceManager.CommonTypes subnamespace
- Moving all common type definitions/parameters into corresponding tsp files for later easy swagger generation
- Adding back compat layers so all existing entities moved from Foundations continue to work (via `alias`)
- Added default logic to add `@useDependency` to CommonType reference via `@armProviderNamespace` and `@armCommonTypeVersion` decorators to prevent breaking change.


pshao25:
I think we should change the swagger emitter from "read metadata from armCommonTypeVersion" to "consume data from compiler output"?

packages/typespec-azure-resource-manager/test/rules/arm-common-types-version.test.ts
```diff
@@ -20,42 +20,6 @@ describe("typespec-azure-resource-manager: ARM common-types version rule", () =>
     );
   });
 
-  it("emits diagnostic when a version in the enum is missing a common type version", async () => {
```
timotheeguerin, line 23: what does this do now?

packages/typespec-azure-resource-manager/test/rules/retry-after.test.ts
```diff
@@ -24,27 +24,40 @@ describe("typespec-azure-resource-manager: retry-after rule", () => {
     await tester
       .expect(
         `
-        @useDependency(Azure.Core.Versions.v1_0_Preview_2)
-        @useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
+        @service({
+          title: "Microsoft.Foo",
+        })
```
timotheeguerin, line 29:
```suggestion
        @service
```

packages/typespec-azure-resource-manager/README.md
```diff
@@ -94,7 +94,7 @@ This decorator is used either on a namespace or a version enum value to indicate
 the version of the Azure Resource Manager common-types to use for refs in emitted Swagger files.
 
 ```typespec
-@Azure.ResourceManager.armCommonTypesVersion(version: valueof string | EnumMember)
+@Azure.ResourceManager.armCommonTypesVersion(version: EnumMember)
```
timotheeguerin, line 97: isn't that a breaking change?

packages/typespec-azure-resource-manager/lib/backcompat.tsp
```diff
@@ -0,0 +1,20 @@
+using Azure.ResourceManager.CommonTypes;
+
+namespace Azure.ResourceManager {
```
timotheeguerin, line 3:
```suggestion
namespace Azure.ResourceManager;
```

file namespace?

packages/typespec-azure-resource-manager/lib/foundations/backcompat.tsp
```diff
@@ -0,0 +1,40 @@
+namespace Azure.ResourceManager.Foundations {
+  // extended-location
+  alias ExtendedLocation = CommonTypes.ExtendedLocation;
+  alias ExtendedLocationType = CommonTypes.ExtendedLocationType;
+
+  // managed-identities
+  alias ManagedServiceIdentity = CommonTypes.ManagedServiceIdentity;
+  #suppress "deprecated" "Need for back compatibility in common-types"
```
markcowl, line 8: Suppression should have the user action, like use the ManagedServiceIdentityProperty mix-in instea

packages/typespec-azure-resource-manager/lib/common-types/extended-location-ref.tsp
```diff
@@ -0,0 +1 @@
+import "./extended-location.tsp";
```
markcowl, line 1: Why is this here?

packages/typespec-azure-resource-manager/lib/common-types/managed-identity-ref.tsp
```diff
@@ -0,0 +1,59 @@
+import "./managed-identity.tsp";
+
+using Azure.ResourceManager.CommonTypes.Private;
+
+namespace Azure.ResourceManager.CommonTypes;
+
+/**
+ * The properties of the managed service identities assigned to this resource.
+ */
+@@armCommonDefinition(ManagedServiceIdentity,
```
markcowl, line 10: I thought we decided not to separate the `@armCommonDefinition` and `@armCommonParameter` decorators into a separate file?

packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
```diff
@@ -1,22 +1,10 @@
 using Azure.ResourceManager.Private;
 
-namespace Azure.ResourceManager.Foundations;
+namespace Azure.ResourceManager.CommonTypes;
```
markcowl, line 3: for 'ManagedServiceIdentityType`, how are we handling the 'space' issue in `systemAssignedIdentity, userAssignedIdentity` - where i initially had a space and is now required to have no space?

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -80,6 +80,40 @@ model ExtensionResource<Properties extends {}> extends Foundations.ExtensionReso
...
 
+//#region Private Link models
+model PrivateLinkResource extends Foundations.ProxyResource {
+  /** Properties of the private link resource. */
+  properties?: PrivateLinkResourceProperties;
+}
+
+/** Properties of a private link resource. */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Matches current common code"
+model PrivateLinkResourceProperties {
```
markcowl, line 91: Why are these here, instead of in the common-types namespace?

packages/typespec-azure-resource-manager/scripts/common-types.tsp
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Supported versions of Azure.ResourceManager building blocks.
+ */
+@doc("Supported versions of Azure.ResourceManager building blocks.")
+enum Versions {
+  @doc("ARM common types v3")
+  v3,
+
+  @doc("ARM common types v3")
```
markcowl, line 24:
```suggestion
  @doc("ARM common types v4")
```

packages/typespec-azure-resource-manager/scripts/common-types.tsp
```diff
@@ -0,0 +1,29 @@
...
+  v4,
+
+  @doc("ARM common types v3")
```
markcowl, line 27:
```suggestion
  @doc("ARM common types v5")
```

packages/typespec-azure-resource-manager/src/common-types.ts
```diff
@@ -56,11 +55,19 @@ export function isArmCommonType(entity: Type): boolean {
 export function $armCommonTypesVersion(
   context: DecoratorContext,
   entity: Namespace | EnumMember,
-  version: string | EnumValue
+  version: EnumMember
```
markcowl, line 58: Can't we look up the appropriate enum value from the string?

packages/typespec-azure-resource-manager/src/common-types.ts
```diff
@@ -56,11 +55,19 @@ export function isArmCommonType(entity: Type): boolean {
...
-  context.program
-    .stateMap(ArmStateKeys.armCommonTypesVersion)
-    .set(entity, typeof version === "string" ? version : version.value.name);
+  context.program.stateMap(ArmStateKeys.armCommonTypesVersion).set(entity, version.name as string);
+
+  if (entity.kind === "Namespace") {
+    const versioned = entity.decorators.find((x) => x.definition?.name === "@versioned");
+    // If it is versioned namespace, we will skip adding @useDependency to namespace
+    if (versioned) {
+      return;
```
markcowl, line 66: This doesn't work in all cases, for example: when I add common-types version to a versioned spec, I can add it once to the namespace for all versions [like this](https://azure.github.io/typespec-azure/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3yvIK3Jlc291cmNlLW1hbmFnZXIiOwoKdXNpbmcgVHlwZVNwZWMuSHR0cDvQFVJlc3TRFVbpAKnIG0HEZS5Db3JlzhJSx3dNxnY7CgovKiogQ29udG9zbyDIHiBQcm92aWRlciDmAJ5tZW50IEFQSS4gKi8KQGFybcggTmFtZXNwYWNlCkBzZXJ2aWNlKHsKICB0aXRsZTogIsdYyC5IdWJDbGllbnQiLAp9KQpA5wFgZWQo5wDBcynFXkNvbW1vbuQA3nPHGSj1AL4uyynoAQRzLnY1KQpu6ACYIE1pY3Jvc29mdC7yAI3vAP5BUEkg5wCTc%2BQA6WVudW0gyFcg5ADYxC4yMDIxLTEwLTAxLXByZXZpZXfINcQ0ICBAdXNlRGVwZW5kZW5jefcAuchWLnYxXzBfUMZIXzEpCiAgYNJhYCwK333ffd99zn0y0X195gEoQegBKusBQyDoAsjkAIBtb2RlbCBFbXBsb3llZSBpcyBUcmFja2Vk6ACCPMgcUHJvcGVydGllcz7lAVkuLukApuQCU1BhcmFtZXRlcskxPjvoAIbJX3DJRNJ8ymDpAbhBZ2Ugb2YgZcg%2F5QEwYWdlPzogaW50MzI76AFkQ2l0edIqY2l0eT86IHN0cuUDZscsUHJvZmls01lAZW5jb2RlKCJiYXNlNjR1cmwi5AFgcMYwPzogYnl0ZXPJSFRoZSBzdGF0dXPES3RoZSBsYXN0IOQAxWF0aW9u5QN5ICBAdmlzaWJpbGl0eSgicmVhZMhVxBjlBABTdGF0xF%2FlAaPMFOkBQMRrzDLFeOUAwmHpAc3Fb0Bscm%2FEO3VzCnVu5AJi0VTlAVzmARLpAqLEX8hHIGNyZcQncmVxdWVzdCBoYXMgYmVlbiBhY2NlcHRlZMRnICBBxw46ICLICyLWUGnEQOQAtOkAwchE7ACcOiAizA%2FaTHVwZGF0xE%2FFQ1XHDjogIsgLyjvpBRrpAMTmANxk5wGaU3VjY2VlZOUAxckM0z%2FFNuQBTWZhaWzJPkbFDTogIsYJ3Dh3YXMgY2FuY2XKPkPHD%2BQFcccL%2FwFAIGRlbGXpAYBExA3mAPnIC%2BQFoe8Db21vduoByukDcU1vdmVSxxXoA2rEc21vdsRoZnJvbSBsb2PmALzFbsQT8QNIyzN0b88xdG%2FKL%2FcAlXNwb25z6wSD5gCWxxbsAJfuA2TFPsVkxnzmAu7NbmludGVyZuQGSU%2FoA4xzIGV4dGVuZHP2B1Muyyl7fQrlBtHII8oby1noAM3mBKtnZXTkAadByjVSZWFk7AUIIOcCe09y5QKn5QHWyy9Dxx1SZXBsYWNlQXN5bmPOP%2BUC99A3UGF0Y2hTzCws8wVOxUDmAj3PQOUCOmVXaXRob3V0T2vTd2xpc3RCecgwR3JvdXDPREzFIlBhcmVudNQ8U3Vic2NyaXDlAhzGO8YzzBnMOegF%2ByBzYW1wbOsDCWFjxUR0aGF05gIG6QW%2BdG8gZGlmZmXkAITvAonFKe4AskHFSO8BN%2BsDFMgN5gKL8wCSSEVBROoF58R%2BY2hlY2vqAKpleGlzdGVu5gdhIMYeRckU7wH7zR3uBzw%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fresource-manager%22%5D%7D%7D)

packages/typespec-azure-resource-manager/src/commontypes.private.decorators.ts
```diff
@@ -0,0 +1,135 @@
...
+  DecoratorContext,
+  Enum,
+  EnumValue,
+  Model,
+  ModelProperty,
+  Program,
+} from "@typespec/compiler";
+import { ArmStateKeys } from "./state.js";
+
+export const namespace = "Azure.ResourceManager.CommonTypes.Private";
```
markcowl, line 11: I don't think we can remove the existing decorators from the existing namespace, this has been used by folks to create common-types that we did not yet have coverage for

markcowl (changes requested):

packages/typespec-azure-resource-manager/test/rules/arm-common-types-version.test.ts
allenjzhang: This is rule is now obsolete as we will add default versions.

packages/typespec-azure-resource-manager/README.md
allenjzhang: Yes. There were handful of using "v5" string. But let me add it back and do a look up.

packages/typespec-azure-resource-manager/lib/common-types/extended-location-ref.tsp
allenjzhang: I just left the `xxx.tsp` and `xxx-ref.tsp` pattern. Once we start generating the common types swagger, we will need to add decorator here.

packages/typespec-azure-resource-manager/lib/common-types/managed-identity-ref.tsp
allenjzhang: I think we need to separate them as not to significantly change the autorest for common-types swagger generation FOR NOW, otherwise $ref will be generated.

packages/typespec-azure-resource-manager/test/rules/arm-common-types-version.test.ts
timotheeguerin: So shouldn't the whole rule be removed? And isn't the point of this rule is we didn't want to use default?

allenjzhang: > And isn't the point of this rule is we didn't want to use default?
My implementation has geared towards no breaking changes. But IF this is indeed a strong goal we have, I can adjust the logic to call @useDependency directly for missing declarations.

packages/typespec-azure-resource-manager/test/rules/retry-after.test.ts
allenjzhang: Just curious, does this matter?

packages/typespec-azure-resource-manager/scripts/common-types.tsp
allenjzhang: Deleting it for now.

packages/typespec-azure-resource-manager/src/common-types.ts
allenjzhang: Logic has been updated.

from @armProviderNamespace:
    - look up current @armCommonTypeVersion on namespace if any
    - Versioned, loop thru each service version enum member
    -       if no @armCommonTypeVersion, add and set version to either the one from step above or default to v3
    - Unversioned, add and set version to either the one from step above or default to v3 to namespace if one doesn't exist

XiaofeiCao:
Would you also help address the visibility discrepancy with swagger too? Thanks!
- https://github.com/Azure/typespec-azure/issues/931

packages/typespec-azure-resource-manager/lib/foundations/backcompat.tsp
allenjzhang: Removed. The CommonType.UserAssignedIdentities has #deprecate which will surface up on the alias.

packages/typespec-azure-resource-manager/test/rules/arm-common-types-version.test.ts
allenjzhang: You feedback is taken, I have put the rule back. It checks the existence of decorator rather than stateMap values.

packages/typespec-azure-resource-manager/src/commontypes.private.decorators.ts
allenjzhang: I am little concern about same name under different namespaces. Did a search, none was found in public main and RPSaaSMaster

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -281,172 +243,339 @@ union Origin {
...
+  @segment("resourceGroups")
+  @pattern("^[-\\w\\._\\(\\)]+$")
+  resourceGroupName: string;
+}
+
+/**
+ * The default subscriptionId parameter type.
+ */
+@doc("The default subscriptionId parameter type.")
+model SubscriptionIdParameter {
```
pshao25, line 508: The description also changed from v4. For the versions before v4, it is "The ID of the target subscription." without "The value must be an UUID."

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -281,172 +243,339 @@ union Origin {
...
+  @typeChangedFrom(Versions.v4, string)
+  @path
+  @segment("subscriptions")
+  @doc("The ID of the target subscription.  The value must be an UUID.")
```
pshao25, line 512: And you have an extra space between these two sentences.

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
allenjzhang: The description emitted is the property @doc. Unfortunately, there is no easy way to version the decorators unless you add and remove the properties themselves. I didn't think it is critical here.

packages/samples/specs/resource-manager/resource-common-properties/common-properties/main.tsp
```diff
@@ -0,0 +1,64 @@
...
+  ...EntityTagProperty;
+  ...ResourceKindProperty;
+  ...ResourcePlanProperty;
+  ...ResourceSkuProperty;
+}
+
+@armResourceOperations
+interface Widgets {
+  get is ArmResourceRead<WidgetResource>;
+  create is ArmResourceCreateOrUpdateAsync<WidgetResource>;
```
markcowl, line 59:
```suggestion
  create is ArmResourceCreateOrReplaceAsync<WidgetResource>;
```

packages/samples/specs/resource-manager/resource-common-properties/managed-identity/main.tsp
```diff
@@ -0,0 +1,58 @@
...
+using TypeSpec.Versioning;
+using Azure.ResourceManager;
+using Azure.Core;
+
+@armProviderNamespace("Contoso.ManagedIdentityExample")
+@service({
+  title: "Contoso.ManagedIdentityExample",
+})
+@OpenAPI.info({
+  version: "2023-03-01-preview",
```
markcowl, line 18: We should not follow this pattern in examples, we should have a Version enum with a single version and decorate the version with `useDependency` and `armCommonTypesVersion`

packages/typespec-azure-resource-manager/lib/common-types/private-links.tsp
```diff
@@ -82,20 +120,9 @@ union PrivateEndpointServiceConnectionStatus {
  * The name of the private endpoint connection associated with the Azure resource.
  * @template Segment The resource type name for private endpoint connections (default is privateEndpointConnections)
  */
-@added(CommonTypes.Versions.v4)
```
markcowl: Do these exist in v3?  Why are we removing these decorators?

packages/typespec-azure-resource-manager/lib/common-types/types-ref.tsp
```diff
@@ -233,6 +295,17 @@ namespace Azure.ResourceManager.CommonTypes;
   Azure.ResourceManager.CommonTypes.Versions.v5
 );
 
+@@resourceParameterBaseFor(ApiVersionParameter.apiVersion,
```
markcowl, line 298: This should not impact autorest serialization, why move this here?

packages/typespec-azure-resource-manager/lib/common-types/types-ref.tsp
```diff
@@ -245,3 +318,97 @@ namespace Azure.ResourceManager.CommonTypes;
   "SubscriptionIdParameter",
   Azure.ResourceManager.CommonTypes.Versions.v5
 );
+@@resourceParameterBaseFor(SubscriptionIdParameter.subscriptionId,
```
markcowl, line 321: same comment - moving this seems unnecessary

packages/typespec-azure-resource-manager/lib/foundations/arm.foundations.tsp
```diff
@@ -44,6 +44,17 @@ model DefaultBaseParameters<Resource extends {}> {
   ...ResourceUriParameter;
 }
 
+/**
+ * Standard type definition for Azure Resource Manager Tags property.
+ *
+ * It is included in the TrackedResource template definition.
+ */
+/** The Azure Resource Manager Resource tags. */
+model ArmTagsProperty {
```
markcowl, line 53: Isn't this a breaking change?  People do add tags to their proxy resources.  Seems like we should have a linting rule for this rather than move the model

packages/typespec-azure-resource-manager/src/lib.ts
```diff
@@ -76,6 +76,12 @@ export const $lib = createTypeSpecLibrary({
         default: paramMessage`No ARM common-types version for this type satisfies the expected version ${"selectedVersion"}.  This type only supports the following version(s): ${"supportedVersions"}`,
       },
     },
+    "arm-common-types-invalid-version": {
+      severity: "error",
+      messages: {
+        default: paramMessage`No ARM common-types version match the version string ${"versionString"}.  The following versions are supported: ${"supportedVersions"}`,
```
markcowl, line 82:
```suggestion
        default: paramMessage`No ARM common-types version matches the version string ${"versionString"}.  The following versions are supported: ${"supportedVersions"}`,
```

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -420,6 +416,40 @@ union CheckNameAvailabilityReason {
   string,
 }
 
+union EncryptionStatus {
```
markcowl, line 419: Needs a description

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -428,29 +458,118 @@ model ApiVersionParameter {
   @query("api-version")
   @minLength(1)
   @doc("The API version to use for this operation.")
-  @resourceParameterBaseFor(
```
markcowl: No need to relocate this, as noted

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/common-types/types-ref.tsp
allenjzhang: I am moving these here to keep the type.tsp pure from decorations as later want to transfer ownership to ARM team.

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
allenjzhang: see above

markcowl (commented):
It looks like the failures are due to a conflict, and just rebuilding and regenerating samples

packages/typespec-azure-resource-manager/lib/foundations/arm.foundations.tsp
allenjzhang: No. ArmTagsProperty was originally in `types.tsp` under common-types folder BUT the namespace was `Foundations`. As I moved everything in types.tsp to under `CommonTypes` namespace, moving this back into foundations.tsp will keep it unchanged. In fact, we may want to consider moving/aliasing it into ARM namespace so that you can directly spread it and limit dependency on Foundations namespace.

packages/typespec-azure-resource-manager/lib/common-types/private-links.tsp
allenjzhang: Private link existed in v3

markcowl (approved):

PR #950, opened 2024-06-04
remove duplicated types in TCGC

ArcturusZhang:


tadelesh (approved):
please add a changelog

PR #955, opened 2024-06-05
Add support for ARM lro options in typespec-autorest

markcowl:


packages/samples/test/output/azure/core/data-plane/api-path-parameter/@azure-tools/typespec-autorest/2022-08-31/openapi.json
```diff
@@ -382,9 +382,7 @@
         },
         "x-ms-long-running-operation-options": {
           "final-state-via": "operation-location",
-          "final-state-schema": {
-            "$ref": "#/definitions/Manufacturer"
-          }
+          "final-state-schema": "#/definitions/Manufacturer"
```
timotheeguerin, line 385: is that change expected?

.chronus/changes/arm-lro-opt-2024-5-5-2-57-50.md
```diff
@@ -0,0 +1,12 @@
...
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-azure-core"
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+- Add support for displaying lro options in typespec-autorest based on lro metadata
+- Add overrides for lro resolution when multiple resolution pathways exist
+- Remove OpenAPI dependencies form ARM LRO templates
```
timotheeguerin, line 12: Ideally this should be creatred as individual changelog files that match their corresponding package, otherwise here all 3 lines will show up in all 3 packages changelog

packages/samples/test/output/azure/core/data-plane/api-path-parameter/@azure-tools/typespec-autorest/2022-08-31/openapi.json
timotheeguerin: hhm, when I commented this was changing from `{$ref: "#/definitions/Manufacturer"}` to `"#/definitions/Manufacturer"`, guessing this was just a commit within this pr that changed?

markcowl: Not expected, but accurate, I think.  Will double-check and update the delete ops to use the right StatusMonitor schema

docs/emitters/typespec-autorest/reference/emitter.md
```diff
@@ -119,3 +119,9 @@ This extension is meant for debugging and should not be depended on.
 **Type:** `boolean`
 
 Create read-only property schema for lro status
+
+### `suppress-lro-options`
```
timotheeguerin, line 123:
```suggestion
### `omit-lro-options`
```
would that name be better and more inline with some other options?

packages/samples/specs/data-plane/api-path-parameter/main.tsp
```diff
@@ -98,7 +101,11 @@ interface Widgets {
 
 interface Manufacturers {
   @doc("Gets status of a Manufacturer operation.")
-  getManufacturerOperationStatus is Operations.GetResourceOperationStatus<Manufacturer>;
+  @sharedRoute
+  getManufacturerDeleteOperationStatus is Operations.GetResourceOperationStatus<
```
timotheeguerin, line 105: who is it sharing route with now?

docs/emitters/typespec-autorest/reference/emitter.md
markcowl: I am cool with that

packages/samples/specs/data-plane/api-path-parameter/main.tsp
markcowl: doesn't need this.

allenjzhang (approved):

packages/samples/specs/data-plane/widget-manager/main.tsp
```diff
@@ -173,7 +173,11 @@ alias Operations = Azure.Core.ResourceOperations<ServiceTraits>;
 interface Widgets {
   // Operation Status
   @doc("Gets status of a Widget operation.")
+  @sharedRoute
```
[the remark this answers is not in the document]
markcowl: This is one way you can think about having a central StatusMonitor that services multiple different operations, with different return values.  Another would be to have a union of or polymorpic result types.  On services, I suspect there is a single endpoint that serves these multiple Status monitors with different result fields, but the operation implementation code is, of course, different for each.

PR #962, opened 2024-06-06
[tcgc] refine logic of core model filtering

tadelesh:
previous logic has two problems:
1. it could not handle any models found in model properties, union variants and so on
2. it will include some of the model used in templated model but not used in real operation

new logic:
1. do not do any filtering when traversal
2. do the filtering at final stage of `getAllModels`
3. set usage to `None` to filter out model not used


iscai-msft (approved):
looks great, thanks @tadelesh

Issue #829, opened 2024-05-14
Provide a generic dictionary of decorators in TCGC for better extensibility


live1206:
### Clear and concise description of the problem

Nowadays when we add a new decorator, TCGC needs to process it and add a new property to SDK model.

We would like to have a generic dictionary to store decorators for emitter to consume, so that we don't need to add a new property to SDK model for better extensibility.
```
decorators
{
  "decoratorName":  string,   // The name should include namespace
  "decoratorArguments": string[]
}
```

Issues to address:
- how to filter which decorators to include in the generic dictionary
- how to present the decorator arguments/values in the dictionary, such as JSON serialization, need to analyze the existing decorators

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #704, opened 2024-04-19
XML support TCGC Implementation


lmazuel:
Make sure TCGC makes available the XML metadata

https://typespec.io/docs/next/libraries/xml/reference

Package is `@typespec/xml`

PR #966, opened 2024-06-06
[tcgc] support generic type decorators in sdk types

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/829, #704 

current design:
1. use whitelist configured in `tspconfig.yaml` to do decorators filtering, support equal/regex comparison
2. use `jsValue` from typespec decorator arguments to get arguments' value
3. only handle arguments value of typing with number/string/boolean/null/tsp enum member value/tsp literal value, other types will add a diagnostic warning


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -257,9 +260,69 @@ export function getSdkTypeBaseHelper<TKind>(
...
+    for (const decorator of type.decorators) {
+      // only process explicitly defined decorators
+      if (decorator.definition) {
+        const decoratorName = `${decorator.definition?.namespace ? getNamespaceFullName(decorator.definition.namespace) + "." : ""}${decorator.definition?.name}`;
+        // white list filtering
+        if (!context.decoratorsWhiteList) {
+          continue;
+        }
+        let found = false;
+        for (const item of context.decoratorsWhiteList) {
```
iscai-msft, line 279: i think it would be better to use `context.decoratorsWhiteList.filter(x =>x === decoratorName || new RegExp(x).test(decoratorName))`, or something like that, i haven't tested it out

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -257,9 +260,69 @@ export function getSdkTypeBaseHelper<TKind>(
...
+    | Record<string, unknown>
+    | Value
+    | unknown[]
+    | string
+    | number
+    | boolean
+    | Numeric
+    | null
+): any {
+  if (typeof arg === "object" && arg !== null && "kind" in arg) {
```
iscai-msft, line 315: I think it would be simpler to type `arg` as a type that includes `{kind: string} | undefined

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -656,7 +656,7 @@ describe("typespec-client-generator-core: package", () => {
 
       const fooBarAccessor = fooClient.methods[0];
       strictEqual(fooBarAccessor.kind, "clientaccessor");
-      strictEqual(fooBarAccessor.crossLanguageDefintionId, "TestService.Foo.Bar");
+      strictEqual(fooBarAccessor.crossLanguageDefintionId, "TestService.Foo.getBar");
```
iscai-msft, line 659: why is the cross language definition id changing here?

iscai-msft (commented):
overall looks great, thanks for your hard work on this!

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: i don't know whether this kind of judgment or use `as` to case to a type to compare kink is better?

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: they should come from some local cache problem, i rollbacked the changes.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -257,9 +260,60 @@ export function getSdkTypeBaseHelper<TKind>(
...
+    if (arg.kind === "EnumMember") {
+      return arg.value ?? arg.name;
+    }
+    if (arg.kind === "String" || arg.kind === "Number" || arg.kind === "Boolean") {
+      return arg.value;
+    }
+  }
+  return arg;
```
live1206, line 314: If `arg` is not among the types handling above, we just pass along it to the emitter?
What if `arg` needs some specific complex logic to process it? Does it mean this decorator should not be included in the white list?

tadelesh: For those types, we do not have a common way to deal with, so just leave it to emitter.
Yes, I believe those decorators should not belong to the usage of generic decorators.

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -3824,6 +3824,69 @@ describe("typespec-client-generator-core: types", () => {
       strictEqual(test.type.values[1].kind, "string");
     });
   });
+  describe("general decorators list", () => {
```
live1206, line 3827: some tests for the xml decorators

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -257,9 +260,60 @@ export function getSdkTypeBaseHelper<TKind>(
     __raw: type,
     deprecation: getDeprecationDetails(context.program, type)?.message,
     kind,
+    decorators: getTypeDecorators(context, type),
   };
 }
 
+export function getTypeDecorators(context: TCGCContext, type: Type): Record<string, Array<any>> {
```
live1206, line 267: should this be `Record<string, Record<string, any>>`?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -77,7 +79,11 @@ export interface SdkOperationGroup {
   service: Namespace;
 }
 
-interface SdkTypeBase {
+interface DecoratedType {
```
live1206, line 82: is type a valid use case?

live1206: we can always add things, should start with smaller scope we are sure of

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -77,7 +79,11 @@ export interface SdkOperationGroup {
...
+  decorators: Record<string, Array<any>>;
+}
+
+interface SdkTypeBase extends DecoratedType {
```
live1206, line 86: is inheritance a good practice here? could this be another property?

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: this is somehow same pattern with typespec core. since lots of type will have decorators, i added it as a separated interface.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: same with previous reply.

tadelesh:
@iscai-msft @live1206 @m-nash could you review again?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -252,12 +255,85 @@ export function getSdkTypeBaseHelper<TKind>(
...
+  context: TCGCContext,
+  type: Type
+): [Record<string, Record<string, any>>, readonly Diagnostic[]] {
+  const diagnostics = createDiagnosticCollector();
+  const retval: Record<string, Record<string, any>> = {};
+  if ("decorators" in type) {
+    for (const decorator of type.decorators) {
+      // only process explicitly defined decorators
+      if (decorator.definition) {
+        const decoratorName = `${decorator.definition?.namespace ? getNamespaceFullName(decorator.definition.namespace) + "." : ""}${decorator.definition?.name}`;
```
ArcturusZhang, line 278: I saw 
```
something ? getNamespaceFullName(something) : ""
```
usages in multiple places, could we just simple have a helper for this?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -252,12 +255,85 @@ export function getSdkTypeBaseHelper<TKind>(
...
+        // white list filtering
+        if (
+          !context.decoratorsWhiteList ||
+          !context.decoratorsWhiteList.some(
+            (x) => x === decoratorName || new RegExp(x).test(decoratorName)
```
ArcturusZhang, line 283: so with this, if I want to specify one certain decorator `foo` in the configuration, I will have to write this:
```
^foo$
```
because `foo` might match other names such as `barfoobaz`
Would this happen? And is this expected?

ArcturusZhang: Also based on your test cases, and since dot `.` is quite common in the name of decorators, if I write this: `TypeSpec.@bar`, would it leads to a regex that matches anything starting with `TypeSpec` and ending with `@bar`? This might match all the decorators with the name `@bar` in all the sibling namespaces under `TypeSpec`, also `TypeSpec2.@bar` maybe
if this is true, this really introduces quite a lot uncertainties

tadelesh: you are right. the two comparisons have overlap. let me change to only allow regex.

haolingdong-msft:
Would like to learn more about the whitelist configuration in `tspconfig.yaml`, should every service team provide the whitelist when writing `tspconfig.yaml`? Will the list be very long and have many duplicate decorators across different service's `tspconfig.yaml`? Is there a default whitelist that covers common decorators, and service team can add additional special decorators if it does not exist.

packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -3824,6 +3824,211 @@ describe("typespec-client-generator-core: types", () => {
...
+      const models = runner.context.experimental_sdkPackage.models;
+      strictEqual(models.length, 1);
+      deepStrictEqual(models[0].decorators["TypeSpec.Xml.@name"], { name: "XmlBook" });
+      deepStrictEqual(models[0].properties[0].decorators["TypeSpec.Xml.@name"], { name: "XmlId" });
+    });
+
+    it("@ns", async function () {
+      runner = await createSdkTestRunner({
+        emitterName: "@azure-tools/typespec-python",
+        "decorators-white-list": ["TypeSpec\\.Xml\\..*"],
```
haolingdong-msft, line 3940: `"TypeSpec\\.Xml\\..*"` -> Are these files contain xml related decorators? I don't see it in this pr.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -252,12 +255,85 @@ export function getSdkTypeBaseHelper<TKind>(
...
+          )
+        ) {
+          continue;
+        }
+
+        retval[decoratorName] = {};
+        for (let i = 0; i < decorator.definition.parameters.length; i++) {
+          if (i >= decorator.args.length) {
+            retval[decoratorName][decorator.definition.parameters[i].name] = undefined;
```
live1206, line 292: Is it possible to skip nullable parameter by using named parameter in typespec decorator?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -252,12 +255,85 @@ export function getSdkTypeBaseHelper<TKind>(
...
+          !context.decoratorsWhiteList ||
+          !context.decoratorsWhiteList.some(
+            (x) => x === decoratorName || new RegExp(x).test(decoratorName)
+          )
+        ) {
+          continue;
+        }
+
+        retval[decoratorName] = {};
+        for (let i = 0; i < decorator.definition.parameters.length; i++) {
```
live1206, line 290: consider to wrap this into a helper method getDecoratorWithArgs for readability

haolingdong-msft: Guess `for (const param of decorator.definition.parameters)` could be better.

tadelesh:
> Would like to learn more about the whitelist configuration in `tspconfig.yaml`, should every service team provide the whitelist when writing `tspconfig.yaml`? Will the list be very long and have many duplicate decorators across different service's `tspconfig.yaml`? Is there a default whitelist that covers common decorators, and service team can add additional special decorators if it does not exist.

tcgc left it to language's emitter. when you call `createSdkContext`, you could add default to `decorators-white-list` config.

haolingdong-msft:
> > Would like to learn more about the whitelist configuration in `tspconfig.yaml`, should every service team provide the whitelist when writing `tspconfig.yaml`? Will the list be very long and have many duplicate decorators across different service's `tspconfig.yaml`? Is there a default whitelist that covers common decorators, and service team can add additional special decorators if it does not exist.
> 
> tcgc left it to language's emitter. when you call `createSdkContext`, you could add default to `decorators-white-list` config.

Emm. Shall TCGC have a default whitelist shared by each language as well?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -242,14 +243,25 @@ export function getSdkBuiltInType(
 }
 
 export function getSdkDurationType(context: TCGCContext, type: Scalar): SdkDurationType {
+  return ignoreDiagnostics(getSdkDurationTypeWithDiagnostics(context, type));
+}
+
+function getSdkDurationTypeWithDiagnostics(
```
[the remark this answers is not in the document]
tadelesh: this is for backward compatibility. tcgc export function without diagnostics before. i don't want to break them. so, for any exported function, there will be a new with diagnostic function.

packages/typespec-client-generator-core/test/types.test.ts
tadelesh: all xml decorators are under `TypeSpec.Xml`

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: no

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: i need index to do cut.

tadelesh: since this function is to get type decorators, i think such short part does need to be abstracted.

haolingdong-msft: I see

tadelesh:
> > > Would like to learn more about the whitelist configuration in `tspconfig.yaml`, should every service team provide the whitelist when writing `tspconfig.yaml`? Will the list be very long and have many duplicate decorators across different service's `tspconfig.yaml`? Is there a default whitelist that covers common decorators, and service team can add additional special decorators if it does not exist.
> > 
> > 
> > tcgc left it to language's emitter. when you call `createSdkContext`, you could add default to `decorators-white-list` config.
> 
> Emm. Shall TCGC have a default whitelist shared by each language as well?

we do not have a clear whitelist for now, so leave it for empty.

packages/typespec-client-generator-core/test/types.test.ts
haolingdong-msft: ok, so this is the way to include all the decorators under `TypeSpec.Xml`? Why there are two escape character `\` (previously I thought it is folder)?

tadelesh: > ok, so this is the way to include all the decorators under `TypeSpec.Xml`? Why there are two escape character `\` (previously I thought it is folder)?

ts escape for `\`

packages/typespec-client-generator-core/src/internal-utils.ts
live1206: Then, can we skip the undefined arguments?
We can direct loop through `decorator.args` instead of `decorator.definition.parameters`, and there is pointless to store the undefined arguments.

tadelesh: i think keep all the decorator info is much better, what do you think?

live1206: Not really, the arguments are not defined in typespec anyway, they will not be used in emitter.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -36,6 +36,7 @@ export interface SdkEmitterOptions {
   "package-name"?: string;
   "flatten-union-as-enum"?: boolean;
   "api-version"?: string;
+  "decorators-white-list"?: string[];
```
m-nash, line 39: If I understand correctly it feels like we are bleeding internal implementation into something the service team needs to understand.  I don't think we want customers to need to know what decorators to choose to include and which ones not to.  This feels like a handshake between TCGC and each emitter and the service team should not be involved?  What if the generator for a given langauge needs a decorator added and the service team doesn't add it in?

tadelesh: though emitter could change it when `createSdkContext`, but i agree it should not be exposed to service. i plan to add it as an option of `createSdkContext`, is that good?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -571,8 +571,11 @@ export function listOperationsInOperationGroup(
   return operations;
 }
 
-interface CreateSdkContextOptions {
+const defaultDecoratorsWhiteList = ["TypeSpec\\.Xml\\..*"];
```
haolingdong-msft, line 574: I would prefer to put the default whitelist into a seperate config file to be easier to find and update. What do others think? I'm open to other opinions.

live1206 (approved):

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -607,6 +610,7 @@ export function createSdkContext<
     __namespaceToApiVersionParameter: new Map(),
     __tspTypeToApiVersions: new Map(),
     __namespaceToApiVersionClientDefaultValue: new Map(),
+    decoratorsWhiteList: options?.decoratorsWhiteList ?? defaultDecoratorsWhiteList,
```
m-nash, line 613: I think the config coming from sdk should be in addition to, not a replacement of.  We could call it something that indicates this like `additionalDecorators`

packages/typespec-client-generator-core/src/interfaces.ts
timotheeguerin: Yeah plus one do not expose that to the user. In the same way you configure the emitter name when creating the context you should be able to configure that

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -77,7 +78,11 @@ export interface SdkOperationGroup {
   service: Namespace;
 }
 
-interface SdkTypeBase {
+interface DecoratedType {
+  decorators: Record<string, Record<string, any>>;
```
jhendrixMSFT, line 82: I get that as decorators are open-ended it's difficult to have strongly typed models. However, this is very open-ended and without any docs/examples, callers are left to examining the contents in the debugger to understand how the data is structured.

E.g. in one of the XML tests I see this.

`TypeSpec.Xml.@attribute`

How is that value formulated, what's the pattern? Can I tell from looking at the docs for that decorator? Is the contract formalized and stable?

tadelesh: yes. the open-ended list is hard to use. but the main purpose for this feature is we want to prevent adding properties to tcgc types for every simple decorator, and let emitter has a way to use new features from different typespec libraries without waiting tcgc support and bump tcgc version. then how to use that feature is left to emitter.

m-nash: @jhendrixMSFT the value of `Typespec.Xml.@attribute` is the fully qualified name of the decorator.  Do you have another suggestion on how to model this?

haolingdong-msft (approved):

packages/typespec-client-generator-core/src/interfaces.ts
jhendrixMSFT: Ok great, thanks for clarifying. Can we document this?

lirenhe (approved):

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: @jhendrixMSFT do you mean i need to doc how each decorator in the list should be used in emitter, iow add link to that decorator doc page?

iscai-msft (approved):
Looks great thanks @tadelesh

packages/typespec-client-generator-core/src/interfaces.ts
jhendrixMSFT: I mean document the structure of the data within the record. i.e. the key is the fully qualified name of a decorator, and the value is the decorator's JSON representation (I assume that's what it is based on the tests). It would also be good to mention that this is constrained by an allow list which means the record will be empty if none of the allowed decorators have been applied (i.e. this doesn't contain all decorators).

tadelesh: Added. Could you help to review again?

jhendrixMSFT: Thanks!

.chronus/changes/decorators_list-2024-5-6-18-38-7.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+export decorators in white list to all sdk types
```
jhendrixMSFT, line 7: allow list

packages/typespec-client-generator-core/src/configs.ts
```diff
@@ -0,0 +1,4 @@
+export const defaultDecoratorsWhiteList = [
```
jhendrixMSFT, line 1: `defaultDecoratorsAllowList`

m-nash (approved):

PR #968, opened 2024-06-06
Update folder name for canonical swagger

dinwa-ms:
ISSUE: since the canonical swagger file version is "canonical", based on current logic, the swagger file folder name is "stable", which should be "canonical".

### DIFF
--- a/packages/typespec-autorest-canonical/src/emitter.ts
+++ b/packages/typespec-autorest-canonical/src/emitter.ts
@@ -84,7 +84,7 @@ async function emitAllServices(
     validateUnsupportedVersioning(program, service.type);
     const context: AutorestEmitterContext = {
       program,
-      outputFile: resolveOutputFile(service, services.length > 1, options),
+      outputFile: resolveOutputFile(service, services.length > 1, options, canonicalVersion),
       service,
       version: canonicalVersion,
       tcgcSdkContext,
@@ -144,11 +144,6 @@ function resolveOutputFile(
       multipleServices || azureResourceProviderFolder
         ? getNamespaceFullName(service.type)
         : undefined,
-    "version-status": azureResourceProviderFolder
-      ? version?.includes("preview")
-        ? "preview"
-        : "stable"
-      : undefined,
     version,
   });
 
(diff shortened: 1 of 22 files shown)


packages/typespec-autorest-canonical/src/emitter.ts
```diff
@@ -84,7 +84,7 @@ async function emitAllServices(
     validateUnsupportedVersioning(program, service.type);
     const context: AutorestEmitterContext = {
       program,
-      outputFile: resolveOutputFile(service, services.length > 1, options),
+      outputFile: resolveOutputFile(service, services.length > 1, options, canonicalVersion),
```
timotheeguerin, line 87: wouldnt it be better to hard code this, instead of treating this as a "version"

dinwa-ms: I removed this version parameter and passed the constant "canonical" to interpolatePath(). Can I ask what do you mean "hard code in the config"? Thanks

timotheeguerin: i mean why are we interpolating "cannonical" as a version, why is it allowed in the output file, the default value should just have canonical included instead of `{version}`

dinwa-ms: I agree with that. I removed that parameter for resolveOutputFile and just used "canonical" in it.

timotheeguerin: I mean there too it shouldn't the user should NOT be allowd to pass `{version}` in output file and have that be interpolated. change the default value of `output-file`

dinwa-ms: update the default of output-file

packages/typespec-autorest-canonical/src/emitter.ts
```diff
@@ -144,12 +143,7 @@ function resolveOutputFile(
       multipleServices || azureResourceProviderFolder
         ? getNamespaceFullName(service.type)
         : undefined,
-    "version-status": azureResourceProviderFolder
-      ? version?.includes("preview")
-        ? "preview"
-        : "stable"
-      : undefined,
-    version,
+    version: canonicalVersion,
```
timotheeguerin, line 146: you can remove that now

dinwa-ms: removed version parameter and update the template config

timotheeguerin (approved):

PR #987, opened 2024-06-12
multipart api design (DO NOT MERGE)

msyyc:
based on https://github.com/microsoft/typespec/issues/3046

### DIFF
--- a/packages/typespec-client-generator-core/multipart_api_design.md
+++ b/packages/typespec-client-generator-core/multipart_api_design.md
@@ -24,7 +24,35 @@ According to https://datatracker.ietf.org/doc/html/rfc7578, multipart request pa
 # Current TCGC API for multipart
 Currently TCGC only has boolean flag [isMultipartFileInput](https://github.com/Azure/typespec-azure/blob/ab7a066d4ac0ae23a40f9ff8f4b6037559bda34c/packages/typespec-client-generator-core/src/interfaces.ts#L368) to distinguish file and non-file. After https://github.com/microsoft/typespec/issues/3046 complete, Typespec permit users to define more info explicitly(e.g. content-type) for each part so boolean flag is not enough.
 
-# proposal about new TCGC API for multipart
+# Multipart in Typespec
+Typespec support two kinds of definition for multipart:
+
+1. Model format
+
+```
+op upload(
+  @header `content-type`: "multipart/form-data",
+  @multipartBody body: {
+    fullName: HttpPart<string>,
+    headShots: HttpPart<Image>[]
+  }
+): void;
+```
+
+2. Tuple format
+
+ ```
+op upload(
+  @header `content-type`: "multipart/form-data",
+  @multipartBody body: [
+    HttpPart<string, #{ name: "fullName" }>,
+    HttpPart<File, #{ name: "file" }>
+  ]
+): void;
+```
+
+# Proposal about new TCGC API for multipart
+Since all language emitters emit multipart body as model format, TCGC will uniform the multipart body type as model, and each model property equals to property of model format or part of tuple format in Typespec.
 
 ```typescript
 exprot interface multipartOptionsType {
--- a/packages/typespec-client-generator-core/multipart_api_design.md
+++ b/packages/typespec-client-generator-core/multipart_api_design.md
@@ -80,6 +80,5 @@ notes:
  (a) serialize array of model as single part or (b) serialize model as single part then send it multi times. With new design, if
  `HttpPart<Model[]>`, multi is false and SDK shall follow (a); if `HttpPart<Model>[]`, multi is true and follow (b)
 - `headers`: equals to custom headers in swagger https://swagger.io/docs/specification/describing-request-body/multipart-requests/  
-- `filename`: Typespec permit author use `httpFile` change requiredness for optional metadata properties, including `filename`.
-If defined, users could explicitly set it as required or not.
-- `contentType`: Typespec permit author use `httpFile` change requiredness for optional metadata properties, including `content-type`.
+- `filename`: When Typespec author use `httpFile` change requiredness for optional metadata properties "filename", this property has value; otherwise it is "undefined".
+- `contentType`: When Typespec author use `httpFile` change requiredness for optional metadata properties "contentType", this property has value; otherwise it is "undefined".
--- a/packages/typespec-client-generator-core/multipart_api_design.md
+++ b/packages/typespec-client-generator-core/multipart_api_design.md
@@ -55,7 +55,7 @@ op upload(
 Since all language emitters emit multipart body as model format, TCGC will uniform the multipart body type as model, and each model property equals to property of model format or part of tuple format in Typespec.
 
 ```typescript
-exprot interface multipartOptionsType {
+export interface multipartOptionsType {
   isNameDefined: boolean; // whether name is defined in Typespec. For multipart/mixed, name may not be defined for some parts
   isFilePart: boolean; // whether this part is for file
   multi: boolean; // whether this part is multi in request payload
--- a/packages/typespec-client-generator-core/multipart_api_design.md
+++ b/packages/typespec-client-generator-core/multipart_api_design.md
@@ -46,7 +46,7 @@ op upload(
   @header `content-type`: "multipart/form-data",
   @multipartBody body: [
     HttpPart<string, #{ name: "fullName" }>,
-    HttpPart<File, #{ name: "file" }>
+    HttpPart<Image, #{ name: "headShots" }>[]
   ]
 ): void;
 ```
--- a/packages/typespec-client-generator-core/multipart_api_design.md
+++ b/packages/typespec-client-generator-core/multipart_api_design.md
@@ -69,7 +69,7 @@ export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
   discriminator: boolean;
   serializedName: string;
   isMultipartFileInput: boolean;  // deprecated
-  multipartOptions: multipartOptionsType; // new options for multipart
+  multipartOptions?: multipartOptionsType; // new options for multipart
   visibility?: Visibility[];
   flatten: boolean;
 }
--- a/packages/typespec-client-generator-core/multipart_api_design.md
+++ b/packages/typespec-client-generator-core/multipart_api_design.md
@@ -58,7 +58,7 @@ Since all language emitters emit multipart body as model format, TCGC will unifo
 export interface multipartOptionsType {
   isNameDefined: boolean; // whether name is defined in Typespec. For multipart/mixed, name may not be defined for some parts
   isFilePart: boolean; // whether this part is for file
-  multi: boolean; // whether this part is multi in request payload
+  isMulti: boolean; // whether this part is multi in request payload
   headers: HeaderProperty[]; // relates to custom header
   filename?: SdkModelPropertyTypeBase;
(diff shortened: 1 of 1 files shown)


packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,84 @@
...
+  multi: boolean; // whether this part is multi in request payload
+  headers: HeaderProperty[]; // relates to custom header
+  filename?: SdkModelPropertyTypeBase;
+  contentType?: SdkModelPropertyTypeBase;
+}
+
+export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
+  kind: "property";
+  discriminator: boolean;
+  serializedName: string;
```
timovv, line 69: `name` is not strictly necessary for non form-data content types such as `multipart/mixed`. Something like this could be in fact valid:

```typespec
op upload(
  @header `content-type`: "multipart/mixed",
  @multipartBody body: [
    HttpPart<string, #{}>,
    HttpPart<File, #{}>
  ]
): void;
```

Have we thought about how we might handle this?

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,84 @@
...
+
+ ```
+op upload(
+  @header `content-type`: "multipart/form-data",
+  @multipartBody body: [
+    HttpPart<string, #{ name: "fullName" }>,
+    HttpPart<File, #{ name: "file" }>
+  ]
+): void;
+```
```
weidongxu-microsoft, line 52: Are 2 format equivalent?

If yes, would you use a same example (with different format)?

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Proposal about new TCGC API for multipart
+Since all language emitters emit multipart body as model format, TCGC will uniform the multipart body type as model, and each model property equals to property of model format or part of tuple format in Typespec.
```
weidongxu-microsoft, line 55: > emit multipart body as model format

What about alias/spread? We support that.

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,84 @@
...
+
+```typescript
+exprot interface multipartOptionsType {
+  isFilePart: boolean; // whether this part is for file
+  multi: boolean; // whether this part is multi in request payload
```
weidongxu-microsoft, line 60: Is this for `string[]` vs `string`?

What's the `type` of this (when `multi=true` -- element or array of element?), in SdkBodyModelPropertyType?

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: Nice hint. For multipart/mixed, the name is not needed so it is also not defined in Typespec. I add `isNameDefined` in multipartOptionsType so that language emitters could get the info there is no name defined.

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: For `httpPart<string>[]`, multi=true and type is string
For `httpPart<string[]>`, multi=false and type is array of string

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: make sense. 
Updated.

packages/typespec-client-generator-core/multipart_api_design.md
weidongxu-microsoft: Is there only one Part be no name? Or multiple Parts is also valid?

I'd agree that if there is no name in Part, the `serializedName` should be undefined/empty, as this represents the wire data.

However, I am not sure whether TCGC would need to fill-in the `name` -- this would be similar to "anonymous model", when typespec does not have a name, TCGC would fill one.
Though, it seems had to figure out a proper name here.

If we won't see a `multipart/mixed` in azure anytime soon, we can have an empty `name`, and delay this discussion to future iterations.

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: If we uniformly emit Multipart body type to model type, it is very similar to common body so it will also support alias/spread.

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,86 @@
...
+    HttpPart<Image, #{ name: "headShots" }>[]
+  ]
+): void;
+```
+
+# Proposal about new TCGC API for multipart
+Since all language emitters emit multipart body as model format, TCGC will uniform the multipart body type as model, and each model property equals to property of model format or part of tuple format in Typespec.
+
+```typescript
+exprot interface multipartOptionsType {
```
weidongxu-microsoft, line 58: nit, maybe call it something like `MultipartHttpPart` -- it seems no need to deviate too much from `HttpOperationPart`

msyyc: I don't have much preference for the type name here, let us see whether there are more comments here.

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: To be honest, I don't know much about `multipart/mixed` and I have not seen any real service adopt this kind of content-type. With `isNameDefined`, at least TCGC have provided the info "name is not defined in Typespec and language may have to decide by yourself how to expose this part". Maybe we could delay the discussion and focus on "multipart/form-data" at first.

packages/typespec-client-generator-core/multipart_api_design.md
weidongxu-microsoft: nit that `exprot` :-)

weidongxu-microsoft:
I don't have major issue with the design.

packages/typespec-client-generator-core/multipart_api_design.md
timovv: I am happy to delay this -- I mainly wanted to bring it up so that we're aware of it. My understanding is that part of the reason for the new TypeSpec API is to allow for different kinds of multipart payloads (not just form data) to be expressed so maybe this is something we will want to support in the future. But if this design for TCGC is focused just on multipart/form-data, I don't have a problem with that.

Using `multipart/mixed` in Azure seems to be rare but it's not unheard of -- as an example, the JS Tables SDK has some custom code handling the content type [here](https://github.com/Azure/azure-sdk-for-js/blob/af883f4bdb91b8d8aee15c45bd9eb85e0c9c351c/sdk/tables/data-tables/src/utils/transactionHelpers.ts).

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,86 @@
...
+  filename?: SdkModelPropertyTypeBase;
+  contentType?: SdkModelPropertyTypeBase;
+}
+
+export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
+  kind: "property";
+  discriminator: boolean;
+  serializedName: string;
+  isMultipartFileInput: boolean;  // deprecated
+  multipartOptions: multipartOptionsType; // new options for multipart
```
chunyu3, line 72: Is this `optional`?

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,86 @@
...
+# Current TCGC API for multipart
+Currently TCGC only has boolean flag [isMultipartFileInput](https://github.com/Azure/typespec-azure/blob/ab7a066d4ac0ae23a40f9ff8f4b6037559bda34c/packages/typespec-client-generator-core/src/interfaces.ts#L368) to distinguish file and non-file. After https://github.com/microsoft/typespec/issues/3046 complete, Typespec permit users to define more info explicitly(e.g. content-type) for each part so boolean flag is not enough.
+
+# Multipart in Typespec
+Typespec support two kinds of definition for multipart:
+
+1. Model format
+
+```
+op upload(
```
chunyu3, line 33: We will not support previous define with `@body` as following?
```
op upload(@header `content-type`: "mulitpart/form-data",
@body body: {
   fileName: string,
   headShots: bytes[]
}
```

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,86 @@
...
+# Proposal about new TCGC API for multipart
+Since all language emitters emit multipart body as model format, TCGC will uniform the multipart body type as model, and each model property equals to property of model format or part of tuple format in Typespec.
+
+```typescript
+exprot interface multipartOptionsType {
+  isNameDefined: boolean; // whether name is defined in Typespec. For multipart/mixed, name may not be defined for some parts
+  isFilePart: boolean; // whether this part is for file
+  multi: boolean; // whether this part is multi in request payload
+  headers: HeaderProperty[]; // relates to custom header
+  filename?: SdkModelPropertyTypeBase;
```
chunyu3, line 63: `filename` and `contentType` is meta-data of file property, Itself is not model `property`. So maybe we shall not define it as `SdkModelPropertyTypeBase`?

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: For now we support this format before all language emitters support `httpPart`.

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: This is TCGC interface which has wrapped original http type and I will make sure it contain all info that language emitters may need.

packages/typespec-client-generator-core/multipart_api_design.md
msyyc: Good question. I think it shall be `optional` since for non-multipart content type there is no need for this property.

packages/typespec-client-generator-core/multipart_api_design.md
lmazuel: `httpPart` is for advanced scenario, we are not planning to remove the `@body` format. Emitter should not care though, it should be transporte for them, they just ask TCGC for the multipart information

packages/typespec-client-generator-core/multipart_api_design.md
```diff
@@ -0,0 +1,86 @@
...
+  ]
+): void;
+```
+
+# Proposal about new TCGC API for multipart
+Since all language emitters emit multipart body as model format, TCGC will uniform the multipart body type as model, and each model property equals to property of model format or part of tuple format in Typespec.
+
+```typescript
+exprot interface multipartOptionsType {
+  isNameDefined: boolean; // whether name is defined in Typespec. For multipart/mixed, name may not be defined for some parts
```
lmazuel, line 59: Why not a string field that may be `undefined` ?

msyyc: `SdkModelPropertyTypeBase` already has required property `name`. If isNamedDefined is true, emitter shall use name; if false, it means typespec doesn't define name and emitter shall not use the name.
NIT: in Typescript, it is impossible to override the property "name" to "optional"

Issue #1010, opened 2024-06-13
Some of our `SdkBuiltInType` is not so "built-in"


ArcturusZhang:
We are categorizing the kind of built-in types into several categories, namely (found in source code [here](https://github.com/Azure/typespec-azure/blob/17238644e6ec27695831f44eceedde54bec7ffb3/packages/typespec-client-generator-core/src/interfaces.ts#L109-L155)):
- SdkIntKindsEnum 
- SdkFloatKindsEnum
- SdkAzureBuiltInStringKindsEnum
- SdkGenericBuiltInStringKindsEnum
- SdkBuiltInKindsMiscellaneousEnum

One of them specifically stating `Azure` therefore I assume others are non-azure namely "real built-in" types.
But there are still some kinds that I could not find in the standard library. I have searched the document, and they do not appear on the ["type relations" graph](https://typespec.io/docs/language-basics/type-relations) or appear in [the list](https://typespec.io/docs/standard-library/built-in-data-types) of built-in types, they are:
- ipAddress (not found anywhere)
- guid (not found anywhere)
- uri (removed in 0.40.0)
- password (found bunch of others but none of them are definitions of the built-in type password)

If these are not supported any more, we should clean them up. As for `uuid`, we should move it to the azure category.

ArcturusZhang:
Resolved by https://github.com/Azure/typespec-azure/pull/1015

Issue #997, opened 2024-06-13
We might should model scalars independently for extensibility


ArcturusZhang:
Azure Core template defines quite a few scalars to define something special to azure, such as `azureLocation`, `armId`, etc.

Currently they are treated as if they are built-in type with their own "kind".
This is not a way that could be extended because every time when there is a new scalar, we will have to update TCGC to adopt it otherwise the new information will be completely discarded, and the downstream SDKs will only get its underlying types.

We could do the following change to make the scalars extensible, such as:
```typescript
interface SdkScalarType extends SdkTypeBase {
  name: string;
  namespace: string;
  underlyingType: SdkBuiltinType; // or call it wireType just as duration and datetime?
}
```
And all those scalars, will be parsed into these new scalar types instead of built-in types, in the downstream SDKs, we could do (taking dotnet as example):
```csharp
public CSharpType CreateType(InputType inputType) => inputType switch {
// omitting others
InputScalarType scalarType => (name, namespace) switch {
	=> ("Azure.Core", "AzureLocation") => typeof(AzureLocation),
	_ => CreateType(scalarType.UnderlyingType),
},
// omitting others
};
```
In this way, the extensible part could only happen in the downstream SDK generator part, and TCGC never needs a new version for such a requirement.

PR #1015, opened 2024-06-14
introduce `name` and `crossLanguageDefinitionId` to `SdkBuiltInType`, `SdkDateTimeType` and `SdkDurationType`

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/1010
Fixes https://github.com/Azure/typespec-azure/issues/997
Fixes https://github.com/Azure/typespec-azure/issues/1106

# Changes and migration guides

## Changes

1. remove the following types from `SdkBuiltInType`:
   - `uuid` (azure type)
   - `ipV4Address` (azure type)
   - `ipV6Address` (azure type)
   - `eTag` (azure type)
   - `armId` (azure type)
   - `azureLocation` (azure type)
   - `password` (removed)
   - `guid` (removed)
   - `uri` (removed)
   - `ipAddress` (removed)
2. `@format` can no longer change a string type to above azure types (`uuid`, `eTag`, etc).
3. add `name`, `crossLanguageDefinitionId` and `baseType` to `SdkBuiltInType`, `SdkDateTimeType` and `SdkDurationType`.
4. now scalars defined using `scalar` keyword will be parsed into either `SdkBuiltInType`, `SdkDateTimeType` or `SdkDurationType` depending on the base type, with its `name` and `crossLanguageDefinitionId`. `@encode` will be added to the scalar type, and will not propagate to its base type.
5. `isSdkFloatKind` now returns `false` for `decimal` and `decimal128`.

## Migration guides for emitters

The major breaking change introduced by this change is the removal of azure related kinds for `SdkBuiltInType`. Therefore for code such as:
```typescript
if (type.kind === "azureLocation")
{
  // do something for azure-location
}
```
should be changed to the following if the emitter is doing something special for azure-location:
```typescript
if (type.kind === "string" && type.crossLanguageDefinitionId === "Azure.Core.azureLocation")
{
  // do something for azure-location
}
```
or just change it to this:
```typescript
if (type.kind === "string")
{
  // treat azure-location as a string
}
```

For those removed kinds, they no longer exist therefore it should be safe just remove related code snippets.

# Summary

This PR includes the following changes:

1. remove the following types from `SdkBuiltInType`:
   - `uuid` (azure type)
   - `ipV4Address` (azure type)
   - `ipV6Address` (azure type)
   - `eTag` (azure type)
   - `armId` (azure type)
   - `azureLocation` (azure type)
   - `password` (removed)
   - `guid` (removed)
   - `uri` (removed)
   - `ipAddress` (removed)
2. `@format` can no longer change a string type to above azure types (`uuid`, `eTag`, etc).
3. add `name`, `crossLanguageDefinitionId` and `baseType` to `SdkBuiltInType`, `SdkDatetimeType` and `SdkDurationType`.
4. now scalars defined using `scalar` keyword will be parsed into either `SdkBuiltInType`, `SdkDatetimeType` or `SdkDurationType` depending on the base type, with its `name` and `crossLanguageDefinitionId`. `@encode` will be added to the scalar type, and will not propagate to its base type.
5. `isSdkFloatKind` now returns `false` for `decimal` and `decimal128`.

# Reasons and motivations

TCGC should be "linguistic neutrality" which does not favor or disfavor any language and only represent the structures defined in a tsp file/files.
Also I think it should be neutral for 3rd party extensions, and azure should be regarded as our first 3rd party extension. Therefore we need a way to be able to handle those azure types without making them special in TCGC's code.

The solution that this PR is applying is similar to the solution we have for models in azure core template.
Instead of creating a special kind for models in azure core template (such as the `ResponseError`, `TrackedResource`, `Resource` or `ExtendedLocation`), we just parse them as regular models with their own name and tsp namespace, so that the downstream genreators could know where these types are defined, and if they want, they could do a filter, to handle those special types.
Same thing could apply to scalars, for a scalar defined in azure core template:
```typespec
namespace Azure.Core;
scalar azureLocation extends string;
```
for those would care this, this stands for an "azure location", for those would not care, this is just a string.
Therefore this PR introduces name and `crossLanguageDefinitionId` to `SdkBuiltInType` just like we have in models, this scalar now will be parsed as:
```
{
	kind: "string",
	name: "azureLocation",
	tspNamespace: "Azure.Core",
	baseType: {
		kind: "string",
		name: "string",
		tspNamespace: "TypeSpec"
	}
}
```
The `crossLanguageDefinitionId` is important for downstream generators to know whether this scalar is defined.
`baseType` here is also important and we must have this, because we do not want to see if users write scalars extending some of the scalars from azure core template, and it ends up with a string.
For example, based on the fact that we have the above `azureLocation` defined in our azure core template, and one could write:
```typespec
scalar myLocation extends azureLocation;

model Foo {
	bar: myLocation;
}
```
in this case, I think the generator should generate the `bar` property of `Foo` model as `AzureLocation` type. But if we do not have the `baseType`, we only have this:
```typespec
{
	kind: "string",
	name: "myLocation",
	tspNamespace: "My.Service"
}
```
which has nothing with `azureLocation` which is not acceptable.


packages/typespec-client-generator-core/test/types.test.ts
```diff
@@ -151,15 +151,9 @@ describe("typespec-client-generator-core: types", () => {
...
           urlScalar: url;
-          uuidScalar: uuid;
-          eTagScalar: eTag;
 
           @format("url")
           urlProperty: string;
-          @format("uuid")
-          uuidProperty: string;
-          @format("eTag")
-          eTagProperty: string;
```
ArcturusZhang, line 162: `uuid` and `eTag` are azure types, we should not support their format by default.

packages/typespec-client-generator-core/test/types/built-in-types.test.ts
```diff
@@ -134,15 +138,9 @@ describe("typespec-client-generator-core: built-in types", () => {
...
         urlScalar: url;
-        uuidScalar: uuid;
-        eTagScalar: eTag;
 
         @format("url")
         urlProperty: string;
-        @format("uuid")
-        uuidProperty: string;
-        @format("eTag")
-        eTagProperty: string;
```
ArcturusZhang, line 145: these are azure tpyes, we should not support their format any more

packages/typespec-client-generator-core/test/types/built-in-types.test.ts
```diff
@@ -3,17 +3,29 @@ import { AzureCoreTestLibrary } from "@azure-tools/typespec-azure-core/testing";
 import { expectDiagnostics } from "@typespec/compiler/testing";
 import { ok, strictEqual } from "assert";
 import { beforeEach, describe, it } from "vitest";
+import { leftOver } from "../../src/interfaces.js";
```
ArcturusZhang, line 6: is there a way that we could verify this without exporting it?
I exported this type only for test, and the type is a never, no one is ever going to use it and it is never useful to anyone

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -104,8 +105,28 @@ export type SdkType =
...
 
+type TypeEquality<T, U> = keyof T extends keyof U
+  ? keyof U extends keyof T
+    ? true
+    : false
+  : false;
+
+// these two vars are used to validate whether our SdkBuiltInKinds are exhaustive for all possible values from typespec
+// if it is not, a typescript compilation error will be thrown here.
+const _: TypeEquality<Exclude<SupportedBuiltInKinds, SdkBuiltInKinds>, never> = true;
```
iscai-msft, line 121: is there a clearer way of writing this? It's quite hard to read.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -104,8 +105,28 @@ export type SdkType =
 export interface SdkBuiltInType extends SdkTypeBase {
   kind: SdkBuiltInKinds;
   encode: string;
+  name: string;
+  tspNamespace?: string;
+  baseType?: SdkBuiltInType;
```
iscai-msft, line 110: what does the `baseType` give us? Won't the base scalar type be the `kind`?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -184,80 +186,162 @@ export function addEncodeInfo(
...
+  kind: "utcDateTime" | "offsetDateTime"
+): [SdkDatetimeType, readonly Diagnostic[]] {
+  const diagnostics = createDiagnosticCollector();
+  const docWrapper = getDocHelper(context, type);
+  const encodeData = getEncode(context.program, type);
+  const baseType = type.baseScalar
+    ? diagnostics.pipe(getSdkDateTimeType(context, type.baseScalar, kind))
+    : undefined;
+  // first we try if this type has an encode on the encode decorator, if not, we get the encode from base type, if still not, we fallback to the default value of datetime rfc3339
+  const encode = encodeData?.encoding ?? baseType?.encode ?? "rfc3339";
```
iscai-msft, line 215: can we keep getting the encode and wire type as a separate function for readability? It's hard to follow with that many question marks

iscai-msft (commented):
Since this is breaking, can you also write a quick migration guide for emitters? Thanks!

Overall I like the direction, thanks for your contribution @ArcturusZhang

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -104,8 +105,28 @@ export type SdkType =
 export interface SdkBuiltInType extends SdkTypeBase {
   kind: SdkBuiltInKinds;
   encode: string;
+  name: string;
```
weidongxu-microsoft, line 108: Is the `name` guaranteed to be same as the name in typespec, for all subclass of `SdkTypeBase`? Or it only be same if under `SdkBuiltInType`?

If latter, should we just call it `tspName`?

ArcturusZhang: I think it might be necessary to have a tspName in the future maybe considering there might be name changes such `@clientName`.
But here we are following the same rule for models. In details:
if this is a type defined somewhere else from the std library, this name represents the `library name`, which is a name modified by various decorators. If this is a type defined in the std library, this is just `type.name`.

packages/typespec-client-generator-core/src/interfaces.ts
ArcturusZhang: base gives us the `B` part in `scalar A extends B`.
This might be important in a case like this:
```typespec
// this is in azure core
scalar azureLocation extends string;

// this is in service spec
scalar computeLocation extends azureLocation;
```
if we do not have the `base`, in the result we will never have anything related with `azureLocation`.
More detailed information for this could be found in the PR description, please take a look as well

packages/typespec-client-generator-core/src/interfaces.ts
ArcturusZhang: I think this is not here for readability. I added this because I think it is necessary to validate our current kind defined in those enums are exhaustive for the kinds defined in std libraries.

We should remove these checks once we get rid of those enums.

I predicated that there should be a readability issue, therefore I added the above comments explaining what these two lines are doing. Please let me know if this still not clear enough

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -154,12 +162,9 @@ enum SdkBuiltInKindsMiscellaneousEnum {
...
 
-export type SdkBuiltInKinds =
-  | keyof typeof SdkBuiltInKindsMiscellaneousEnum
-  | keyof typeof SdkIntKindsEnum
-  | keyof typeof SdkFloatKindsEnum
-  | keyof typeof SdkGenericBuiltInStringKindsEnum
-  | keyof typeof SdkAzureBuiltInStringKindsMapping;
+export type SdkBuiltInKinds = Exclude<IntrinsicScalarName, SdkBuiltInKindsExcludes> | "any";
+
+export type SdkBuiltInKindsExcludes = "utcDateTime" | "offsetDateTime" | "duration";
```
ArcturusZhang, line 167: so I think this should not be exported. I will make a change to remove the export.

packages/typespec-client-generator-core/src/types.ts
ArcturusZhang: Yeah - this might be messy.
I will make a function for this so that it could be more clear

packages/typespec-client-generator-core/src/interfaces.ts
weidongxu-microsoft: Probably not going to block the PR. Just want to check if we've considered some corner cases.

Could somebody do this? (it is valid in playground)

```ts
@friendlyName("url")
scalar uri extends string;

model Resource {
  prop1: uri;
}
```

packages/typespec-client-generator-core/src/types.ts
ArcturusZhang: This part has been refactored.

ArcturusZhang:
> Since this is breaking, can you also write a quick migration guide for emitters? Thanks! We also need approval from all languages
> 
> Overall I like the direction, thanks for your contribution @ArcturusZhang

Usually where we write the migration? In the changelog?

ArcturusZhang:
@iscai-msft I added some guidance in the changelog.

packages/typespec-client-generator-core/test/types/built-in-types.test.ts
haolingdong-msft: Why do we remove below cases? 
```
uuidScalar: uuid;
eTagScalar: eTag;
```

haolingdong-msft:
@ArcturusZhang  I understand and like your your 2 and 3 points in your description. Would like to understand more on 1, why do we remove those types?

packages/typespec-client-generator-core/src/interfaces.ts
ArcturusZhang: I do not think any one needs to ever remember this
> if this is a type defined somewhere else from the std library, this name represents the library name, which is a name modified by various decorators. If this is a type defined in the std library, this is just type.name.

I am saying this because I am doing this in the code, but there should not be such a rule because no one could change the name of a type from the std library, I did not call the `getLibraryName` when constructing them. I could do that but I think it is meaningless because no one could change it.

I could align the implementation so that there is no such rule.

ArcturusZhang:
> @ArcturusZhang I understand and like your your 2 and 3 points in your description. Would like to understand more on 1, why do we remove those types?

several reasons:
1. because we have the name and namespace, we already have the ability to represent azure stuffs, we do not need duplicates.
2. removing them is "making our library consistent with rules for scalar types which built up from std types"
3. like I have mentioned in the motivation:
> Also I think it should be neutral for 3rd party extensions, and azure should be regarded as our first 3rd party extension. Therefore we need a way to be able to handle those azure types without making them special in TCGC's code.
we should not make "azure" special in such a fundamental library if we do not have to. This should benefit in the future if we need to trim azure off from TCGC because we already did it.

packages/typespec-client-generator-core/src/interfaces.ts
weidongxu-microsoft:
```
@@friendlyName(TypeSpec.duration, "dur");

@encode(DurationKnownEncoding.seconds, float32)
scalar myDuration extends duration;
```

image

weidongxu-microsoft: What I mean is, give the behavior of this `name`, it would be much safer to dev to just use `__raw.name` if we really want to know the typespec name of the type.

I am OK with your PR. Just saying given the complication of this `type.name`, I won't really using it to check typespec name.

ArcturusZhang: `raw` is a workaround, we could use it for now, but we cannot use it forever, therefore we definitely need something else to represent the original name for a type.
If we could even add decorators to the std types, I will make a change to always call `getLibraryName` to be align. It may not make sense but i think we must align because no one could remember those complicated rules.

ArcturusZhang: per offline discussion, we should be able to identify types from template using `crossLanguageDefinitionId`, therefore we do not need to have the original name?

ArcturusZhang: Oh - after I checked my code, there is no such rule in the first place. All the `name` is constructed by `getLibraryName` even for those primitive types.

.chronus/changes/add-scalar-type-2024-5-14-13-23-7.md
```diff
@@ -0,0 +1,49 @@
...
+The major breaking change introduced by this change is the removal of azure related kinds for `SdkBuiltInType`. Therefore for code such as:
+```typescript
+if (type.kind === "azureLocation")
+{
+  // do something for azure-location
+}
+```
+should be changed to the following if the emitter is doing something special for azure-location:
+```typescript
+if (type.kind === "string" && type.crossLanguageDefinitionId === "Azure.Core.azureLocation")
```
weidongxu-microsoft, line 36: Does dev need to visit the `baseType` in some case?

e.g. if tsp

```ts
scalar myuuid extends Azure.Core.uuid;

model Resource {
  resourceId: myuuid;
}
```

I assume we still treat `myuuid` as UUID?

haolingdong-msft:
I think Java will still need the `tspNamespace` info to calculate our schema's namespace. [[code](https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/utils.ts#L27)] @weidongxu-microsoft please correct me if I'm wrong.

weidongxu-microsoft:
> I think Java will still need the `tspNamespace` info to calculate our schema's namespace. [[code](https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/utils.ts#L27)] @weidongxu-microsoft please correct me if I'm wrong.

We can have another discussion or PR to solve for Java.

Dapeng's PR is handling a different problem.

haolingdong-msft:
> > I think Java will still need the `tspNamespace` info to calculate our schema's namespace. [[code](https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/utils.ts#L27)] @weidongxu-microsoft please correct me if I'm wrong.
> 
> We can have another discussion or PR to solve for Java namespace problem.
> 
> Dapeng's PR handles a different problem.

Yeah, sure, just saw `tspNamespace` is removed, so raise this concern.

packages/typespec-client-generator-core/src/interfaces.ts
weidongxu-microsoft: A good use of `never` :-)

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -42,6 +42,20 @@ import {
...
+  raw?: Type
+): SdkBuiltInType {
+  return {
+    __raw: raw,
+    kind: kind,
+    name: kind,
+    encode: encode ?? kind,
+    crossLanguageDefinitionId: `TypeSpec.${kind}`,
+  };
+}
```
weidongxu-microsoft, line 57: I am not sure whether this is the best solution.

At least, the `__raw` here does not exist.

Should we use the real TypeSpec builtin type? e.g.
`program.getGlobalNamespaceType().namespaces.get("TypeSpec").scalars.get("string")`

ArcturusZhang: OK if we could have this then great, previously I do not know we could just sommon a type from typespec.
This function serves the purpose where we are building literals, or building operations, we just need to come up with a std type. Before this change, we are doing it in this way:
```
type: {
...getXXXSdkBase(context, type, "string"),
encode: getEncodeHelper(...)
}
```
now since we have quite a few properties in the BuiltInType, it is not so convenient if we still construct it from scratch when we need it, therefore I created this function

ArcturusZhang: but if we could do it in a more official way, then great, I will try if this could work.

.chronus/changes/add-scalar-type-2024-5-14-13-23-7.md
ArcturusZhang: I think we should have an offline discussion for this topic with Chenjie, I have disagreement with him on this on a particular case for encoding.
But in general, the best practice now looks like:
```
function getType(type: SdkType): YourType
{
	if (type.kind === "string")
	{
		// recursively find if any level we have a concerned definition id
		const result = findForSpecial(type);
		if (result) return result;
		return usualString;
	}
	// omit others
}
function findForSpecial(type: SdkBuiltInType & { kind: "string" }): YourType | undefined {
	if (type.crossLanguageDefinitionId === "Something") return something;
	
	if (type.baseType) return findForSpecial(type.baseType);
	
	return undefined;
}
```
I admit this looks weird. Therefore i think we might need an offline discussion for this.

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: I'm still not sure we need `baseType`, at least to start with. What does knowing that it extends from `azureLocation` give you? I'd prefer if we don't add it unless it's absolutely necessary, since `baseType` is also confusing in this context. To me, the `baseType` would eventually be string here, bc `string` is the true underlying base type of both `azureLocation` and `compulteLocation`

iscai-msft (approved):
thanks @ArcturusZhang

packages/typespec-client-generator-core/src/interfaces.ts
ArcturusZhang: the value of `baseType` here is that in the above case, we want to generate `AzureLocation` for this `computeLocation`, because the `extends` means "is a", we should not generate a `string` in this case.
Therefore we need the base type to let us know this information, otherwise we just lose the information of azure location here.

tadelesh (approved):

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -118,8 +119,28 @@ export type SdkType =
 export interface SdkBuiltInType extends SdkTypeBase {
   kind: SdkBuiltInKinds;
   encode: string;
+  name: string;
```
lirenhe, line 122: Could you update https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/doc/types.tsp to make the doc consistent?

haolingdong-msft (approved):

packages/typespec-client-generator-core/src/interfaces.ts
ArcturusZhang: it has not been updated for a long time, and I have asked Chenjie, it might have been abandoned.
Therefore I will not update it otherwise I will have to catch up quite a few other changes.

tadelesh:
@ArcturusZhang could you also address this [issue](https://github.com/Azure/typespec-azure/issues/1106) in this pr?

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: current doc is: https://azure.github.io/typespec-azure/docs/howtos/DataPlane%20Generation%20-%20DPG/07tcgcTypes
could you help to update this and delete the types.tsp?

ArcturusZhang: done

.chronus/changes/add-scalar-type-2024-5-14-13-23-7.md
ArcturusZhang: this part has been included in the migration guide in the changelog

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -134,30 +156,20 @@ enum SdkIntKindsEnum {
...
   float64 = "float64",
+}
+
+enum SdkFixedPointKindsEnum {
   decimal = "decimal",
   decimal128 = "decimal128",
 }
 
-const SdkAzureBuiltInStringKindsMapping = {
-  uuid: "uuid",
```
MaryGao, line 146: May not relevant to this pr but I notice uuid is azure scope concept but url is not, i was wondering if uuid also is a general concept and should be in our bulit-in type in compiler?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -261,7 +260,7 @@ export function getSdkTypeBaseHelper<TKind>(
   return diagnostics.wrap({
     __raw: type,
     deprecation: getDeprecationDetails(context.program, type)?.message,
-    kind,
+    kind: kind,
```
MaryGao, line 263: nit, the same as but simpler
```suggestion
    kind,
```

packages/typespec-client-generator-core/src/interfaces.ts
ArcturusZhang: now if this `isStdType` function from compiler returns `true` for one scalar, it will be a builtin type kind.
Namely things defined [here](https://github.com/microsoft/typespec/blob/222fd806a018faff841837ac740c10c25222ef97/packages/compiler/lib/intrinsics.tsp#L9) and [here](https://github.com/microsoft/typespec/blob/222fd806a018faff841837ac740c10c25222ef97/packages/compiler/lib/std/types.tsp#L8)

I have similar opinion that uuid might be general enough to be in the standard library. But unfortunately it is not right now.
Please feel free to open issues in typespec repo for this.
And I have added some mechanism that once their kind for std type has new variant, we will know from the TCGC side (get compilation errors)

lirenhe (approved):

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -1340,9 +1350,10 @@ export interface SdkConstantType extends SdkTypeBase {
 export interface SdkUnionType extends SdkTypeBase {
   name: string;
   // determines if the union name was generated or not
-  generatedName: boolean;
+  isGeneratedName: boolean;
```
iscai-msft, line 1353: this is another breaking change, can you make a separate pr for this change, or at least call it out in the changeset?

ArcturusZhang: this is not a breaking change I believe.
I did not change this API in this PR. I only updated this file (a document) to align the latest APIs in our code.

iscai-msft: ohh i see, didn't realize it was a doc file. All good then

core
tadelesh, line 1: need to remove this change

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -435,27 +435,6 @@ describe("typespec-client-generator-core: decorators", () => {
...
-      const sdkPackage = runner.context.sdkPackage;
-      strictEqual(sdkPackage.clients.length, 1);
-      const mainClient = sdkPackage.clients[0];
-      strictEqual(mainClient.methods.length, 1);
-
-      const clientAccessor = mainClient.methods[0];
-      strictEqual(clientAccessor.kind, "clientaccessor");
-      strictEqual(clientAccessor.response.kind, "client");
-      strictEqual(clientAccessor.response.name, "ClientModel");
-    });
```
tadelesh, line 458: please revert this change.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -620,14 +764,7 @@ function getSdkEnumValueType(
...
   const diagnostics = createDiagnosticCollector();
   let kind: "string" | "int32" | "float32" = "string";
-  let type: EnumMember | UnionVariant;
   for (const value of values) {
-    if ((value as EnumMember).kind) {
-      type = value as EnumMember;
-    } else {
-      type = (value as UnionEnumVariant<string> | UnionEnumVariant<number>).type;
-    }
-
```
tadelesh, line 630: why remove this part?

Issue #963, opened 2024-06-06
TCGC, support list all nested sub clients under a client 


haolingdong-msft:
TCGC provides helper function to list all operation groups in `listOperationGroups` [function](https://github.com/Azure/typespec-azure/blob/ab7a066d4ac0ae23a40f9ff8f4b6037559bda34c/packages/typespec-client-generator-core/src/decorators.ts#L485):. But when adopting `sdkPackage` and `SdkClientType`, there no similar helper function. Could TCGC export function to list nested sub clients?

```ts
/**
 * List all the operation groups inside a client or an operation group. If ignoreHierarchy is true, the result will include all nested operation groups.
 *
 * @param context TCGCContext
 * @param group Client or operation group to list operation groups
 * @param ignoreHierarchy Whether to get all nested operation groups
 * @returns
 */
export function listOperationGroups(
  context: TCGCContext,
  group: SdkClient | SdkOperationGroup,
  ignoreHierarchy = false
):
```

e.g. client A contains two subclients, B and C, client C contains client D. The helper function's signature is something like 
`listSubClients(client: SdkClientType, listNestedClients: boolean): SdkClientType[]`
calling `listSubClients(ClientA, true)` will return client B,C,D.
calling `listSubClients(ClientA, false)` will return client B,C.

haolingdong-msft:
@tadelesh, thanks for the fix! But this issue is to add a helper function to allow list all the nested sub clients under a client. e.g. client A contains two subclients, B and C, client C contains client D. The helper function's signature is something like `listSubClients(client: SdkClientType, listNestedClients: boolean): SdkClientType[]`
calling `listSubClients(ClientA, true)` will return client B,C,D.
calling `listSubClients(ClientA, false)` will return client B,C.

It is similar as the existing helper function listOperationGroups [code](https://github.com/Azure/typespec-azure/blob/ab7a066d4ac0ae23a40f9ff8f4b6037559bda34c/packages/typespec-client-generator-core/src/decorators.ts#L485)

I think expose top level client is a seperate requirement that not related to this.

PR #1033, opened 2024-06-18
[tcgc] only expose top level client in SdkPackage

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1040

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -620,7 +620,7 @@ export function getSdkPackage<
   return {
     name: getClientNamespaceString(context)!,
     rootNamespace: getClientNamespaceString(context)!,
-    clients: listClients(context).map(c => diagnostics.pipe(createSdkClientType(context, c))),
+    clients: listClients(context).map((c) => diagnostics.pipe(createSdkClientType(context, c))),
     models: modelsAndEnums.filter((x): x is SdkModelType => x.kind === "model"),
     enums: modelsAndEnums.filter((x): x is SdkEnumType => x.kind === "enum"),
     diagnostics: diagnostics.diagnostics,
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -573,7 +573,6 @@ function createSdkClientType<
     // eslint-disable-next-line deprecation/deprecation
     arm: client.kind === "SdkClient" ? client.arm : false,
   };
-  context.__clients!.push(sdkClientType);
   return diagnostics.wrap(sdkClientType);
 }
 
@@ -613,7 +612,6 @@ export function getSdkPackage<
   TServiceOperation extends SdkServiceOperation,
 >(context: SdkContext<TOptions, TServiceOperation>): SdkPackage<TServiceOperation> {
   const diagnostics = createDiagnosticCollector();
-  context.__clients = new Array<SdkClientType<TServiceOperation>>();
   populateApiVersionInformation(context);
   const modelsAndEnums = diagnostics.pipe(getAllModelsWithDiagnostics(context));
   const crossLanguagePackageId = diagnostics.pipe(getCrossLanguagePackageId(context));
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -616,14 +616,11 @@ export function getSdkPackage<
   context.__clients = new Array<SdkClientType<TServiceOperation>>();
   populateApiVersionInformation(context);
   const modelsAndEnums = diagnostics.pipe(getAllModelsWithDiagnostics(context));
-  for (const client of listClients(context)) {
-    createSdkClientType(context, client);
-  }
   const crossLanguagePackageId = diagnostics.pipe(getCrossLanguagePackageId(context));
   return {
     name: getClientNamespaceString(context)!,
     rootNamespace: getClientNamespaceString(context)!,
-    clients: Array.from(context.__clients.values()),
+    clients: listClients(context).map(c => diagnostics.pipe(createSdkClientType(context, c))),
     models: modelsAndEnums.filter((x): x is SdkModelType => x.kind === "model"),
     enums: modelsAndEnums.filter((x): x is SdkEnumType => x.kind === "enum"),
     diagnostics: diagnostics.diagnostics,
--- a/packages/typespec-client-generator-core/test/decorators.test.ts
+++ b/packages/typespec-client-generator-core/test/decorators.test.ts
@@ -21,7 +21,13 @@ import {
   shouldGenerateConvenient,
   shouldGenerateProtocol,
 } from "../src/decorators.js";
-import { SdkClientType, SdkHttpOperation, SdkMethodResponse, SdkOperationGroup, UsageFlags } from "../src/interfaces.js";
+import {
+  SdkClientType,
+  SdkHttpOperation,
+  SdkMethodResponse,
+  SdkOperationGroup,
+  UsageFlags,
+} from "../src/interfaces.js";
 import { getCrossLanguageDefinitionId, getCrossLanguagePackageId } from "../src/public-utils.js";
 import { getAllModels } from "../src/types.js";
 import { SdkTestRunner, createSdkContextTestHelper, createSdkTestRunner } from "./test-host.js";
@@ -3306,7 +3312,8 @@ describe("typespec-client-generator-core: decorators", () => {
       strictEqual(clientAccessor.name, "getInterfaceV2");
       deepStrictEqual(clientAccessor.apiVersions, ["v2"]);
 
-      const interfaceV2 = versioningClient.methods.find((x) => x.kind === "clientaccessor")?.response as SdkClientType<SdkHttpOperation>;
+      const interfaceV2 = versioningClient.methods.find((x) => x.kind === "clientaccessor")
+        ?.response as SdkClientType<SdkHttpOperation>;
       ok(interfaceV2);
       strictEqual(interfaceV2.methods.length, 1);
 
--- a/packages/typespec-client-generator-core/test/decorators.test.ts
+++ b/packages/typespec-client-generator-core/test/decorators.test.ts
@@ -21,7 +21,7 @@ import {
   shouldGenerateConvenient,
   shouldGenerateProtocol,
 } from "../src/decorators.js";
-import { SdkMethodResponse, SdkOperationGroup, UsageFlags } from "../src/interfaces.js";
+import { SdkClientType, SdkHttpOperation, SdkMethodResponse, SdkOperationGroup, UsageFlags } from "../src/interfaces.js";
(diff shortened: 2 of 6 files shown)


packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3284,7 +3290,7 @@ describe("typespec-client-generator-core: decorators", () => {
         `
       );
       const sdkPackage = runner.context.experimental_sdkPackage;
-      strictEqual(sdkPackage.clients.length, 2);
+      strictEqual(sdkPackage.clients.length, 1);
```
qiaozha, line 3293: Is the clients.length 2 before because we treated namespace Versioning and interface InterfaceV2 as the sub clients ?  
I am curious what the expected client constructor for IntefaceV2 is now that it's a sub client for Versioning?

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3306,7 +3312,8 @@ describe("typespec-client-generator-core: decorators", () => {
       strictEqual(clientAccessor.name, "getInterfaceV2");
```
qiaozha, line 3312: should we add linter rule for the name if it's going to be a subclient of a client? Another question is since we are using interface name as the sub client name, would it be confusing to add Client suffix if we do add that linter rule? 
Also, how can we tell if we want this interface to be a subclient of the current client of just an operation group of the current client?

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3306,7 +3312,8 @@ describe("typespec-client-generator-core: decorators", () => {
...
       deepStrictEqual(clientAccessor.apiVersions, ["v2"]);
 
-      const interfaceV2 = sdkPackage.clients.find((x) => x.name === "InterfaceV2");
+      const interfaceV2 = versioningClient.methods.find((x) => x.kind === "clientaccessor")
+        ?.response as SdkClientType<SdkHttpOperation>;
```
qiaozha, line 3316: in the case that InterfaceV2 is a sub client of VersioningClient, how do we expect customer to initalize the client InterfaceV2 ? is it something like 
`new VersioningClient("endpoint").getInterfaceV2()` or `new VersioningClient().getInterfaceV2("endpoint")` I wonder if we need to have a way to specify where the initialization properties should belong to ?

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -620,12 +622,17 @@ describe("typespec-client-generator-core: package", () => {
...
-      const barClient = sdkPackage.clients.filter((c) => c.name === "Bar")![1];
+      const mainClient = sdkPackage.clients[0];
+      const fooClient = mainClient.methods.find(
+        (m) => m.kind === "clientaccessor" && m.name === "getFoo"
+      )?.response as SdkClientType<SdkHttpOperation>;
+      const fooBarClient = fooClient.methods.find((m) => m.kind === "clientaccessor")
+        ?.response as SdkClientType<SdkHttpOperation>;
+      const barClient = mainClient.methods.find(
+        (m) => m.kind === "clientaccessor" && m.name === "getBar"
+      )?.response as SdkClientType<SdkHttpOperation>;
```
qiaozha, line 635: the interface Bar inside namespace Foo is also called getBar? will it be problematic for naming conflicts with the client options?

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2671,10 +2677,11 @@ describe("typespec-client-generator-core: package", () => {
       `
       );
       const sdkPackage = runnerWithCore.context.experimental_sdkPackage;
-      strictEqual(sdkPackage.clients.length, 2);
+      strictEqual(sdkPackage.clients.length, 1);
```
qiaozha, line 2680: curious why the client.length is 2 before in this case?

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -616,14 +616,11 @@ export function getSdkPackage<
...
   const modelsAndEnums = diagnostics.pipe(getAllModelsWithDiagnostics(context));
-  for (const client of listClients(context)) {
-    createSdkClientType(context, client);
-  }
   const crossLanguagePackageId = diagnostics.pipe(getCrossLanguagePackageId(context));
   return {
     name: getClientNamespaceString(context)!,
     rootNamespace: getClientNamespaceString(context)!,
-    clients: Array.from(context.__clients.values()),
+    clients: listClients(context).map((c) => diagnostics.pipe(createSdkClientType(context, c))),
```
qiaozha, line 623: maybe we should remove context.__clients ? I don't see where this value being used somewhere else, maybe I miss something ?

qiaozha (commented):
I am curious how `@client` decorator would impact the listClients result and the hierachy structure between those clients?

packages/typespec-client-generator-core/test/decorators.test.ts
tadelesh: yes. previously, we will expose all the sub clients. for now, versioning client has a method `getInterfaceV2` which response is the interfacev2 client. since the client's initialization of interfacev2 is private, it should only be initialized with its parent client.

packages/typespec-client-generator-core/test/decorators.test.ts
tadelesh: what's lint rule? i don't quite get you. this name is the method name, not the client name. there is another issue to track adding `groupPath` in legacy `SdkOperationGroup` type for the new interface. And operation group in the new interface is also a client, but with private initialization.

packages/typespec-client-generator-core/test/decorators.test.ts
tadelesh: if the client has no public initialization, it should be initialized from parent.

tadelesh: this will be a tom requirement.

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: in previous reply, i think we need to find a way to show the path, so emitter could have an easy way to deal with this.

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: see this:
```
@route("/widgets")
    @tag("Widgets")
    interface Widgets {
      ${code}
    }`);
    }
```

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2953,7 +2960,9 @@ describe("typespec-client-generator-core: package", () => {
         sdkPackage.models.map((x) => x.name).sort(),
         ["CheckupCollectionWithNextLink", "Checkup", "PetStoreError", "CheckupUpdate"].sort()
       );
-      const createOrUpdate = sdkPackage.clients[0].methods[0];
+      const client = sdkPackage.clients[0].methods.find((x) => x.kind === "clientaccessor")
+        ?.response as SdkClientType<SdkHttpOperation>;
```
haolingdong-msft, line 2964: Do we have test case for non-initializable subclients? Just curious what will `sdkPackage.clients` look like and how to initialize subclient.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -573,7 +573,6 @@ function createSdkClientType<
     // eslint-disable-next-line deprecation/deprecation
     arm: client.kind === "SdkClient" ? client.arm : false,
   };
-  context.__clients!.push(sdkClientType);
```
haolingdong-msft, line 576: curious why do we delete this?

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -21,7 +21,13 @@ import {
   shouldGenerateConvenient,
   shouldGenerateProtocol,
 } from "../src/decorators.js";
-import { SdkMethodResponse, SdkOperationGroup, UsageFlags } from "../src/interfaces.js";
+import {
+  SdkClientType,
+  SdkHttpOperation,
+  SdkMethodResponse,
+  SdkOperationGroup,
```
haolingdong-msft, line 28: seems we will not use `SdkOperationGroup` in `sdkPackage.clients`, do we still need this import?

haolingdong-msft:
Thanks for the quick action! Overall looks good to me, just leave a few minor  comments.

packages/typespec-client-generator-core/test/package.test.ts
tadelesh: i think all current sub clients are non-initializable, they all have initialization with private access.

packages/typespec-client-generator-core/src/package.ts
tadelesh: it is useless to cache clients since we only expose top level in the sdkpacakge.

packages/typespec-client-generator-core/test/decorators.test.ts
tadelesh: this import is for other tests of `@operationGroup` decorator.

packages/typespec-client-generator-core/test/package.test.ts
haolingdong-msft: Yeah, but TCGC should consider this case as well, we can add those cases when we met.

packages/typespec-client-generator-core/test/decorators.test.ts
haolingdong-msft: ok, for `listClients()`?

haolingdong-msft:
PS: Please make sure that merging this pr will not close this https://github.com/Azure/typespec-azure/issues/963. Thanks.
image

packages/typespec-client-generator-core/test/decorators.test.ts
iscai-msft: we still have the `@operationGroup` decorator so we still need to support it with tests

packages/typespec-client-generator-core/test/package.test.ts
iscai-msft: non-initializable subclients are operation groups, which is what we currently support. we have a bunch of tests for this rn, so no need to add more

packages/typespec-client-generator-core/test/package.test.ts
iscai-msft: we are only listing top-level clients in `.clients` now, that's the change we are making in this PR. There is still the interior client, it's just going to be listed as a subclient on the top-level client

packages/typespec-client-generator-core/test/decorators.test.ts
iscai-msft: `getInterfaceV2` isn't the name of the subclient, it's the name of the client accessor method. We have linter rules when people use the `@client` decorator

iscai-msft (approved):

PR #1038, opened 2024-06-18
[tcgc] allow csv list of scopes

### LINKED ISSUE #978: Accept scope as a list

iscai-msft:
fixes #978


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -77,8 +78,9 @@ function setScopedDecoratorData(
   const targetEntry = context.program.stateMap(key).get(target);
   // If target doesn't exist in decorator map, create a new entry
```
tadelesh, line 79: could you check the logic of target exists? i think that branch does not handle the scope with `,`

tadelesh (approved):

PR #1045, opened 2024-06-19
filter api versions for GA

### LINKED ISSUE #965: [tcgc] provide an API to list all supported api-versions for a SDK
Now when we support versioning, .NET and Java need to generate [ServiceVersion in client](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-versioning-select-api-version). It may not be same as the list of api-versions in main.tsp, in released SDK, the ServiceVersion only contains stable api-versions.

The supported api-versions should follow below rules:

- The api-versions not later than the SDK target api-version, by default the SDK target api-version is the latest one.
- Comply with [Azure API guideline](https://azure.github.io/azure-sdk/general_design.html#service-api-versions): so

1. A stable version of client library(SDK) will target the latest GA api-version by default, and only contains GA api-versions in `ServiceVersion` enum
2. A public beta version(preview version) of client library will target the latest public preview API version by default. It contains all api-versions in `ServiceVersion` enum.

Those rules are for all languages, so it is better for TCGC to provide an API to list out all the supported Api-version for a SDK.

iscai-msft:
fixes #965


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -36,6 +36,7 @@ export interface SdkEmitterOptions {
   "package-name"?: string;
   "flatten-union-as-enum"?: boolean;
   "api-version"?: string;
+  "flavor"?: "azure";
```
tadelesh, line 39: since it is an emitter behavior to filtering preview version, i think we could add an option when create sdk context which could infer how to do preview filtering with regex. that could be more flexible i believe.

iscai-msft: what are your suggestions? I'm not sure I'm following

tadelesh: i suggest to add an option [here](https://github.com/Azure/typespec-azure/blob/0b1afda6a4531f1e0e458e34538958f890129810/packages/typespec-client-generator-core/src/decorators.ts#L580). the option is a regex for filtering preview version.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -432,3 +434,23 @@ export function getAnyType(): SdkBuiltInType {
...
+    context.apiVersion !== undefined &&
+    context.apiVersion !== "latest" &&
+    context.apiVersion !== "all"
+  ) {
+    const index = sdkVersionsEnum.values.findIndex((v) => v.value === context.apiVersion);
+    if (index >= 0) {
+      sdkVersionsEnum.values = sdkVersionsEnum.values.slice(0, index + 1);
+    }
+  }
+  const defaultApiVersion = getDefaultApiVersion(context, client.service);
```
chunyu3, line 452: Is the `defaultApiVersion` not the latest item of `sdkVersionsEnum`? We may not need to  retrieve again from service?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -432,3 +434,23 @@ export function getAnyType(): SdkBuiltInType {
...
+  if (!defaultApiVersion?.value.endsWith("-preview")) {
```
chunyu3, line 453: This is the rule for `Azure`. Shall we filter out the non-azure scenario?

packages/typespec-client-generator-core/src/interfaces.ts
chunyu3: Maybe we need to define
```
interface CreateSdkContextOptions {
  readonly versionStrategy?: {
    strategy: "ignore|filter",
    filterFunc: (defaultApiversion: string, versions: VersionEnum[]) => void
  }
}
```

packages/typespec-client-generator-core/src/internal-utils.ts
iscai-msft: @chunyu3 we need to use the output from `defaultApiVersion` before we create our `sdkVersionsEnum`, since we use it to create the enum. Would be nice if we could remove it, but I don't think we can

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: i think that could work, and we could have the default regex value be `"-preview"`

iscai-msft: @chunyu3 that would be a breaking change, but I do like the idea of grouping the versioning strategy together. I looked at it seems no one is using this option. I'm going to go ahead and try to follow your suggestion, and people can leave comments on it

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1572,6 +1573,7 @@ export function getAllModelsWithDiagnostics(
       const sdkVersionsEnum = diagnostics.pipe(
         getSdkEnumWithDiagnostics(context, versionMap.getVersions()[0].enumMember.enum)
       );
+      filterApiVersionsInEnum(context, client, sdkVersionsEnum);
       if (
```
tadelesh, line 1577: duplicate code

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -530,6 +533,30 @@ export function getHttpOperationResponseHeaders(
   return headers;
 }
 
+export function filterApiVersionsInEnum(
```
tadelesh, line 536: it seems this func has many duplicate code with `getDefaultApiVersion`, can we do soem consolidation?

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3472,6 +3472,73 @@ describe("typespec-client-generator-core: decorators", () => {
...
+        `
+      );
+      const sdkVersionsEnum = runner.context.sdkPackage.enums[0];
+      strictEqual(sdkVersionsEnum.name, "Versions");
+      strictEqual(sdkVersionsEnum.usage, UsageFlags.ApiVersionEnum);
+      strictEqual(sdkVersionsEnum.values.length, 1);
+      strictEqual(sdkVersionsEnum.values[0].value, "2024-10-01");
+    });
+    it("default latest preview version with GA", async () => {
+      await runner.compile(
```
haolingdong-msft, line 3509: I'm a bit confused. What is the difference between the two test cases? Seems the versions are the same? And I did not find the version set into emitter options.

iscai-msft: we assign value to the ordering, because we expect users to input versioning in the correct order. In the first one, I inputted a preview version that occurred before the GA version, while in the second test, I put in a preview version that occurs after the GA version.

Since we default the api version to be the last in the list, this changes the default api version from a GA one in the first test, to a preview one in the second test. Therefore in the first test, we only expose a GA version, while in the second test, we expose both the preview and the GA version

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -530,6 +533,38 @@ export function getHttpOperationResponseHeaders(
   return headers;
 }
 
+export function removeVersionsPreviousToExplicitlySpecified(
```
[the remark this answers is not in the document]
iscai-msft: no if you explicitly specified version `v3`, this would remove versions `v1` and `v2`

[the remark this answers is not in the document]
chunyu3: API: `Array.splice(start, deleteCount)` - means remove `deleteCount` items from array start from the item `start`
So
`versions.splice(index + 1, versions.length - index - 1);` means remove `versions.length - index - 1` items from index `index + 1`

So seems in this function, it should remove the versions after specified version.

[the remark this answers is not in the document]
tadelesh: yes, it is used to remove all the version bigger than specific version. let me change the name.

packages/typespec-client-generator-core/test/decorators.test.ts
haolingdong-msft: I see, do we want to add test cases where user explicitly set API version in emitter options?

tadelesh (approved):

chunyu3 (approved):

Issue #1049, opened 2024-06-21
[Bug]: [TCGC][ARM] client customization in client.tsp for ARM does not have sdkContext.arm=true


weidongxu-microsoft:
### Describe the bug

Use case https://gist.github.com/weidongxu-microsoft/fa420c35cf0611a62a8d155771dd3425?permalink_comment_id=5095904#gistcomment-5095904

Customization in client.tsp for ARM would start with

```ts
@client({
  name: "AvsClient", 
  service: Microsoft.AVS
})
@useDependency(Microsoft.AVS.Versions.v2023_09_01)
namespace Customizations;
```

`sdkContext.arm` from TCGC would be false in this case.

PS: we cannot add `@armProviderNamespace("Microsoft.AVS")` on namespace in client.tsp, it would cause error
```
error: @azure-tools/typespec-azure-resource-manager/single-arm-provider - Only one @armProviderNamespace can be declared in a typespec spec at once. @ C:/github/azure-sdk-for-java/sdk/avs/azure-resourcemanager-avs/TempTypeSpecFiles/Microsoft.AVS/client.tsp
```

### Reproduction

Use client.tsp like https://gist.github.com/weidongxu-microsoft/fa420c35cf0611a62a8d155771dd3425?permalink_comment_id=5095904#gistcomment-5095904 in https://github.com/Azure/azure-rest-api-specs/tree/main/specification/vmware/Microsoft.AVS

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1050, opened 2024-06-21
tcgc, arm, set context.arm for explicit client 

weidongxu-microsoft:
cause: `SdkContext.arm` is not set, in the case of explicit client.

fix https://github.com/Azure/typespec-azure/issues/1049


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -282,6 +282,9 @@ export function listClients(context: TCGCContext): SdkClient[] {
   const explicitClients = [...listScopedDecoratorData(context, clientKey)];
   if (explicitClients.length > 0) {
     context.__rawClients = getClientsWithVersioning(context, explicitClients);
+    for (const client of context.__rawClients) {
+      context.arm = isArm(client.service);
+    }
```
weidongxu-microsoft, line 287: Here uses the same logic as 
https://github.com/Azure/typespec-azure/blob/45c186059741b396d451099548969a8e1671ae7a/packages/typespec-client-generator-core/src/decorators.ts#L294-L303

Assuming we won't have mixed service (arm and non-arm).

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -282,6 +282,9 @@ export function listClients(context: TCGCContext): SdkClient[] {
   const explicitClients = [...listScopedDecoratorData(context, clientKey)];
   if (explicitClients.length > 0) {
     context.__rawClients = getClientsWithVersioning(context, explicitClients);
+    for (const client of context.__rawClients) {
+      context.arm = isArm(client.service);
```
tadelesh, line 292: I think if one client is arm, then context.arm should be arm.

tadelesh (approved):

PR #1065, opened 2024-06-24
[CanonicalEmitter] Add option and isArmCommonType check

dinwa-ms:
This PR addressed 2 issues:
1. add option "use-read-only-status-schema" to fix ProvisioningStateMustBeReadOnly bug; 
The bug is mentioned here https://github.com/Azure/azure-rest-api-specs/discussions/29471 and the suggested walkaround is set option "use-read-only-status-schema" to true.
Canonical emitter doesn't have this option. So add this option to the emitter.
image

2. add isArmCommonType check to avoid decorator validation in canonical emitter
Canonical emitter will generate warning when @typeChangedFrom decorator is used. To make sure this not apply to common type, add isArmCommonType check


markcowl (approved):
I think I would remove the optionalk altogether as dicussed.  Also, looks like you need to remove the test that checks for behavior when read-onlly-status-schema is turned off.

Otherwise, looks good.

PR #1068, opened 2024-06-24
remove unused isErrorOrChildOfError

iscai-msft:
Right after we added this function, `isErrorModel` from `@typespec/compiler` would check inheritance for error. I've checked and no one uses this. Since it was never used, I'm going to go ahead and remove it


tjprescott (approved):

catalinaperalta (approved):
Looks like you also need to clean up an import

PR #1069, opened 2024-06-24
[tcgc] add `.Error` to `UsageFlags`

### LINKED ISSUE #917: [TCGC] Support 'Exception' usage on `SdkModelType`/`SdkEnumType`
Java needs to know if the model's usage is exception or not, e.g. not generate exception model. Currently we did special handling in emitter to see if a model is exception or not. But we should not track usage by ourselves after migrating to use TCGC usage. So it's better that TCGC can provide an `Exception` usage.

Our handling in emitter: https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L1643-L1650

iscai-msft:
fixes #917


tadelesh (dismissed):

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1285,6 +1285,9 @@ function updateTypesFromOperation(
         );
         if (generateConvenient) {
           updateUsageOfModel(context, UsageFlags.Output, sdkType);
+          if (sdkType.kind === "model" && isErrorModel(context.program, innerResponse.body.type)) {
+            updateUsageOfModel(context, UsageFlags.Error, sdkType);
+          }
```
tadelesh, line 1290: is this the best place to set error usage? if we just want to label a model with error usage only when the model has `@error` annotation. we should do this in the end of `getAllModels` or during `getSdkModelType` which could fully cover any corner case.

iscai-msft: actually the reason I put it here was because I only wanted to set something to usage `error` if it was decorated with an `@error` decorator and explicitly used as an error output. I'm thinking that might be too complicated though, so I'm ok with just adding usage is error if the `@error` decorator was applied. Let me know your thoughts

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -600,4 +596,6 @@ export enum UsageFlags {
   MultipartFormData = 1 << 5,
   // Used in spread.
   Spread = 1 << 6,
+  // Output will also be set when Error is set.
+  Error = 1 << 6,
```
tadelesh, line 600:
```suggestion
  // Output will also be set when Error is set.
  Error = 1 << 7,
```

packages/typespec-client-generator-core/test/types/model-types.test.ts
```diff
@@ -1384,11 +1385,15 @@ describe("typespec-client-generator-core: model types", () => {
     const models = getAllModels(runner.context);
     strictEqual(models.length, 1);
     strictEqual(models[0].kind, "model");
-    strictEqual(models[0].isError, true);
-    const rawModel = models[0].__raw;
+    strictEqual((models[0].usage & UsageFlags.Error) > 0, true);
```
tadelesh, line 1388:
```suggestion
    ok(models[0].usage & UsageFlags.Error);
```

tadelesh (approved):

PR #1076, opened 2024-06-25
[tcgc] add example model in tcgc

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/974

design doc: https://gist.github.com/tadelesh/a61883724a5cd072ef18f500826c7a7f


packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,630 @@
...
+  const diagnostics = createDiagnosticCollector();
+
+  const examples = diagnostics.pipe(await loadExamples(context, client));
+  const methodQueue = [...client.methods];
+  while (methodQueue.length > 0) {
+    const method = methodQueue.pop()!;
+    if (method.kind === "clientaccessor") {
+      methodQueue.push(...method.response.methods);
+    } else {
+      const operationId = resolveOperationId(context.program, method.__raw!).toLowerCase();
```
tadelesh, line 151: one concern here is whether tcgc need to see `@example` defined in `typespec-autorest`.

tadelesh:
this pr contains breaking change: `createSdkContext` change from sync to async.

packages/typespec-autorest/src/emit.ts
```diff
@@ -98,7 +98,7 @@ export async function getAllServicesAtAllVersions(
   program: Program,
   options: ResolvedAutorestEmitterOptions
 ): Promise<AutorestServiceRecord[]> {
-  const tcgcSdkContext = createSdkContext(
+  const tcgcSdkContext = await createSdkContext(
```
iscai-msft, line 101: won't this be breaking for all language emitters because we now have to add async? It's a small break, but I think we should call it out as such

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,630 @@
...
+  if (!context.examplesDirectory) {
+    return diagnostics.wrap(new Map());
+  }
+
+  const allApiVersions = client.apiVersions;
+  let apiVersion = context.apiVersion;
+  if (apiVersion === "all") {
+    return diagnostics.wrap(new Map());
+  }
+  if (apiVersion === "latest" || apiVersion === undefined || !allApiVersions.includes(apiVersion)) {
```
iscai-msft, line 60: can we make this a helper function, since I think we're doing similar apiVersion checks throughout the code

packages/typespec-client-generator-core/src/example.ts
iscai-msft: is it a concern because we can't use decorators from `typespec-autorest`? If so, the `.decorators` PR should resolve this issue, right?

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,630 @@
...
+  const diagnostics = createDiagnosticCollector();
+
+  if (isSdkIntKind(type.kind) || isSdkFloatKind(type.kind)) {
+    return getSdkBaseTypeExample("number", type as SdkType, example, relativePath);
+  } else if (type.kind === "string" || type.kind === "bytes") {
+    return getSdkBaseTypeExample("string", type as SdkType, example, relativePath);
+  } else if (type.kind === "boolean") {
+    return getSdkBaseTypeExample("boolean", type as SdkType, example, relativePath);
+  } else if (
+    type.kind === "password" ||
```
iscai-msft, line 357: since @ArcturusZhang is working on changing the kinds here, I would recommend going based off of enums and typechecking functions instead of directly listing them out. Just need to make sure that these changes are consistent with the changes that @ArcturusZhang is introducing

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,630 @@
...
+        typeof type.value as "string" | "number" | "boolean",
+        type,
+        example,
+        relativePath
+      );
+    } else {
+      addExampleValueNoMappingDignostic(diagnostics, example, relativePath);
+      return diagnostics.wrap(undefined);
+    }
+  } else if (type.kind === "enum") {
```
iscai-msft, line 400: i think using a `switch` statement instead of a bunch of `else if`s would be cleaner

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -627,3 +628,13 @@ export function getHttpOperationWithCache(
   context.httpOperationCache.set(operation, httpOperation);
   return httpOperation;
 }
+
+/**
+ * Get the examples for a given http operation.
+ */
+export function getHttpOperationExamples(
```
iscai-msft, line 635: do we want to expose examples through a helper function, or make them a property on our sdk operation / sdk method types?

iscai-msft: oh wait, that's already the case. Why are we exposing `getHttpOperationExamples` in `public-utils` then?

iscai-msft (commented):
Looks amazing, thank you so much for your hard work @tadelesh

packages/typespec-autorest/src/emit.ts
tadelesh: yes. it will break all emitters. also, i don't think there is a good way to work around this breaking. i will let all impacted emitter be aware of this before merge.

packages/typespec-client-generator-core/src/example.ts
tadelesh: yes. that will make it possible to get example value without import the lib. thanks for reminder.

packages/typespec-client-generator-core/src/example.ts
tadelesh: i'll update the logic after his pr merged.

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: js rlc does not use tcgc common interface. also, .net and java want to get example before they fully switch to tcgc operation interface. this helper is for them.

packages/typespec-client-generator-core/src/example.ts
tadelesh: i went through all the current spec and it seems there is no actual usage for `@example`. i decided to leave it for now.

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,623 @@
...
+interface LoadedExample {
+  readonly relativePath: string;
+  readonly file: SourceFile;
+  readonly data: any;
+}
+
+async function loadExamples<TServiceOperation extends SdkServiceOperation>(
+  context: TCGCContext,
+  client: SdkClientType<TServiceOperation>
+): Promise<[Map<string, Record<string, LoadedExample>>, readonly Diagnostic[]]> {
```
ArcturusZhang, line 49: well it is confusing if an api is returning a map.
How about we return an array of something?
Same comment to the `Record<string, LoadedExample>>`

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -599,3 +601,105 @@ export enum UsageFlags {
   // Input will also be set when MultipartFormData is set.
   MultipartFormData = 1 << 5,
 }
+
+export type SdkExampleBase = {
```
ArcturusZhang, line 605: why this is type not interface?

ArcturusZhang: also should this be exported? looks like an internal thing.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -599,3 +601,105 @@ export enum UsageFlags {
...
+export interface SdkExampleTypeBase {
+  kind: string;
+  type: SdkType | SdkModelPropertyType;
+  value: unknown;
+}
+
+export interface SdkStringExample extends SdkExampleTypeBase {
+  kind: "string";
+  type: SdkBuiltInType | SdkDateTimeType | SdkDurationType | SdkEnumType | SdkConstantType;
+  value: string;
```
ArcturusZhang, line 655: the type here does not really make sense, because builtintype also includes the numbers.

Could this work?
```suggestion
  type: SdkBuiltInType & { kind: "string" } | SdkDateTimeType | SdkDurationType | SdkEnumType | SdkConstantType;
  value: string;
```

ArcturusZhang: same for the numbers and boolean.

ArcturusZhang: if we have my PR on scalars merged, here could be more clear because those based on string types will always have the kind of `string` (such as `azurelocation`)

tadelesh: i just want to list all possible types here. i believe i could have some optimization after your pr merged.

packages/typespec-client-generator-core/src/example.ts
tadelesh: the map is for the efficiency for finding the examples of an operation. since it is not exported, i just added some doc for it.

packages/typespec-autorest/src/emit.ts
weidongxu-microsoft: Yeah, java calls it in `constructor` :-(

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,631 @@
...
+}
+
+export async function handleClientExamples(
+  context: TCGCContext,
+  client: SdkClientType<SdkServiceOperation>
+): Promise<[void, readonly Diagnostic[]]> {
+  const diagnostics = createDiagnosticCollector();
+
+  const examples = diagnostics.pipe(await loadExamples(context, client));
+  const methodQueue = [...client.methods];
```
weidongxu-microsoft, line 146: Be careful on the `client.methods` and its operationId.

When client.tsp defines the client or operation group, the operationId could be the `<interface>_<op>` in client.tsp, instead of routes.tsp.

Just make sure we've tested such case.

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,631 @@
...
+ * Load all examples for a client
+ * 
+ * @param context 
+ * @param client 
+ * @returns a map of all operations' examples, key is operation's operation id, 
+ * value is a map of examples, key is example's title, value is example's details
+ */
+async function loadExamples<TServiceOperation extends SdkServiceOperation>(
+  context: TCGCContext,
+  client: SdkClientType<TServiceOperation>
```
weidongxu-microsoft, line 56: This method seems only use `apiVersion` from the `client`. It seems misleading to send a `client`.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -599,3 +601,105 @@ export enum UsageFlags {
   // Input will also be set when MultipartFormData is set.
   MultipartFormData = 1 << 5,
 }
+
+interface SdkExampleBase {
+  kind: string;
+  name: string;
```
weidongxu-microsoft, line 607: So, all ##Example is indexed by this "name=title"?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -517,3 +525,55 @@ export function getAnyType(
...
+ * Resolve the OpenAPI operation ID for the given operation using the following logic:
+ * - If @operationId was specified use that value
+ * - If operation is defined at the root or under the service namespace return `<operation.name>`
+ * - Otherwise(operation is under another namespace or interface) return `<namespace/interface.name>_<operation.name>`
+ *
+ * @param context TCGC context
+ * @param operation Operation
+ * @returns Operation ID in this format `<name>` or `<group>_<name>`
+ */
+export function resolveOperationId(context: TCGCContext, operation: Operation) {
```
weidongxu-microsoft, line 546: Any difference from `resolveOperationId` in typspec-openapi ?

weidongxu-microsoft:
One pending issue is about the filename. Since now we checkout tsp into Temp folder, we are not able to get a relative path (starts from `/specification/...`). Though this would be a much later issue.

packages/typespec-client-generator-core/src/example.ts
tadelesh: what do you mean misleading? it is not exported and only for internal logic.

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: oh, forgot to removed.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: yes. the name is the title of the example.

packages/typespec-autorest/src/emit.ts
tadelesh: will it be a problem for java to change to async?

weidongxu-microsoft: Probably not blocking problem. Just need to make some (non-trivial) refactor.

packages/typespec-client-generator-core/src/example.ts
weidongxu-microsoft: This is just a comment on dev for method, as e.g. a string can get the job done, why sending a whole client.
It kind of making the method hard to understand (what's it do with the client?).

Just a dev perspective about code readability.

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,637 @@
...
+  parameters: SdkHttpParameter[],
+  example: any,
+  relativePath: string
+): [SdkHttpParameterExample[], readonly Diagnostic[]] {
+  const diagnostics = createDiagnosticCollector();
+  const parameterExamples = [] as SdkHttpParameterExample[];
+  if ("parameters" in example && typeof example.parameters === "object") {
+    for (const name of Object.keys(example.parameters)) {
+      const parameter = parameters.find(
+        (p) => (p.kind !== "body" && p.serializedName === name) || p.name === name
```
weidongxu-microsoft, line 231: nit, just a suggestion, no action needed.

I assume you do `p.name === name` for BODY.

But in TypeSpec, now it is much harder to know the name of parameter of BODY.

Java implement would do a fallback to "body" in example JSON, as this is most observed.

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,637 @@
...
+  const diagnostics = createDiagnosticCollector();
+  if (!context.examplesDirectory) {
+    return diagnostics.wrap(new Map());
+  }
+
+  const apiVersion = getValidApiVersion(context, client.apiVersions);
+
+  const exampleDir = apiVersion
+    ? resolvePath(context.examplesDirectory, apiVersion)
+    : resolvePath(context.examplesDirectory);
```
weidongxu-microsoft, line 67: This may fail in SDK repo.

Java did a fallback to this
```ts
        exampleDir = version
          ? resolvePath(program.projectRoot, examplesDirectory, version)
          : resolvePath(program.projectRoot, examplesDirectory);
```

tadelesh: i suppose the config should be `"{project-root}/examples"`?

weidongxu-microsoft: Yeah, `"{project-root}/examples"` would also do.

But from what I saw on tspconfig, when writing this to typespec-autorest, service would just write "examples".

tadelesh: will java fail to get example for `"{project-root}/examples"` according to your code snippet?

weidongxu-microsoft: Yes, we do fallback to the that equivalent to `{project-root}/{whatever-dev-write-ion-exampleDirectory}`.

I am also fine that we require dev to write `"{project-root}/examples"`

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -621,10 +615,15 @@ export function createSdkContext<
     __namespaceToApiVersionParameter: new Map(),
     __tspTypeToApiVersions: new Map(),
     __namespaceToApiVersionClientDefaultValue: new Map(),
+    examplesDirectory: context.options["examples-directory"],
     decoratorsAllowList: [...defaultDecoratorsAllowList, ...(options?.additionalDecorators ?? [])],
     previewStringRegex: options?.versioning?.previewStringRegex || /-preview$/,
   };
   sdkContext.sdkPackage = getSdkPackage(sdkContext);
+  for (const client of sdkContext.sdkPackage.clients) {
+    diagnostics.pipe(await handleClientExamples(sdkContext, client));
```
MaryGao, line 624: Will any exception in handleClientExamples break the createSdkContext?

Does that mean if we don't care about example we don't need to wait this promise?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -604,3 +606,105 @@ export enum UsageFlags {
...
+
+export interface SdkHttpParameterExample {
+  parameter: SdkHttpParameter;
+  value: SdkTypeExample;
+}
+
+export interface SdkHttpResponseExample {
+  response: SdkHttpResponse;
+  headers: SdkHttpResponseHeaderExample[];
+  value?: SdkTypeExample;
```
MaryGao, line 632: Is the value here body example value?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -604,3 +606,105 @@ export enum UsageFlags {
...
+  kind: "union";
+  type: SdkUnionType;
+  value: unknown;
+}
+
+export interface SdkModelExample extends SdkExampleTypeBase {
+  kind: "model";
+  type: SdkModelType;
+  value: Record<string, SdkTypeExample>;
+  additionalProperties?: Record<string, SdkTypeExample>;
```
MaryGao, line 709: why additionalProperties

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: all exception will be in diagnostics, will not block `createSdkContext`

the example will be added in each operation, so it is impossible to do sync way.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: want to keep consistent for all other example type, value is represent the example value.

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -0,0 +1,641 @@
...
+
+      if (!map.has(example.operationId.toLowerCase())) {
+        map.set(example.operationId.toLowerCase(), {});
+      }
+      const examples = map.get(example.operationId.toLowerCase())!;
+
+      if (example.title in examples) {
+        diagnostics.add(
+          createDiagnostic({
+            code: "duplicate-example-file",
```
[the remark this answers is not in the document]
tadelesh: for swagger, title is the map key, so title should be unique for different examples

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: it represent any additional value for model with additional properties.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: sync offline, will refine the interface.

weidongxu-microsoft (approved):

MaryGao (approved):

msyyc (approved):

PR #1081, opened 2024-06-26
Fix ArmResourcePatch templates and incorporate common-types updates Fix #859

### LINKED ISSUE #859: Fix ArmResourcePatch* templates to use the resource directly in request body
- The ArmResourcePatch* templates do not accommodate resources with a required properties bag.  They cause an error 'properties' already exist when passed in a resource with a required properties bag. Need to open this up to allow BrownField resources.

markcowl:


markcowl:
The typespec-next PR for this is here: https://github.com/Azure/azure-rest-api-specs/pull/29617

was able to make these changes with a search and replace and then a tsp format:

```
search: ^    azure-resource-provider-folder: "resource-manager"
replace: 
    azure-resource-provider-folder: "resource-manager"
    emit-common-types-schema: "reference-only"
in: specification/**/tspconfig.yaml

search: ArmResourcePatch(Sync|Async)(\s*)<([\s\r\n]*)(\w+)([\s\r\n]*),([\s\r\n]*)(\w+)
replace:ArmCustomPatch$1$2<$3$4$5,$6Azure.ResourceManager.Foundations.ResourceUpdateModel<$4, $7>
in specification/**/*.tsp
```

.chronus/changes/fix-patch-2024-5-13-3-36-28.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+Fix #383: Generate patch models for common-types
```
timotheeguerin, line 8:
```suggestion
Fix generate patch models for common-types
```
not sure this will render nicely in the changelog and chronus will auto link PRs already

.chronus/changes/resource-patch-2024-5-27-2-24-51.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Fix ArmResourcePatch templates and incorporate common-types updates Fix #859
```
timotheeguerin, line 8:
```suggestion
Fix ArmResourcePatch templates and incorporate common-types updates
```

.vscode/launch.json
```diff
@@ -144,6 +144,26 @@
         "order": 2
       }
     },
+    {
```
timotheeguerin, line 147: was this just use for testing?

packages/samples/test/output/azure/resource-manager/arm-library/@azure-tools/typespec-autorest/openapi.json
```diff
@@ -1764,6 +1764,167 @@
         }
       }
     },
+    "Azure.ResourceManager.CommonTypes.ManagedServiceIdentityType": {
```
timotheeguerin, line 1767: is this expected to not use the common type anymore? is it because it should be a "patch" model with everything optional?

packages/typespec-autorest/src/lib.ts
```diff
@@ -193,6 +199,14 @@ const EmitterOptionsSchema: JSONSchemaType<AutorestEmitterOptions> = {
       description:
         "Determine whether and how to emit x-ms-long-running-operation-options for lro resolution",
     },
+    "emit-common-types-schema": {
+      type: "string",
+      enum: ["reference-only", "for-visibility-changes"],
```
timotheeguerin, line 204:
```suggestion
      enum: ["never", "for-visibility-changes"],
```

I think that makes more sense with the option name, alternatively could call teh option: `common-types-schema`

packages/samples/test/output/azure/resource-manager/arm-library/@azure-tools/typespec-autorest/openapi.json
timotheeguerin: forgot to delete

pnpm-lock.yaml
timotheeguerin, line 1: feel like there shouldn't be change to this file at all, maybe was a previous merge issue

core
timotheeguerin, line 1: probably also shouldn't need change here

packages/typespec-autorest/src/lib.ts
markcowl: I was thinking that we could use 'all' as a setting for actually emitting the common types schema in the future, but 'never' makes sense as well.

allenjzhang (approved):

PR #1087, opened 2024-06-26
Publish/tcgc

iscai-msft:


tadelesh (approved):

markcowl (commented):
The only potential problem is that, since the main branch has had changes since the release, if any API changes from other packages were consumed in TCGC, they won't be released yet.

I think this is most likely for core typespec changes.  The only way to be really sure is to target the fixes and this PR at the release branch instead of main.

tadelesh:
> The only potential problem is that, since the main branch has had changes since the release, if any API changes from other packages were consumed in TCGC, they won't be released yet.
> 
> I think this is most likely for core typespec changes. The only way to be really sure is to target the fixes and this PR at the release branch instead of main.

yeah, that's the potential risk that release a tcgc hotfix from main. we decided to switch to dev and release from a hotfix branch after next typespec main release.

Issue #960, opened 2024-06-06
[TCGC] support advanced feature of multipart


msyyc:
After https://github.com/microsoft/typespec/issues/3046, Typespec already supports complete multipart feature, and TCGC needs to adopt it.
```[tasklist]
### Bug in Typespec
- [ ] https://github.com/microsoft/typespec/issues/3532
- [ ] https://github.com/microsoft/typespec/issues/3644
- [ ] https://github.com/microsoft/typespec/issues/3779
```
```[tasklist]
### Feature
- [ ] https://github.com/Azure/typespec-azure/issues/608
- [ ] https://github.com/Azure/typespec-azure/issues/607
```

```[tasklist]
### TCGC API design
- [ ] https://github.com/Azure/typespec-azure/pull/987
```

```[tasklist]
### TCGC Implementation
- [ ] https://github.com/Azure/typespec-azure/pull/1090
```

```[tasklist]
### Cadl-Ranch Test
- [ ] https://github.com/Azure/cadl-ranch/pull/613
```

msyyc:
Currently TCGC only has boolean flag [isMultipartFileInput](https://github.com/Azure/typespec-azure/blob/ab7a066d4ac0ae23a40f9ff8f4b6037559bda34c/packages/typespec-client-generator-core/src/interfaces.ts#L368), which is not enough to support some new features of multipart in Typespec. As first step, we shall design new TCGC API for multipart.

PR #1090, opened 2024-06-27
[Multipart] Support model format for `@multipartBody`

msyyc:
- fixes part of https://github.com/Azure/typespec-azure/issues/960
- Pending on https://github.com/microsoft/typespec/pull/3676

Context: 
Typespec supports 3 kinds of format to define multipart:
1. common
```

op upload(
  @header `content-type`: "multipart/form-data",
  @body body: {
    basic: string,
    headShots: bytes[],
  }
): void;
```

2. advanced model format
```
op upload(
  @header `content-type`: "multipart/form-data",
  @multipartBody body: {
    fullName: HttpPart<string>,
    headShots: HttpPart<bytes[]>
  }
): void;
```

3. advanced array format
```
op upload(
  @header `content-type`: "multipart/form-data",
  @multipartBody body: [
    // single
    HttpPart<string, #{ name: "fullName" }>,
    HttpPart<bytes[], #{ name: "headShots" }>,
  ]
): void;
```
- This PR is implementation for API design https://github.com/Azure/typespec-azure/pull/987, mainly for 1 and 2 format. Format 3 will be implemented in another PR.


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -391,11 +391,30 @@ export type SdkModelPropertyType =
   | SdkBodyParameter
   | SdkHeaderParameter;
 
+export interface MultipartOptionsType {
+  // whether name is defined in Typespec. For multipart/mixed, name may not be defined for some parts
+  isNameDefined: boolean;
+  // whether this part is for file
+  isFilePart: boolean;
+  // whether this part is multi in request payload
+  multi: boolean;
```
weidongxu-microsoft, line 400: full name: multiple or allowMultiple?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1089,16 +1104,68 @@ export function getSdkModelPropertyType(
       })
     );
   }
-  return diagnostics.wrap({
+  const multiPartOption: MultipartOptionsType = {
+    isNameDefined: httpOperationPart ? httpOperationPart.name !== undefined : true,
+    isFilePart: isBytesInput && operationIsMultipart,
+    multi: httpOperationPart ? httpOperationPart.multi : isRawBytesArray,
+    headers: [], // TODO
```
weidongxu-microsoft, line 1111: why expose the API when it is not implemented?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1060,27 +1071,31 @@ export function getSdkModelPropertyTypeBase(
...
     );
   }
   // Currently we only recognize bytes and list of bytes as potential file inputs
+  const httpPart = getHttpPart(context.program, type.type);
+  const isRawBytesArray = base.type.kind === "array" && base.type.valueType.kind === "bytes";
   const isBytesInput =
     base.type.kind === "bytes" ||
-    (base.type.kind === "array" && base.type.valueType.kind === "bytes");
+    isRawBytesArray ||
+    (httpPart !== undefined && isOrExtendsHttpFile(context.program, httpPart.type));
```
weidongxu-microsoft, line 1098: For the preparation codes, I would recommend we separate the "single" case and "multipart" case.

For "single", file etc. is via guesswork. For "multipart", file/contentType etc. is clear from HttpPart.
I'd prefer not to mix the guesswork and the clear data back into TCGC code.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1089,16 +1104,68 @@ export function getSdkModelPropertyType(
...
+      : undefined,
+  };
+  base.isMultipartFileInput = multiPartOption.isFilePart;
+  base.multipartOptions = operationIsMultipart ? multiPartOption : undefined;
+  if (operationIsMultipart && base.type.kind === "array") {
+    if (httpOperationPart === undefined) {
+      // for "images: bytes[]", return type shall be "bytes" instead of "bytes[]"" and "multipartOptions.multi" shall be true
+      if (base.type.valueType.kind === "bytes") {
+        base.type = base.type.valueType;
+      }
```
weidongxu-microsoft, line 1128: I think typespec dev asks this behavior for all types.

packages/typespec-client-generator-core/src/interfaces.ts
MaryGao, line 400: just curious why we need this? I mean we already under the interface `MultipartOptionsType`.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -391,11 +391,30 @@ export type SdkModelPropertyType =
...
+  // relates to custom header
+  headers: SdkHeaderParameter[];
+  // undefined if filename is not set explicitly in Typespec
+  filename?: SdkModelPropertyType;
+  // undefined if contentType is not set explicitly in Typespec
+  contentType?: SdkModelPropertyType;
```
MaryGao, line 406: I may ask if TCGC would calculate default contentType for us.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -391,11 +391,30 @@ export type SdkModelPropertyType =
   | SdkBodyParameter
   | SdkHeaderParameter;
 
+export interface MultipartOptionsType {
+  // whether name is defined in Typespec. For multipart/mixed, name may not be defined for some parts
+  isNameDefined: boolean;
```
weidongxu-microsoft, line 396: If we haven't decided what to do with the case of no name, we may not want to expose this now.

E.g. we may want to do
```
name = <generated-name>
generatedName = true
serializedName = ""
```
in property. And this be `serializedNameDefined`.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1089,16 +1104,68 @@ export function getSdkModelPropertyType(
...
+      ? diagnostics.pipe(getSdkModelPropertyType(context, httpOperationPart.filename, operation))
+      : undefined,
+    contentType: httpOperationPart?.body.contentTypeProperty
+      ? diagnostics.pipe(
+          getSdkModelPropertyType(context, httpOperationPart.body.contentTypeProperty, operation)
+        )
+      : undefined,
+  };
+  base.isMultipartFileInput = multiPartOption.isFilePart;
+  base.multipartOptions = operationIsMultipart ? multiPartOption : undefined;
```
weidongxu-microsoft, line 1122: Do we need to update `base.serializedName` to `httpOperationPart.name`? Are `httpOperationPart.name` the wire name?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1089,16 +1104,68 @@ export function getSdkModelPropertyType(
...
+  }
+  return diagnostics.wrap(undefined);
+}
+
+export function getSdkModelPropertyType(
+  context: TCGCContext,
+  type: ModelProperty,
+  operation?: Operation,
+  httpOperationPart?: HttpOperationPart
+): [SdkModelPropertyType, readonly Diagnostic[]] {
```
weidongxu-microsoft, line 1142: Mostly for @tadelesh  Why is `getSdkModelPropertyType` export? Is there other lib that use this API.

Ask because here its signature is changed, which means a breaks if any lib uses it.

packages/typespec-client-generator-core/src/interfaces.ts
weidongxu-microsoft: backward compitibility?

Also, I am not sure if `MultipartOptionsType` exists, if "single" (I assume our existing cadl-ranch cases is such).

packages/typespec-client-generator-core/src/interfaces.ts
msyyc: This PR doesn't contain case that name is null so I will remove this property for now.

packages/typespec-client-generator-core/src/interfaces.ts
msyyc: With its meaning, it equals to "multiple"

msyyc: It is easy to understand with design doc: https://github.com/Azure/typespec-azure/pull/987/

packages/typespec-client-generator-core/src/types.ts
msyyc: removed

packages/typespec-client-generator-core/src/types.ts
msyyc: No, at least with model format, it is impossible to change property name with HttpPart: 
image.

weidongxu-microsoft: I mean, which is the `name` to send in request? From the tsp I would assume it be `myPart`, not `myPropertyPart`.

weidongxu-microsoft: Also, in your doc this is an array without `myPropertyPart`?

image

msyyc: With array format, the only way to define serialized name of part is with `#{name: "fullName"}`; with model format, there is already property name which shall be same with serialized name of part so `#{name: "fullName"}` doesn't work. @timotheeguerin please correct me if wrong.

packages/typespec-client-generator-core/src/types.ts
msyyc: Make sense. I make refactor and the logic shall be more clear now.

packages/typespec-client-generator-core/src/types.ts
msyyc: Updated.

packages/typespec-client-generator-core/src/interfaces.ts
msyyc: Added `defaultContentTypes`. Only when `HttpPart` set, Typespec complier provides these values so TCGC doesn't calculate it but just pass it from Typespec compiler.

packages/typespec-client-generator-core/src/types.ts
msyyc: I understand that this function shall not be exposed.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1060,45 +1070,102 @@ export function getSdkModelPropertyTypeBase(
...
-  if (isSdkHttpParameter(context, type)) return getSdkHttpParameter(context, type, operation!);
-  // I'm a body model property
-  let operationIsMultipart = false;
-  if (operation) {
+  if (httpOperationPart) {
+    // body decorated with @multipartBody
+    const httpPart = getHttpPart(context.program, type.type);
+    base.multipartOptions = {
+      isFilePart:
+        isBytesInput ||
```
weidongxu-microsoft, line 1090:
```suggestion
        (!httpPart && isBytesInput) ||
```

We should not consider this `isBytesInput` if we have HttpPart.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1060,45 +1070,102 @@ export function getSdkModelPropertyTypeBase(
...
+          createDiagnostic({
+            code: "encoding-multipart-bytes",
+            target: type,
+          })
+        );
+      }
+      base.multipartOptions = {
+        isFilePart: isBytesInput,
+        multi: base.type.kind === "array",
+        defaultContentTypes: [],
```
weidongxu-microsoft, line 1122: Not sure whether TCGC want to provide the `defaultContentTypes`. E.g. nothing for text, `application/octet-stream` for file.

packages/typespec-client-generator-core/src/types.ts
msyyc: `httpOperationPart` doesn't have property to tell us whether this part is for File or not. we have to Judge it by ourselves.

packages/typespec-client-generator-core/src/types.ts
msyyc: Now only JS needs it. If more languages need, TCGC could help calculate it for default values.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -855,7 +858,14 @@ export function getClientTypeWithDiagnostics(
     case "Model":
       retval = diagnostics.pipe(getSdkArrayOrDictWithDiagnostics(context, type, operation));
       if (retval === undefined) {
-        retval = diagnostics.pipe(getSdkModelWithDiagnostics(context, type, operation));
+        const httpPart = getHttpPart(context.program, type);
```
msyyc, line 861: For `HttpPart<T>`, TCGC need `getHttPart` to get original `T` wrapped by `HttpPart`.

packages/typespec-client-generator-core/src/interfaces.ts
msyyc: I will keep it for JS for now and other languages just don't need to consume `defaultContentTypes`

packages/typespec-client-generator-core/src/types.ts
msyyc: https://github.com/Azure/typespec-azure/pull/1090#discussion_r1660521295

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1060,45 +1070,102 @@ export function getSdkModelPropertyTypeBase(
...
+          )
+        : undefined,
+      defaultContentTypes: httpOperationPart.body.contentTypes,
+    };
+  } else if (operation) {
+    // common body
     const httpOperation = getHttpOperationWithCache(context, operation);
-    operationIsMultipart = Boolean(
+    const operationIsMultipart = Boolean(
       httpOperation && httpOperation.parameters.body?.contentTypes.includes("multipart/form-data")
```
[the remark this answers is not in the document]
msyyc: For now TCGC only supports `multipart/form-data`. And we could add support for mixed in the future if needed.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1060,45 +1070,102 @@ export function getSdkModelPropertyTypeBase(
   });
 }
 
-export function getSdkModelPropertyType(
+function updateMultiPartInfo(
   context: TCGCContext,
   type: ModelProperty,
-  operation?: Operation
-): [SdkModelPropertyType, readonly Diagnostic[]] {
+  base: SdkBodyModelPropertyType,
```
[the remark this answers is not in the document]
msyyc: Similar with https://github.com/Azure/typespec-azure/pull/1090#discussion_r1658102298. If no language emitters adopt this function, we shall not expose it.
`httpOperationPart` is provided by http package which contains some useful info of each part of multipart.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -391,11 +391,28 @@ export type SdkModelPropertyType =
   | SdkBodyParameter
   | SdkHeaderParameter;
 
+export interface MultipartOptionsType {
```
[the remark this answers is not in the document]
msyyc: Will take into consideration.

packages/typespec-client-generator-core/src/interfaces.ts
msyyc: - For `addresses: HttpPart<AddressModel[]>`, TCGC API:
```
"name": "addresses",
"type": {"kind": "array"}
"multi": false,
```
It means this part could contain array of model
- For `addresses: HttpPart<AddressModel>[]`, TCGC API:
```
"name": "addresses",
"type": {"kind": "model"}
"multi": true,
```
It means each part could only contain content of one model.

packages/typespec-client-generator-core/src/types.ts
msyyc: It is already used by some language emitters. The new added signature is optional and doesn't change old logic.

packages/typespec-client-generator-core/src/types.ts
weidongxu-microsoft: I am strongly against doing guesswork in the `HttpPart` logic.

If dev write `HttpPart<File>`, it is a file. If not, it is not.

One step back, if you really want to guess, it should be on `content-type`, not the type of the body.
Similar to "single", if `@body body: bytes`, it is still `application/json`, unless you specific e.g. `content-type=impage/png`.

timotheeguerin: `HttpPart<bytes>` is a binary part and should be treated as such, File allows you to more finely control the filename and content type but they should been seen as the same over the wire. The goal is to still provide a good experience for the base types and not require always to go with the more precise one

The also produce the exact same openapi3 https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwpAc2VydmljZSh7CiAgdGl0bGU6ICJXaWRnZXQgU8YbIiwKfSkKbmFtZXNwYWNlIERlbW%2FHGzsKCm9wIHRlc3QoCiAgQG11bHRpcGFydEJvZHkgYm9keTogxFggIGE6IMRwUGFydDxieXRlcz47xRhiyxhGaWxlxRd9LAopOiB2b2lkOwo%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D

weidongxu-microsoft: So, `HttpPart<{@body body: bytes, @header contentType: "text/plain"}>` is also a file?

https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwpAc2VydmljZSh7CiAgdGl0bGU6ICJXaWRnZXQgU8YbIiwKfSkKbmFtZXNwYWNlIERlbW%2FHGzsKCm9wIHRlc3QoCiAgQG11bHRpcGFydEJvZHkgYm9keTogxFggIGE6IMRwUGFydDx7QMQaxx9ieXRlcywgQGhlYWRlciBjb250ZW505AClOiAidGV4dC9wbGFpbiJ9PjvFSWLLSUZpbGXFF30sCik6IHZvaWQ7Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D

msyyc: I think weidong's concern make sense. It will be hard to judge whether it is file part since we may need judge it according to content-type and data type. So maybe `Only File stand for file part` is proper which is easy and explicit.
However, according to Typespec design doc, both `HttpPart<Bytes>` and `HttpPart<File>` stand for file part:

image

Add @timotheeguerin @bterlson for confirmation

packages/typespec-client-generator-core/src/interfaces.ts
lmazuel: Isn't the information redondant with the `kind` in `type`?

packages/typespec-client-generator-core/src/types.ts
msyyc: It is bug https://github.com/microsoft/typespec/issues/3779

packages/typespec-client-generator-core/src/types.ts
msyyc: After discussion, we decide to treat `HttpPart<bytes>` as file part. Thanks all!

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1109,15 +1198,19 @@ function addPropertiesToModelType(
   operation?: Operation
 ): [void, readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
-  for (const property of type.properties.values()) {
+  const httpOpParts = getHttpOperationParts(context, operation);
```
tadelesh, line 1201: can we put the entries of adding multi-part info into http operation handling logic in http.ts? i'm trying to not to add too much protocol special logic in the type handling process (some refactoring work for tcgc). if a request is multi-part, then add new info to sdkbodymodelpropertytype.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1061,45 +1071,124 @@ export function getSdkModelPropertyTypeBase(
...
+  } else if (type.kind === "model") {
+    if (type.__raw && isOrExtendsHttpFile(context.program, type.__raw)) {
+      // Http<File>
+      return true;
+    }
+    // HttpPart<{@body body: bytes}> or HttpPart<{@body body: File}>
+    const body = type.properties.find((x) => x.kind === "body");
+    if (body) {
+      return isFilePart(context, body.type);
+    }
```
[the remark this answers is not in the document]
msyyc: It is valid. For example, if users want to define custom header, they can define `HttpPart<{@body body: File, @header myHeader: string}>`

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -387,11 +387,28 @@ export type SdkModelPropertyType =
...
+  // whether this part is for file
+  isFilePart: boolean;
+  // whether this part is multi in request payload
+  multi: boolean;
```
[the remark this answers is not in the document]
msyyc: renamed to `isMulti`

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -387,11 +387,28 @@ export type SdkModelPropertyType =
   | SdkBodyParameter
   | SdkHeaderParameter;
 
+export interface MultipartOptionsType {
```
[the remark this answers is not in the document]
msyyc: Updated

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -856,7 +859,14 @@ export function getClientTypeWithDiagnostics(
     case "Model":
       retval = diagnostics.pipe(getSdkArrayOrDictWithDiagnostics(context, type, operation));
       if (retval === undefined) {
-        retval = diagnostics.pipe(getSdkModelWithDiagnostics(context, type, operation));
+        const httpPart = getHttpPart(context.program, type);
```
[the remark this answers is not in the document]
msyyc: For `HttpPart<bytes>`, compiler gives Model Type whose name is "HttpPart" but actually TCGC only want "bytes" wrapped by "HttpPart", so we need special logic here. Then the result is `modelsMap["HttpPart<bytes>"] = SdkType<bytes>`

packages/typespec-client-generator-core/src/types.ts
msyyc: According to our design https://github.com/Azure/typespec-azure/pull/987, multipart info is bound with property of model instead of operation level. As declared in https://github.com/Azure/typespec-azure/pull/1090#discussion_r1682069461, if no special logic in getClientType, TCGC will produce some unexpected sdkType named `HttpPart`.

tadelesh: emmm, is it possible to remove the httppart wrapper when handling model property since it is kind of special body model property instead of remove the wrapper in get client type?

msyyc: After offline discussion with @tadelesh, `getClientType` will not be used for type of ModelProperty so I will try to centralize the logic in one place.

tadelesh (approved):

PR #1099, opened 2024-07-01
Add pipeline to detect breaking change in swagger

pshao25:
We want to ensure TypeSpec won't break generated swagger prior to its release. One solution is what this PR is doing: every time when the TypeSpec related change is in progress, we could run this pipeline to see if it impacts the existing swagger.

This [script](https://github.com/Azure/azure-rest-api-specs/pull/29646) is a prerequisite because I call it to regen swaggers. I'm not calling existing TypeSpec-Validation.ps1 because I don't want to fail this script if there is any change, we want to see the diff.

After run this pipeline [here](https://dev.azure.com/azure-sdk/internal/_build?definitionId=7053&_a=summary), there will be a draft PR like [this](https://github.com/Azure/azure-rest-api-specs/pull/29698). Ignore the script added, we will not have this change after [this](https://github.com/Azure/azure-rest-api-specs/pull/29646) merged.

How to run the pipeline: type in your PR head ref 
image


eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,59 @@
+trigger: none
+
+resources:
+  repositories:
+    - repository: azure-rest-api-specs
+      type: github
+      name: Azure/azure-rest-api-specs
+      endpoint: azure
+
```
MaryGao, line 9:
```suggestion
variables:
  NODE_OPTIONS: --max_old_space_size=16384
```

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+              nodeVersion: "18.x"
+              workingDirectory: $(Build.SourcesDirectory)/typespec-azure
+          - script: pnpm run build
+            displayName: Build
+            workingDirectory: $(Build.SourcesDirectory)/typespec-azure
+
+          #Step 2: Update swagger by TypeSpec in Step 1
+          - checkout: azure-rest-api-specs
+            fetchDepth: 1
+          - pwsh: |
```
timotheeguerin, line 36: We already have this logic for the cadl-ranch e2e test, could we reuse that instead of having to maintain an extra place and deal with powershell

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+              $packageJson.devDependencies["@typespec/http"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/http"
+              $packageJson.devDependencies["@typespec/openapi"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/openapi"
+              $packageJson.devDependencies["@typespec/openapi3"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/openapi3"
+              $packageJson.devDependencies["@typespec/rest"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/rest"
+              $packageJson.devDependencies["@typespec/versioning"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/versioning"
+
+              $packageJson | ConvertTo-Json -Depth 100 | Out-File -Path "package.json" -Encoding utf8 -NoNewline -Force
+            displayName: "Update package.json in azure-rest-api-specs"
+            workingDirectory: $(Build.SourcesDirectory)/azure-rest-api-specs
+          - script: npm install
```
timotheeguerin, line 55:
```suggestion
          - script: npm install --no-package-lock
```

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+        pool:
+          name: azsdk-pool-mms-ubuntu-2004-general
+          vmImage: ubuntu-20.04
+        steps:
+          # Step 1: Build TypeSpec
+          - checkout: self
+            submodules: true
+          - template: /eng/pipelines/templates/install.yml
+            parameters:
+              nodeVersion: "18.x"
```
timotheeguerin, line 27:
```suggestion
```

don't respecify here, this is otherwise an extgra place we'll need to upgrade, use the default

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+          - template: /eng/common/pipelines/templates/steps/git-push-changes.yml@azure-sdk-tools
+            parameters:
+              BaseRepoBranch: auto-update-swagger-$(Build.BuildNumber)
+              BaseRepoOwner: azure-sdk
+              CommitMsg: Update swagger with TypeSpec built from $(Build.BuildNumber)
+              TargetRepoOwner: Azure
+              TargetRepoName: azure-rest-api-specs
+              WorkingDirectory: $(Build.SourcesDirectory)/azure-rest-api-specs
+              ScriptDirectory: $(Build.SourcesDirectory)/azure-sdk-tools/eng/common/scripts
+          - task: PowerShell@2
```
timotheeguerin, line 79: How is creating a PR going to help if everything was reverted here? What is the purpose?

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+              $packageJson.devDependencies["@typespec/rest"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/rest"
+              $packageJson.devDependencies["@typespec/versioning"] = "file:$(Build.SourcesDirectory)/typespec-azure/core/packages/versioning"
+
+              $packageJson | ConvertTo-Json -Depth 100 | Out-File -Path "package.json" -Encoding utf8 -NoNewline -Force
+            displayName: "Update package.json in azure-rest-api-specs"
+            workingDirectory: $(Build.SourcesDirectory)/azure-rest-api-specs
+          - script: npm install
+            displayName: Install updated TypeSpec dependencies
+            workingDirectory: $(Build.SourcesDirectory)/azure-rest-api-specs
+          - pwsh: ./eng/scripts/TypeSpec-Generate-Swagger.ps1 -CheckAll
```
timotheeguerin, line 58: Where is this script comming from I don't see it in azure-rest-api-specs.

`TypeSpec-Validation.ps1` is the one that does the regen and checks

pshao25: Yeah I put details in the description that it comes from this [PR](https://github.com/Azure/azure-rest-api-specs/pull/29646). 

`TypeSpec-Validation.ps1` would throw error if there is any change, which is not we are expecting. We hope to get the diff.

eng/pipelines/pr-breaking-check.yml
pshao25: What you mean by "everything was reverted here"? I only revert package.json and package-lock.json, which is not what we care about. We care about swagger change. An example is [here](https://github.com/Azure/azure-rest-api-specs/pull/29698).

timotheeguerin: I see, so from experience with the spec repo most of the change we need to apply are typespec changes in the typespec-next branch so wondering if this just does a small portion of the solution

timotheeguerin: also because of hte typespec-next branch I feel like this is just going to fail anyway if you run on main

pshao25: I didn't find the typespec-next branch you are saying. Could you point me to it?

I want to clarify the user story here:
1. Developer does some change to typespec-azure and send out a PR with PR number like `abc`.
2. Developer go to this pipeline UI and type in the target branch is `ref/pulls/abc/head` or `ref/pulls/abc/merge`. Then run. You could see the screenshot in the description.

Therefore, our target is WIP change, instead of the code already merged into main.

timotheeguerin: typespec-next branch https://github.com/Azure/azure-rest-api-specs/tree/typespec-next
See for example the last release the changes involved https://github.com/Azure/azure-rest-api-specs/pull/29440

There is many little things that would not count as breaking change but still need a little update to the tsp/swagger like formatting.

This pipeline would basically only work up to the first change.

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+              CommitMsg: Update swagger with TypeSpec built from $(Build.BuildNumber)
+              TargetRepoOwner: Azure
+              TargetRepoName: azure-rest-api-specs
+              WorkingDirectory: $(Build.SourcesDirectory)/azure-rest-api-specs
+              ScriptDirectory: $(Build.SourcesDirectory)/azure-sdk-tools/eng/common/scripts
+          - task: PowerShell@2
+            displayName: Create pull request
+            inputs:
+              pwsh: true
+              filePath: $(Build.SourcesDirectory)/azure-sdk-tools/eng/common/scripts/Submit-PullRequest.ps1
```
weshaggard, line 83: Why are we creating these PR's against the spec repo? We don't wish to start creating a lot of PRs. Are there other ways we can get the validation we are after?

eng/pipelines/pr-breaking-check.yml
```diff
@@ -0,0 +1,94 @@
...
+              arguments: >
+                -RepoOwner "Azure"
+                -RepoName "azure-rest-api-specs"
+                -BaseBranch "main"
+                -PROwner "azure-sdk"
+                -PRBranch "auto-update-swagger-$(Build.BuildNumber)"
```
weshaggard, line 89: This naming pattern for a branch will create a branch/PR per build in the specs repo which is something we should definitely avoid, that will likely cause way too much noise.

weshaggard:
@hallipr how does this related to the other validation pipelines you are doing for typespec and autorest?

eng/pipelines/pr-breaking-check.yml
hallipr: In the typespec preview pipelines, we set the branch name early in the build via inline powershell.   We attempt to reuse the same PR branch name for the same source PR, scheduled runs use a single branch name and all other runs will use a unique dynamic name:

[azure-sdk-tools / archetype-autorest-preview.yml](https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/stages/archetype-autorest-preview.yml)

```yaml
    - pwsh: |
        $sourceBranch = '$(Build.SourceBranch)'
        $buildReason = '$(Build.Reason)'
        $buildNumber = '$(Build.BuildNumber)'

        if ($buildReason -eq 'Schedule') {
          $branchName = 'auto-update-autorest-scheduled'
        } elseif ($sourceBranch -match "^refs/pull/(\d+)/(head|merge)$") {
          $branchName = "auto-update-autorest-pr-$($Matches[1])"
        } else {
          $branchName = "auto-update-autorest-$buildNumber"
        }

        Write-Host "Setting variable 'branchName' to '$branchName'"
        Write-Host "##vso[task.setvariable variable=branchName;isOutput=true]$branchName"
      displayName: Set branch name
      name: set_branch_name
```

weshaggard: That would definitely be better if we need a PR. However I'd still like to understand if a PR into the specs repo is needed to accomplish the testing we are looking for.

eng/pipelines/pr-breaking-check.yml
pshao25: I see. I understand it is very possible (even maybe 100%) we expect to change tsp before each release. But our intention is to uncover all the unexpected change in swagger, which might come from bug or lack of consideration. Every time you are not confident in your change, you could run the pipeline **before** merge. The granularity is for each RP. We might expect tsp change for each release, but we won't expect tsp change for each PR, right?

eng/pipelines/pr-breaking-check.yml
pshao25: We want to have a visualized way to see the diff. We could reduce the number by the way Patrick put above. Or maybe we could even put the PRs in azure-rest-api-specs-pr if your concern is you don't want to confuse customers.

> If it is needed who will be responsible for monitoring it and closing it.

@hallipr do you know how it is monitored in azure-sdk-for-net?

pshao25:
> @hallipr how does this related to the other validation pipelines you are doing for typespec and autorest?

I don't need too many machines to regen for this case. I only need one. Therefore, I didn't publish any packages (which is a must in the other pipelines because they need parallel regen). I just simply clone typespec repo and azure-rest-api-specs repo. Let the versions in azure-rest-api-specs point to the local files of typepsec directly. I think this simplified process is difficult to fit into the pattern of other pipelines.

eng/pipelines/pr-breaking-check.yml
timotheeguerin: Ok but if you test against main you will always have diff from previous PRs that might have had expected change as well

eng/pipelines/pr-breaking-check.yml
weshaggard: You can have a visual change by pushing to a branch without needing to create a PR. Creating a PR causes a lot of other things to kick off that we should avoid if they aren't needed.

Noise inside of the private repo is still a concern.

pshao25: How about we just create a closed PR? It sounds really wired to me that pushing to a branch which is almost 100% going to create a PR.

eng/pipelines/pr-breaking-check.yml
markcowl: I feel like we get this protection from the azure-rest-api-specs typespec-next ci, which automatically takes up the developer versions and validates them against typespec-next.

It would be *nice* to have a more automated way to check a particular PR against typespec-next directly

markcowl (commented):
I would like to understand the intention of this job, and how it improves on the current typespec-next ci,  which tests the existing specs against each development version of TypeSpec

eng/pipelines/pr-breaking-check.yml
raych1: @pshao25 we can compose the URL to view the difference between two branches like this `https://github.com/[username]/[repository]/compare/[base-branch]...[compare-branch]` instead of creating PRs.

eng/pipelines/pr-breaking-check.yml
pshao25: Could you point me to typespec-next ci? Actually I want to understand how it works.

pshao25:
> I would like to understand the intention of this job, and how it improves on the current typespec-next ci, which tests the existing specs against each development version of TypeSpec

It starts from [0.55 upgrade](https://github.com/Azure/azure-rest-api-specs/pull/28569/files#diff-14231e689eb8e9684b376014f78b6a83dc45dd144b4f97918a7bf123b7ff9951). We found that this upgrade removed the “readOnly: true” from “nextLink”, which causes breaking change in the generated SDK. Therefore, we are exploring some way to uncover such change as early as possible. This pipeline is what we prefer. We described to @allenjzhang in the TypeSpec sync meeting and he is supportive of it.

eng/pipelines/pr-breaking-check.yml
timotheeguerin: It is just teh `TypeSpecValidation - All` pipeline from teh spec repo that is run in the typespec-next branch on every commit as well as as soon as one of the typespec and typespec azure pipeline finish https://dev.azure.com/azure-sdk/public/_build?definitionId=6143&_a=summary

eng/pipelines/pr-breaking-check.yml
hallipr: If the purpose of the branch (or pr) is to use github's UI to visualize a diff, I like Ray's suggestion to just use a ref comparison instead of a PR.   We could even add that compare URL to the pipeline output so it's clickable from the ADO UI

markcowl:
> > I would like to understand the intention of this job, and how it improves on the current typespec-next ci, which tests the existing specs against each development version of TypeSpec
> 
> It starts from [0.55 upgrade](https://github.com/Azure/azure-rest-api-specs/pull/28569/files#diff-14231e689eb8e9684b376014f78b6a83dc45dd144b4f97918a7bf123b7ff9951). We found that this upgrade removed the “readOnly: true” from “nextLink”, which causes breaking change in the generated SDK. Therefore, we are exploring some way to uncover such change as early as possible. This pipeline is what we prefer. We described to @allenjzhang in the TypeSpec sync meeting and he is supportive of it.

This was a *planned and approved breaking change*,  it was not a mistake, and it was detected by the existing typespec validation.  I think the issue may be just communication - breaking changes are approved by the TypeSpec breaking change board and the release notes for each release contain details on changes.

markcowl:
@pshao25 Is there a reason to keep this PR around, can we delete it?

pshao25:
> @pshao25 Is there a reason to keep this PR around, can we delete it?

I'm OK to close. Is any reason we have to delete?

Issue #858, opened 2024-05-17
[Linting rules] Add a rule to warn not to use .Private namespaces


allenjzhang:
### Clear and concise description of the problem

A few specs are including `.Private` from either Azure.Core or ARM libraries. We should warn about not using these.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1105, opened 2024-07-02
Remove 'x-ms-client-flatten' in typespec-azure-resource-manager templates and make it conditional in autorest

### LINKED ISSUE #709: Make 'x-ms-client-flatten' in typespec-azure-resource-manager templates conditional
This requires updating the dependencies so that typespec-autorest depends on typespec-azure-resource-manager

- [x] Add a new decorator `@conditionaFlattening` which marks a resource as flattened, depending on emitter settings.  and an accessor isConditionallyFlattened, both of these apply to model properties
- [x] Replace all template instances using `@OpenApi.extension("x-ms-client-flatten": true) with the `@conditionalFlattening` decorator.
- [x] Add a new option in typespec-autorest `arm-resource-flattening` with values `true` and `false`,  defaulting to false.  If this option is true, use the `isConditionallyFlattened` accessor to determine whether to add an `x-ms-client-flatten`:` true` to the property representation in OpenAPI
- [ ] Update azure-rest-api-specs, with the new flag set to true in ARM projects

allenjzhang:
Closes #709


.chronus/changes/azhang_RemoveClientFlatten-2024-6-1-16-58-12.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: breaking
+packages:
+  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+x-ms-client-flatten extension on some of resource properties property is now configurable to be emitted by autorest emitter. Default is false which will skip emission of that extension.
```
timotheeguerin, line 8:
```suggestion
`x-ms-client-flatten `extension on some of resource properties property is now configurable to be emitted by autorest emitter. Default is false which will skip emission of that extension.
```

can you also describe what is breaking and how one would fix it as its not clear to me from this text

packages/typespec-azure-resource-manager/src/commontypes.private.decorators.ts
```diff
@@ -127,6 +127,12 @@ export function $armCommonDefinition(
   storeCommonTypeRecord(context, entity, "definitions", definitionName, version, referenceFile);
 }
 
+/**
+ * Specify the ARM commont type version reference for a particular spec version or namespace.
+ * @param {DecoratorContext} context DecoratorContext object
+ * @param {enumType} entity Decorator target type. Must be `Model`
+ * @returns {void}
```
timotheeguerin, line 134: nit but I don't think its ncessary to include the tyupes in the jsdoc with typescript

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -350,3 +350,18 @@ function hasProperty(program: Program, model: Model): boolean {
...
+ * Please DO NOT USE in RestAPI specs.
+ * Internal decorator that deprecated direct usage of `x-ms-client-flatten` OpenAPI extension.
+ * It will programatically enabled/disable client flattening with @flattenProperty with autorest
+ * emitter flags to maintain compatibility in swagger.
+ */
+export function $conditionalClientFlatten(context: DecoratorContext, entity: ModelProperty) {
+  context.program.stateMap(ArmStateKeys.armConditionalClientFlatten).set(entity, true);
+}
+
+export function isConditionallyFlattened(program: Program, entity: ModelProperty): boolean {
```
timotheeguerin, line 364: would naming this legacyFlattened be better?

packages/typespec-azure-resource-manager/lib/private.decorators.tsp
```diff
@@ -66,3 +66,11 @@ extern dec defaultResourceKeySegmentName(
...
  */
 extern dec enforceConstraint(target: Operation | Model, sourceType: Model, constraintType: Model);
+
+/**
+ * Please DO NOT USE in RestAPI specs.
+ * Internal decorator that deprecated direct usage of `x-ms-client-flatten` OpenAPI extension.
+ * It will programatically enabled/disable client flattening with @flattenProperty with autorest
+ * emitter flags to maintain compatibility in swagger.
+ */
+extern dec conditionalClientFlatten(target: ModelProperty);
```
qiaozha, line 76: wonder if we could add linter rule to prevent customer from using it in rest api specs?

packages/typespec-azure-resource-manager/src/private.decorators.ts
qiaozha: if legacy is likely referring to the legacy client? I'd like to point it out that the legacy client behavior is different from language to language, for example  Go doesn't support flatten before. while other languages like dotnet, js does.

packages/typespec-azure-resource-manager/lib/private.decorators.tsp
allenjzhang: There is a work item to warn on using Private namespace in spec. #858

packages/typespec-azure-resource-manager/src/commontypes.private.decorators.ts
allenjzhang: You meant the @param?

timotheeguerin: oh wait I misread. Just remove the wrapping `{}`
usualy in JS doc the syntax is
```
@param foo {Type} Dec of param
```

but in tsdoc is should just be 
```
@param foo Dec of param
```

packages/typespec-azure-resource-manager/src/private.decorators.ts
allenjzhang: Per conversation with Laurent, this function is intended for `autorest`. We are done away with flattening by default. If folks wants to flatten, instead of relying this decorator, spec writer can use `@@flattenProperty(XX.properties)` in `client.tsp`. Let's keep the name as it is.

timotheeguerin (approved):

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -359,3 +359,18 @@ export function isAzureResource(program: Program, resourceType: Model): boolean
   const isResourceBase = program.stateMap(ArmStateKeys.azureResourceBase).get(resourceType);
   return isResourceBase ?? false;
 }
+
+/**
+ * Please DO NOT USE in RestAPI specs.
+ * Internal decorator that deprecated direct usage of `x-ms-client-flatten` OpenAPI extension.
+ * It will programatically enabled/disable client flattening with @flattenProperty with autorest
```
tadelesh, line 366: @allenjzhang i do not see where to set `@flattenProperty` decorator, but only see the set of openapi2 `x-ms-client-flatten`. am i miss sth?

Issue #1110, opened 2024-07-02
[Bug]: Wrong `null` tagging for the type of property `userAssignedIdentities` for ManagedServiceIdentity in common types


MaryGao:
### Describe the bug

The type of property `userAssignedIdentities` for ManagedServiceIdentity should be `Record<UserAssignedIdentity | null>` but it is `Record<UserAssignedIdentity> | null`.

current typespec: https://github.com/Azure/typespec-azure/blob/fe90e201b7ad8bda7f98e786debfae37c025ee98/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp#L26C4-L27

swagger definition: https://github.com/Azure/azure-rest-api-specs/blob/87a08b955c257c773a3bd42553c718d4b1092955/specification/common-types/resource-management/v5/managedidentity.json#L15

### Reproduction

[playground](https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgQSB7DQogIGZvbz86IFJlY29yZDxzdHJpbmc%2BIHwgbnVsbDvEIGJhctAgxx8%2BOw0KfQ0K&e=%40typespec%2Fopenapi3&options=%7B%7D)

see swagger definition: https://github.com/Azure/azure-rest-api-specs/blob/87a08b955c257c773a3bd42553c718d4b1092955/specification/common-types/resource-management/v5/managedidentity.json#L15

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1111, opened 2024-07-02
Fix [Bug]: Wrong `null` tagging in ManagedServiceIdentity

MaryGao:
fixes https://github.com/Azure/typespec-azure/issues/1110

to match with the swagger definition https://github.com/Azure/azure-rest-api-specs/blob/87a08b955c257c773a3bd42553c718d4b1092955/specification/common-types/resource-management/v5/managedidentity.json#L9-L16

### DIFF
--- a/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
+++ b/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
@@ -24,7 +24,7 @@ model ManagedServiceIdentity {
 
   /** The identities assigned to this resource by the user. */
   @typeChangedFrom(Versions.v5, Record<UserAssignedIdentity>)
-  userAssignedIdentities?: Record<UserAssignedIdentity> | null;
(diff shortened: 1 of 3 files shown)


packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp
```diff
@@ -24,7 +24,7 @@ model ManagedServiceIdentity {
 
   /** The identities assigned to this resource by the user. */
   @typeChangedFrom(Versions.v5, Record<UserAssignedIdentity>)
-  userAssignedIdentities?: Record<UserAssignedIdentity> | null;
+  userAssignedIdentities?: Record<UserAssignedIdentity | null>;
```
allenjzhang, line 27: Should update `model UserAssignedIdentities is Record<UserAssignedIdentity> with | null` as well.

allenjzhang (approved):

timotheeguerin:
Can you regenerate the new common types and add the command to the regen-samples so it’s verified in the ci

Issue #1096, opened 2024-07-01
[Bug][TCGC]: No compilation error on naming conflicts with clientName decorator


live1206:
### Describe the bug

When a model is renamed to a name conflicting with existing model, TCGC should throw compilation error

### Reproduction

[...]

PR #1119, opened 2024-07-03
Report diagnostics on `@clientName` conflicts

live1206:
Resolves https://github.com/Azure/typespec-azure/issues/1096


packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,42 @@
...
+  const [models, diagnostics] = getAllModelsWithDiagnostics(tcgcContext);
   if (diagnostics.length > 0) {
     program.reportDiagnostics(diagnostics);
   }
+
+  validateModels(program, models);
+}
+
+function validateModels(program: Program, models: (SdkModelType | SdkEnumType)[]) {
+  const modelNameMap = new Map<string, Set<SdkModelType | SdkEnumType>>();
```
timotheeguerin, line 19: Btw we have DuplicateTracker until in the compiler that is quite usufruit for this

live1206: applied

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -410,17 +410,27 @@ export interface TCGCContext {
...
     __namespaceToApiVersionClientDefaultValue: new Map(),
   };
 }
 
+function getEmitterName(program: Program): string {
+  if (program.emitters.length > 0 && program.emitters[0].metadata.name !== undefined) {
+    return program.emitters[0].metadata.name;
+  }
+
+  // This is a fallback, but it should only be hit for testing
```
live1206, line 430: While the compiler loadEmitter [here](https://github.com/microsoft/typespec/blob/fe6d24aa8e80bd5028797cdcfed6194d90c298de/packages/compiler/src/core/program.ts#L598-L602), it tries to resolve the library.
But we don't really have a library for the emitter in place while testing, so just add this fallback to make the test work.

timotheeguerin (changes requested):
Doing this implicit parsing of emitter is exactly what you shouldn’t do. This break if you don’t run emitter or run multiple.

you need to expose all the data to this internal function(don’t expose it outside of the package) and the function should then be able to check all scopes

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2701,6 +2702,161 @@ describe("typespec-client-generator-core: decorators", () => {
...
+          code: "@azure-tools/typespec-client-generator-core/duplicate-model-name",
+        }
+        ]
+      );
+    });
+  
+    it("duplicate model client name with csharp language scopes", async () => {
+      const diagnostics = await runner.diagnose(
+      `
+      @service({
```
timotheeguerin, line 2817: simplify those by just passing `@service` without args

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2701,6 +2702,161 @@ describe("typespec-client-generator-core: decorators", () => {
...
+    it("duplicate enum client name with all language scopes", async () => {
+      const diagnostics = await runner.diagnose(
+      `
+      @service({
+        title: "Contoso Widget Manager",
+      })
+      namespace Contoso.WidgetManager;
+      
+      @error
+      model Error {
```
timotheeguerin, line 2762: I think a lot of those test have quite a few uncessary items, can it be simplified to just have the models needed, we don't need to make a full example service per test.

tadelesh:
> Doing this implicit parsing of emitter is exactly what you shouldn’t do. This break if you don’t run emitter or run multiple.
> 
> you need to expose all the data to this internal function(don’t expose it outside of the package) and the function should then be able to check all scopes

agreed. need to check in each scope for duplication.

live1206:
> > Doing this implicit parsing of emitter is exactly what you shouldn’t do. This break if you don’t run emitter or run multiple.
> > you need to expose all the data to this internal function(don’t expose it outside of the package) and the function should then be able to check all scopes
> 
> agreed. need to check in each scope for duplication.

@timotheeguerin @tadelesh What do you mean by "check all scopes" and "check each scope"?
If we take a look at the definition of [LanguageScopes](https://github.com/Azure/typespec-azure/blob/c826ffc9f385de5456cc033bf391d327d37fbf20/packages/typespec-client-generator-core/src/interfaces.ts#L587) for `clientName`, it can be any string. Or, we only care about `csharp, java, js, python and go`?

Another way I can think of is to get all the defined language scopes within `program.stateMap`.

tadelesh:
> > > Doing this implicit parsing of emitter is exactly what you shouldn’t do. This break if you don’t run emitter or run multiple.
> > > you need to expose all the data to this internal function(don’t expose it outside of the package) and the function should then be able to check all scopes
> > 
> > 
> > agreed. need to check in each scope for duplication.
> 
> @timotheeguerin @tadelesh What do you mean by "check all scopes" and "check each scope"? If we take a look at the definition of [LanguageScopes](https://github.com/Azure/typespec-azure/blob/c826ffc9f385de5456cc033bf391d327d37fbf20/packages/typespec-client-generator-core/src/interfaces.ts#L587) for `clientName`, it can be any string. Or, we only care about `csharp, java, js, python and go`?
> 
> Another way I can think of is to get all the defined language scopes within `program.stateMap`.

get all scopes' renaming from state map,  and throw error if any scope has duplicate naming.

live1206:
> > > > Doing this implicit parsing of emitter is exactly what you shouldn’t do. This break if you don’t run emitter or run multiple.
> > > > you need to expose all the data to this internal function(don’t expose it outside of the package) and the function should then be able to check all scopes
> > > 
> > > 
> > > agreed. need to check in each scope for duplication.
> > 
> > 
> > @timotheeguerin @tadelesh What do you mean by "check all scopes" and "check each scope"? If we take a look at the definition of [LanguageScopes](https://github.com/Azure/typespec-azure/blob/c826ffc9f385de5456cc033bf391d327d37fbf20/packages/typespec-client-generator-core/src/interfaces.ts#L587) for `clientName`, it can be any string. Or, we only care about `csharp, java, js, python and go`?
> > Another way I can think of is to get all the defined language scopes within `program.stateMap`.
> 
> get all scopes' renaming from state map, and throw error if any scope has duplicate naming.

Again, please define `all scopes` here, I am taking the defined language scopes from `program.stateMap` now.

tadelesh:
yes. all existed scopes in program.statemap.

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: simplified

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: simplified

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -529,3 +529,32 @@ export function getHttpOperationResponseHeaders(
   }
   return headers;
 }
+
+export class DuplicateTracker<K, V> {
+  #entries = new Map<K, Set<V>>();
```
live1206, line 534: Can't use the existing DuplicateTracker, because we need `Set<V>` as value instead of `V[]`.
Otherwise, we will throw errors on the same model multiple times.

live1206:
@tadelesh @timotheeguerin Comments resolved, please take another look.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -48,12 +48,23 @@ import { getLibraryName } from "./public-utils.js";
 import { getSdkEnum, getSdkModel, getSdkUnion } from "./types.js";
 
 export const namespace = "Azure.ClientGenerator.Core";
-const AllScopes = Symbol.for("@azure-core/typespec-client-generator-core/all-scopes");
+export const AllScopes = Symbol.for("@azure-core/typespec-client-generator-core/all-scopes");
```
timotheeguerin, line 51: I would watch out what you export here without changing the `index.ts` this is now accessible to everyone consuming the package.

This means that things you might consider internal could be seen as breaking for someone that took dependency on it.

I would recommend being explicit in `index.ts` of what is exported that way this is not exposed. Or reorganize a little bit the files to be sure anything in decorators.ts is meant to be accesible outside of this package

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,52 @@
...
-  const [_, diagnostics] = getAllModelsWithDiagnostics(tcgcContext);
-  if (diagnostics.length > 0) {
-    program.reportDiagnostics(diagnostics);
+  const languageScopes = getDefinedLanguageScopes(program);
+  validateModels(program, tcgcContext, languageScopes);
+}
+
+function getDefinedLanguageScopes(program: Program): Set<string | symbol> {
+  const languageScopes = new Set<string | symbol>();
+  const stateMap = program.stateMap(createStateSymbol("clientName"));
```
timotheeguerin, line 14: I would recommend not getting the data here but keeping accesor to this state with types in the same place. Just makes sure you don't have 5 different places that assume the type of the state.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -159,6 +159,12 @@ export const $lib = createTypeSpecLibrary({
         default: `Cannot pass an empty value to the @clientName decorator`,
       },
     },
+    "duplicate-model-name": {
+      severity: "error",
+      messages: {
+        default: `Duplicate model name: "${"name"}"`,
```
timotheeguerin, line 165: Can this also specify the scope, I think otherwise it is going to be confusing.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -159,6 +159,12 @@ export const $lib = createTypeSpecLibrary({
         default: `Cannot pass an empty value to the @clientName decorator`,
       },
     },
+    "duplicate-model-name": {
```
timotheeguerin, line 162: Is this really just only applicable to models? What about:
- properties
- enum
- union
- interfaces
- operations, etc.
basically everything that clientName affect

packages/typespec-client-generator-core/src/decorators.ts
live1206: Moving `AllScopes` to `internal-utils` instead

packages/typespec-client-generator-core/src/lib.ts
live1206: I thought scope is not needed in the first place because it is kind of obvious.
And if we want to add scope to the error message, it will report multiple times for the same duplicates.
But maybe it is more clear to state the scope in the message and throw the same duplicate per scope.

packages/typespec-client-generator-core/src/validate.ts
live1206: I don't quite follow here, I need to get the `clientName` state map in order to get the defined language scopes within it.
Are you saying I should only get the `clientName` state map in a central place and get defined scopes prepared in advance, and pass that along to this function instead?

packages/typespec-client-generator-core/src/internal-utils.ts
live1206: use the default one since we throw per scope.

packages/typespec-client-generator-core/src/decorators.ts
live1206: moved

packages/typespec-client-generator-core/src/lib.ts
live1206: added

packages/typespec-client-generator-core/src/lib.ts
live1206: Good point!
Based on the [Type definition](https://github.com/microsoft/typespec/blob/76bbb4041372f9183be44cd9268d4f4a3c0b34eb/packages/compiler/src/core/types.ts#L112-L136), which can be applied with `clientName`, I summarized the check rule after some discussion with @tadelesh

```
Unique per namespace:
- Namespace
- Interface
- Model & Enum & Union
- Operation

Unique within a specific model:
- ModelProperty

Unique within a specific enum:
- EnumMember

Unique within a Union:
- UnionVariant

```

tadelesh: scalar is missing

live1206: implemented with above rules.

packages/typespec-client-generator-core/src/validate.ts
timotheeguerin: yeah meant that, only have a single place that create the state symbol and provide typed function to use that elsewhere

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2694,6 +2694,7 @@ describe("typespec-client-generator-core: decorators", () => {
 
       strictEqual(runner.context.sdkPackage.clients[0].methods[0].parameters[0].name, "body");
     });
+
```
timotheeguerin, line 2697: doesn't work if there is no service
[playground](https://cadlplayground.z22.web.core.windows.net/typespec-azure/prs/1119/?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsKCnVzaW5nIEHENC5DxSVHyCQuQ29yZTsKCgptb2RlbCBGb28ge30KCkDGTE5hbWUoIkZvbyIpxyFCYXLEIQ%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -159,6 +159,12 @@ export const $lib = createTypeSpecLibrary({
         default: `Cannot pass an empty value to the @clientName decorator`,
       },
     },
+    "duplicate-name": {
+      severity: "error",
+      messages: {
+        default: paramMessage`Duplicate name: "${"name"}" in language scope: "${"scope"}"`,
```
timotheeguerin, line 165: I think there is maybe a little more improvement to the message or target that could be done here. 

Trying this out in the [playground](https://cadlplayground.z22.web.core.windows.net/typespec-azure/prs/1119/?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsKCnVzaW5nIEHENC5DxSVHyCQuQ29yZTsKCkBzZXJ2aWNlCm5hbWVzcGFjZSBDb250b3NvLldpZGdldE1hbmFnZXI7Cgptb2RlbCBGb28ge30KCkDGdk5hbWUoIkZvbyIpxyFCYXLEIQ%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)

it is maybe a bit confusing when you look at `Foo` why it is a duplicate name.

Wondering if 
1. you should target the `@clientName` decorator that introduce the conflict when applicable instead of the type, this would make it much clearer specially when you have multiple. Could also make the message more targetted then `Client name  "${"name"}" is conflicting  in language scope: "${"scope"}"`
2. when targetting the other type that might have a conflict but doesn't have `@clientName` maybe could use a different message that would indicate that somewhere else you have a `@clientName`

.chronus/changes/throw-client-name-conflicts-2024-6-5-13-7-58.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Fix to throw on clientName conflicts per language scope
```
timotheeguerin, line 7:
```suggestion
---
changeKind: feature
packages:
  - "@azure-tools/typespec-client-generator-core"
---

Report diagnostics on `@clientName` conflicts 
```

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: Had a chat with @tadelesh, if there is no service, tcgc will not apply the decorators.
Do we still need to handle the conflicts for this case?

timotheeguerin: I feel like so as there is use case for model only generation

packages/typespec-client-generator-core/src/lib.ts
live1206: refined, throw with different error message separately for decorator and non-decorator.

packages/typespec-client-generator-core/src/validate.ts
live1206: moved `clientNameKey` to `internal-utils`

packages/typespec-client-generator-core/test/decorators.test.ts
tadelesh: @live1206 refer this [code](https://github.com/Azure/typespec-azure/blob/e2a28c6311616df88525730bc82604a8c4d7085c/packages/typespec-autorest/src/emit.ts#L110-L113) for service traversal

live1206: This example just reminds me that we should take consideration of versioning as well.

timotheeguerin: The example in the playground now shows a weird place for one of  the errors

packages/typespec-client-generator-core/src/lib.ts
timotheeguerin: much nicer, i think this might have introduced a bug though on targetting the one with clientName

packages/typespec-client-generator-core/src/index.ts
```diff
@@ -3,3 +3,4 @@ export * from "./interfaces.js";
 export * from "./lib.js";
 export * from "./public-utils.js";
 export * from "./types.js";
+export * from "./validate.js";
```
timotheeguerin, line 6:
```suggestion
export { $onValidate } from "./validate.js";
```

know you don't have precedent for that in this package but its good to try to only export what you need so you don't export somethihng meant private and breaking consumer by mistake

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,177 @@
...
+    }
+    for (const languageScope of Object.keys(value)) {
+      languageScopes.add(languageScope);
+    }
+  }
+  return languageScopes;
+}
+
+function validateClientNames(program: Program, tcgcContext: TCGCContext, scope: string | symbol) {
+  const services = listServices(program);
```
timotheeguerin, line 43: Feel like you should just be able to omit all here and just always start at the global namespace as you recurrsively look into down

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,177 @@
...
+            format: { name, scope },
+            target: item,
+          });
+        } else {
+          reportDiagnostic(program, {
+            code: "duplicate-client-name",
+            messageId: "nonDecorator",
+            format: { name, scope },
+            target: item,
+          });
```
timotheeguerin, line 172:
```suggestion
      const scopeStr = scope === AllScopes ? "AllScopes" : scope;
      if (item.kind === "Decorator") {
        reportDiagnostic(program, {
          code: "duplicate-client-name",
          format: { name, scope: scopeStr },
          target: item,
        });
      } else {
        reportDiagnostic(program, {
          code: "duplicate-client-name",
          messageId: "nonDecorator",
          format: { name, scope: scopeStr },
          target: item,
        });
```
can remove duplication here i think

live1206:
```
const scopeStr = scope === AllScopes ? "AllScopes" : scope;
```
I tried this before, this doesn't really work.
The type of `scopeStr` here is `string | symbol`, we can't simply put it into for message.

```
const scopeStr = typeof scope === "string" ? scope : "AllScopes";
```
This could work since it ensures the type is `string`, but it will assign `AllScopes` under unexpected conditions.
So, I would prefer to keep the redundant code to make things work properly.

packages/typespec-client-generator-core/src/validate.ts
live1206: You are right, that would just work

packages/typespec-client-generator-core/src/validate.ts
timotheeguerin: are you sure you did that correctly? this works fine for me with a symbol

image

this is just a **lot** of duplicate code

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: I checked the target as below, it is the `clientName` decorator, wondering why it is pointing to the namespace instead...
image

timotheeguerin: Oh it’s because you target the decorator decoration not the application

packages/typespec-client-generator-core/src/validate.ts
live1206: OK, it's safe to case `string | symbol` to `string` in this case anyway, updated it.

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: could you elaborate on the definition of `application`?

packages/typespec-client-generator-core/src/validate.ts
timotheeguerin: It’s but the code should always be a string here and typescript is smart enough to understand that

packages/typespec-client-generator-core/test/decorators.test.ts
timotheeguerin: You are targeting the extern dec clientName

you should target the decorator application(call to @clientName) node that should be on the type having the issue

packages/typespec-client-generator-core/src/validate.ts
live1206: Apparently, typescript is not smart enough to understand this if I remove the casting, got error as below
```
src/validate.ts:146:27 - error TS2322: Type 'string | symbol' is not assignable to type 'string'.
```

timotheeguerin: So as shown above it works, please check you have the right type to start with

timotheeguerin: If the type is symbol | string then this whole think is wrong anyway

timotheeguerin: The type should be typeof AllScopes | string

live1206: `AllScopes` is a value of `Symbol`, not a type though

timotheeguerin: That’s why there is a typeof in front, please compare with my screenshot above

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: updated

packages/typespec-client-generator-core/src/validate.ts
live1206: updated

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: The bug is resolved in the [playground](https://cadlplayground.z22.web.core.windows.net/typespec-azure/prs/1119/?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsKCnVzaW5nIEHENC5DxSVHyCQuQ29yZTsKCgptb2RlbCBGb28ge30KCkDGTE5hbWUoIkZvbyIpxyFCYXLEIQ%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D).
Only versioning handling is left for this now.

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,205 @@
...
+  program: Program,
+  tcgcContext: TCGCContext,
+  scope: string | typeof AllScopes,
+  namespace: Namespace
+) {
+  const apiVersions = getVersions(program, namespace)[1]
+    ?.getVersions()
+    ?.map((x) => x.value);
+  if (apiVersions !== undefined) {
+    const versionProjections = buildVersionProjections(program, namespace);
```
timotheeguerin, line 55: The versioning should be done differently this is extremely clostly it will project the whole program for every version and do that again in the emitters.  Reminder that all those things need to be able to run very fast to get IDE support

It also will cause duplicate failures [See playground](https://cadlplayground.z22.web.core.windows.net/typespec-azure/prs/1119/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3ytjLWNsaWVudC1nZW5lcmF0b3LINgp1c2luZyBBxDQuQ8UlR8gkLkNvcmU7xyJUeXBlU3BlYy5IdHRw0RVSZXN00RVW6QDKyBvHZ8tXyhIuVHJhaXRzOwoKQHNlcnZpY2UKQOcBDGVkKENvbnRvc28uV2lkZ2V0TWFuYWdlcshicykKbmFtZXNwYWNlINUqOwoKLyoqIFRoySAgxiAgxyEgx3Qgx3MuICovCmVudW0gyGQgewogIHYxLMQGMiwKfQoKbW9kZWwgRm9vIHt9CgpA5gFjTmFtZSgiRm9vIinHIUJhcsQh&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fdata-plane%22%5D%7D%7D)

It would be better to ignore the versioning case for now if its too complex than doing this.

timotheeguerin: this is also probably not working when involving multiple service getting versioned individually.

In the versioning library all the validation is done by checking the data directly and checking if that name changed for example instead of traversing `n` graphs.

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,205 @@
...
+    string,
+    Type | DecoratorExpressionNode | AugmentDecoratorStatementNode
+  >,
+  scope: string | typeof AllScopes
+) {
+  for (const [name, duplicates] of duplicateTracker.entries()) {
+    for (const item of duplicates) {
+      const scopeStr = scope === AllScopes ? "AllScopes" : scope;
+      // If the item is a decorator application node
+      if (item.kind === 5) {
```
timotheeguerin, line 189: replace 5 with the `SyntaxKind.` enum value

live1206: resolved

packages/typespec-client-generator-core/src/validate.ts
live1206: > this is also probably not working when involving multiple service getting versioned individually.

I don't imagine how it can't be handled, but I agree we should skip it for now if projecting the whole program is extremely costly. Reverted the change for versioning.

> In the versioning library all the validation is done by checking the data directly and checking if that name changed for example instead of traversing n graphs

Just out of curiosity, could you point me to the place where versioning library is checking this?

timotheeguerin: Versioning validation here https://github.com/microsoft/typespec/blob/main/packages/versioning/src/validate.ts

and not really as you can see here versioning is incredibly complex(and still full of bugs)

packages/typespec-client-generator-core/src/validate.ts
timotheeguerin: I still see this, can't you just call 
```
validateClientNamesPerNamespace(program, tcgcContext, scope, program.getGlobalNamespaceType());
```

live1206: > I still see this, can't you just call
> 
> ```ts
> validateClientNamesPerNamespace(program, tcgcContext, scope, program.getGlobalNamespaceType());
> ```

Think this doesn't handle multiple services.

timotheeguerin: This will traverse every namespaces, if not you are missing a recursion anyway to look at sub namespace or the service

live1206: I was wondering whether or not 1 namespace is guaranteed to have 1 namespace.
If so, I am fine with this change.

live1206: changed it back, please let me know if anything is left within this PR.

timotheeguerin: having a service or not feels irrelevant of the use of `@clientName`. Everything is under the global namespace in the type graph

timotheeguerin (approved):

packages/typespec-client-generator-core/src/validate.ts
live1206: right, this is traversing everything within a namespace not a service.

tadelesh:
hold merge before next release of tcgc

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -1,12 +1,160 @@
...
+    for (const languageScope of Object.keys(value)) {
+      languageScopes.add(languageScope);
+    }
+  }
+  return languageScopes;
+}
+
+function validateClientNamesPerNamespace(
+  program: Program,
+  tcgcContext: TCGCContext,
```
tadelesh, line 46: duplication, you could get program from context.

tadelesh (approved):

packages/typespec-client-generator-core/src/validate.ts
live1206: good point, removed the duplicate

PR #1122, opened 2024-07-03
Fix regen of common types no validated in CI

timotheeguerin:


allenjzhang:
As discussed, can you move `x-ms-identifier` from `types.tsp` into the supplement file here.

.chronus/changes/fix-common-types-gen-2024-6-3-16-15-21.md
```diff
@@ -0,0 +1,8 @@
+---
```
allenjzhang, line 1: As discussed, can you move x-ms-identifier from types.tsp into the supplement file.

allenjzhang (approved):

PR #1123, opened 2024-07-03
Replace direct OpenAPI extension reference to add `x-ms-azure-resource` with internal decorator

allenjzhang:
Removed direct reference to OpenAPI extension `x-ms-azure-resource` in ARM library and replaced with `@Azure.ResourceManager.Private.azureResourceBase` decorator. It is only used internally on base resource types. `autorest` emitter has been updated to check the decorator and still emit `x-ms-azure-resource` extension in swagger.


.chronus/changes/azhang_RemoveAzureResource-2024-6-3-12-45-15.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: internal
```
timotheeguerin, line 2: doesn't feel like an internal change

packages/typespec-azure-resource-manager/src/private.decorators.ts
```diff
@@ -350,3 +350,12 @@ function hasProperty(program: Program, model: Model): boolean {
   if (model.baseModel) return hasProperty(program, model.baseModel);
   return false;
 }
+
+export function $azureResourceBase(context: DecoratorContext, resourceType: Model) {
+  context.program.stateMap(ArmStateKeys.azureResourceBase).set(resourceType, true);
+}
+
+export function isAzureResourceBase(program: Program, resourceType: Model): boolean {
```
timotheeguerin, line 358: isn't it `isAzureResource` more accuratE?>

chrisradek: +1 - not sure what the distinction would be with `base`

.chronus/changes/azhang_RemoveAzureResource-2024-6-3-12-45-15.md
allenjzhang: As we expect all ARM spec writers to extend from `Tracked/Proxy` or extreme cases `Resource`, spec writer should not use this nor the need to add `@extension("x-ms-azure-resource", true)` themselves. Hence I placed it in the `Private` namesapce. With the pending linting rule on `no reference Private namespace`, it is a pure internal matter, from outside (spec write et al) looking in, No?

timotheeguerin: i mean this expose a new api, if that change is internal this means it shouildn't technically rquire a new release which isn't accurate here

timotheeguerin (approved):

PR #1129, opened 2024-07-08
[tcgc] add `UsageFlags.Json`

### LINKED ISSUE #896: TCGC: Add Support for Json Usage Flag for a Given Model
The .net generator requires knowing when a model is used in an operation that had content type to `application/json` to aid with generating the serialization code for a model. The existing usage flags should be extended to support an additional `json` usage flag to indicate when a model is used in an operation who's content type includes `application/json`.

iscai-msft:
fixes #896


haolingdong-msft:
Double confirm: the flag will be added to all the models used by operation whose content type is `application/json`, right?

m-nash (approved):

PR #1146, opened 2024-07-10
Added back logic to keep doc unchanged

allenjzhang:


packages/typespec-azure-resource-manager/lib/decorators.tsp
```diff
@@ -125,12 +125,14 @@ extern dec armResourceCollectionAction(target: Operation);
...
  * @param parentFriendlyTypeName Friendly name for parent.
+ * @param updateOperationName Whether to update the operation name in addition to the doc. Default: true
  */
 extern dec armRenameListByOperation(
   target: Operation,
   resourceType: Model,
   parentTypeName?: valueof string,
-  parentFriendlyTypeName?: valueof string
+  parentFriendlyTypeName?: valueof string,
+  updateOperationName?: valueof boolean
```
timotheeguerin, line 135: this feels a little weird to have that flag when the decorator point is doing that

timotheeguerin: also should we not have just removed it?

allenjzhang: True. Since it has two parts, doc and operation name and now I am reusing it to only do doc for part of the call.

allenjzhang: Alternative is do another private decorator and invoke the original internally with the optional flag.

timotheeguerin: yeah my vote would be 
1. remove this decorator
2. make a new internal one that does the doc part

packages/typespec-azure-resource-manager/lib/private.decorators.tsp
```diff
@@ -71,3 +71,16 @@ extern dec enforceConstraint(target: Operation | Model, sourceType: Model, const
...
  */
 extern dec defaultListByOperationDoc(target: Operation, resource: Model);
+
+/**
+ * Marks the operation as being a collection action
+ * @param resourceType Resource
+ * @param parentTypeName: Parent type name.
+ * @param parentFriendlyTypeName Friendly name for parent.
+ */
+extern dec armRenameListByOperation(
```
timotheeguerin, line 81: could this omit the rename part and be called liek the previous decorator you added instead?

timotheeguerin (approved):

qiaozha (approved):

PR #1148, opened 2024-07-11
add `@override` to override default method signature generation

### LINKED ISSUE #1054: [tcgc] add `overrideClientMethod` decorator

We want the ability in `client.tsp` to override a generated client method. When we switch to `...` always meaning spread, we will also need the ability to generate options bags for languages that prefer options bags. Adding a decorator like `overrideClientMethod` will allow authors to clearly state their preferred signature for specific language(s). This will also allow patched customizations without having to craft an entire client to modify one operation

```tsp
// main.tsp
namespace MyService;

model Params = {
  foo: string;
  bar: string;
}

op myOperation(...Params): void;

// client.tsp
namespace MyCustomizations;

op myOperationCustomization(@bodyRoot params: Params): void;

@@overrideClientMethod(MyService.myOperation, MyCustomizations.myOperationCustomization);
```

### Checklist

iscai-msft:
fixes #1054


packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3688,4 +3688,40 @@ describe("typespec-client-generator-core: decorators", () => {
...
+        }
+
+        op original(...Params): void;
+        `,
+        `
+        namespace MyCustomizations;
+        model NewModel {
+          params: MyService.Params;
+        }
+        op customization(...NewModel): void;
```
weidongxu-microsoft, line 3710: @tadelesh 

Would this `op customization` be picked by TCGC as client method?

iscai-msft: it isn't because it's not under a client

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -991,3 +992,68 @@ export function $clientName(
...
+  properties.forEach((value, key) => {
+    // If the property is of type 'model', recurse into its properties
+    if (value.type.kind === "Model") {
+      paramNames = paramNames.concat(collectParamNames(value.type.properties));
+    } else {
+      paramNames.push(key);
+    }
+  });
+
+  return paramNames;
```
tadelesh, line 1011: using name for validation is not safe. i suggest to use model property instance to do the validation.

tadelesh:
another concern is what weidong mentioned in the issue: https://github.com/Azure/typespec-azure/issues/1054#issuecomment-2225461649

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1038,3 +1039,81 @@ export function getClientNameOverride(
...
+const overrideClientMethodKey = createStateSymbol("overrideClientMethod");
+
+// Recursive function to collect parameter names
+function collectParams(properties: RekeyableMap<string, ModelProperty>): ModelProperty[] {
+  let params: ModelProperty[] = [];
+
+  properties.forEach((value, key) => {
+    // If the property is of type 'model', recurse into its properties
+    if (value.type.kind === "Model") {
+      params = params.concat(collectParams(value.type.properties));
```
tadelesh, line 1052: will this cause infinite loop if cyclic ref?

iscai-msft: good catch, let me make sure we're not duplicating these properties

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -4256,4 +4256,265 @@ describe("typespec-client-generator-core: decorators", () => {
...
+
+    it("regrouping", async () => {
+      const mainCode = `
+        @service
+        namespace MyService;
+        model Params {
+          foo: string;
+          bar: string;
+          fooBar: string;
+        }
```
weidongxu-microsoft, line 4398: Can we mix a `@query` or `@header` into this model? It is pretty common to have a POST that contains body and a few query param or header param.

One uncertainty is whether the logic of `compareModelProperties` can handle some query/header that was not explicitly defined in service typespec.
For example, service API may have conditional headers like `ifMatch`. And these would be from typespec-azure-core [trait](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/traits.tsp#L163-L176), where [headers](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/models.tsp#L136-L157)
<-- it is fine we handle this later. We can get the PR merged for simple "revert the spread" scenario.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/test/decorators.test.ts
tadelesh: a related concern is that is the code could support override templated operation like `op scheduleRepairs is Operations.ResourceAction<Widget, RepairInfo, RepairResult>;`

.chronus/changes/override_client_method-2024-6-10-21-0-35.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+add `@overrideClientMethod` decorator that allows authors to explicitly describe their desired client method
```
srnagar, line 7: Update name of the decorator to `@override`.

docs/howtos/DataPlane Generation - DPG/08methodInputs.mdx
```diff
@@ -961,22 +961,17 @@ func (client *Client) ScheduleRepairs(ctx context.Context, widgetName string, pr
 </TabItem>
 </Tabs>
 
-If you want to keep the model, you could use a wrapper to explicit set the body to prevent spread.
+If you want to keep the model, you have two options to prevent spreading:
+
+1. `@override`
+2. If you don't want to do client customizations, you could use a wrapper to explicit set the body to prevent spread.
```
srnagar, line 967:
```suggestion
2. If you don't want to do client customizations, you could use a wrapper to explicitly set the body to prevent spread.
```

docs/libraries/typespec-client-generator-core/reference/decorators.md
```diff
@@ -450,6 +450,47 @@ Create a ClientGenerator.Core operation group out of a namespace or interface
...
+ bar: string;
+}
+op myOperation(...Params): void; // by default, we generate the method signature as `op myOperation(foo: string, bar: string)`;
+
+// client.tsp
+namespace MyCustomizations;
+
+op myOperationCustomization(@bodyRoot params: Params): void;
+
+@@override(MyService.myOperation, MyCustomizations.myOperationCustomization);
```
srnagar, line 490: So, the name of the operation in client.tsp doesn't matter if it's used as an override? Codegen will use the operation name defined in main.tsp?

packages/typespec-client-generator-core/README.md
```diff
@@ -457,6 +458,47 @@ Create a ClientGenerator.Core operation group out of a namespace or interface
...
+ bar: string;
+}
+op myOperation(...Params): void; // by default, we generate the method signature as `op myOperation(foo: string, bar: string)`;
+
+// client.tsp
+namespace MyCustomizations;
+
+op myOperationCustomization(@bodyRoot params: Params): void;
+
+@@override(MyService.myOperation, MyCustomizations.myOperationCustomization);
```
srnagar, line 498: It might be good to add an example that scopes it to a specific language.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -159,6 +159,12 @@ export const $lib = createTypeSpecLibrary({
         default: `Cannot pass an empty value to the @clientName decorator`,
       },
     },
+    "override-method-parameters-mismatch": {
+      severity: "error",
+      messages: {
+        default: paramMessage`Method "${"methodName"}" is not directly referencing the same parameters as in the original operation. There needs to be a 1-1 mapping.`,
```
srnagar, line 165: Can we also provide more details in the diagnostics on which params are missing/new in the overridden method?

srnagar (commented):
Update the PR title as well to use the new decorator name.

docs/libraries/typespec-client-generator-core/reference/decorators.md
iscai-msft: yep the name in client.tsp doesn't matter, it will just use the service name

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -159,6 +159,12 @@ export const $lib = createTypeSpecLibrary({
         default: `Cannot pass an empty value to the @clientName decorator`,
       },
     },
+    "override-method-parameters-mismatch": {
+      severity: "error",
+      messages: {
+        default: paramMessage`Method "${"methodName"}" is not directly referencing the same parameters as in the original operation. The original method has parameters "${"originalParameters"}", while the override method has parameters "${"overrideParameters"}".`,
```
srnagar, line 165: Nice! Thanks for updating this message as it makes troubleshooting a lot easier now.

srnagar (approved):
LGTM!

Issue #1027, opened 2024-06-17
[Linter] Add warning or error if spec is not versioned


mikeharder:
If a resource-manager spec doesn't specify `@versioned`, TypeSpec generates swagger under a folder named `0000-00-00`, which may be confusing to spec authors.  It would be helpful to raise a warning or error suggesting to add `@versioned`.

[...]

PR #1154, opened 2024-07-11
Add new linting rule requiring versioning

timotheeguerin:
fix [#1027](https://github.com/Azure/typespec-azure/issues/1027)


packages/samples/specs/resource-manager/arm-library/main.tsp
```diff
@@ -17,12 +17,19 @@ using Azure.ResourceManager;
...
 namespace Microsoft.LibraryTest;
 
+/** Contoso API versions */
+enum Versions {
+  /** 2021-10-01-preview version */
+  @useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v5)
+  `2021-10-01-preview`,
+}
+
```
chrisradek, line 32: There was some copy pasting I think:
```suggestion
/** ARM Library API versions */
enum Versions {
  /** 2021-09-21-preview version */
  @useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v5)
  `2021-09-21-preview`,
}

```
I think we can delete the `@OpenAPI.info` decorator as well? (should we warn if that's present with `@versioned`?)

timotheeguerin: yeah should definitely delete it from the samples and yeah maybe but in general for azure we don't want people to use openapi decorators at all

chrisradek (approved):

PR #1155, opened 2024-07-11
[tcgc] make endpoint overridable

### LINKED ISSUE #1153: Always make the `baseUrl` overridable in clients
Tcgc issue from the discusssion here https://github.com/Azure/typespec-azure/issues/217

As discussed the `baseUrl` (`https://foo.com/bar`) should always be settable in clients as an option.

Issue was also filed to see if http library should also provide a different view https://github.com/microsoft/typespec/issues/3821 in the future

> 
> Just to clarify here too but it should be the baseUrl that is overridable not the host(Basically the entire thing that `@server` would set. Without that you are blocking many scenarios still:
> 
> - `http:// localhost`
> - different hosted path

iscai-msft:
fixes #1153


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -508,7 +508,33 @@ function getSdkEndpointParameter(
         );
       }
     }
-    optional = Boolean(servers[0].url.length && templateArguments.every((param) => param.optional));
+    if (templateArguments.length === 0) {
+      templateArguments.push({
+        kind: "path",
+        name: "endpoint",
```
timotheeguerin, line 514: baseUrl?

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -508,7 +508,33 @@ function getSdkEndpointParameter(
...
+        urlEncode: false,
+        optional: false,
+        serializedName: "endpoint",
+        correspondingMethodParams: [],
+        type: {
+          kind: "string",
+          encode: "string",
+          decorators: [],
+        },
+        clientDefaultValue: servers[0].url,
```
timotheeguerin, line 527: isn't that breaking if this has parameters?

iscai-msft: I only add if the url is not parameterized

packages/typespec-client-generator-core/src/package.ts
iscai-msft: yep that's fair

packages/typespec-client-generator-core/src/package.ts
timotheeguerin: this feels like it defeat the point though, if someone does this 
```
@server("https://{region}.myservice.com", "", {region: string})
```
I still would want to be able to provide by local test server/some staging url

timotheeguerin: this is I guess my whole point on why this make 0 sense to have as a compiler/http library api

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -511,13 +511,13 @@ function getSdkEndpointParameter(
     if (templateArguments.length === 0) {
       templateArguments.push({
         kind: "path",
-        name: "endpoint",
+        name: "baseUrl",
```
timotheeguerin, line 514: I guess if endpoint is the whole thing too and how you use to call it its fine

packages/typespec-client-generator-core/src/package.ts
weidongxu-microsoft: Just a comment, that does not affect the discussion here.

The reason why we need `endpoint` in ARM, is that `@server("https://management.azure.com")` is just wrong.
That URL is only valid for Azure Global.
Azure US Gov, Azure Germany, Azure China all uses different endpoint. E.g. Azure Gov is `"https://management.usgovcloudapi.net/"`.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -508,7 +508,33 @@ function getSdkEndpointParameter(
...
+        isApiVersionParam: false,
+        apiVersions: context.__tspTypeToApiVersions.get(client.type)!,
+        crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, client.service)}.baseUrl`,
+        decorators: [],
+      });
+    }
+    optional = Boolean(
+      servers[0].url.length &&
+        templateArguments.every((param) => param.clientDefaultValue !== undefined)
+    );
```
weidongxu-microsoft, line 537: quick question, what this `optional` for? And why it now relates to `clientDefaultValue`?

weidongxu-microsoft (commented):
LGTM

packages/typespec-client-generator-core/src/package.ts
timotheeguerin: > (though this sound a bit arrogant as we are forcing something to dev -- regardless of how they write the typespec, and their only choice otherwise is not to use our emitter) 
I would argue strongly for the opposite, as a spec writer I wrote the server(s) I want my user to use. But I wouldn't want to describe my dev/stagging/test server in the spec but still have the option to do so. If you feel strongly about this behavior you can always provide an option to disable it. This should be an option when creating a client, user can use it or not.

And yeah would argue the same for code generated from OpenAPI this is the same case.


For the Arm specific case however I kinda agree that relying on this functionality might not be enough. The way I see it is an escape hatch which means a client would provide it but it might not be obvious.

Though I would also compare that to the Github API which is about the same scenario where you have your own hosted entreprise github with a different endpoint

```ts
new Octokit() // Github public endpoint
new Octokit({
  baseUrl: "https://myentreprise.hostedgithub.com",
})
```

so I would be fine additionally from this feature updating the arm library to provide either a list of all the cloud(not ideal) or just say that the host is configurable but the default is arm but I think it really depends on what is the experience for this escape hatch. This will also be a BREAKING change to the swagger which all use the public cloud today

weidongxu-microsoft: > you can always provide an option to disable it. This should be an option when creating a client, user can use it or not.

Yeah, agree this may be a better solution. Codegen can refine that later.

PS: dev can override the endpoint anyway, in current SDK, with the common HttpPipelinePolicy (though it looks bad) e.g.
```java
        new ##ClientBuilder()
            .addPolicy((context, policy) -> {
                context.getHttpRequest().setUrl(context.getHttpRequest().getUrl().toString().replace("http://localhost:3000/", "http://localhost:3001/"));
                return policy.process();
            })
            .buildClient()
```

haolingdong-msft (approved):

packages/typespec-client-generator-core/src/package.ts
timotheeguerin: yeah that does feel like a little too much here, but if you had 
```
new ClientBuilder().overrideBaseUrl("https://localhost:30001")
```

instead of explicitly allowing it in the constructor then I think that would satisfy

iscai-msft:
For this PR, I'm going to hold on merging it before this release because there's still open discussion about what the best way to do this is. In the meantime, I want to confirm that no language is blocked right now, correct? It's just that emitters are having to hardcode information about ARM.

haolingdong-msft:
> For this PR, I'm going to hold on merging it before this release because there's still open discussion about what the best way to do this is. In the meantime, I want to confirm that no language is blocked right now, correct? It's just that emitters are having to hardcode information about ARM.

Hi @iscai-msft, I'm afraid it blocks Java adopting TCGC sdkpackage, I guess it also blocks other languages, as it will impact all management-plane generated SDKs. So if TCGC will not do this and we would like to adopt TCGC sdkpackage, each language emitter will do their own handling.

[API change](https://apiview.dev/Assemblies/Review/e0cf256f814e456bac0f0d751bd4aa9f/99297657ec5a4c28a177ad9d3a984044?diffRevisionId=b792a3ef67e044379ff8407f58c99dd1&diffOnly=True) if we do not make endpoint overridable.
image

iscai-msft:
@haolingdong-msft can java do their own handling for this release? I would use this code snippet:

```typescript

function serializeClient(context: SdkContext, client: SdkClientType) {
  ...
  for (param of client.initialization.parameters) {
    if (param.kind === "endpoint" && context.arm) {
      param.clientDefaultValue = param.serverUrl;
      param.serverUrl = "{baseUrl}";
      param.templateArguments = [{
          "baseUrl",
          isGeneratedName: true,
          description: "Service host",
          kind: "path",
          onClient: true,
          urlEncode: false,
          optional: false,
          serializedName: "baseUrl",
          correspondingMethodParams: [],
          type: {
            kind: "string",
            encode: "string",
            decorators: [],
          },
          isApiVersionParam: false,
          apiVersions: context.__tspTypeToApiVersions.get(client.type)!,
          crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, client.service)}.baseUrl`,
          decorators: [],
        },
      ]
    }
  }
}

function mainFunc(context: SdkContext) {
  for (const client of sdkPackage.clients) {
    serializeClient(context, client);
  }
}

```

We just want to make sure everyone's on the same page before committing this to tcgc. Agree that this isn't very nice code, but with this it shouldn't be blocking

haolingdong-msft:
> ```ts
> templateArguments 
> ```

It's fine, I will do the handling in Java emitter first. Thanks

tadelesh:
@iscai-msft is there any update for this? for typespec like follows, it seems there is no default value from current templated argument
```
@server(
  "{endpoint}",
  "Test server endpoint",
  {
    endpoint: string = "http://localhost:3000",
  }
)
```

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -419,35 +420,35 @@ function getSdkEndpointParameter(
...
-          type: getTypeSpecBuiltInType(context, "string"),
-          isApiVersionParam: false,
-          apiVersions: context.__tspTypeToApiVersions.get(client.type)!,
-          crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, client.service)}.endpoint`,
-          decorators: [],
-        },
-      ],
-      decorators: [],
-    };
+    type = defaultOverridableEndpointType;
```
tadelesh, line 451: since we decided to use union as the endpoint type, for multi server url, why not also use union to generate all the defined endpoint?

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -973,15 +989,18 @@ These are parameters to client initialization and method son the client. These w
 
 #### SdkEndpointParameter
 
-An `SdkEndpointParameter` represents a parameter to a client's endpoint. If the client is not parameterized, the endpoint parameter represents the entire endpoint input for the client. For parameterized clients, there will be as many `SdkEndpointParameter`s as needed to parameterize the client endpoint.
+An `SdkEndpointParameter` represents a parameter to a client's endpoint.
 
-```ts
+TCGC will always give it to you as overridable: If the server URL is a constant, we will return a templated endpoint with a default value of the constant server URL.
+In the case where the endpoint has extra template arguments, the type is a union of a completely-overridable endpoint, and an endpoint that accepts template arguments.
```
srnagar, line 995: We should also document the behavior when there are multiple servers - one with a default value and one with templated url.

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -1407,7 +1426,7 @@ export type AccessFlags = "internal" | "public";
 ```ts
 export interface SdkEndpointType {
   kind: "endpoint";
-  serverUrl?: string;
+  serverUrl: string;
```
srnagar, line 1429: Is this a breaking change for any of the emitters?

srnagar (approved):

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
iscai-msft: No, this was just a fix in documentation of our type: the `serverUrl` has been required this whole time

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -413,78 +414,113 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
...
+    serverUrl: "{endpoint}",
+    templateArguments: [
+      {
+        name: "endpoint",
+        isGeneratedName: true,
+        description: "Service host",
+        kind: "path",
+        onClient: true,
+        urlEncode: false,
+        optional: false,
```
tadelesh, line 435: it seems all the endpoint param is not optional? then param with default value is also not optional, is it by design?

haolingdong-msft (approved):

packages/typespec-client-generator-core/src/package.ts
iscai-msft: yeah, i realized the endpoint is always required to be sent to the service: there might be a client default value, but that client default value still needs to be sent

PR #1161, opened 2024-07-12
Optional template parameter to customize the optionality of ARM resource `properties` field

### LINKED ISSUE #131: Customization for arm build-in model's visibility and optional setting for resource `properties` property

allenjzhang:
Added an optional template parameter on `TrackedResource`, `ProxyResource`, and `ExtensionResource` ARM templates that allows brownfield services to customize the optionality of the ARM resource `properties` field.

Closes #131

### DIFF
--- a/docs/libraries/azure-resource-manager/reference/data-types.md
+++ b/docs/libraries/azure-resource-manager/reference/data-types.md
@@ -465,14 +465,15 @@ Concrete extension resource types can be created by aliasing this type using a s
 See more details on [different Azure Resource Manager resource type here.](https://azure.github.io/typespec-azure/docs/howtos/ARM/resource-type)
 
 ```typespec
-model Azure.ResourceManager.ExtensionResource<Properties>
+model Azure.ResourceManager.ExtensionResource<Properties, OptionalProperties>
 ```
 
 #### Template Parameters
 
-| Name       | Description                                                           |
-| ---------- | --------------------------------------------------------------------- |
-| Properties | A model containing the provider-specific properties for this resource |
+| Name               | Description                                                                                                                                    |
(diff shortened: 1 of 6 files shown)


docs/libraries/azure-resource-manager/reference/data-types.md
```diff
@@ -631,14 +632,15 @@ Concrete proxy resource types can be created by aliasing this type using a speci
 See more details on [different Azure Resource Manager resource type here.](https://azure.github.io/typespec-azure/docs/howtos/ARM/resource-type)
 
 ```typespec
-model Azure.ResourceManager.ProxyResource<Properties>
+model Azure.ResourceManager.ProxyResource<Properties, OptionalProperties>
```
timotheeguerin, line 635: Would `PropertiesOptional` be more accurate in the name?

timotheeguerin (approved):

markcowl (approved):

heaths:
Does this only affect ARM templates, though? We need it for CP as well.

timotheeguerin:
> @markcowl, @allenjzhang does this only affect ARM templates, though? We need it for CP as well.

Can you say more? This problem is very specific to the arm library

PR #1166, opened 2024-07-15
Fix #1048 Allow void in final result for Azure-AsyncOperation header

### LINKED ISSUE #1048: AzureAsyncOperationHeader template should accept 'void'  for FinalResult
The type restriction only accepts a model, but should accept a model or void type.

markcowl:


packages/typespec-azure-resource-manager/lib/responses.tsp
```diff
@@ -138,7 +138,7 @@ model ArmOperationStatus<
 model ArmAsyncOperationHeader<
   StatusMonitor extends {} = ArmOperationStatus,
   UrlValue extends string = string,
-  FinalResult extends {} = never
+  FinalResult extends {} | void = never
```
timotheeguerin, line 141: Is void resulting in different than never

markcowl: Yes,  never causes the decorator not to execute; void has it execute with value void.  The difference in the LRO metadata is subtle, but important

markcowl: In terms of the lro, never means the last call is to the polling endpoint,  'void'  means the endpoint is dereferenced again after polling completes and returns an empty body response

timotheeguerin (approved):

PR #1192, opened 2024-07-17
have createSdkContext create sdkPackage, createTcgcContext not

iscai-msft:


packages/typespec-autorest-canonical/src/emitter.ts
```diff
@@ -27,6 +27,7 @@ import {
   getVersion,
 } from "@typespec/versioning";
 import { AutorestCanonicalEmitterOptions, reportDiagnostic } from "./lib.js";
+import { TCGCContext } from "../../typespec-client-generator-core/dist/src/internal-utils.js";
```
timotheeguerin, line 30: this also need to be exported

packages/typespec-autorest/src/utils.ts
```diff
@@ -14,12 +14,13 @@ import {
 } from "@typespec/compiler";
 import { getOperationId } from "@typespec/openapi";
 import { pascalCase } from "change-case";
+import { TCGCContext } from "../../typespec-client-generator-core/dist/src/internal-utils.js";
```
timotheeguerin, line 17: same here

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -18,7 +22,38 @@ import {
   HttpVerb,
   Visibility,
 } from "@typespec/http";
-import { TCGCContext } from "./internal-utils.js";
+import { TspLiteralType } from "./internal-utils.js";
+
+export interface TCGCContext {
```
timotheeguerin, line 27: is there too many props here? I don't see them set in the createTCGCcontecxt?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -18,7 +22,38 @@ import {
   HttpVerb,
   Visibility,
 } from "@typespec/http";
-import { TCGCContext } from "./internal-utils.js";
```
timotheeguerin, line 21: I don't see the other one being removed

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: it currently is only adding the required props, i have optional props that are more for creating the sdk package, so I'm trying to have `createTcgcContext` create the minimal interface needed

packages/typespec-client-generator-core/src/validate.ts
```diff
@@ -16,12 +16,13 @@ import {
...
-import { getClientNameOverride } from "./decorators.js";
-import { AllScopes, clientNameKey, createTCGCContext, TCGCContext } from "./internal-utils.js";
+import { getClientNameOverride, createTCGCContext } from "./decorators.js";
+import { AllScopes, clientNameKey } from "./internal-utils.js";
 import { reportDiagnostic } from "./lib.js";
+import { TCGCContext } from "./interfaces.js";
 
 export function $onValidate(program: Program) {
-  const tcgcContext = createTCGCContext(program);
+  const tcgcContext = createTCGCContext(program, "@azure-tools/typespec-client-generator-core");
```
timotheeguerin, line 25: I feel like it is good that this was not provided

packages/typespec-client-generator-core/src/interfaces.ts
timotheeguerin: but wouldn't you want that on the `SDKContext` then?

packages/typespec-autorest/src/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { SdkContext, getClientNameOverride } from "@azure-tools/typespec-client-generator-core";
+import { getClientNameOverride, TCGCContext } from "@azure-tools/typespec-client-generator-core";
```
timotheeguerin, line 1:
```suggestion
import { getClientNameOverride, type TCGCContext } from "@azure-tools/typespec-client-generator-core";
```

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: ok that's fair, let me go ahead and do that. thanks!

packages/typespec-client-generator-core/src/validate.ts
iscai-msft: why would it be good if it wasn't applied?

timotheeguerin: We are technically in a stage where there is no specific scope so this is kinda lying and might have unexpected consequences

iscai-msft:
@timotheeguerin, circling back to the optional params in the SdkContext interface, they are actually needed in `getAllModels`, which TCGCContext needs to be able to be passed to (even though we aren't automatically calling it in `createTcgcContext`. The reason they're optional is because they're created on the fly. I've moved them back to `TcgcContext`, let me know if there's a better typescript design here that could happen though

timotheeguerin (approved):

PR #1198, opened 2024-07-18
Migrate azure core library to use generated decorator signature

timotheeguerin:
Step to having all our libraries use tspd to generate the signatures and make sure they are in sync.

There is quite a lot of times we expected the wrong type which would cause a crash. This fix a few to get the more precise type in the TypeSpec if it was going to crash originally. Otherwise expand the type in the TS


chrisradek (approved):
Yay for auto-generated defs!

PR #1204, opened 2024-07-19
[tcgc] add xml usage and fix generic decorator issue

tadelesh:
for xml support

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -603,4 +603,6 @@ export enum UsageFlags {
   Error = 1 << 7,
   // Set when model is used in conjunction with an application/json content type.
   Json = 1 << 8,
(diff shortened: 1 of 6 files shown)


iscai-msft (approved):

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -603,4 +603,6 @@ export enum UsageFlags {
   Error = 1 << 7,
   // Set when model is used in conjunction with an application/json content type.
   Json = 1 << 8,
+  // Set when model is used in conjunction with an application/xml content type.
+  Xml = 1 << 8,
```
jhendrixMSFT, line 607: Shouldn't this be `1 << 9`?

ArcturusZhang: I found this as well - I will open a PR to fix this.

PR #1208, opened 2024-07-19
[tcgc] add linter rulesets framework

### LINKED ISSUE #1147: [tcgc] adding linter rules for languages to tcgc

Currently, we don't have a unified approach for throwing linter warnings that language SDKs will run into. Since we want these linter warnings to come up during definition time, we agree that it would be best to include these linter warnings in tcgc. That being said, we should still have a mapping for what warnings apply to which languages. This mapping will be internal, so we can figure out when to throw, and properly use information from `scope` to determine whether the issue has been fixed. One idea is to have an exclusion and inclusion list for languages

When generating `azure`, we want to run linter warnings for all languages, since it is within Azure guidelines to generate all major languages. When generating `unbranded`, we will only run language-specific warnings. This is because we don't want to force an unbranded person only generating c# to have to deal with python-specific warnings.

I'll start off by designing a framework that will allow easy mapping to languages, and easily allow languages to opt in or opt out of rules. One good rule of thumb is, if it's a warning log, it will be a good idea to move it to be an actual linter warning

### Checklist

iscai-msft:
fixes #1147


docs/libraries/typespec-client-generator-core/rules/require-client-suffix.md
```diff
@@ -0,0 +1,44 @@
...
+@@client(MyService);
+```
+
+#### ✅ Ok
+
+Either you can rename the client using input param `{"name": "MyClient"}`, or if you are completely recreating a client namespace in your `client.tsp`, you can rely on that namespace to end in `Client`
+
+```ts
+// client.tsp
+import "@azure-tools/typespec-client-generator-core";
```
timotheeguerin, line 36: For the example I would omit those imports and using just focus on the actual problem

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -3,25 +3,13 @@ import { createTypeSpecLibrary, paramMessage } from "@typespec/compiler";
...
       },
     },
-    "unknown-client-format": {
-      severity: "error",
-      messages: {
-        default: paramMessage`Client format "${"format"}" is unknown. Known values are "${"knownValues"}"`,
-      },
-    },
-    "incorrect-client-format": {
+    "invalid-client-format": {
```
timotheeguerin, line 12: Is this actually used?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -929,7 +922,7 @@ export const $access: AccessDecorator = (
 ) => {
   if (typeof value.value !== "string" || (value.value !== "public" && value.value !== "internal")) {
     reportDiagnostic(context.program, {
-      code: "access",
+      code: "incorrect-access",
```
timotheeguerin, line 925: isn't it `invalid-access` in lib.ts that should make the build fail, doesn't it?

packages/typespec-client-generator-core/test/rules/require-client-suffix.test.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  BasicTestRunner,
+  createLinterRuleTester,
+  LinterRuleTester,
+} from "@typespec/compiler/testing";
+import { beforeEach, describe, it } from "vitest";
+import { requireClientSuffixRule } from "../../src/rules/require-client-suffix.js";
+import { createTcgcTestRunner } from "../test-host.js";
+
+describe("typespec-client-generator-core: client names end with 'Client'", () => {
```
timotheeguerin, line 10: you can remove the top level describe now, with vitgest it adds extra nesting otherwise

docs/libraries/typespec-client-generator-core/rules/require-client-suffix.md
```diff
@@ -0,0 +1,44 @@
...
+import "./main.tsp";
+
+using Azure.ClientGenerator.Core;
+
+@client({service: MyService})
+namespace MyClient {
```
timotheeguerin, line 42: I would also show more example like you can use `@clientName`, ect.

And you can create 2 category to still provide user with the best of the option 
- ✅ Recommended
- ✅ Ok

packages/typespec-client-generator-core/src/lib.ts
iscai-msft: it's used when we're checking the input of [`@clientFormat`](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/decorators.ts#L793). Is this valid?

docs/libraries/typespec-client-generator-core/reference/linter.md
```diff
@@ -0,0 +1,31 @@
...
+Available ruleSets:
+
+- `@azure-tools/typespec-client-generator-core/all`
+- `@azure-tools/typespec-client-generator-core/best-practices:csharp`
+
+## Rules
+
+| Name                                                                 | Description                                                             |
+| -------------------------------------------------------------------- | ----------------------------------------------------------------------- |
+| `@azure-tools/typespec-client-generator-core/require-client-suffix`  | Client names should end with 'Client'.                                  |
```
m-nash, line 30: can we specify which ruleset each of these are in?

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. Generic rule that applies to all emitted languages
+2. Specific rule that applies to a subset of all language: it can even apply to just one language
+
+The process for adding a rule starts off the same for both: for language-specific rules, there is an extra step
+
+1. Write the rule in `typespec-azure/packages/typespec-client-generator-core/src/rules/[rule-name].rule.ts
+2. Add reference to the rule in the `rules` array in [`typespec-azure/packages/typespec-client-generator-core/src/linter.ts`][tcgc-linter]
+   - This will automatically add it to a ruleset called `:all` for `@azure-tools/typespec-client-generator-core`
+3. Add the rule to the enable list for [`data-plane.ts`][data-plane-ruleset] and/or [`resource-manager.ts`][resource-manager-ruleset] in the [rulesets][rulesets] package. You can set `enable` to `false` here, if you want to delay enabling
```
m-nash, line 15: Do we really want to have different rules for data plane and mgmt plane?  I know we are less strict with mgmt plane, but aren't the recommendations the same for both?

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+
+**If you are adding a language-specific rule**, you will also need this extra step 4. Add reference to the rule in the `[language]Rules` array in [`typespec-azure/packages/typespec-client-generator-core/src/linter.ts`][tcgc-linter]
+
+For Azure generations then, all rules, including all language-specific rules, will be run on the specs.
```
m-nash, line 19: I think we want to limit this to the Tier 1 languages not necessarily `all` languages?  Should we specify this?  I assume Go could add their own rules in here and those wouldn't apply unless generating Go even for azure?

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+For unbranded generations, since we've added the rules into specific `best-practices:[language]` rulesets, you can explicitly specify a subset of rules in your `tsp-config.yaml`, i.e. if I only want Python best-practices, I could add this in my `tsp-config.yaml`:
```
m-nash, line 20: This might create a bad situation where someone can turn off csharp rules but in reality they cannot.  I think we might want to consider what flexibility a user would have over tcgc rules when running in azure context.

docs/libraries/typespec-client-generator-core/reference/linter.md
```diff
@@ -0,0 +1,31 @@
...
+  extends:
+    - "@azure-tools/typespec-client-generator-core/all"
+```
+
+## RuleSets
+
+Available ruleSets:
+
+- `@azure-tools/typespec-client-generator-core/all`
+- `@azure-tools/typespec-client-generator-core/best-practices:csharp`
```
m-nash, line 24: What does best practice mean here?  Does it mean this is optional?  If so is there another category which is not optional?

docs/libraries/typespec-client-generator-core/reference/linter.md
```diff
@@ -0,0 +1,31 @@
...
+linter:
+  extends:
+    - "@azure-tools/typespec-client-generator-core/all"
+```
+
+## RuleSets
+
+Available ruleSets:
+
+- `@azure-tools/typespec-client-generator-core/all`
```
m-nash, line 23: I am not sure `all` would ever be useful when considering non tier 1 languages as well as required rules vs best practices?  Should we be a bit more granular on what `all` means?

docs/libraries/typespec-client-generator-core/rules/property-name-conflict.md
```diff
@@ -0,0 +1,27 @@
...
+title: "property-name-conflict"
+---
+
+```text title="Full name"
+@azure-tools/typespec-client-generator-core/property-name-conflict
+```
+
+Verify that there isn't a name conflict between a property in the model, and the name of the model itself
+
+#### ❌ Incorrect
```
m-nash, line 11: I think we should use Invalid over Incorrect.

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
iscai-msft: That's the current setup of `azure-core-rulesets`, where you explicitly list rules for the [data plane](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-rulesets/src/rulesets/data-plane.ts) and [mgmt](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-rulesets/src/rulesets/resource-manager.ts) rulesets

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
iscai-msft: That's fair, let me update this to say that language-agnostic rules, and tier-1 languages will be run

docs/libraries/typespec-client-generator-core/reference/linter.md
iscai-msft: `all` isn't a specificly-defined ruleset, it's the standard name for all of the rules existing in a package

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
timotheeguerin: there is some rules that management plane specific and some that shouldn't apply to mamangement plane. But the majority should apply to both

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
timotheeguerin: every azure service is enforced to have one of the 2 rules set enabled by the spec repo pipeline.

docs/libraries/typespec-client-generator-core/reference/linter.md
timotheeguerin: That file is auto generated with the list so would need to update the ref doc generator. If you have some idea on what'd you like to see feel free to suggest a new style but this should work for a generic linter

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
iscai-msft: With rulesets, we have to give users the ability to turn them off because they are ultimately warnings. After talking with @timotheeguerin, you can't have error linter rules, only warnings. If there is a case where you explicitly want to error in csharp for a csharp-specific issue, the csharp emitter itself can throw a fatal error with `reportDiagnostic`, but the fatal error should have an analog in the linter warnings in tcgc that says "if you do this in csharp and ignore me, the csharp emitter will fail"

docs/libraries/typespec-client-generator-core/reference/linter.md
iscai-msft: best practices is what I've called the ruleset for now, open to suggestions! I think the advice from the typespec team has been to use the word "best-practices", @timotheeguerin is this correct?

docs/libraries/typespec-client-generator-core/rules/property-name-conflict.md
timotheeguerin: Yeah just pattern from how eslint document its rules, e.g. https://typescript-eslint.io/rules/class-literal-property-style

But if we feel like changing then we can discuss the common format we want

docs/libraries/typespec-client-generator-core/reference/linter.md
srnagar: Aren't these rulesets optional and can be opted out?

docs/libraries/typespec-client-generator-core/reference/linter.md
```diff
@@ -0,0 +1,31 @@
...
+- `@azure-tools/typespec-client-generator-core/best-practices:csharp`
+
+## Rules
+
+| Name                                                                 | Description                                                             |
+| -------------------------------------------------------------------- | ----------------------------------------------------------------------- |
+| `@azure-tools/typespec-client-generator-core/require-client-suffix`  | Client names should end with 'Client'.                                  |
+| `@azure-tools/typespec-client-generator-core/property-name-conflict` | Avoid naming conflicts between a property and a model of the same name. |
```
srnagar, line 31: Is this a required rule or is it a recommendation to avoid property and model having the same name.

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
+# Contributing Linter Rules to the @azure-tools/typespec-client-generator-core Package
+
+For information about linter rules in typespec in general, view [here][generic-linter]
+
+In the `@azure-tools/typespec-client-generator-core` library, we have two main types of rules:
+
+1. Generic rule that applies to all emitted languages
+2. Specific rule that applies to a subset of all language: it can even apply to just one language
```
srnagar, line 8:
```suggestion
2. Specific rule that applies to a subset of all languages: it can even apply to just one language
```

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
srnagar: Do we have a single place to add a rule that applies to both or do we need to add the rule explicitly to both lists?

docs/libraries/typespec-client-generator-core/rules/property-name-conflict.md
```diff
@@ -0,0 +1,27 @@
...
+
+```ts
+model Widget {
+  widget: string;
+}
+```
+
+#### ✅ Ok
+
+Using items from a private namespace within the same library is allowed.
```
srnagar, line 21: This description doesn't seem to match the sample.

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+
+Finally, we recommend that every warning or error you throw in your language emitter has a corresponding warning in TCGC. This is part of our shift-left policy, so tsp authors can catch these potential pitfalls earlier in the process.
+
+### Links
+
+[generic-linter]: https://typespec.io/docs/next/extending-typespec/linters "Generic Linter Docs"
+[tcgc-linter]: https://github.com/typespec-azure/packages/typespec-client-generator-core/src/linter.ts "Linter TS File"
+[rulesets]: https://github.com/typespec-azure/packages/typespec-azure-rulesets "Rulesets package"
+[data-plane-ruleset]: https://github.com/typespec-azure/packages/typespec-azure-rulesets/src/rulesets/data-plane.ts "Data Plane Ruleset"
+[resource-manager-ruleset]: https://github.com/typespec-azure/packages/typespec-azure-rulesets/src/rulesets/resource-manager.ts "Resource Manager Ruleset"
```
srnagar, line 36: These links don't resolve. They are missing `/Azure/` after `https://github.com` and also `blob/main/` after `typespec-azure/`

docs/libraries/typespec-client-generator-core/rules/CONTRIBUTING.md
srnagar, line 15: I don't see step 3 done in this PR. Will this be added in a follow-up PR?

docs/libraries/typespec-client-generator-core/reference/linter.md
timotheeguerin: oh I don't think we have a strong opinion on that name, we just have a placeholder `@typespec/best-practices` package that would contain misc linter to write good typespec.

I would say name this however makes the most sense. But I would note that I don't think this really should spend too much time on this as for all of azure they MUST use either the `data-plane` or `resource-manager` ruleset in teh spec repo. And I assume all those rules for all those language will be enabled.

I expect those more targeted ruleset might be more useful when we have the non azure version of tcgc(e.g. `@typespec/client`) where external customer might decide which language they care about.

docs/libraries/typespec-client-generator-core/reference/linter.md
timotheeguerin: all rules are in a sense optional its up to the user to enable them(directly or via a ruleset). For azure specs we require that the `data-plane` or `resource-manager` ruleset is used.

packages/typespec-client-generator-core/test/rules/require-client-suffix.test.ts
```diff
@@ -0,0 +1,81 @@
+import {
+  BasicTestRunner,
+  createLinterRuleTester,
+  LinterRuleTester,
+} from "@typespec/compiler/testing";
+import { beforeEach, describe, it } from "vitest";
+import { requireClientSuffixRule } from "../../src/rules/require-client-suffix.rule.js";
+import { createSdkTestRunner } from "../test-host.js";
+
+describe("typespec-client-generator-core: require-client-suffix", () => {
```
timotheeguerin, line 10: nit: no need to add those top level describe with vitest, they just add extra nesting in the UI

packages/typespec-client-generator-core/src/lib.ts
timotheeguerin: I don't see what it is used, is it still there? the file you link doesn't have reference to that code

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -699,6 +692,186 @@ export function shouldGenerateConvenient(context: TCGCContext, entity: Operation
   return value ?? Boolean(context.generateConvenienceMethods);
 }
 
+<<<<<<< HEAD
```
timotheeguerin, line 695: merge conflict?

packages/typespec-client-generator-core/src/lib.ts
timotheeguerin: I don't see it used in main anymore, could it have been removed

packages/typespec-client-generator-core/test/rules/property-name-conflict.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+it("emit warning if property name conflicts with model name", async () => {
+  await tester.expect(`model Foo {foo: string}`).toEmitDiagnostics({
+    code: "@azure-tools/typespec-azure-core/property-name-conflict",
+    message: `Property 'foo' having the same name as its enclosing model will cause problems with C# code generation. Consider renaming the property directly or using the @clientName("newName", "csharp") decorator to rename the property for C#.`,
+  });
+});
+
+// TODO: reenable when rule is moved to tcgc and can resolve if clientName was set
+it.skip(`is valid if conflict resolved through @clientName("newName", "csharp")`, async () => {
```
timotheeguerin, line 30: i think those todo can be actioned on now?

timotheeguerin: should you be removing that rule from azure core too now?

docs/libraries/typespec-client-generator-core/rules/property-name-conflict.md
timotheeguerin, line 1: those needs to be moved to the correct place

packages/typespec-azure-rulesets/src/rulesets/data-plane.ts
```diff
@@ -41,6 +41,8 @@ export default {
     "@azure-tools/typespec-azure-core/no-private-usage": true,
     "@azure-tools/typespec-azure-core/friendly-name": true,
     "@azure-tools/typespec-azure-core/no-query-explode": true,
+    "@azure-tools/typespec-client-generator-core/require-client-suffix": true,
+    "@azure-tools/typespec-client-generator-core/property-name-conflict": true,
```
timotheeguerin, line 45: yeah also this is actually included in this PR, so either disable for now or need to fix the spec repos.

Also can you move into its own section(just comment above)

PR #1223, opened 2024-07-23
Report error diagnostic when trying to flattening a model with polymorphsim

### LINKED ISSUE #959: [TCGC] propertyFlatten not apply to property of polymorphic type

Context: at present, `@flattenProperty` had to be applied to template property. e.g. https://github.com/Azure/azure-rest-api-specs/blob/1418642cf8a57954c781c26c759fd4aef39a1287/specification/vmware/Microsoft.AVS/client.tsp#L19

For a property of polymorphic type, it is not possible to flatten the model, as polymorphic is runtime behavior.
https://github.com/Azure/azure-rest-api-specs/blob/1418642cf8a57954c781c26c759fd4aef39a1287/specification/vmware/Microsoft.AVS/models.tsp#L1715-L1744

Therefore, either the decorator, or the TCGC, may want to not set `SdkBodyModelPropertyType.flatten=true`.

Alternatively, this can be done in language emitter part (Java does this), as e.g. one would still encounter it from Swagger. However, the flatten may already be done by m4.
https://github.com/Azure/azure-rest-api-specs/blob/32f566eb38af9593cfc1ed183362471c50d67fb2/specification/vmware/resource-manager/Microsoft.AVS/stable/2023-09-01/vmware.json#L6514-L6516

### Checklist

archerzz:
resolve #959


.chronus/changes/tcgc-forbid-polymorphism-flatten-2024-6-23-22-12-38.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+throw error when trying to flattening a model with polymorphsim
```
timotheeguerin, line 7:
```suggestion
Report error diagnostic when trying to flattening a model with polymorphism
```

do not call reporting diagnostic throw as this is something you should absolutely not do in TypeSpec libraries.

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2545,6 +2545,34 @@ describe("typespec-client-generator-core: decorators", () => {
         code: "decorator-wrong-target",
       });
     });
+
+    it("throws error when used on a polymorphism type", async () => {
+      const diagnostics = await runner.diagnose(`
+        @service({})
```
timotheeguerin, line 2551:
```suggestion
        @service
```

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2545,6 +2545,34 @@ describe("typespec-client-generator-core: decorators", () => {
...
+
+          @test
+          @discriminator("kind")
+          model Model2{
+            kind: string;
+          }
+
+          @test
+          @route("/func1")
+          op func1(@body body: Model1): void;
```
timotheeguerin, line 2568: is the route relevant to this test, can it be removed

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -166,6 +166,12 @@ export const $lib = createTypeSpecLibrary({
         nonDecorator: paramMessage`Client name: "${"name"}" is defined somewhere causing nameing conflicts in language scope: "${"scope"}"`,
       },
     },
+    "flatten-polymorphism": {
+      severity: "error",
+      messages: {
+        default: `Flattening polymorphic types is not supported.`,
```
weidongxu-microsoft, line 172: nit, "Cannot flatten property of polymorphic type"?

weidongxu-microsoft (approved):

archerzz:
I changed the merge target branch from `main` to `release/july-2024`, per suggestion.

iscai-msft:
@timotheeguerin should this be an error, or should it be a linter warning? It is a bad thing to generate, but I'm not sure we should completely blow up generation in this case

timotheeguerin:
I think it's fine to be an error if it's run in the validator as it would count as a validator of the for @flatten decorator. If run in the preprocessor then no. And people will not see it in the ide too

timotheeguerin (approved):

Issue #1229, opened 2024-07-24
Add a `@useSystemTextJsonConverter` decorator for csharp only


live1206:
### Clear and concise description of the problem

In https://github.com/Azure/autorest.csharp/issues/3469, we need to flag if a model needs specific handling of JSON converter.
So that, we would like to add a csharp only decorator `@hasJSONConverter` in TCGC, but with in a different namespace `Azure.ClientGenerator.Core.CSharp`

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1230, opened 2024-07-24
Add `@useSystemTextJsonConverter` decorator for csharp

live1206:
https://github.com/Azure/typespec-azure/issues/1229


packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -419,3 +419,19 @@ extern dec access(
  */
 #deprecated "@flattenProperty decorator is not recommended to use."
 extern dec flattenProperty(target: ModelProperty, scope?: valueof string);
+
+namespace Azure.ClientGenerator.Core.CSharp {
```
timotheeguerin, line 423: This should just be CSharp here otherwise you redefine the entire structure again

timotheeguerin: I.e the current namespace here is 
Azure.ClientGenerator.Core.Azure.ClientGenerator.Core.CSharp

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -419,3 +419,19 @@ extern dec access(
...
+   *
+   * @example
+   * ```typespec
+   * @hasJSONConverter
+   * model MyModel {
+   *   prop: string;
+   * }
+   * ```
+   */
+  extern dec hasJSONConverter(target: Model, scope?: "csharp");
```
timotheeguerin, line 436: I think `@hasJsonConverter` would be more correct with the naming standards we follow

packages/typespec-client-generator-core/test/types/general-decorators-list.test.ts
```diff
@@ -327,4 +327,29 @@ describe("typespec-client-generator-core: general decorators list", () => {
       ]);
     });
   });
+
+  describe("csharp only decorator", () => {
```
timotheeguerin, line 331: Is it on purpose that there is no actual accessor at all?

packages/typespec-client-generator-core/lib/decorators.tsp
timotheeguerin: i have actually no idea how it is working right now

live1206: I think we should have the complete namespace in the TypeSpec.
The existing definition of decorators in Azure.Core:
https://github.com/Azure/typespec-azure/blob/b8c96e8ce51b517421c424348b1d9d55d4465961/packages/typespec-azure-core/lib/decorators.tsp#L3
and 
https://github.com/Azure/typespec-azure/blob/b8c96e8ce51b517421c424348b1d9d55d4465961/packages/typespec-azure-core/lib/decorators.tsp#L139

And when we set the namespace for the implementation, it's like you suggested we should not only set the sub-namespace, like 
https://github.com/Azure/typespec-azure/blob/b8c96e8ce51b517421c424348b1d9d55d4465961/packages/typespec-azure-core/src/decorators.ts#L1524
I did the same thing in this PR.

timotheeguerin: The example you link use block namespace not file namespace. What you have is a bug see this pr https://github.com/microsoft/typespec/pull/4050

I recommend you move it to another file if you feel like putting the whole namespace.

packages/typespec-client-generator-core/lib/decorators.tsp
live1206: OK, renamed.

packages/typespec-client-generator-core/test/types/general-decorators-list.test.ts
live1206: Not quite follow here, what do you mean by "no actual accessor"?

timotheeguerin: There is no api to get the data apart from the generic decorator collector

live1206: Right. the only usage  I can see is to flow into generic decorator list as a flag for the generator.

packages/typespec-client-generator-core/lib/decorators.tsp
live1206: I see your point now.
I would prefer to keep it nested, and update it after your fix merged.

live1206: updated

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1056,3 +1057,14 @@ export function getClientNameOverride(
 ): string | undefined {
   return getScopedDecoratorData(context, clientNameKey, entity, languageScope);
 }
+
+const hasJsonConverterKey = createStateSymbol("hasJsonConverterKey");
+export const $hasJsonConverter: DecoratorFunction = (
+  context: DecoratorContext,
+  entity: Model,
+  scope?: LanguageScopes
```
timotheeguerin, line 1065: Does it make sense to have scopes here?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1056,3 +1057,14 @@ export function getClientNameOverride(
...
+) => {
+  setScopedDecoratorData(context, $hasJsonConverter, hasJsonConverterKey, entity, true, scope);
```
timotheeguerin, line 1067: Do you even need to set the data then if you are never gonna read it

packages/typespec-client-generator-core/src/decorators.ts
live1206: not needed

packages/typespec-client-generator-core/src/decorators.ts
live1206: you are right, it can just be empty

timotheeguerin (approved):

live1206:
Given the timeline: https://github.com/Azure/autorest.csharp/issues/3469#issuecomment-2243538254, we would like to get this out in the next release.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1056,3 +1057,10 @@ export function getClientNameOverride(
 ): string | undefined {
   return getScopedDecoratorData(context, clientNameKey, entity, languageScope);
 }
+
+export const $hasJsonConverter: DecoratorFunction = (
```
m-nash, line 1061: should take language scope and default the value to csharp.  this way in the off chance some other language wants to use it they can by simply changing the scope in tsp.

packages/typespec-client-generator-core/test/types/general-decorators-list.test.ts
m-nash, line 331: add test cases to verify that if you don't pass in scope param it doesn't get applied to java but does get applied to csharp.  Also add test case that verifies the opposite that if you override scope to java it applies to java but not csharp.

packages/typespec-client-generator-core/src/decorators.ts
timotheeguerin: Decorator is in the csharp namespace so this feels confusing to offer as an option and would be even more confusing if passing an other emitter was becoming a thing

live1206: Based on the usage of this decorator, it just flows to the generic decorator list, the scope parameter here actually can't control if this shows up in the generic decorator list.
If we take a look at the implementation of generic decorators [here](https://github.com/Azure/typespec-azure/blob/45eab1e41e71cf4156a6675c97192b637859a649/packages/typespec-client-generator-core/src/internal-utils.ts#L275C17-L308), it takes `decorators` from the type and there is no way to filter decorators based on the scope since scope is a possible parameter of the decorator, and it is stored in the statemap [here ](https://github.com/Azure/typespec-azure/blob/45eab1e41e71cf4156a6675c97192b637859a649/packages/typespec-client-generator-core/src/decorators.ts#L123-L124).

So, if a user defines a decorator, it will flow into the generic decorator list no matter if scope is defined.
The specific language emitter will decide if they will ignore this decorator or not. 
Therefore, I don't think we need scope as parameter for this case.

iscai-msft (approved):

packages/typespec-client-generator-core/src/decorators.ts
live1206: > This will be problematic. We definitely should not put decorators in the decorator list that were not scoped for the current emitter. This should get fixed for sure, if this is the way it works already can you open a follow up issue and link it here?

@m-nash https://github.com/Azure/typespec-azure/issues/1295 for a follow-up fix.

live1206: > If we want to rethink this design we need to make sure we have a way to start to use the decorator for another language in the future? Can we safely move a decorator to a new namespace? Do we leave it in the current namespace for back compat but simply add the option scope param at that point? Do we think that this out weighs the simplicity of having all decorators in tcgc be consistent?

@m-nash 
Can we put this decorator still in the tcgc root namespace and add a default scope of `csharp` for it?
So that, when the scope of generic decorator list is fixed, this will apply and the other languages will be able to reuse it if needed.

m-nash: I think it would be confusing to have to memorize which decorators in root have which default scopes.  The idea here is that if its in the csharp sub namespace it has default csharp scope.  This is simple to remember.

live1206: Try to summarize the options we have:
- define decorator in sub-namespace with a scope parameter
   Con: As @timotheeguerin pointed out it is confusing that we allow other languages could use this decorator.
- define decorator in sub-namespace, if other languages want to use this decorator, we add it to another sub-namespace. I just realized this is not possible, because we can't set multiple namespace for the decorator in the implementation.
-  define decorator in root namespace with scope default to `csharp` 
   Con: As @m-nash pointed out it is confusing for user to memorize which decorators are default to which scope.
- define decorator in sub-namespace, if other languages want to reuse it we put it to the root namespace with scope parameter. 
  Con: This would be a breaking change for the decorator. 
  Pro: We have a clear boundary for the decorators, we can just filter out the generic decorator lists based on the sub-namespace for emitters.

live1206: After discussion, we decide to have it in the root namespace and scope default to all.

.chronus/changes/json-converter-2024-6-24-16-48-41.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Add `@hasJsonConverter` for csharp only to indicate if JSON converter is needed
```
m-nash, line 7: lets rename to @useSystemTextJsonConverter to make it more explicit what this is used for.

m-nash (approved):

packages/typespec-client-generator-core/test/types/general-decorators-list.test.ts
live1206: Will be covered in https://github.com/Azure/typespec-azure/issues/1295

.chronus/changes/json-converter-2024-6-24-16-48-41.md
live1206: renamed

timotheeguerin: I'll say it feels quite weird having this at the root of Tcgc namespace with this name that is very csharp specific

live1206: I think this is indeed csharp specific only decorator.
@timotheeguerin Are you implying we should keep it within csharp sub-namespace if we ensure it will not be used for other languages?

timotheeguerin: Yeah just feels like this is not super compatible with the new approach of using a shared name and scope

live1206:
merge to august release branch instead in https://github.com/Azure/typespec-azure/pull/1336

.chronus/changes/json-converter-2024-6-24-16-48-41.md
iscai-msft: We have decided that language-specific decorators will go in the main namespace because of grow-up, but they will have to specify the scope, and we have an [issue](https://github.com/Azure/typespec-azure/issues/1334) to have linter warnings if a decorator is applied to emitters that won't do anything with it.

My question with this PR is more: I thought we had decided that there was a use case for Java here too, and that there was going to be a redesign of this decorator (most likely just a rename), that would make it compatible with Java's scenario as well

PR #1237, opened 2024-07-25
[tcgc] release hotfix

tadelesh:
This TCGC hotfix contains two breaking changes:
1. `createSdkContext` changed to async function
- Emitter should use await to get the result
- The impact is limited to API only, no behavior change.
2. Built-in types in TCGC has some refactor which removed some kinds of built-in types
- Refer [this](https://github.com/Azure/typespec-azure/pull/1015) to see how to do migration.
- The impact is limited because if emitter does not care about different kinds of string types, they only need to remove these removed kinds.

This version also has:
1. Support swagger example to be represented in TCGC types, any questions please ping @tadelesh 
2. Support new `@multipartBody` feature from TypeSpec, any questions please ping @msyyc 
3. Fix of `@clientName` on sub-clients, cc: @haolingdong-msft 
4. Fix some features used for XML, cc: @jhendrixMSFT


haolingdong-msft (approved):

lirenhe:
When this PR merged, the regen pipeline of Nightly builds will be impacted by 1-2 days waiting all emitters to apply this change.
cc @timotheeguerin , @lmazuel

timotheeguerin:
Please revert the changes to other packages

timotheeguerin (commented):
This is quite bad that you have to do breaking changes in a patch release and this is affecting the rest of typespec packages. Can this not be done in main

qiaozha (approved):

tadelesh:
> This is quite bad that you have to do breaking changes in a patch release and this is affecting the rest of typespec packages. Can this not be done in main

There are lots of feedback from emitter to not to combine typespec breaking with tcgc breaking which makes emitter's upgrade really hard. So, we decided to do this in release branch.
After the release, I'll do the backmerge to main asap. Is there anything else I do not know that will cause problems?

timotheeguerin (changes requested):
Can you check the effect on the spec repo for those changes first as this is bypassing the typespec-next tests. I also I am not at all a fan of the change that was done to autorest emitter as we did a complete opposite change in main. Why is the `createTcgcContext` async, how much slower is the compilation of autorest right now due to that.

I think you need to start being a little more careful about breaking change when we have a much higher bar for TypeSpec and the autorest emitter is affected by tcgc, you should maybe then prioritize splitting the preprocessor(which you can update whenever you want) from the decorators which should follow the typespec release cycle.

And why is this not done as before just in the main branch but releasing earlier?

packages/typespec-autorest/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log - @azure-tools/typespec-autorest
 
+## 0.44.1
+
+### Features
+
+- [#1076](https://github.com/Azure/typespec-azure/pull/1076) create tcgc context change to async
```
timotheeguerin, line 7:
```suggestion
- [#1076](https://github.com/Azure/typespec-azure/pull/1076) Respond to TCGC breaking changes
```

packages/typespec-client-generator-core/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log - @azure-tools/typespec-client-generator-core
 
+## 0.44.2
+
+### Bug Fixes
+
+- [#1231](https://github.com/Azure/typespec-azure/pull/1231) fix the duplicate usageflags values for json and xml
+- [#1203](https://github.com/Azure/typespec-azure/pull/1203) have @clientName work for operation groups as well
```
timotheeguerin, line 8:
```suggestion
- [#1203](https://github.com/Azure/typespec-azure/pull/1203) have `@clientName` work for operation groups as well
```

packages/typespec-client-generator-core/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+- [#1222](https://github.com/Azure/typespec-azure/pull/1222) Fix `@clientName` conflict validation that operation defined within an interface
```
timotheeguerin, line 9:
```suggestion
- [#1222](https://github.com/Azure/typespec-azure/pull/1222) Validate `@clientName` conflict for operations inside interface
```

.chronus/changes/add-scalar-type-2024-5-14-13-23-7.md
timotheeguerin, line 1: a new Pr was also merged since then so this should be updated to include the new changelog

tadelesh:
> Can you check the effect on the spec repo for those changes first as this is bypassing the typespec-next tests. I also I am not at all a fan of the change that was done to autorest emitter as we did a complete opposite change in main. Why is the `createTcgcContext` async, how much slower is the compilation of autorest right now due to that.
> 
> I think you need to start being a little more careful about breaking change when we have a much higher bar for TypeSpec and the autorest emitter is affected by tcgc, you should maybe then prioritize splitting the preprocessor(which you can update whenever you want) from the decorators which should follow the typespec release cycle.
> 
> And why is this not done as before just in the main branch but releasing earlier?

I also prefer the change to sperate create tcgc context and create sdk context in main. Do you think I could check-pick that from main to release? We do not want this breaking to be in the same release cycle of TypeSpec.

The reason why we change to use release is we want to separate the thing that we want to release along with typesspec release cycle and the thing we want to have on-demand release. And the perfect solution is what you said, split the decorator thing and sdk thing. @iscai-msft we may need to prioritize it.

timotheeguerin:
hhm ok then yeah please cherry pick teh change izzy made first then and update the changelogs to reflect that.
And do test this is not breaking azure rest api spec main branch(you should as well do a Pr there to bump the version after the release)

tadelesh:
> hhm ok then yeah please cherry pick teh change izzy made first then and update the changelogs to reflect that. And do test this is not breaking azure rest api spec main branch(you should as well do a Pr there to bump the version after the release)

what is the best way to validate this release does not have breaking for spec main? i tried the artifact from the ci of this pr locally and compile with one arm lib, all works fine. i don't know if this could work in pr: https://github.com/Azure/azure-rest-api-specs/pull/29965.

timotheeguerin:
You use the artifact from the pr that are changed in the package.json there and then run pwsh ./Eng/scripts/TypespecValidation.ps1 -CheckAll

Or you make a pr with your changes to run the ci(might also need to update the ci script) there is a section in the contributing.md

tadelesh:
> You use the artifact from the pr that are changed in the package.json there and then run pwsh ./Eng/scripts/TypespecValidation.ps1 -CheckAll
> 
> Or you make a pr with your changes to run the ci(might also need to update the ci script) there is a section in the contributing.md

thanks for the guide. the validation passed: https://github.com/Azure/azure-rest-api-specs/pull/29965/checks?check_run_id=27918314359. is there any blocker for the release? i'll do the back-merge and update version for spec repo after the release.

timotheeguerin (approved):
Thanks @tadelesh, I would recommend we look into splitting this package into: 
- `@azure-tools/typespec-client-generator-core`
- `@azure-tools/typespec-client-generator-preprocessor`
- 
To simplify those processes and issues like that in the future

PR #1240, opened 2024-07-25
Backmerge tcgc(0.44.2) and autorest (0.44.1) hotfix

tadelesh:


pnpm-lock.yaml
timotheeguerin, line 1: I don't think supposed to be that many change to that file here, can you try reseting to the main version and running `pnpm install`

tadelesh: i've rerun the `pnpm install` but still many changes. i don't know if it comes from add openapi dep in tcgc. i've compared other files in tcgc and all are good compared with the release branch.

tadelesh: @timotheeguerin could you help to merge this?

timotheeguerin: Did you reset first from the version from main

timotheeguerin: It's something we have to be careful as it likely revert the recent upgrade of dependencies

tadelesh: resolved.

timotheeguerin (approved):

PR #1244, opened 2024-07-26
cut off the base type calculation when this type is a std type

ArcturusZhang:
In previous version, the base type on `SdkBuiltInType` will give you every details on those std type, you could even see:
```
int32 -> int64 -> integer -> numeric
```

This is completely redundant.
This PR cuts off those redundant base type calculation once we know that the current type is a std type.
Also adds a few new test cases to test those very basic types.

### DIFF
--- a/packages/typespec-client-generator-core/test/types/built-in-types.test.ts
+++ b/packages/typespec-client-generator-core/test/types/built-in-types.test.ts
@@ -3,6 +3,7 @@ import { AzureCoreTestLibrary } from "@azure-tools/typespec-azure-core/testing";
 import { expectDiagnostics } from "@typespec/compiler/testing";
 import { ok, strictEqual } from "assert";
 import { beforeEach, describe, it } from "vitest";
+import { SdkBuiltInType } from "../../src/interfaces.js";
 import { getAllModels } from "../../src/types.js";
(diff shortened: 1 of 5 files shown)


archerzz (approved):

packages/typespec-client-generator-core/test/types/built-in-types.test.ts
```diff
@@ -124,6 +212,7 @@ describe("typespec-client-generator-core: built-in types", () => {
     strictEqual(sdkType.valueType.crossLanguageDefinitionId, "TestService.Base64UrlBytes");
     strictEqual(sdkType.valueType.baseType?.kind, "bytes");
     strictEqual(sdkType.valueType.baseType.encode, "base64");
+    strictEqual(sdkType.valueType.baseType.baseType, undefined);
```
lirenhe, line 215: Can we build one more case that 'baseType.baseType' is not undefined?

tadelesh (approved):

packages/typespec-client-generator-core/test/types/built-in-types.test.ts
ArcturusZhang: we have such cases in other test cases because the base type is not introduced in this PR.

PR #1247, opened 2024-07-26
Uptake uri templates

timotheeguerin:
Core PR https://github.com/microsoft/typespec/pull/3932


chrisradek (approved):

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -596,7 +596,11 @@ function getCollectionFormat(
         ? getHeaderFieldOptions(program, type)
         : undefined
   )?.format;
-  if (tspCollectionFormat === "form" || tspCollectionFormat === "simple") {
+  if (
+    tspCollectionFormat === "form" ||
+    tspCollectionFormat === "simple" ||
+    tspCollectionFormat === "csv"
```
msyyc, line 602: Why add filter logic for `csv` here?

timotheeguerin: Because those are all the default and I tried to make it so there would not be a diff in the code model

timotheeguerin: but yeah I think this was added first in order to not have `csv` show up all the time but changed it in the http library to only populate format: when the type was an array to more closely replicate when you would have used format.

PR #1248, opened 2024-07-26
Refactor parameter handling in swagger emitter

timotheeguerin:
When making the URI template PR it was a bit hard to deal with the mess and better to split the refactoring

Also fix issue where `multi` wasn't even allowed for header part https://github.com/Azure/typespec-azure-pr/issues/3825


packages/typespec-autorest/src/openapi.ts
```diff
@@ -1240,66 +1244,166 @@ export async function getOpenAPIForService(
     }
   }
 
-  function getOpenAPI2Parameter<T extends OpenAPI2ParameterType>(
+  function getOpenAPI2PArameterBase(
```
chrisradek, line 1247: Minor:
```suggestion
  function getOpenAPI2ParameterBase(
```

chrisradek (approved):

Issue #1060, opened 2024-06-24
feat(tcgc): provide origin of `SdkClientType`


archerzz:
### Clear and concise description of the problem

An `SdkClientType` can derive from either a namespce or an operation group. Some languages (like csharp) have different processing logic for those two kinds. We need TCGC to provide that metadata somewhere from `SdkClientType`.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

iscai-msft:
Can you explain this more clearly @archerzz? Are you tryign to determine whether a client is a top-level client? I'm not sure what you mean if a client is derived from an "operation group"

archerzz:
> Can you explain this more clearly @archerzz? Are you tryign to determine whether a client is a top-level client? I'm not sure what you mean if a client is derived from an "operation group"

@iscai-msft sort of. Here is an example. In `autorest.csharp`, we have a test case which defines a sub namespace: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/UnbrandedProjects/Platform-OpenAI-TypeSpec/audio/operations.tsp#L11
```typescript
namespace OpenAI;
@route("/audio")
namespace Audio {
```

And in typespec, it's parsed as an `SdkOperatoinGroup` (I'm not clear of the magic here, but it happens).
![image](https://github.com/Azure/typespec-azure/assets/4516/7d6b600a-5872-4abd-ab84-9318dd80fdec)

So, I think we should have similar info in TCGC.

haolingdong-msft:
For Java, we will also need the `groupPath` info to build the operation group's name.

tadelesh:
final solution: provide parent info in `SdkClientType` and emitter could use the hierarchal client's name to know the path

PR #1253, opened 2024-07-29
[tcgc] client related features

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1060
resolve: https://github.com/Azure/typespec-azure/issues/963


tadelesh:
@haolingdong-msft @archerzz @chunyu3 please review and check if it is enough for your language.
also, cc: @iscai-msft

haolingdong-msft:
Thanks for the fix. Having `parent` should be enough for Java's current need. Our need for now is just to build the sub client's name

iscai-msft (approved):

PR #1256, opened 2024-07-29
Backmerge/release/july 2024 2024 07 29

iscai-msft:


pnpm-lock.yaml
timotheeguerin, line 1: please fix this file, shoul be 10-20lines changed
reset it to main then `pnpm install`

timotheeguerin (approved):

Issue #1268, opened 2024-07-31
Add `resourceType` scalar in `typespec-azure-core` library


ArcturusZhang:
We also need a `resourceType` scalar to make a specific scalar to be the arm resource type.
DotNet SDK will generate `resourceType` into `Azure.Core.ResourceType`.
Other languages could simply generate it as `string`.

The definition could be
```
scalar resourceType extends string;
```

And the `TrackedResource`, `ProxyResource` template needs to be updated accordingly - their type property should be `resourceType`.

markcowl:
Also, this should change the type of the allowedResourceTypes array in the `armResourceIdentifier` template

markcowl:
est: 5

PR #1272, opened 2024-07-31
add `armResourceType` scalar in `typespec-azure-core`

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/1268


tadelesh:
@allenjzhang could you help to review?

timotheeguerin (changes requested):
We should go through some design discussion to approve this new type

ArcturusZhang:
> We should go through some design discussion to approve this new type

Yeah I assume we should go through design discussions before we could merge this. What is the process of that?

timotheeguerin:
We can talk about it tomorrow in dpg meeting(or later one if we can't) don't think this needs to be brought up to the Wednesday one

ArcturusZhang:
@timotheeguerin any update on this?

.chronus/changes/add-resource-type-2024-6-31-14-6-51.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-core"
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add resourceType scalar, and changed some properties to be resourceType instead of string
```
timotheeguerin, line 8:
```suggestion
Add `resourceType` scalar, and changed some properties to be resourceType instead of string
```

packages/typespec-azure-core/lib/models.tsp
```diff
@@ -349,6 +349,17 @@ scalar eTag extends string;
...
+/**
+ * Represents an Azure Resource Type.
+ *
+ * @example
+ *
+ * ```
+ * Microsoft.Network/virtualNetworks/subnets
+ * ```
+ */
+scalar resourceType extends string;
```
timotheeguerin, line 361: Sorry, I was OOF the last 2 weeks, has there been agreement we needed this and this was the name we wanted?

ArcturusZhang: Oh - I do not know. Do we need to bring up this topic in one of our typespec sync up meeting or your internal typespec sync up meeting?
In those meetings that I have participated in, this has not been a topic there.
Or maybe we could put this as an agenda in our next typespec sync up meeting on your wednesday night.

markcowl: I think we probably need to differentiate this as we did with armResourceIdentifier - armResourceType.  

There is definitely use for this in ARM, and since the type is needed to properly specify the allowedResourceType in armResourceIdentifier, it seems appropriate to put this here.

ArcturusZhang: Yeah I agree the name should be `armResourceType` which makes more sense.
The second part is that you suggest we add a generic argument for allowed resource type? like this?
```
scalar armResourceType<AllowedResourceTypes extends ArmResourceIdentifierAllowedResource[] = never> extends string {}
```
I think this looks weird because this type is modeling the resource type of myself, if there is a limited list of resource types that this field is supporting, should we define a extensible enum based on resource type like this?
```
union SupportedResourceType extends armResourceType {
	"Microsoft.Compute/virtualMachines"
}
```
Well but I could see values for this because we could update the definition of `Resource` and `TrackedResource` to let them accept another argument of resource type like this:
```
model TrackedResource<Properties extends Model, ResourceType extends ArmResourceIdentifierAllowedResource> {
id: armResourceIdentifier;
type: armResourceType<ResourceType>;

properties: Properties;
}
```
so that we finally have explicitly defined resource type on resources, currently arm resources could have resource types but they are implicitly - we can infer them from the uri paths we built for resource related operations.

timotheeguerin: yeah I think using a union here instead to be more precise makes more sense instead of the `armREsourceIdentifier` which does that for a subset of the value.

and `+1` on the name being `armResourceType`

packages/samples/common-types/gen.ts
```diff
@@ -86,6 +86,7 @@ function cleanupDocument(original: OpenAPI2Document): OpenAPI2Document {
 
   replaceUuidRefs(document, "Azure.Core.uuid");
   replaceUuidRefs(document, "Azure.Core.azureLocation");
+  replaceUuidRefs(document, "Azure.Core.armResourceType");
```
markcowl, line 89: What does this actually do? It seems like this should have some impact on a generated common-types swagger, but I don't see it in this PR

markcowl (commented):
@AlitzelMendez Any comments on the common-types/gen.ts ?

packages/samples/common-types/gen.ts
ArcturusZhang: this line is added to keep this file is unchanged.
I followed the same pattern for azurelocation and uuid above.
Without this, the generated swagger will look like this:
"property": {
"$ref": "#/definitions/armResourceType"
},
// in the definitions
"armResourceType": {
"type": "string"
}

markcowl: so, this method should just be named 'omitScalarSchema` :-)

markcowl (approved):

packages/samples/common-types/gen.ts
ArcturusZhang: LOL indeed - so this method was first used on uuid and later we found we should apply that to more types but we did not rename the method.

PR #1279, opened 2024-07-31
[Canonical Emitter] Fix included versions value in canonical emitter

dinwa-ms:
Fix the included versions value in canonical emitter. 

It returned the version name, should return version value instead. Updated the test case.

### DIFF
--- a/packages/typespec-autorest-canonical/src/emitter.ts
+++ b/packages/typespec-autorest-canonical/src/emitter.ts
@@ -97,7 +97,7 @@ async function emitAllServices(
     const result = await getOpenAPIForService(context, options);
     const includedVersions = getVersion(program, service.type)
       ?.getVersions()
-      ?.map((item) => item.value);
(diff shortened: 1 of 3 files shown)


packages/typespec-autorest-canonical/src/emitter.ts
```diff
@@ -97,7 +97,7 @@ async function emitAllServices(
     const result = await getOpenAPIForService(context, options);
     const includedVersions = getVersion(program, service.type)
       ?.getVersions()
-      ?.map((item) => item.name);
+      ?.map((item) => item.value);
```
timotheeguerin, line 100: the value could also be undefined need to `?? item.name`

dinwa-ms: update the value

timotheeguerin (approved):

dinwa-ms:
@timotheeguerin Thanks for the approval. Can you help merge the PR when you get a chance?

PR #1281, opened 2024-08-01
[Multipart][TCGC] Support @multipartBody for `bodyParam` of `SdkHttpOperation`

msyyc:
supplement for PR https://github.com/Azure/typespec-azure/pull/1090
- Support @multipartBody for `bodyParam` of `SdkHttpOperation`
- Only property of multipart body shall be added `multipartOptions` and the property of base model shall have no `multipartOptions`

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -31,7 +31,6 @@ import {
 } from "@typespec/compiler";
 import {
   Authentication,
-  HttpOperation,
   HttpOperationPart,
   Visibility,
   getAuthentication,
@@ -1260,7 +1259,7 @@ function updateMultiPartInfo(
   context: TCGCContext,
   type: ModelProperty,
   base: SdkBodyModelPropertyType,
-  operation: Operation,
+  operation: Operation
 ): [void, readonly Diagnostic[]] {
   const httpOperationPart = getHttpOperationPart(context, type, operation);
   const diagnostics = createDiagnosticCollector();
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -1261,7 +1261,6 @@ function updateMultiPartInfo(
   type: ModelProperty,
   base: SdkBodyModelPropertyType,
   operation: Operation,
-  httpOperation: HttpOperation
 ): [void, readonly Diagnostic[]] {
   const httpOperationPart = getHttpOperationPart(context, type, operation);
   const diagnostics = createDiagnosticCollector();
@@ -1287,6 +1286,7 @@ function updateMultiPartInfo(
     }
   } else {
     // common body
+    const httpOperation = getHttpOperationWithCache(context, operation);
     const operationIsMultipart = Boolean(
       httpOperation && httpOperation.parameters.body?.contentTypes.includes("multipart/form-data")
     );
@@ -1344,7 +1344,7 @@ export function getSdkModelPropertyType(
       httpOperation.parameters.body.type === type.model
     ) {
       // only add multipartOptions for property of multipart body
-      diagnostics.pipe(updateMultiPartInfo(context, type, result, operation, httpOperation));
+      diagnostics.pipe(updateMultiPartInfo(context, type, result, operation));
     }
   }
   return diagnostics.wrap(result);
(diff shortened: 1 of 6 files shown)


packages/typespec-client-generator-core/test/types/model-types.test.ts
```diff
@@ -1502,4 +1502,71 @@ describe("typespec-client-generator-core: model types", () => {
...
+        model MultiPartRequest{
+          id?: HttpPart<string>;
+          profileImage: HttpPart<bytes>;
+          address: HttpPart<Address>;
+          picture: HttpPart<File>;
+        }
+        @post
+        op upload(@header contentType: "multipart/form-data", @multipartBody body: MultiPartRequest): void;
+        `);
+    const formDataMethod = runner.context.sdkPackage.clients[0].methods[0];
```
haolingdong-msft, line 1520: The test looks good to me. 
Would like to know the behavior of inheritance case, do we want to add a test case to cover case like below?
```ts
model InheritFile extends File {
  contentType?: string;
  filename: string;
}

op uploadHttpPart(
  @multipartBody body: {
    fileData1: HttpPart<InheritFile>;
  },
): OkResponse;
```

msyyc: this case is to make sure TCGC parse @multipartBody correctly and inheritance shall not influence the result.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1328,7 +1329,15 @@ export function getSdkModelPropertyType(
...
   };
   if (operation) {
-    diagnostics.pipe(updateMultiPartInfo(context, type, result, operation));
+    const httpOperation = getHttpOperationWithCache(context, operation);
+    if (
+      type.model &&
+      httpOperation.parameters.body &&
+      httpOperation.parameters.body.type === type.model
+    ) {
+      // only add multipartOptions for property of multipart body
```
weidongxu-microsoft, line 1338: Not sure how this connected. Is it not `body.bodyKind === "multipart"`?

msyyc: It is another fix 
image
Without the fix, TCGC will loop all the related properties of model including base model under multipart operation.

weidongxu-microsoft: What I mean, is, why no condition of 
```
httpOperation.parameters.body.bodyKind === "multipart"
```

weidongxu-microsoft: Ok, it needs to include the normal `@body`...

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/types.ts
tadelesh: will this work for the implicit/explicit spread without `HttpPart`?
example 1: https://github.com/Azure/typespec-azure/blob/2a5768ad218fad53c205d0317b3d10280f30686f/packages/typespec-client-generator-core/test/package.test.ts#L3095-L3113
example 2: `op test(part1: string, part2: bytes, @header contentType: "multipart/form-data"): void;`

weidongxu-microsoft: Ah, yeah, the spread.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1251,7 +1252,8 @@ function updateMultiPartInfo(
   context: TCGCContext,
   type: ModelProperty,
   base: SdkBodyModelPropertyType,
-  operation: Operation
+  operation: Operation,
+  httpOperation: HttpOperation
```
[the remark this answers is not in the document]
msyyc: `updateMultiPartInfo` is just Internal function and now that we already get `httpOperation`, it is better to pass it to `updateMultiPartInfo` instead of call getHttpOperationCache to get it.

packages/typespec-client-generator-core/src/types.ts
msyyc: Just add some assert to example 1 and it works; there is already existing [test case ](https://github.com/Azure/typespec-azure/blob/2a5768ad218fad53c205d0317b3d10280f30686f/packages/typespec-client-generator-core/test/types/multipart-types.test.ts#L68-L95) for example 2

weidongxu-microsoft: Ok. `body.type === type.model` likely covers `body.property === type` case.

tadelesh (approved):

PR #1303, opened 2024-08-06
[tcgc] add namespace to usage

### LINKED ISSUE #1300: Have `@usage` decorator apply to namespaces

We have `@usage` for individual models and properties: we want to be able to apply this globally through a decorator that takes in a namespace / interface. More granularity down the line wins out

```md
| Decorator State                   | Model with Public Usage | Model with Private Usage | Model without Usage |
|-----------------------------------|-------------------------|--------------------------|---------------------|
| `@usage(Usage.None)`              | Used                    | Used                     | Not Used            |
| `@usage(Usage.Input/Usage.Output)`| Used                    | Used                     | Used                |
```

### Checklist

iscai-msft:
fixes #1300


packages/typespec-client-generator-core/README.md
```diff
@@ -565,7 +565,7 @@ you need to take care of all related models/enums.
 
 ##### Target
 
-`Model | Enum | Union`
+`Model | Enum | Union | Namespace`
```
srnagar, line 568: Could you also please add some documentation on what it means for usage to be applied to `Namespace`? For Model, Enum and Union it's straightforward but for Namespace it might not be obvious what this means. So, the table you had in the other issue might be worth adding here.

srnagar (approved):

joheredi (approved):

timotheeguerin (approved):

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -925,6 +925,13 @@ export function getUsageOverride(
   return getScopedDecoratorData(context, usageKey, entity);
 }
 
+export function getDefaultUsage(context: TCGCContext, namespace?: Namespace): UsageFlags {
+  if (!namespace) {
+    return UsageFlags.None;
+  }
+  return getScopedDecoratorData(context, usageKey, namespace) ?? UsageFlags.None;
+}
+
```
tadelesh, line 934: why not merge this logic to `getUsageOverride`? i don't think it's kind of default usage, but kind of usage override on wide scale.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1704,9 +1705,9 @@ function handleServiceOrphanType(context: TCGCContext, type: Model | Enum | Unio
     const sdkType = diagnostics.pipe(getClientTypeWithDiagnostics(context, type));
     updateUsageOfModel(context, UsageFlags.Input | UsageFlags.Output, sdkType);
   }
-  if (getAccessOverride(context, type) !== undefined) {
-    const sdkType = diagnostics.pipe(getClientTypeWithDiagnostics(context, type));
-    updateUsageOfModel(context, UsageFlags.None, sdkType);
+  const sdkType = diagnostics.pipe(getClientTypeWithDiagnostics(context, type));
+  if ((sdkType.kind === "model" || sdkType.kind === "enum") && sdkType.usage === UsageFlags.None) {
+    updateUsageOfModel(context, getDefaultUsage(context, type.namespace), sdkType);
```
tadelesh, line 1710:
```suggestion
  const sdkType = diagnostics.pipe(getClientTypeWithDiagnostics(context, type));
  updateUsageOfModel(context, UsageFlags.None, sdkType);
```

all logic could be covered by `getUsageOverride`.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -925,6 +925,13 @@ export function getUsageOverride(
...
 export function getUsage(context: TCGCContext, entity: Model | Enum): UsageFlags {
```
tadelesh, line 935: found another logic that we forgot. for union, since it could be union as enum, then it could also have usage.

packages/typespec-client-generator-core/src/types.ts
tadelesh: also, if not put the ns override into `getUsageOverride`, then only orphan model will take `@usage` on ns.

tadelesh (changes requested):

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: good point, let me fix this

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2464,6 +2464,80 @@ describe("typespec-client-generator-core: decorators", () => {
...
+          @route("/one")
+          op one(@body body: InputModel): OutputModel;
+
+          @route("/two")
+          op two(@body body: RoundtripModel): RoundtripModel;
+        }
+      `)) as { OrphanModel: Model; InputModel: Model; OutputModel: Model; RoundtripModel: Model };
+      strictEqual(getUsage(runner.context, OrphanModel), UsageFlags.Input | UsageFlags.Output);
+      // this is set to input and ouput because of the namespace override
+      strictEqual(
```
iscai-msft, line 2503: @tadelesh can you confirm that the usage for `InputModel` and `OutputModel` should be both `Usage.input` and `Usage.output` because of the `@usage` decorator?

tadelesh: yes. i think the `@usage` on ns override their usage.

tadelesh (approved):

PR #1305, opened 2024-08-06
Add namespace as target for @access decorator

### LINKED ISSUE #1299: Have `@access` decorator apply to namespaces

We have `@access` for individual models and properties: we want to be able to apply this globally through a decorator that takes in a namespace / interface. More granularity down the line wins out

```md
| Decorator State            | Model with Public Usage  | Model with Private Usage | Model without Usage |
|----------------------------|--------------------------|--------------------------|---------------------|
| `@access(public)`          | Public                   | Public                   | Public              |
| `@access(private)(default)`| Public                   | Private                  | Private             |
```

### Checklist

joheredi:
Fixes #1299

### DIFF
--- a/packages/typespec-client-generator-core/lib/decorators.tsp
+++ b/packages/typespec-client-generator-core/lib/decorators.tsp
@@ -261,8 +261,8 @@ enum Access {
 /**
  * Set explicit access for operations, models and enums.
  * When setting access for namespaces,
- * the access the access info will be propagated to the model children of the namespace
- * if the model has an access override, the model override takes precedence.
+ * the access info will be propagated to the models defined in the namespace.
+ * If the model has an access override, the model override takes precedence.
  * When setting access for models,
  * the access info wll not be propagated to models' properties, base models or sub models.
  * When setting access for an operation,
--- a/packages/typespec-client-generator-core/lib/decorators.tsp
+++ b/packages/typespec-client-generator-core/lib/decorators.tsp
@@ -260,6 +260,9 @@ enum Access {
 
 /**
  * Set explicit access for operations, models and enums.
+ * When setting access for namespaces,
+ * the access the access info will be propagated to the model children of the namespace
+ * if the model has an access override, the model override takes precedence.
  * When setting access for models,
  * the access info wll not be propagated to models' properties, base models or sub models.
  * When setting access for an operation,
@@ -398,7 +401,7 @@ enum Access {
  * ```
  */
 extern dec access(
-  target: Model | Operation | Enum | Union,
+  target: Model | Operation | Enum | Union | Namespace,
   value: EnumMember,
   scope?: valueof string
 );
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -968,7 +968,7 @@ export function getAccess(
 ) {
   const override = getAccessOverride(context, entity);
   if (override) {
-    return override || "public";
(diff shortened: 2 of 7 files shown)


iscai-msft (approved):

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -260,6 +260,9 @@ enum Access {
 
 /**
  * Set explicit access for operations, models and enums.
+ * When setting access for namespaces,
+ * the access the access info will be propagated to the model children of the namespace
+ * if the model has an access override, the model override takes precedence.
```
tadelesh, line 265:
```suggestion
 * the access info will be propagated to the models defined in the namespace.
 * If the model has an access override, the model override takes precedence.
```

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -951,17 +951,23 @@ export const $access: AccessDecorator = (
...
+
+  return accessOverride;
 }
 
 export function getAccess(
   context: TCGCContext,
-  entity: Model | Enum | Operation | Union
-): AccessFlags {
+  entity: Model | Enum | Operation | Union | Namespace
+) {
```
tadelesh, line 968: this func is used for get final calculation result of access, so i don't think we should add namespace

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -951,17 +951,23 @@ export const $access: AccessDecorator = (
...
   const override = getScopedDecoratorData(context, accessKey, entity);
```
tadelesh, line 969:
```suggestion
  const override = getAccessOverride(context, entity);
```

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -951,17 +951,23 @@ export const $access: AccessDecorator = (
...
-  if (override || entity.kind === "Operation") {
+  if (override) {
     return override || "public";
```
tadelesh, line 971:
```suggestion
    return override;
```

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -970,15 +976,20 @@ export function getAccess(
...
-    case "Union":
+    case "Union": {
       const type = getSdkUnion(context, entity);
       if (type.kind === "enum" || type.kind === "model") {
         return type.access;
       }
       return "public";
+    }
+    case "Operation":
+    case "Namespace":
```
tadelesh, line 987:
```suggestion
    case "Operation":
```

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -1690,6 +1690,36 @@ describe("typespec-client-generator-core: decorators", () => {
...
+          @test
+          model Test {
+            prop: string;
+          }
+        `)) as { Test: Operation };
+  
+        const actual = getAccess(runner.context, Test);
+        strictEqual(actual, "internal");
+      })
+    })
```
tadelesh, line 1722: also need test for operation

tadelesh (changes requested):

tadelesh (approved):

packages/typespec-client-generator-core/test/decorators.test.ts
joheredi: Added additional tests

PR #1327, opened 2024-08-08
Fix collectionFormat for "csv"

msyyc:


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -596,11 +596,7 @@ function getCollectionFormat(
         ? getHeaderFieldOptions(program, type)
         : undefined
   )?.format;
-  if (
-    tspCollectionFormat === "form" ||
-    tspCollectionFormat === "simple" ||
-    tspCollectionFormat === "csv"
-  ) {
+  if (tspCollectionFormat === "form" || tspCollectionFormat === "simple") {
```
msyyc, line 599: I think logic for "csv" is added with the careless in https://github.com/Azure/typespec-azure/pull/1247/files#r1708791042. Add @timotheeguerin to confirm

tadelesh: it seems previous logic is wrong for `simple`, which cause this unconsciously false change.

tadelesh: we could revert it first and i'll change the logic when adopting 0.59.

tadelesh (approved):

timotheeguerin (approved):

packages/typespec-client-generator-core/src/http.ts
timotheeguerin: Yeah that was a mistake then, was trying to make the format behave as closely as possible as it was when using the new form.  But did also a change in the http library to only set `format: csv` when teh type of the property is an array later and could have removed that. I think its another issue where the default is not respected as well. And yeah the previous logic was borken as @tadelesh is saying `form` is the equivalent of `multi` and `simple` of `csv` and the default (undefined means simple/csv)

msyyc:
done by https://github.com/Azure/typespec-azure/pull/1330

Issue #1322, opened 2024-08-08
The `filePath` in examples should be relative


ArcturusZhang:
The `filePath` in examples should be relative to the `project-root` or some other kind of root to keep the data/behavior consistent across different machines/agents

PR #1328, opened 2024-08-08
[tcgc] change example file path to relative file path

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1322

cc: @ArcturusZhang 

the example's file path changed to be the relative path compared with `examples-directory` config.


packages/typespec-client-generator-core/src/example.ts
```diff
@@ -197,7 +194,7 @@ function handleHttpOperationExamples(
       kind: "http",
       name: title,
       description: title,
-      filePath: example.file.path,
+      filePath: example.relativePath,
```
timotheeguerin, line 197: where do those examples come from?

tadelesh: from config of `examples-directory`, this `filePath` is a relative path to this config for each of the example.

timotheeguerin: Does this config has the format: absolute-path so the compiler validate that for you. Relative path are a nightmare to understand to what and we agreed a while back to not have any new thing using it

tadelesh: how to define the option defined in the tcgc which is not an emitter?

timotheeguerin: where is it setup then? its not the emitter sending it?
For doc on how to use it https://typespec.io/docs/next/extending-typespec/emitters#absolute-path

timotheeguerin: talked to Izzy and we looked at the code, the confusion comes from teh fact that you hardcode that directly. WHere then it makes a lot more sense to hard code to `{project-root}/examples`. However that directory is configurable by the autorest emitter so it is not necessarly always in this folder FYI.

tadelesh: i don't quite understand what you mean. `examples-directories` will be configured in each language's emitter. and the `filePath` here is to show the relative path of one example to that config path. so, join `examples-directories` and `filePath` will get the full absolute path of the example.

timotheeguerin: this part `examples-directories will be configured in each language's emitter.`. There is 2 things about it
1. Need to make sure you use `format: absolute-path` here so this issue doesn't keep happening
2. This is a lot of duplication, every emitter will need to respecify this property in the tspconfig.yaml
   - not much we can do abotu that right now

tadelesh: for 1, since it is a config in emitter, i'll let all languages to use the right format, anything could be done in tcgc?

timotheeguerin: Yeah no just more of a FYI to make sure everyone does the right thing

weidongxu-microsoft:
What do we get, in a typical MGMT codegen?

In full path, Java emitter current logic is

tsp-location gives
```
"specification/standbypool/StandbyPool.Management"
```

tcgc gives
```
C:/github/azure-sdk-for-java/sdk/standbypool/azure-resourcemanager-standbypool/TempTypeSpecFiles/StandbyPool.Management/examples/2023-12-01-preview/StandbyVirtualMachinePools_Update.json
```

We find the last segment from former (`StandbyPool.Management`) and compare it with latter, resolve to
```
specification/standbypool/StandbyPool.Management/examples/2023-12-01-preview/StandbyVirtualMachinePools_Update.json
```

packages/typespec-client-generator-core/src/example.ts
tadelesh: i found in current [autorest emitter](https://github.com/Azure/typespec-azure/blob/5edf99de54d93a4ad04303ae5dca3d507081d557/packages/typespec-autorest/src/lib.ts#L152), the `examples-directories` not set format to be `absolute-path`. can we change that? if so, i believe we will enforce all pr merged to azure spec repo to set it with absolute path.

tadelesh:
> What do we get, in a typical MGMT codegen?
> 
> In full path, Java emitter current logic is
> 
> tsp-location gives
> 
> ```
> specification/standbypool/StandbyPool.Management
> ```
> 
> tcgc gives
> 
> ```
> C:/github/azure-sdk-for-java/sdk/standbypool/azure-resourcemanager-standbypool/TempTypeSpecFiles/StandbyPool.Management/examples/2023-12-01-preview/StandbyVirtualMachinePools_Update.json
> ```
> 
> We take the last segment from former (`StandbyPool.Management`) and find same segment in latter, resolve to
> 
> ```
> specification/standbypool/StandbyPool.Management/examples/2023-12-01-preview/StandbyVirtualMachinePools_Update.json
> ```

tsp-location:
```
specification/standbypool/StandbyPool.Management
```

examples-directory config:
```
C:/github/azure-sdk-for-java/sdk/standbypool/azure-resourcemanager-standbypool/TempTypeSpecFiles/StandbyPool.Management/examples
```

tcgc:
```
2023-12-01-preview/StandbyVirtualMachinePools_Update.json
```

weidongxu-microsoft:
> tsp-location:
> 
> ```
> specification/standbypool/StandbyPool.Management
> ```
> 
> examples-directory config:
> 
> ```
> C:/github/azure-sdk-for-java/sdk/standbypool/azure-resourcemanager-standbypool/TempTypeSpecFiles/StandbyPool.Management/examples
> ```
> 
> tcgc:
> 
> ```
> 2023-12-01-preview/StandbyVirtualMachinePools_Update.json
> ```

So if I concat `examples-directory` and tcgc result, I would get back the current full path
```
C:/github/azure-sdk-for-java/sdk/standbypool/azure-resourcemanager-standbypool/TempTypeSpecFiles/StandbyPool.Management/examples/2023-12-01-preview/StandbyVirtualMachinePools_Update.json
```

Sure, I am good with this.

ArcturusZhang:
In general, I think TCGC should not make the assumption that it will only apply to `azure-rest-api-specs` or our SDK repos.
It should be generic enough to fit every situation.

tadelesh:
> In general, I think TCGC should not make the assumption that it will only apply to `azure-rest-api-specs` or our SDK repos. It should be generic enough to fit every situation.

current impl is not only for mpg, it is relay on the `examples-directory` config to find example and in each example, the `filePath` is relative to this config.

packages/typespec-client-generator-core/src/example.ts
timotheeguerin: yeah we have issue to do that, just it is a breaking change as this existed before `absolute-path` was present
the goal is to rename it to `examples-dir`

timotheeguerin: which would be nice if you named it that way as well as we have been naming things with `-dir` for directories

tadelesh: got it, let me do the change.

tadelesh:
@ArcturusZhang @weidongxu-microsoft @MaryGao if this change is ok for your language, could you help to approve this pr?

weidongxu-microsoft (approved):

weidongxu-microsoft (approved):

ArcturusZhang (approved):

Issue #1326, opened 2024-08-08
[Bug]: tcgc, unexpected warning on operationGroup within client with scope


weidongxu-microsoft:
### Describe the bug

Lots of warning like
```
warning @azure-tools/typespec-client-generator-core/client-service: Client "FaceList" is not inside a service namespace. Use @client({service: MyServiceNS}
```
on client.tsp with scope on `@client` and `@operationGroup` without scope.

e.g.
```
@client({
  name: "FaceServiceClient",
  service: Face,
}, "csharp,python")
@client({
  name: "administration.FaceServiceClient",
  service: Face,
}, "java")
namespace FaceServiceClient {
  @operationGroup
  @clientName("FaceListClientImpl", "csharp")
  interface FaceList {
    ...
```

See "Repreduction" on the source.

---

Not sure what is the current logic of the diagnostic. I assume it should not warn if I run emitter for "java", as it is scoped in both `@client` and `@operationGroup`.

It is fine to warn on "javascript" as it is not scoped in `@client`.

### Reproduction

Source client.tsp https://github.com/Azure/azure-rest-api-specs/blob/29f19b6d4c30dda233f5f69a28e69b8a47e01a0d/specification/ai/Face/client.tsp#L119-L121

It is in PR https://github.com/Azure/azure-rest-api-specs/pull/29178, at commit https://github.com/Azure/azure-rest-api-specs/pull/29178/commits/29f19b6d4c30dda233f5f69a28e69b8a47e01a0d

Run emitter on this client.tsp

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1338, opened 2024-08-09
[tcgc] consider scope when find service of a client

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1326


weidongxu-microsoft (approved):
What would VS code plugin report (where no emitter in scope)?

iscai-msft (approved):

Issue #1195, opened 2024-07-18
Add a linter rule to discourage using @useRef decorator


MaryGao:
In typespec service team could use `@useRef` to generate new swagger similar to their old swagger(if in that swagger they have some references). 

Here is an example from Azure Fleet, if we try to generate SDK from typespec directly we will get an empty model for `ApiError` the referenced swagger definition would only work for OpenAPI emitter. In that case we should report this out to ask service team to migrate referenced swagger to typespec also. 

Once the service team finish their migration there is no need to refer any swagger definition. So it's better to have a linter rule to warn the usage for `@useRef` out.

[tsp definition](https://github.com/Azure/azure-rest-api-specs/blob/db2f049580ad55549ce9a34635ab58bc5a27c814/specification/azurefleet/AzureFleet.Management/common.tsp#L18-L19)
```
@doc("ApiError for Fleet")
@useRef("../../../../../compute/resource-manager/Microsoft.Compute/common-types/v1/common.json#/definitions/ApiError")
model ApiError {}
```
[generated swagger](https://github.com/Azure/azure-rest-api-specs/blob/db2f049580ad55549ce9a34635ab58bc5a27c814/specification/azurefleet/resource-manager/Microsoft.AzureFleet/preview/2024-05-01-preview/azurefleet.json#L984-L985)
```
"error": {
    "$ref": "../../../../../compute/resource-manager/Microsoft.Compute/common-types/v1/common.json#/definitions/ApiError",
    "description": "Error Information when `operationStatus` is `Failed`",
    "readOnly": true
  }
```

timotheeguerin:
Agreed, I think though this should just be a rule to say do not use anything from autorest library(and openapi)

PR #1342, opened 2024-08-09
Update no-operationId rule to a more generic no-openapi

timotheeguerin:
fix [#1195](https://github.com/Azure/typespec-azure/issues/1195)


markcowl (approved):
This looks good.  But I don't think we can do this until we have a solution for x-ms-identifiers in ARM

timotheeguerin:
> This looks good. But I don't think we can do this until we have a solution for x-ms-identifiers in ARM

I actually have an exception in place in the rule just for that right now so I think we should merge that and we can remove the exception when we find that solution

markcowl:
**Describe the breaking change**
Expand existing linter rule that discouraged use of operationId decorator to any openapi decorators (@extension, @useRef, etc.)
Those decorators do not affect other TypeSpec emitters so they are just a sign of someone trying to match existing swagger during conversion.

**Back-compat design to avoid immediate breaking**
All existing specs are corrected

**Impact**
This change affecting:

- [x] ARM
- [x] Data-plane
- [x] Existing specs: ~30
- [ ] New specs
- [ ] Emitters
- [ ] SDK

**Targeted Sprint**
September

**Additional information**
This is part of a DPG-requested epic to remove all dependecies on OpenAPI libraries from specs

Issue #1264, opened 2024-07-30
[tcgc] add test for encode as string


iscai-msft:
### Clear and concise description of the problem

We've pre-emptively added support for encode int as string in TCGC, however the TCGC compiler hasn't added full support for `@encode(string)` yet. Once this is in main, we will add a test in TCGC to verify the end to end works

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1350, opened 2024-08-12
tcgc, bug fix on encode string

weidongxu-microsoft:
Problem on TCGC 0.45.1 here https://github.com/Azure/cadl-ranch/pull/656#discussion_r1713119741

link https://github.com/Azure/typespec-azure/issues/1264
link https://github.com/Azure/typespec-azure/issues/1179

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -138,15 +138,7 @@ function getAnyType(context: TCGCContext, type: Type): [SdkBuiltInType, readonly
 
 function getEncodeHelper(context: TCGCContext, type: Type, kind: string): string {
   if (type.kind === "ModelProperty" || type.kind === "Scalar") {
-    const encode = getEncode(context.program, type);
-    if (encode?.encoding) {
-      return encode.encoding;
-    }
-    if (encode?.type) {
-      // if we specify the encoding type in the decorator, we set the `.encode` string
-      // to the kind of the encoding type
-      return getSdkBuiltInType(context, encode.type).kind;
-    }
+    return getEncode(context.program, type)?.encoding || kind;
   }
   return kind;
 }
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -215,6 +216,11 @@ export function addEncodeInfo(
       innerType.encode = "bytes";
     }
   }
+  if (isSdkIntKind(innerType.kind)) {
+    if (encodeData && "encode" in innerType) {
+      innerType.encode = getEncodeHelper(context, type, innerType.kind);
```
tadelesh, line 221: i believe `getEncodeHelper` is not used to add encode from typespec, but just a kind of initialization. `addEncodeInfo` is the right place to handle real encode. you may move izzy's change to that function to here.

weidongxu-microsoft: Moved the code from `getEncodeHelper` to `addEncodeInfo`.

Added a test case to verify `@encode` on scalar still works.

tadelesh (approved):

packages/typespec-client-generator-core/src/types.ts
timotheeguerin: tcgc doesn't support it shouldn't be a reason to not add a scenario. This is a valid scenario which is also the most intuitive way to use `@encode`

weidongxu-microsoft: Timothee, not sure I understand your comment.

I've added case of both encode on property
```
      model Test {
        @encode(string)
        value: safeint;
      }
```
and encode on scalar
```
      @encode(string)
      scalar int32EncodedAsString extends int32;

      model Test {
        value: int32EncodedAsString;
      }
```

I assume this is all the cases we had on "integer encode as string'?

timotheeguerin: im sorry i commented on the wrong pr after reading the message in the email, forget, does this pr actually fix this?

timotheeguerin: yeah was about that one so forget what I said https://github.com/Azure/cadl-ranch/pull/656 seems like you are fixing it

haolingdong-msft (approved):

PR #1352, opened 2024-08-12
Add a troubleshooting section for the converter guidance

pshao25:
As titled.

### DIFF
--- a/docs/migrate-swagger/01-get-started.md
+++ b/docs/migrate-swagger/01-get-started.md
@@ -77,10 +77,12 @@ This is the probably most critical step of the conversion. As you have pre-sorte
 
 ### Error "does not have a GetAll method"
 
-If converter fails withe error 
+If converter fails withe error
+
 ```
 fatal   | The ResourceCollection RestorePointCollection (RequestPath: {Certain request path}) does not have a GetAll method
 ```
+
 you could bypass it by the configuration in this [page](https://github.com/Azure/autorest.csharp/blob/feature/v3/docs/mgmt/readme.md#list-exception).
 
 ## How to Get Help
--- a/docs/migrate-swagger/01-get-started.md
+++ b/docs/migrate-swagger/01-get-started.md
@@ -73,6 +73,16 @@ This is the probably most critical step of the conversion. As you have pre-sorte
 
 - Review CI checks such as breaking changes and other failures.
 
+## Troubleshooting
+
+### Error "does not have a GetAll method"
+
+If converter fails withe error 
+```
+fatal   | The ResourceCollection RestorePointCollection (RequestPath: {Certain request path}) does not have a GetAll method
+```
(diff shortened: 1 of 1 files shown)


docs/migrate-swagger/01-get-started.md
```diff
@@ -73,6 +73,18 @@ This is the probably most critical step of the conversion. As you have pre-sorte
...
+
+```
+fatal   | The ResourceCollection RestorePointCollection (RequestPath: {Certain request path}) does not have a GetAll method
+```
+
+you could bypass it by the configuration in this [page](https://github.com/Azure/autorest.csharp/blob/feature/v3/docs/mgmt/readme.md#list-exception).
+
 ## How to Get Help
 
 - Ask questions in the [TypeSpec Discussions Channel](https://teams.microsoft.com/l/channel/19%3a906c1efbbec54dc8949ac736633e6bdf%40thread.skype/TypeSpec%2520Discussion%2520%25F0%259F%2590%25AE?groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47)
```
lirenhe, line 90: Can we use aka.ms link here?

docs/migrate-swagger/01-get-started.md
```diff
@@ -73,6 +73,18 @@ This is the probably most critical step of the conversion. As you have pre-sorte
 
 - Review CI checks such as breaking changes and other failures.
 
+## Troubleshooting
+
+### Error "does not have a GetAll method"
+
+If converter fails withe error
```
lirenhe, line 80: Typo?  'withe'

docs/migrate-swagger/01-get-started.md
```diff
@@ -73,6 +73,18 @@ This is the probably most critical step of the conversion. As you have pre-sorte
...
+
+```
+fatal   | The ResourceCollection RestorePointCollection (RequestPath: {Certain request path}) does not have a GetAll method
+```
+
+you could bypass it by the configuration in this [page](https://github.com/Azure/autorest.csharp/blob/feature/v3/docs/mgmt/readme.md#list-exception).
```
lirenhe, line 86: No quite sure about the action here.
By checking the link, do you suggest to update the config or update the Swagger to solve the problem?

docs/migrate-swagger/01-get-started.md
pshao25: We will delete this workaround after we remove the .net dependency. I prefer not to add an aka item.

ArthurMa1978: Here is the short link for this channel https://aka.ms/typespec/discussions.
We need support TypeSpec migration for quite some time, so this line should remain in the document.

pshao25:
Synced offline. We will add suppression by default. https://github.com/Azure/azure-sdk-tools/pull/8827

PR #1357, opened 2024-08-12
Azure Core linter rules docs and cleanup - Pass 1

timotheeguerin:
Going through the Azure core linter rule and adding docs + cleaning up 

- `use-standard-operations`: Add docs for linter rule and cleanup a lot of things not needed since we moved to the linter system
- `spread-discriminated-model`: Add docs
- `prevent-format`: Add docs and cleanup
- `no-nullable`: Add Docs


docs/libraries/azure-core/rules/no-multiple-discriminator.md
```diff
@@ -0,0 +1,45 @@
+---
+title: "no-multiple-discriminator"
+---
+
+```text title="Full name"
+@azure-tools/typespec-azure-core/no-multiple-discriminator
+```
+
+Using a nested polymophic relationship is now allowed in Azure services. Most JSON serializer and deserializer do not support this feature.
```
markcowl, line 9:
```suggestion
Using a nested polymophic relationship is not allowed in Azure services. Most JSON serializers and deserializers do not support this feature.
```

docs/libraries/azure-core/rules/use-standard-operations.md
```diff
@@ -0,0 +1,36 @@
...
+Azure Data Plane services should use standard operations defined in the Azure Core library.
+
+#### ❌ Incorrect
+
+```tsp
+op myResourceRead(): MyResource;
+op myResourceCreate(@body resource: MyResource);
+```
+
+Using operations from `Azure.Core.Foundations` it not acceptable either.
```
markcowl, line 18:
```suggestion
Using operations from `Azure.Core.Foundations` is not acceptable either.
```

markcowl (approved):
Always nice when you can remove code labeled with 'hack, hack, hack'

Issue #1348, opened 2024-08-12
[docs] Improve docs for dataplane code generation for Long running operations.


lirenhe:
Let's have the document to clarify how service teams should author TypeSpec for LRO and what the Code generated from that TypeSpec.

PR #1361, opened 2024-08-13
docs, add DPG LRO doc

weidongxu-microsoft:
fix https://github.com/Azure/typespec-azure/issues/1348

I am using same op/model from cadl-ranch https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/lro/standard/main.tsp


docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
```diff
@@ -0,0 +1,188 @@
...
+  NoRepeatableRequests &
+  NoClientRequestId>;
+
+@resource("users")
+@doc("Details about a user.")
+model User {
+  @key
+  @visibility("read")
+  @doc("The name of user.")
+  name: string;
```
MaryGao, line 26: Is it possible we make this name as optional not required? Since User would be used both input and output model, and the splitting model introducing by visibility is still under discussion. JS may not handle this required read-only yet.

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+```typespec
+alias ResourceOperations = global.Azure.Core.ResourceOperations<NoConditionalRequests &
+  NoRepeatableRequests &
+  NoClientRequestId>;
+
+@resource("users")
+@doc("Details about a user.")
+model User {
+  @key
```
MaryGao, line 23: Could you help me understand what the @key here is for? Can we remove this?

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
```diff
@@ -0,0 +1,188 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+# Long-Running Operations
```
MaryGao, line 4: Some services may have requirement to expose polling operation, should we have an example for that case?

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/apicenter/ApiCenter.DataApi/resources/ApiDefinition.tsp#L27

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
weidongxu-microsoft: I think `@resource` must have a `@key`

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
weidongxu-microsoft: guide https://azure.github.io/typespec-azure/docs/getstarted/azure-core/step04

```
@doc("A widget.")
@resource("widgets")
model Widget {
  @key("widgetName")
  @doc("The widget name.")
  @visibility("read")
  name: string;

  @doc("The ID of the widget's manufacturer.")
  manufacturerId: string;
}
```

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
weidongxu-microsoft: I am currently not sure whether SDK reviewers want to have this API in SDK API.

Does JS have some way to get the "id" easily and allow customer call this GET API with this "id"?

In Java, it is not easy for customer to get the "id"... (in standard LRO, customer can read the "id" from the poll response, and use that -- but this is not intuitive)

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
weidongxu-microsoft: If you have strong opinion, maybe we can see if we remove `@visibility("read")` and add `@path`

MaryGao: Got it, no strong opinion here since it is also in typespec guide i think this is more js issue.

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
MaryGao: I am asking because I notice when I searched the `LongRunningResource` I found almost operations would define their polling operation e.g @pollingOperation(OperationStatuses.get) but not every exposed it in client.tsp.

So I was wondering if this is common case or should their polling operation be removed?

weidongxu-microsoft: It depends. In my opinion, if they use the standard LRO here, `@pollingOperation` should be removed.

But if they use their customized LRO, they may need it.

MaryGao: Got it and i am okay not adding customized polling here if this is not common case.

weidongxu-microsoft: I am checking in channel about what is the best practice.
https://teams.microsoft.com/l/message/19:16f8f1ed70d044139be153707bdaee0e@thread.skype/1723790726909?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1723790726909&teamName=Azure%20SDK&channelName=TypeSpec%20Engineering&createdTime=1723790726909

iscai-msft (approved):

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
```diff
@@ -0,0 +1,190 @@
...
+  NoRepeatableRequests &
+  NoClientRequestId>;
+
+@resource("users")
+@doc("Details about a user.")
+model User {
+  @key
+  @visibility("read")
+  @doc("The name of user.")
+  name: string;
```
lirenhe, line 26: As the name is defined as read-only, why people can set its value through the SDK?

weidongxu-microsoft: In request, the name is on path (not body).

iscai-msft: @weidongxu-microsoft should we set it as readonly, or only as a key

weidongxu-microsoft: discussion with Mary here https://github.com/Azure/typespec-azure/pull/1361#discussion_r1718025420

Currently I am using the resource definition pattern in typespec-azure get-started doc
https://azure.github.io/typespec-azure/docs/getstarted/azure-core/step04

The alternative is to remove the `@visibility` and add a `@path`. Though this use may trigger a warning (probably depends on the template), e.g. 
```
path property will be ignored as it is inside of a @body property. Use @bodyRoot instead if wanting to mix.TypeSpec(@typespec/http/metadata-ignored)
```

The `@path` may also be involved a consistency discussion, as I remember currently we filter the `@path` (query/header too) from model/class.

docs/howtos/DataPlane Generation - DPG/10longRunningOperations.mdx
weidongxu-microsoft: Mark's reply says they required the pollingOperation. I will add a `getOperationStatus` but comment that SDK may not want to include it.

I remember a long discussion months before, about whether emitter include the op in `@pollingOperation`. I remember the decision is to handle it like any other operation.
But there is way to exclude it. One be client.tsp `@client` omit it. Another to define it in another namespace (I didn't see any service did this, though).

PR #1363, opened 2024-08-13
[tcgc] support uri template

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1320
resolve: https://github.com/Azure/typespec-azure/issues/1317
resolve: #1316
resolve: #1319


packages/typespec-client-generator-core/test/packages/utils.ts
```diff
@@ -0,0 +1,54 @@
+import { strictEqual } from "assert";
+import {
```
haolingdong-msft, line 2: name the file to utils.test.ts?

haolingdong-msft: ok, I see, this is just a util for testing, not test on the utils, then the name should be fine.

packages/typespec-client-generator-core/test/packages/parameters.test.ts
```diff
@@ -0,0 +1,1005 @@
...
+  SdkHttpOperation,
+  SdkPathParameter,
+  SdkQueryParameter,
+  SdkServiceMethod,
+} from "../../src/interfaces.js";
+import { SdkTestRunner, createSdkTestRunner } from "../test-host.js";
+import { getServiceMethodOfClient, getServiceWithDefaultApiVersion } from "./utils.js";
+
+describe("typespec-client-generator-core: parameters", () => {
+  let runner: SdkTestRunner;
```
haolingdong-msft, line 15: The test files under `packages` folder are all used to test `sdkpackage`? Shall we move `package.test.ts` under this `package` folder as well?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -479,11 +479,19 @@ export interface SdkQueryParameter extends SdkModelPropertyTypeBase {
...
 }
 
 export interface SdkPathParameter extends SdkModelPropertyTypeBase {
   kind: "path";
+  /**
+   * @deprecated This property is deprecated. Use `allowReserved` instead.
+   * @param serviceParam
+   */
   urlEncode: boolean;
+  explode?: boolean;
```
lirenhe, line 492: Why explode/allowReserved are defined as nullable?  What if this value is null?

Seems we got this value from the compiler, @timotheeguerin, what is the meaning of null for those properties?

packages/typespec-client-generator-core/test/packages/parameters.test.ts
```diff
@@ -0,0 +1,1005 @@
...
+
+  beforeEach(async () => {
+    runner = await createSdkTestRunner({ emitterName: "@azure-tools/typespec-python" });
+  });
+
+  it("path basic", async () => {
```
haolingdong-msft, line 21: Seems most of tests are moved from `package.test.ts`, could you please highlight the part of tests you added?

packages/typespec-client-generator-core/test/packages/parameters.test.ts
tadelesh: yes. i'll do this step by step. i don't want to add too many refactor code in this pr, just move the parameter's tests first.

packages/typespec-client-generator-core/test/packages/parameters.test.ts
```diff
@@ -0,0 +1,1005 @@
...
+    const sdkPackage = runner.context.sdkPackage;
+    const method = getServiceMethodOfClient(sdkPackage);
+    strictEqual(method.parameters.length, 0);
+    strictEqual(method.response.type, undefined);
+    strictEqual(method.operation.parameters.length, 0);
+    strictEqual(method.operation.responses.get(200)?.headers.length, 0);
+    strictEqual(method.operation.responses.get(200)?.type, undefined);
+  });
+
+  describe("uri template related", () => {
```
tadelesh, line 742: @haolingdong-msft and other reviewers, please only review the tests below.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -543,6 +551,7 @@ export interface SdkHttpOperation extends SdkServiceOperationBase {
   __raw: HttpOperation;
   kind: "http";
   path: string;
+  uriTemplate: string;
```
haolingdong-msft, line 554: What does emitter suppose to do with `uriTemplate` currently? I think we only care about `allowReserved` and `explode` for now, this field is for the whole `uriTemplate` feature support in the future? Please correct me if I'm wrong.

Also is this a must-have field, or optional field? `uriTemplate?: string` or `uriTemplate: string`

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -479,11 +479,19 @@ export interface SdkQueryParameter extends SdkModelPropertyTypeBase {
...
 export interface SdkPathParameter extends SdkModelPropertyTypeBase {
   kind: "path";
+  /**
+   * @deprecated This property is deprecated. Use `allowReserved` instead.
+   * @param serviceParam
+   */
   urlEncode: boolean;
+  explode?: boolean;
+  style?: "simple" | "label" | "matrix" | "fragment" | "path";
+  allowReserved?: boolean;
```
haolingdong-msft, line 494: Is `allowReserved` only for path parameter? Can query parameter define this? I did not find it is only applied to path parameter in [RFC](https://datatracker.ietf.org/doc/html/rfc6570#page-22) and [issue](https://github.com/microsoft/typespec/issues/3736).
image

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: 1. correct
2. required field (same with typespec core)

packages/typespec-client-generator-core/test/packages/parameters.test.ts
```diff
@@ -0,0 +1,1005 @@
...
+      `);
+      const sdkPackage = runner.context.sdkPackage;
+      const method = getServiceMethodOfClient(sdkPackage);
+      strictEqual(method.operation.path, "/explicit/{param}");
+      strictEqual(method.operation.uriTemplate, "/explicit/{param}");
+      const param = method.operation.parameters[0] as SdkPathParameter;
+      strictEqual(param.allowReserved, false);
+      strictEqual(param.style, "simple");
+      strictEqual(param.explode, false);
+    });
```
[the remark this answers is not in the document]
tadelesh: the former use uri template and not use `@path` to label path param, the later use `@path`.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: i infer this conclusion from typespec's design. @timotheeguerin could you confirm?

timotheeguerin: Yeah query params cannot but I think after some investigation is is kinda pointless as to skip encoding for query params

packages/typespec-client-generator-core/src/interfaces.ts
timotheeguerin: Right now you can do whatever you want to support whatever you want. But the longer term goal is you use the uni template to produce the uri by using an implementation of the algorithm(not very complex)

packages/typespec-client-generator-core/src/interfaces.ts
timotheeguerin: I thought the compiler was always setting it. But default explode is false default style is simple. Default is not allow reserved: true

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -1040,13 +1044,16 @@ Currently we only support HTTP service parameters.
...
+export type CollectionFormat = "multi" | "csv" | "ssv" | "tsv" | "pipes" | "simple" | "form";
 
 export interface SdkQueryParameter extends SdkModelPropertyTypeBase {
   kind: "query";
+  // the collection format to use to separate each value in array of a query parameter
   collectionFormat?: CollectionFormat;
   serializedName: string;
   correspondingMethodParams: SdkModelPropertyType[];
+  // if true, send each value in array or object as a separate query parameter
+  explode?: boolean;
```
lirenhe, line 1056: Shall we also update here as it is not nullable for the TCGC model.

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -1070,7 +1077,17 @@ export interface SdkHeaderParameter extends SdkModelPropertyTypeBase {
 ```ts
 export interface SdkPathParameter extends SdkModelPropertyTypeBase {
   kind: "path";
-  urlEncode: boolean;
+  // when interpolating this parameter in the case of array or object, expand each value using the given style
+  explode?: boolean;
```
lirenhe, line 1081: same above

lirenhe (approved):
Some update of mdx needed, other parts look good.

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -1040,13 +1044,16 @@ Currently we only support HTTP service parameters.
 This represents an HTTP query parameter.
 
 ```ts
-export type CollectionFormat = "multi" | "csv" | "ssv" | "tsv" | "pipes";
+export type CollectionFormat = "multi" | "csv" | "ssv" | "tsv" | "pipes" | "simple" | "form";
```
weidongxu-microsoft, line 1047: nit, could we have a similar list as path? I am not sure whether I know what's for `simple` and `form`.

```
  // - simple: no separator
  // - label: . as separator
  // - matrix: ; as separator
  // - fragment: # as separator
  // - path: / as separator
```

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -128,26 +133,26 @@ function getSdkHttpParameters(
...
               expectedType: "body",
-              actualType: getParamResponse.kind,
+              actualType: bodyParam.kind,
             },
           })
         );
         return diagnostics.wrap(retval);
       }
-      retval.bodyParam = getParamResponse;
+      retval.bodyParam = bodyParam;
```
weidongxu-microsoft, line 155: What are we modifying here? I didn't see the relationship with uri template.

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -292,75 +297,90 @@ function addContentTypeInfoToBodyParam(
...
+ */
 export function getSdkHttpParameter(
   context: TCGCContext,
-  type: ModelProperty,
+  param: ModelProperty,
   operation?: Operation,
+  httpParam?: HttpOperationParameter,
   location?: "path" | "query" | "header" | "body"
-): [SdkHttpParameter, readonly Diagnostic[]] {
+): [SdkHttpParameter | void, readonly Diagnostic[]] {
```
weidongxu-microsoft, line 315: On which case would `void` be returned? What does it mean for a parameter to be void?

What do you expect emitter do on this `void`?

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
tadelesh: added, `simple` is same as `csv` and `form` is same as `multi`

packages/typespec-client-generator-core/src/http.ts
tadelesh: just a naming refactor

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
```diff
@@ -891,7 +891,10 @@ export interface SdkHttpOperation {
   kind: "http";
   // raw HTTP operation output from typespec/http
   __raw: HttpOperation;
+  // route path for the target operation. TypeSpec suggest to use `uriTemplate` instead as `path` will not work for complex cases like not-escaping reserved chars.
```
iscai-msft, line 894: Should we deprecate path? Or wait until typespec deprecated path as well

packages/typespec-client-generator-core/src/interfaces.ts
haolingdong-msft: @timotheeguerin, in the RFC there are examples that `allowReserved` can be applied to query parameters. May I know why query params cannot `allowReserved` and it is pointless? 
https://datatracker.ietf.org/doc/html/rfc6570#page-22
image

/cc @srnagar

docs/howtos/DataPlane Generation - DPG/07tcgcTypes.mdx
tadelesh: after typespec, i believe.

packages/typespec-client-generator-core/src/interfaces.ts
MaryGao: @timotheeguerin we have query with `"x-ms-skip-url-encoding": true` in [swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/web/resource-manager/Microsoft.Web/stable/2020-06-01/Recommendations.json#L476-L481). I guess for service team side this will be helpful in paging filter in which customers would input the already encoded string.

PR #1368, opened 2024-08-13
Add new option, deprecate old one and make default valuable for example directory

### LINKED ISSUE #1173: [Bug]: Specs can fail to load examples when compiled from outside spec folder
### Describe the bug

Specs can fail to load examples when compiled from outside the spec folder (e.g. `tsp compile path/to/spec`), but succeed when compiled from inside the spec folder (e.g. `cd path/to/spec; tsp compile .`).

Related: https://github.com/microsoft/typespec/issues/2241, https://github.com/Azure/typespec-azure-pr/issues/3182

### Reproduction

```
$ gh repo clone azure/azure-rest-api-specs
$ cd azure-rest-api-specs
$ npm ci

$ npx tsp compile specification/oracle/Oracle.Database.Management
warning @azure-tools/typespec-autorest/example-loading: Skipping example loading from examples/2023-09-01-preview because there was an error reading the directory.

$ cd specification/oracle/Oracle.Database.Management
$ npx tsp compile .
<no warning>
```

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).

timotheeguerin:
fix #1173
fix azure/typespec-azure-pr#2754


packages/typespec-autorest/src/lib.ts
```diff
@@ -42,10 +42,17 @@ export interface AutorestEmitterOptions {
 
   /**
    * Directory where the examples are located.
-   * @default `{cwd}/examples`
```
timotheeguerin, line 45: This wasn't actually true, if the examples directory wasn't set it wasn't loading examples at all.
I think it is better that we try to always load from `{project-root}/examples` as every azure service has this structure and would allow all tspconfig to be simplified by one config option

packages/typespec-autorest/test/decorators.test.ts
```diff
@@ -287,7 +287,7 @@ describe("typespec-autorest: decorators", () => {
     it("duplicate example file", async () => {
       const host = await createAutorestTestHost();
       const runner = await createAutorestTestRunner(host, {
-        "examples-directory": resolveVirtualPath("./examples"),
+        "examples-dir": resolveVirtualPath("./examples"),
```
markcowl, line 290: Seems like we should add a test dir with no configuration setting for examples to protect the default behavior (loading from "./examples") - we could duplicate this test with no examples-dir setting, for example

markcowl (changes requested):
Mostly looks good.  Given that we weren't really loading form './examples' before, feels like we should protect this behavior with a test

packages/typespec-autorest/test/decorators.test.ts
timotheeguerin: added more test

markcowl (approved):

PR #1377, opened 2024-08-15
[Tcgc] Fix getLibraryName for anonymous model which is derived from template

msyyc:
After https://github.com/microsoft/typespec/pull/4144, anonymous model type may also have templateMapper arguments. For this scenario, `getLibraryName` shall not try to generate name from arguments, otherwise may generate model with duplicated name.


tadelesh (approved):

weidongxu-microsoft (approved):

weidongxu-microsoft:
Do we need same fix in release/august branch?

Issue #1235, opened 2024-07-25
[tcgc] doc representation in tcgc


tadelesh:
In TypeSpec, we have `@doc` and `@summary` to give a documentation string for a type.
In TCGC, we have some logic to convert them into two type's attribute: `description` and `details`, the logic is:
1. if there is no `@summary`, then `@doc` content will be put into `description`
2. if there is `@summary`, then `@doc` content will be put into `details` and `@summary` content will be put into `description`

Since I received some feedbacks to keep the original TypeSpec doc structure, I want to revisit the design in this issue.

@haolingdong-msft @archerzz could you share your feedback here?

@m-nash Izzy told me the original design is from you, could you share the thinking?

FYI: @jhendrixMSFT @Azure/dpg-devs 

### create new doc and summary from @doc and @summary and deprecate the previous description and details ###

haolingdong-msft:
I would slightly prefer keeping the original output from TCGC, i.e. `description` and `summary`.
As Java codegen will have some handling using `description` or `summary`, and using the current TCGC's `details` and `description, we can't distinghuish `summary` and `codegen`. So we will meet some generated code description change, and we need to update our codegen logics. e.g https://github.com/Azure/autorest.java/pull/2861#discussion_r1683754341

It's fine to update our codegen logics, but just if we can avoid additional efforts, it's better.

archerzz:
C# is using a different logic for operations. See the codes: https://github.com/Azure/autorest.csharp/blob/302040fc2e59e65e316956fc8ad117169cf5247a/src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs#L62-L63
```csharp
public string? SummaryText => Summary.IsNullOrEmpty() ? Description : Summary;
public string? DescriptionText => Summary.IsNullOrEmpty() || Description == Summary ? string.Empty : Description;
```

I think the above logic is:
- `SummaryText`: It's printed into `<Summary>` in the XML comments.
  - use `Summary` if it's not empty
  - Otherwise use `Description`
- `DescriptionText`: It's printed into `<Remarks>` in the XML comments.
  - Use `Description` if it's not empty and it's not the same as `Summary`

Previously, `Summary` comes from `@summary`, and `Description` comes from `@doc`.

archerzz:
[...]

PR #1379, opened 2024-08-15
[tcgc] add `doc` and `summary` to tcgc types

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1235


packages/typespec-client-generator-core/test/types/duration-type.test.ts
```diff
@@ -115,8 +115,10 @@ describe("typespec-client-generator-core: duration types", () => {
     strictEqual(sdkType.kind, "array");
     strictEqual(sdkType.valueType.kind, "duration");
     strictEqual(sdkType.valueType.name, "Float32Duration");
-    strictEqual(sdkType.valueType.description, "title");
-    strictEqual(sdkType.valueType.details, "doc");
+    strictEqual(sdkType.valueType.description, "title"); // eslint-disable-line deprecation/deprecation
```
lirenhe, line 118: Can we test with different locales like Chinese etc?

tadelesh: is it kind of test for typespec, not for tcgc?

iscai-msft (approved):

packages/typespec-client-generator-core/test/types/duration-type.test.ts
tadelesh: added.

Issue #1132, opened 2024-07-08
[TCGC] - use original model for simple spread instead of creating a new anonymous model


joheredi:
### Clear and concise description of the problem

There are scenarios where tracking the source model of anonymous models is beneficial. For instance, in TypeScript code generation, this feature would enable us to map serializer/deserializer types between RLC (Rest Level Client) and Modular. This functionality is crucial for bridging gaps between RLC and Modular types.

Example
Consider the following TypeSpec model:

```ts
model Foo {
   bar: {
      baz: string;
   }
} 
```

With the ability to track the source model, we could implement the following TypeScript code:

```ts
declare function deserializeFooBar(input: FooRest["bar"]): BarModel;
```

Here, BarModel is generated from TCGC, but there is no equivalent interface in RLC as it doesn't generate anonymous models.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

tadelesh:
could you clarify the source meaning? for example, if anonymous model is in parameter, what should the source model be? other cases are model property, return type, etc.

iscai-msft:
@ArcturusZhang has the following comment:

For this typespec:
```typespec
model Bar {}
model Baz {}
op foo(...Bar): Baz;
```
the TCGC result gives us a `FooRequest` as the model of body parameter. Since it is speading model `Bar`, could we get back the model `Bar` as the body parameter in this case?
This help us reduce the size of the generated assembly.

We agree that there are cases that returning the original model is not possible, such as an operation with spread of two different models, in those cases, we could still do the `FooRequest` as body because it is combining the two models.
But we would like to get the model back if the model is there.

tadelesh:
> @ArcturusZhang has the following comment:
> 
> For this typespec:
> 
> ```
> model Bar {}
> model Baz {}
> op foo(...Bar): Baz;
> ```
> 
[...]

PR #1388, opened 2024-08-16
[tcgc] spread behavior change

tadelesh:
use original model for body parameter in `SdkServiceOperation` if it is from a simple spread.

simple spread means spread model without any `@header`, `@query` or `@path` property and the spread model is the only source of http body.

examples, 

`op test(...Model): void;`

`op test(@header h: string, @query q: string, ...Model): void;`

these two operation will have `Model` as body parameter.

```
op test(...A): void;
model A {
  @query
  q: string;
  prop: string;
}
```
this operation will have an anonymous model with only `prop` property and generated name `TestRequest` as body parameter.

resolve: https://github.com/Azure/typespec-azure/issues/1132

cc: @ArcturusZhang @joheredi


iscai-msft (approved):

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -546,3 +553,41 @@ export function isXmlContentType(contentType: string): boolean {
...
+
+/**
+ * If body is from spread, then it does not directly from a model property.
+ * @param httpBody
+ * @param parameters
+ * @returns
+ */
+export function isHttpBodySpread(
+  httpBody: HttpOperationBody | HttpOperationMultipartBody,
+  parameters: Model
```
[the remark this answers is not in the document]
tadelesh: good catch, when i change the function logic, forgot to change the signature.

tadelesh:
@ArcturusZhang  @joheredi please review and check if it could satisfy your language's requirement, thanks.

ArcturusZhang:
Do we have a dev version so that we could see if this change could meet our requirement?
The code here LGTM

tadelesh:
> Do we have a dev version so that we could see if this change could meet our requirement? The code here LGTM

before this pr merge, you could use the ci build artifact. after merge, you could use the dev version.

tadelesh:
@ArcturusZhang have you tried this in .net emitter? if all behavior is good, could you approve this pr?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -546,3 +553,38 @@ export function isXmlContentType(contentType: string): boolean {
...
+ * @param type
+ * @returns
+ */
+export function getHttpBodySpreadModel(type: Model): Model {
+  if (type.sourceModels.length === 1 && type.sourceModels[0].usage === "spread") {
+    const innerModel = type.sourceModels[0].model;
+    // for case: `op test(...Model):void;`
+    if (innerModel.name !== "" && innerModel.properties.size === type.properties.size) {
+      return innerModel;
+    }
```
weidongxu-microsoft, line 578: Did we consider the case that the `Model` here include `@path` `@query` etc.?

tadelesh: nope, see pr description for details.

packages/typespec-client-generator-core/test/decorators.test.ts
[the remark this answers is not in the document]
tadelesh: added more tests

packages/typespec-client-generator-core/src/internal-utils.ts
weidongxu-microsoft: Wondering which line exclude this scenario. `innerModel.properties.size === type.properties.size`?

If `Model` contains `@query`, would the size be different?

weidongxu-microsoft (approved):

Issue #1351, opened 2024-08-12
[docs] Improve docs for dataplane code generation for Etag headers


lirenhe:
Please provide doc on how codegen would handle etag handers.

PR #1393, opened 2024-08-19
doc for etag

msyyc:
fix https://github.com/Azure/typespec-azure/issues/1351

### DIFF
--- a/docs/howtos/Client Generation/11etag.mdx
+++ b/docs/howtos/Client Generation/11etag.mdx
@@ -1,0 +1,64 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+# Conditional Request Header
+
+This doc details what emitters will generate for conditional request header
+
+
+##  `If-Match` and `If-None-Match`
+<Tabs>
+<TabItem value="typespec" label="TypeSpec" default>
+
+```typespec
+op get(
+  @header("If-Match")
+  @doc("The request should only proceed if an entity matches this string.")
+  ifMatch?: string;
+
+  @header("If-None-Match")
+  @doc("The request should only proceed if no entity matches this string.")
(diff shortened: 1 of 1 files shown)


docs/howtos/Client Generation/11etag.mdx
```diff
@@ -0,0 +1,64 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+# Conditional Request Header
+
+This doc details what emitters will generate for conditional request header
+
+## `If-Match` and `If-None-Match`
```
XiaofeiCao, line 8: How about `If-Modified-Since` and `If-Unmodified-Since`?

I think we have cases of either 2 or 4 headers.

docs/howtos/Client Generation/11etag.mdx
```diff
@@ -0,0 +1,64 @@
...
+```typespec
+op get(
+  @header("If-Match")
+  @doc("The request should only proceed if an entity matches this string.")
+  ifMatch?: string,
+
+  @header("If-None-Match")
+  @doc("The request should only proceed if no entity matches this string.")
+  ifNoneMatch?: string,
+): void;
```
XiaofeiCao, line 22: `get` returning `void` seems weird...

docs/howtos/Client Generation/11etag.mdx
msyyc: Will add for the other 2 headers later.

docs/howtos/Client Generation/11etag.mdx
msyyc: response in this case doesn't matter, but GET without response indeed be weird... Let me add a response type.

weidongxu-microsoft (approved):

XiaofeiCao (approved):

Issue #1290, opened 2024-08-05
If access/usage are used at the same time, access is not propagated anymore


lmazuel:
Today we have the following TCGC behavior:
- If access(public) is used on a model, propagate to all included models
- If usage() is used on a model, don't propagate (too complex)

Thought, we have an interesting side effect, that if both `access` and `usage` are used at the same time on a model, suddenly the access propogation doesn't work. 

I expect that the two features are indenpendant, and that if used together. we keep having access propagated.

CC @l0lawrence @richardpark-msft

tadelesh:
refer doc: https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators#@Azure.ClientGenerator.Core.access. `@access` override on model will not be propagated. only `@access(internal)` on operation will impact models' access calculation.
we decided to not to do any propagation for `@access` and `@usage` on model and let user to take responsible for that if they want to change the default calculation.
do we have a real case that are confusing? i could explain the calculation logic and see if we need to do some change.

lmazuel:
Why would we not propagate access on models? If you have A has a property on B, and you make A public, having B as private is literally creating invalid SDK in the first place.

tadelesh:
propagation for `@access` will cause conflict that we could not determine whether the access should be `public` or `private` (a simple example as follows). so if user want to override access, we want them to do the cascade. we just want to reduce the complexity for tcgc calculation.
```
@access(private)
model A {
  prop: B
}

@access(public)
model B {
  prop: A
}

```

richardpark-msft:
@tadelesh, this seems like a good situation to either error out if there are conflicting directives. My main concern here is that a developer (like me!) would have no idea we're relying on an implementation detail of resolution.

@lmazuel mentioned that he didn't think this situation was an error, so it might be more idiomatic to perhaps emit a warning? 

I think the key here is that two people, rather experienced in this area, missed this bug and it took awhile for us to figure out what happened.

lmazuel:
> propagation for `@access` will cause conflict that we could not determine whether the access should be `public` or `private` (a simple example as follows). so if user want to override access, we want them to do the cascade. we just want to reduce the complexity for tcgc calculation.
> 
> ```
> @access(private)
> model A {
>   prop: B
> }
> 
> @access(public)
> model B {
>   prop: A
> }
> ```

[...]

PR #1395, opened 2024-08-19
[tcgc] usage and access override propagation

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1290
**calculation logic is [here](https://microsoft.sharepoint.com/:fl:/s/b1efc18e-ee6e-4d7f-823e-936d90153230/EY9i28kbSQpEsrYzxCpSaDkB7DoyVcAi56TkDgqTufQzpQ?e=uZ9HPz&nav=cz0lMkZzaXRlcyUyRmIxZWZjMThlLWVlNmUtNGQ3Zi04MjNlLTkzNmQ5MDE1MzIzMCZkPWIlMjFtUEQtbDRVcV9rU1J0Y1M0Z3dHekJPcC1xWHBqbjJwQXVPak1tWDRneGxZM2JIMlRJX0RXUmJxWGhLY0Qzal9EJmY9MDFUSlhZVVdVUE1MTjRTRzJKQkpDTEZOUlRZUVZGRTJCWiZjPSUyRiZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUZ0VUVRdGJEUlZjVjlyVTFKMFkxTTBaM2RIZWtKUGNDMXhXSEJxYmpKd1FYVlBhazF0V0RSbmVHeFpNMkpJTWxSSlgwUlhVbUp4V0doTFkwUXphbDlFZkRBeFZFcFlXVlZYVTA5V1JFUktTMEZDVEZJMVEwc3pTRTlTVWt4U1ZVY3lTbEUlM0QlMjIlMkMlMjJpJTIyJTNBJTIyNDgxYzVmMjQtMGM5Yi00ZjgxLWFlOTktYWM2MTJhMjg1NzdlJTIyJTdE)**


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1475,37 +1419,44 @@ function updateUsageOfModel(
...
+    // usage set
+    type.usage |= value;
   } else {
-    type.usage |= usage;
+    // access set
+    if (!type.__accessSet || type.access !== "public") {
+      type.access = value;
+      type.__accessSet = true;
+    }
+    if (options.isOverride) {
```
iscai-msft, line 1450: I think another option here so we don't need `__accessSet` is still setting it to some kind of default, like undefined, and at the end before we return, all `undefined`s can be set to `"public"`, since we're already iterating over them

iscai-msft (approved):

packages/typespec-client-generator-core/src/types.ts
tadelesh: i tried to do that at first. but it should either make the interface access to be | undefined or use as magic to force set to the type. maybe the later one is better?

iscai-msft: ok no worries, that's a good point about how we don't want `undefined` to be a valid value for `access`

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -198,6 +198,18 @@ export const $lib = createTypeSpecLibrary({
         default: `Cannot flatten property of polymorphic type.`,
       },
     },
+    "conflict-access-override": {
+      severity: "warning",
+      messages: {
+        default: `@access override conflicts with the access calculated from operation or other @access override.`,
```
iscai-msft, line 204: i think it would be good to just add the name of the model that is conflicting

tadelesh: the warning info is binding to the model type, and the diagnostics will help to show where the warning is come from, so i did not add name in the message.

Issue #1391, opened 2024-08-19
[TCGC] `api-version` parameter should not be `onClient===true` if the server is non-versioned


haolingdong-msft:
Case: 
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/server/versions/not-versioned/main.tsp#L33

This is a non-versioned client, and I think the `api-version` should not be on client, but TCGC mark it as` onClient`.
TCGC output:
![image](https://github.com/user-attachments/assets/f6cdb889-a25b-4b1c-9274-b41505b67387)

Seems  a bug in TCGC?

iscai-msft:
Our current behavior is that we will always elevate `api-version` onto the client. It does make sense that versioning corresponds to setting api versioning on clients, but I do think we should talk about this in scrum on weds. Adding that to the agenda list

haolingdong-msft:
Sure, let's have discussion how the generated sdk should look like for non-versioned clients, I think we should not add api-version to the client parameter, otnerwise  it is not **non-versioned** client.
For Java, if it is a non-versioned client, we will not add api-version parameter to the client. How about other languages? 
/cc @iscai-msft @tadelesh @archerzz @chunyu3 @qiaozha @MaryGao

archerzz:
C# doesn't check `onClient`. C# emitter will pass the `api-version` method parameter to the generator. And generator will check if there is already a client parameter `api-version`. If so, the method parameter will be suppressed: https://github.com/Azure/autorest.csharp/blob/bb1c49bfd05512775ed080679152e5cad1f64211/src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs#L291-L301

What's `onClient` for? Does it indicate that the query parameter already exists on client level?

haolingdong-msft:
Hi @archerzz , from my understanding, `onClient` is used to decide whether the parameter should be on client or not. And emitter is expected to use this value to decide whether to put the parameter on client. This is also related with Izzy's latest [pr ](https://github.com/Azure/typespec-azure/pull/1398) as well,  which adds client parameters to the client initialization object. Those parameters will also have `onClient==true` on `SdkHttpOperation`.

haolingdong-msft:
Summary on 9/5 meeting:
1. TCGC will mark `onClient` as `false` for api-version in non-versioned clients.
2. (TBD) what should we set for `isApiVersion` flag on the parameter.

archerzz:
[...]

PR #1398, opened 2024-08-19
[tcgc] add `@clientInitialization` decorator

### LINKED ISSUE #914: TCGC: @methodLocation decorator
Is it possible to move parameters that persist on each operation to the client instead with a decorator of some sort? In EventGrid this shows up in the case of
 
client()
publish(topic_name, events)
receive(topic_name, subscription_name, max_events)
ack(topic_name, subscription_name, lock_tokens)
 
to 
 
client(topic_name, subscription_name)
publish(events)
receive(max_events)
ack(lock_tokens)

typespec spec: https://github.com/Azure/azure-rest-api-specs/blob/9df71d5a717e4ed5e6728e7e6ba2fead60f62243/specification/eventgrid/Azure.Messaging.EventGrid/main.tsp#L261

iscai-msft:
fixes #914


packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -4743,4 +4743,256 @@ describe("typespec-client-generator-core: decorators", () => {
...
+
+      const methods = client.methods;
+      strictEqual(methods.length, 2);
+      const download = methods[0];
+      strictEqual(download.name, "download");
+      strictEqual(download.kind, "basic");
+      strictEqual(download.parameters.length, 2);
+
+      const blobNameParam = download.parameters.find((x) => x.name === "blobName");
+      ok(blobNameParam);
```
chunyu3, line 4911: parameter `blobName` is bump to client, so parameter `blobName` is not the paramter of the `download` method

tadelesh (commented):
three concerns:
1. can we have a test with `@client` and `@operationGroup`? 
2. i'm wondering if we need to add a lint to disallow promoting method param to client if it is not in all client's operation?
3. with this decorator, is it possible to remove the hard code for promoting `subscriptionId` cause it is not already client parameter in brownfield?

iscai-msft:
> three concerns:
> 
> 1. can we have a test with `@client` and `@operationGroup`?
> 2. i'm wondering if we need to add a lint to disallow promoting method param to client if it is not in all client's operation?
> 3. with this decorator, is it possible to remove the hard code for promoting `subscriptionId` cause it is not already client parameter in brownfield?

1. Yep, let me add this test
2. I think this is a fair linter-rule to add for unbranded, but it's complicated because of ARM's behavior. Currently the `list` operations don't need `subscriptionId` but all ARM sdks still have `subscriptionId` as a client-level param
3. Yeah that is down the line, where ARM can auto-apply the `@clientInitialization` decorators for `subscriptionId`

chunyu3 (commented):
When the parameter `blobName` is bumped to client-level, there will no `blobName` in method (SdkMethod), but it will still in Operation (method.Operation). Right?
But we need to know that `blobName` is bumped to client-level, so that when we handle this parameter we will find it in Client. It is better to have a flag to indicate that the operation parameter is moved to client, such as Location="client"

iscai-msft:
@chunyu3 it will have property `.onClient` set to `true`

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3785,4 +3785,282 @@ describe("typespec-client-generator-core: decorators", () => {
...
+      strictEqual(blobNameParam.onClient, true);
+      strictEqual(blobNameParam.optional, false);
+      strictEqual(blobNameParam.kind, "method");
+
+      const blobClient = getBlobClient.response;
+
+      strictEqual(blobClient.kind, "client");
+      strictEqual(blobClient.name, "BlobClient");
+      strictEqual(blobClient.initialization.access, "public");
```
iscai-msft, line 3891: my understanding was that `interface`s mapped to `operation groups` when no decorators were set, but it seems that the initialization is still public. That's totally fine, I jsut want to clarify my understanding @tadelesh. That's why I updated my code so that we only throw an error if there are new client params introduced for ogs explicitly titled with `@operationGroup`

tadelesh: i think whether explicitly set or not, the behavior should be same. why op from interface infer could be public and `@operationGroup` be internal?

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
```diff
@@ -433,6 +433,36 @@ export type UseSystemTextJsonConverterDecorator = (
...
+ *
+ * @@clientInitialization(MyService, MyServiceClientOptions)
+ * // The generated client will have `blobName` on it. We will also
+ * // elevate the existing `blobName` parameter to the client level.
+ * ```
+ */
+export type ClientInitializationDecorator = (
+  context: DecoratorContext,
+  target: Namespace | Interface,
+  options: Model,
```
tadelesh, line 462: is it possible that we have a consolidated decorator signature like `@override` if we want to make this decorator flexible like redefine the initialization?

packages/typespec-client-generator-core/test/decorators.test.ts
iscai-msft: i'm not sure, that's the current behavior though, so I think there's something wrong with current logic. Let me take a look today

weidongxu-microsoft: I kind of agree with Chenjie, that they be same.

A different scenario would be namespace nested under namespace (which is the subclient? -- do we have similar way to specify this with explicit `@client`)
```
namespace Client {
  namespace SubClient {
  }
}
```

docs/libraries/typespec-client-generator-core/reference/decorators.md
```diff
@@ -378,6 +417,46 @@ op myOperationCustomization(params: Params): void;
...
+op upload(blobName: string): void;
+
+// client.tsp
+namespace MyCustomizations;
+model MyServiceClientOptions {
+  blob: string;
+}
+
+@@clientInitialization(MyService, MyServiceClientOptions)
+@@paramAlias(MyServiceClientOptions.blob, "blobName")
```
weidongxu-microsoft, line 454: Does this syntac only cover `blobName` in op (like `op upload(blobName: string)`), or both this and `blob` (e.g. `op download(blob: string`?)

From your test, it appears to be latter?

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3785,4 +3785,334 @@ describe("typespec-client-generator-core: decorators", () => {
...
+      const methods = client.methods;
+      strictEqual(methods.length, 2);
+      const download = methods[0];
+      strictEqual(download.name, "download");
+      strictEqual(download.kind, "basic");
+      strictEqual(download.parameters.length, 1);
+
+      const downloadBlobNameParam = download.parameters.find((x) => x.name === "blobName");
+      ok(downloadBlobNameParam);
+      strictEqual(downloadBlobNameParam.onClient, true);
```
weidongxu-microsoft, line 4106: From this test, it appears that TCGC automatically rename the `blob` in `download` to `blobName`, according to `@paramAlias`?

So emitter would only need to map it by name.

Is it correct understanding?

weidongxu-microsoft (approved):
LGTM, having a few query though.

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3785,4 +3785,334 @@ describe("typespec-client-generator-core: decorators", () => {
...
+        op download(@path blobName: string): void;
+        `,
+        `
+        namespace MyCustomizations;
+
+        model MyClientInitialization {
+          blobName: string;
+        }
+
+        @@clientInitialization(MyService, MyCustomizations.MyClientInitialization);
```
haolingdong-msft, line 3804: So from TCGC's response perspective, there is not concept of `clientInitialization`, emitter just need to process the `SdkParameter` when process each operation. Does my understanding correct?

packages/typespec-client-generator-core/test/decorators.test.ts
weidongxu-microsoft: We may need to be extra careful, if service write e.g.

```
model RequestBody {
  blobName: string;
  otherProperties: ...;
}

op action(@path blob: string, ...RequestBody);
```

packages/typespec-client-generator-core/test/decorators.test.ts
weidongxu-microsoft:
```
    initialization: diagnostics.pipe(getSdkInitializationType(context, client)),
```

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
iscai-msft: I'm not quite sure what you're suggesting, are you suggesting we use `@override` instead of `@clientInitialization`? If so, in some ways they're similar, but I feel since there is no original "client initialization" that is defined in the tsp, using the decorator `@override` wouldn't make sense to users

docs/libraries/typespec-client-generator-core/reference/decorators.md
iscai-msft: Both! It's saying `blob` and `blobName` in paramters refer to the same thing

packages/typespec-client-generator-core/test/decorators.test.ts
iscai-msft: This is because we have a bug where we're still adding some client params to the method signature. That fix has been merged and I'm going to fix the test in this PR. Good catch!

packages/typespec-client-generator-core/test/decorators.test.ts
iscai-msft: @haolingdong-msft the model in `@clientInitialization` gets added as the `initialization` property on the `SdkClientType`. They are treated as method parameters, so we don't have a concept of them being query / body etc: that we can find out in `method.operation` from the params being mapped back to them

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -989,3 +1006,68 @@ export const $useSystemTextJsonConverter: DecoratorFunction = (
...
+  if (hasExplicitClientOrOperationGroup(context)) {
+    isOg = context.program.stateMap(operationGroupKey).get(target) !== undefined;
+  }
+  if (isOg) {
+    if (
+      target.namespace &&
+      context.program.stateMap(clientInitializationKey).get(target.namespace) &&
+      context.program.stateMap(clientInitializationKey).get(target.namespace)[AllScopes] === options
+    ) {
+      setScopedDecoratorData(context, $override, clientInitializationKey, target, options, scope);
```
tadelesh, line 1028: is `$override` a typo?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -989,3 +1006,68 @@ export const $useSystemTextJsonConverter: DecoratorFunction = (
...
+    } else {
+      reportDiagnostic(context.program, {
+        code: "assigning-public-params-to-internal-client",
+        format: { name: target.name },
+        target: context.decoratorTarget,
+      });
+    }
+  }
+  setScopedDecoratorData(context, $override, clientInitializationKey, target, options, scope);
```
tadelesh, line 1037: add `else` to prevent set after diagnostic?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -989,3 +1006,68 @@ export const $useSystemTextJsonConverter: DecoratorFunction = (
...
+  }
+  if (isOg) {
+    if (
+      target.namespace &&
+      context.program.stateMap(clientInitializationKey).get(target.namespace) &&
+      context.program.stateMap(clientInitializationKey).get(target.namespace)[AllScopes] === options
+    ) {
+      setScopedDecoratorData(context, $override, clientInitializationKey, target, options, scope);
+    } else {
+      reportDiagnostic(context.program, {
```
tadelesh, line 1030: for .net, they could support sub client with param, and they will generate such api for face. so, do we really need to add this check?

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -526,7 +546,7 @@ export function getCorrespondingMethodParams(
       );
       return diagnostics.wrap([]);
     }
-    return diagnostics.wrap([context.__subscriptionIdParameter]);
+    return diagnostics.wrap(subId ? [subId] : []);
   }
 
   // to see if the service parameter is a method parameter or a property of a method parameter
```
tadelesh, line 552: from this logic, we could not pop a method's model param's property to client, right?

tadelesh (commented):
can we add a test of redefining the client hierarchy in client.tsp and then add `@clientInitialization` to the new defined client? it seems will work well from the code logic, but i'm not sure if there is any corner case.

tadelesh (approved):
overall lgtm. one thing is could you also check if we could solve this: https://github.com/Azure/typespec-azure/issues/1391 in this pr?

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
```diff
@@ -433,6 +433,67 @@ export type UseSystemTextJsonConverterDecorator = (
...
+ * namespace MyCustomizations;
+ * model MyServiceClientOptions {
+ *   blob: string;
+ * }
+ *
+ * @@clientInitialization(MyService, MyServiceClientOptions)
+ * @@paramAlias(MyServiceClientOptions.blob, "blobName")
+ *
+ * // The generated client will have `blobName` on it. We will also
+ * // elevate the existing `blob` parameter to the client level.
```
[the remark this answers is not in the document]
iscai-msft: yep thanks for the catch!

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -87,13 +92,18 @@ function getScopedDecoratorData(
   return retval[AllScopes]; // in this case it applies to all languages
 }
 
-function listScopedDecoratorData(context: TCGCContext, key: symbol): any[] {
+function listScopedDecoratorData(
+  context: { program: Program },
+  key: symbol,
+  emitterName?: string
+): any[] {
```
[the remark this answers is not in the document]
iscai-msft: let me roll back this change, I needed it at one point, but you're right we don't need to modify this code

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: let me go ahead and get rid of this check. This is explicitly for operation groups, not initializable subclients, but it is making the code more complicated than it needs to be, so let me roll this back

packages/typespec-client-generator-core/src/http.ts
iscai-msft: yes we can, this is returning corresponding method params, which in this case we're returning a client param

PR #1405, opened 2024-08-21
Update CODEOWNERS

mario-guerra:
Adding myself as a code owner so I can stay on top of issues and PRs filed.


markcowl (approved):

timotheeguerin:
just waiting for resolution of dicussion on the same PR for typespec, I don't think this is necsssary and will not necessarly achieve the expected result anwway

PR #1410, opened 2024-08-22
fix(tcgc): set sdk method body parameter encode with http content type

### LINKED ISSUE #1084: [Bug][TCGC]: inconsistent `encode` between SdkMethodParameter and HttpOperationParameter
### Describe the bug

Check this spec: https://github.com/Azure/autorest.csharp/blob/6d5884d0b9b1367f5d36ab648aecab41baccf33b/test/TestProjects/MediaTypes-TypeSpec/MediaTypes-TypeSpec.tsp#L29

The request body is defined as `bytes`.

Parsing the spec, and check the type of both `SdkMethodParameter` and `HttpOperationParamter`, you'll find their `encode` is different. The one in `SdkMethodParameter` is correct (e.g. `base64`).

### SdkMethodParameter
image

### HttpOperationParameter
image

### Reproduction

Just try to parse the definition above.

### Checklist

- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)

archerzz:
- after fetching the sdk method body parameter, reset its encode according to the underlying http request content-type
- add unit test

fix #1084

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -74,6 +74,7 @@ import {
   getLibraryName,
 } from "./public-utils.js";
 import {
+  addEncodeInfo,
   getAllModelsWithDiagnostics,
   getClientTypeWithDiagnostics,
   getSdkCredentialParameter,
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -247,6 +248,19 @@ function getSdkBasicServiceMethod<TServiceOperation extends SdkServiceOperation>
   const serviceOperation = diagnostics.pipe(
     getSdkServiceOperation<TServiceOperation>(context, operation, methodParameters)
   );
+  // set the correct encode for body parameter according to the content-type
```
archerzz, line 251: Another way to set the `encode` is trying to get the http parameter type encode value. That's quite ugly because you have to do a lot of type detection since here we only have `SdkType`.

iscai-msft (approved):

PR #1415, opened 2024-08-22
Add modelAsStruct decorator

JoshLove-msft:
Contributes to https://github.com/microsoft/typespec/issues/3619


m-nash (commented):
Can we add some tests before merging.
This looks good to me, @iscai-msft can you take a look as well?
@srnagar would java ever use something like this and if so does the name make sense?

srnagar:
@m-nash No, Java doesn't have a struct. So, all models are classes. So, we don't have a usecase for this decorator in Java.

packages/typespec-azure-core/lib/decorators.tsp
```diff
@@ -12,6 +12,11 @@ namespace Azure.Core {
    */
   extern dec pagedResult(entity: Model);
 
+  /**
```
iscai-msft, line 15: should this be going in azure core?

JoshLove-msft: Ah I guess it is not at all Azure specific. Would it be better in client-generator-core?

iscai-msft: I would def put it in tcgc then, because we have more control here and we can always elevate it later if need be. @m-nash is this good with you

PR #1418, opened 2024-08-23
[tcgc] don't require api-version in tcgc override if from core

iscai-msft:


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -63,7 +63,7 @@ export function getDefaultApiVersion(
  * @param parameter
  * @returns
  */
-export function isApiVersion(context: TCGCContext, type: { name: string }): boolean {
+export function isApiVersion(type: { name: string }): boolean {
```
iscai-msft, line 66: breaking change: if i can figure out how to better check for core params, then won't need to make this change

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1072,7 +1072,9 @@ function collectParams(
       if (value.type.kind === "Model") {
         collectParams(value.type.properties, params);
       } else {
-        params.push(value);
+        if (!isApiVersion(value)) {
```
iscai-msft, line 1075: ideally, I want to check if this is from a core parameter. This is hard, I tried doing `isAzureCoreModel(value.sourceModel.model)`, but the issue is it seems to be catching even `foo` to be of an `Azure.Core` model. @tadelesh do you have any recs?

tadelesh: `isAzureCoreModel` function miss one logic: it should filter all templated instances model which belong to the service but with core namespace.

iscai-msft: ok let me try it this way today

tadelesh:
@iscai-msft it seems you changed this pr to merge to release branch which includes a lot of main branch commits?

iscai-msft:
@trangevi going to create a new pr for this to main, since I accidentally trashed this PR and we're going to have our main release go out in like a week

PR #1435, opened 2024-08-27
[tcgc] don't require core-introduced params with `@override`

iscai-msft:
With `@override`, we require a 1:1 mapping between the actual parameters and the parameters passed in your method signature override. However, this causes issues with `core`-introduced parameters because of template-related issues. On a higher level though, I think it's fair to not require `core` params in the method signature, since we already hide these parameters in the method signature, and these parameters will still be eventually set to the service

cc @trangevi

### DIFF
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -914,6 +914,11 @@ function collectParams(
           params.push(value);
         } else if (!sourceProp.model) {
           params.push(value);
+        } else {
+          // eslint-disable-next-line no-console
+          console.log(
+            `We are not counting "${sourceProp.name}" as part of a method parameter because it's been added by Azure.Core templates`
+          );
         }
       }
     }
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -59,6 +59,7 @@ import {
   AllScopes,
   clientNameKey,
   getValidApiVersion,
+  isAzureCoreModel,
   parseEmitterName,
 } from "./internal-utils.js";
 import { createStateSymbol, reportDiagnostic } from "./lib.js";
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -905,7 +906,15 @@ function collectParams(
       if (value.type.kind === "Model") {
         collectParams(value.type.properties, params);
       } else {
-        params.push(value);
+        let sourceProp = value;
+        while (sourceProp.sourceProperty) {
+          sourceProp = sourceProp.sourceProperty;
+        }
+        if (sourceProp.model && !isAzureCoreModel(sourceProp.model)) {
```
tadelesh, line 913: for templated type in core lib like [`Page`](https://github.com/Azure/typespec-azure/blob/955fc9534c2f7153d45c9fbe5743baa0c104d803/packages/typespec-azure-core/lib/models.tsp#L17), it will also be a core model judged by `isAzureCoreModel`. i don't know if it will happen for param, but i think we should fix the `isAzureCoreModel`.
```
@useDependency(Versions.v1_0_Preview_2)
@server("http://localhost:3000", "endpoint")
@service()
namespace My.Service;
@resource("widgets")
model Widget {
  @key("widgetName")
  @visibility("read")
  name: string;
  manufacturerId: string;
}

op listWidgets is ResourceList<Widget>;
```

tadelesh (commented):
another concern is for custom traits, do we also want to ignore these params?

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: i'm not able to repro this issue with the test you linked. Since I still don't fully understand the issue and how to repro it, I'm going to go ahead and merge this pr. Do you have another repro link? I can take a look at this tomorrow. Thanks @tadelesh!

srnagar, line 913: Can we add a log when we skip core models? It shouldn't happen often, but when it does, it might confuse the user. So, having a log might be useful.

srnagar (approved):

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: it's fine. since your purpose is just to ignore the parameter from core model spread, the logic should be good. what i'm saying is for `isAzureCoreModel`, sth like `Azure.Core.Page<Widget>` will also be considered as a core model which should be wrong.

tadelesh (approved):

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: I still can't repro it. I've added a test to the pr and it's still just showing a single model, feel free to add to it after I merge this PR

PR #1442, opened 2024-08-28
fix(doc): problematic polling operation for lro delete

archerzz:
Currently in the offical guide, the same polling operation is shared by all lro operations, which could cause problems for some emitters since the lro metadata contains value which doesn't fit for lro delete. See https://github.com/Azure/autorest.csharp/issues/4827

This commit will create a separate polling operation which fits lro delete. The same definition is already available in https://github.com/Azure/typespec-azure/blob/955fc9534c2f7153d45c9fbe5743baa0c104d803/packages/samples/specs/data-plane/trait-versioning/main.tsp#L72-L77


markcowl (approved):

docs/getstarted/azure-core/step06.md
```diff
@@ -28,7 +33,7 @@ interface Widgets {
   createOrUpdateWidget is Operations.LongRunningResourceCreateOrUpdate<Widget>;
```
weidongxu-microsoft, line 33: @markcowl 
Is it OK for us to replace this example with `LongRunningResourceCreateOrReplace`?

As guideline
[⛔](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-no-patch-lro) DO NOT implement PATCH as an LRO. If LRO update semantics are required, implement it using the [LRO POST action pattern](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-existing-resource) .

archerzz: I updated according to Weidong's comment. I can roll back to previous comment if needed.

weidongxu-microsoft (approved):

XiaofeiCao (commented):
Update content below to use `createOrReplace` as well?
https://github.com/Azure/typespec-azure/blob/1c38fb845b1fcb80231063ae64adfb77ea1947ac/docs/getstarted/azure-core/step06.md?plain=1#L39

XiaofeiCao (approved):

markcowl (approved):
OK with the added changes

Issue #1380, opened 2024-08-15
[typespec-azure-core] Add new model `MultiPartFile` with required metadata


msyyc:
typespec/http has native model [`File`](https://github.com/microsoft/typespec/blob/2eac90cd1e33c39d9e3e32e00d00920f7f1d606a/packages/http/lib/http.tsp#L110-L114) with optional metadata `contentType` and `filename` while azure services all ask filename as required. Currently typespec author has to define the model by themselves like:
```
model FileRequiredMeta extends File {
    filename: string
}
```
After discussion, most devs pretend to add the model with required metadata to make it more convenient for typespec author. What's more, SDK could add new Type in core library for this type.

msyyc:
(1) Solution 1: define the model specificly like:
```
model MultiPartFile {
    contents: bytes;
   filename: string;
   contentType: string;
}
```
And we also need to supply a function (e.g. isMultiPartFile) like [isOrExtendsHttpFile](https://github.com/microsoft/typespec/blob/2eac90cd1e33c39d9e3e32e00d00920f7f1d606a/packages/http/src/private.decorators.ts#L60) of `typespec/http`
(2) Solution 2: extends from File of typespec/http like:
```
model FileRequiredMeta extends File {
    filename: string;
    contentType: string;
}
```
With this way, no need new function of solution 1

weidongxu-microsoft:
```
namespace Azure.Core;

model File extends TypeSpec.Http.File {
    filename: string;
}
```

weidongxu-microsoft:
For 1, I assume you need to add the `@Private.httpFile`?

msyyc:
@weidongxu-microsoft If we adopt same name, wouldn't it cause model name conflict? 
````
using Azure.Core;
using Typespec.Http;
```

msyyc:
> For 1, I assume you need to add the `@Private.httpFile`?

I think it is not recommended to use decorator under `Private` namespace so I don't add this private decorator. And that's why solution1 need TCGC provide additional function like [isOrExtendsHttpFile](https://github.com/microsoft/typespec/blob/2eac90cd1e33c39d9e3e32e00d00920f7f1d606a/packages/http/src/private.decorators.ts#L60)

weidongxu-microsoft:
> @weidongxu-microsoft If we adopt same name, wouldn't it cause model name conflict?
> 
> ```
> using Azure.Core;
> using Typespec.Http;
> ```

[...]

PR #1458, opened 2024-09-02
Add multipartfile model in Azure.Core

msyyc:
fix https://github.com/Azure/typespec-azure/issues/1380


packages/typespec-azure-core/lib/models.tsp
```diff
@@ -400,3 +400,9 @@ model ArmResourceIdentifierAllowedResource {
    */
   scopes?: ArmResourceDeploymentScope[];
 }
+
+@doc("Used in file part of multipart request body")
+model MultiPartFile extends File {
+  filename: string;
+  contentType: string;
```
msyyc, line 407: context: https://github.com/Azure/typespec-azure/issues/1380#issue-2467620574

docs/libraries/azure-core/reference/data-types.md
```diff
@@ -220,6 +220,21 @@ model Azure.Core.MaxPageSizeQueryParameter
...
+
+```typespec
+model Azure.Core.MultiPartFile
+```
+
+#### Properties
+
+| Name        | Type     | Description |
+| ----------- | -------- | ----------- |
+| filename    | `string` |             |
```
iscai-msft, line 235: can you add descriptions for these

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
```diff
@@ -454,6 +455,52 @@ describe("typespec-client-generator-core: multipart types", () => {
     strictEqual(fileRequiredFileName.multipartOptions.contentType.optional, false);
   });
 
+  it("with MultiPartFile of Azure.Core", async function () {
```
iscai-msft, line 458: we also want tests in `azure-core` for this

docs/libraries/azure-core/reference/data-types.md
msyyc: added

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
msyyc: added

packages/typespec-azure-core/lib/models.tsp
```diff
@@ -400,3 +400,12 @@ model ArmResourceIdentifierAllowedResource {
    */
   scopes?: ArmResourceDeploymentScope[];
 }
+
+@doc("Used in file part of multipart request body")
+model MultiPartFile extends File {
```
timotheeguerin, line 405: not sure this name makes much sense in describing that everything is required. The fact that it sent over multipart(which is already the main use case of the standard `File`) doesn't imply requireness of filename or contentType

packages/typespec-azure-core/lib/models.tsp
```diff
@@ -400,3 +400,12 @@ model ArmResourceIdentifierAllowedResource {
...
+  @doc("The file name in file part of multipart request body")
+  filename: string;
+
+  @doc("The content type in file part of multipart request body")
```
timotheeguerin, line 409: use doc comments

packages/typespec-azure-core/test/models.test.ts
timotheeguerin, line 1: rename to `types/<whatever-new-name>.test.ts` like we have for `arm-resource-identifier`

packages/typespec-azure-core/test/models.test.ts
```diff
@@ -0,0 +1,30 @@
+import { expectDiagnosticEmpty } from "@typespec/compiler/testing";
+import { getHttpPart, isOrExtendsHttpFile } from "@typespec/http";
+import { ok, strictEqual } from "assert";
+import { describe, it } from "vitest";
+import { getOperations } from "./test-host.js";
+
+describe("typespec-azure-core: models", () => {
```
timotheeguerin, line 7: remove top level describe

packages/typespec-azure-core/test/models.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    ok(operations.length === 1);
+    ok(operations[0].parameters.body);
+    strictEqual(operations[0].parameters.body.bodyKind, "multipart");
+    strictEqual(operations[0].parameters.body.type.kind, "Model");
+    const fileHttpPart = operations[0].parameters.body.type.properties.get("file");
+    ok(fileHttpPart);
+    const file = getHttpPart(runner.program, fileHttpPart.type);
+    ok(file !== undefined);
+    ok(isOrExtendsHttpFile(runner.program, file.type));
+    expectDiagnosticEmpty(diagnostics);
```
timotheeguerin, line 28: this is uncessary

packages/typespec-azure-core/lib/models.tsp
msyyc: I ever thought about some names:
- `FileWithRequiredMeta`: it seems too long
- `FileDetails`: not specific
Any idea?

packages/typespec-azure-core/lib/models.tsp
msyyc: updated

packages/typespec-azure-core/test/models.test.ts
msyyc: updated

packages/typespec-azure-core/test/models.test.ts
msyyc: updated

packages/typespec-azure-core/test/models.test.ts
msyyc: I am not familiar with the test framework but this case calls async function `getOperations` which can not be used directly in `describe`.

timotheeguerin: Just this line and the closing }) keep the content at the root. That was added when we were using mocha and all tests are merged together so needed some way to have a unique top level. But vitest already scope per file

packages/typespec-azure-core/lib/models.tsp
iscai-msft: `FileWithRequiredMetadata` seems ok to me, but yeah it's definitely a balance between being too long and being descriptive enough.

`RequiredFile`? tbh I don't love that

timotheeguerin: `ExplicitFile`?
`FileWithRequiredMetadata` seems ok as well

msyyc: Updated with `FileWithRequiredMetadata`

packages/typespec-azure-core/test/models.test.ts
msyyc: I see.
Updated.

markcowl (commented):
The Azure.Core part looks good.  Need someone from TCGC to sign off the TCGC changes

packages/typespec-azure-core/lib/models.tsp
weidongxu-microsoft: Why contentType is also required? Are user **required** to fill-in the contentType?

msyyc: `Should we have e.g. contentType: "application/octet-stream", if this must be a required property?`
-> It is OK for Python since the default content-type in RFC is also `application/octect-stream`. @chunyu3 / @MaryGao / @qiaozha what's your opinion?

weidongxu-microsoft: Java probably do not want to handle this `contentType: string` as Python.

Making it "required" here basically deprive service dev from actually saying that "contentType" must be filled by user.
If Java add a client default value of "application/octect-stream" in this case of `contentType: string`, how does service dev write in TypeSpec, to have Java not providing this client default value?

weidongxu-microsoft: Java would typically want 

> this contentType must to be sent on wire, but if user does not set anything, SDK would use "application/octet-stream"

I assume this cannot be expressed in TypeSpec without client default value.

But the point, is that I didn't see why the `contentType` property be different from the definition in `TypeSpec.Http.File`. If `contentType` be required here, it should be required in `TypeSpec.Http.File` too.

msyyc: So Java hopes the core model only change `filename` to required.

iscai-msft: this model is explicitly for a file that has more restrictions on it than a normal `File`. It extends from `File` but overrides the properties to be required instead of optional

packages/typespec-azure-core/test/types/models.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  ok(operations.length === 1);
+  ok(operations[0].parameters.body);
+  strictEqual(operations[0].parameters.body.bodyKind, "multipart");
+  strictEqual(operations[0].parameters.body.type.kind, "Model");
+  const fileHttpPart = operations[0].parameters.body.type.properties.get("file");
+  ok(fileHttpPart);
+  const file = getHttpPart(runner.program, fileHttpPart.type);
+  ok(file !== undefined);
+  ok(isOrExtendsHttpFile(runner.program, file.type));
```
iscai-msft, line 25: can you check the optionality of the `contentType` and `filename` properties? They should be required

packages/typespec-azure-core/lib/models.tsp
msyyc: We could add 2 models for customer:
- `FileWithRequiredMetadata`: both filename and contentType are required (already in this PR)
- `FileWithRequiredFileName`: only filename is required

Then azure service team could choose any one which meets their requirements.

packages/typespec-azure-core/test/types/models.test.ts
msyyc: I added the check in TCGC test: https://github.com/Azure/typespec-azure/pull/1458/files#diff-5d0dbc69414fb42ee89e9e13679560f649d290268dca89d259d5567f0174a7ab

packages/typespec-azure-core/lib/models.tsp
weidongxu-microsoft: I'd assume Java azure-core would only add one class.

loop @srnagar for opinion as well.

msyyc: @srnagar What's your opinion about https://github.com/Azure/typespec-azure/pull/1458#discussion_r1753022700?

Issue #1462, opened 2024-09-03
The name of example related types should emphasize the `examples` and `example values`


ArcturusZhang:
For instance, `SdkStringExample` should actually be `SdkStringExampleValue` to reduce confusion.

These `examples` and `example values` category has been in the example types because the example will extend from `SdkExampleBase` and those value types do not.

Issue #933, opened 2024-05-29
`values` makes no sense in `SdkUnionType`


ArcturusZhang:
We have the following definition for union type in TCGC:

```typescript
export interface SdkUnionType extends SdkTypeBase {
	name: string;
	isGeneratedName: string;
	kind: "union";
	values: SdkType[];
}
```

it would make more sense if we change `values` to `valueTypes` or `itemTypes` or `variantTypes` (I prefer `variantTypes`), because the siblings of an union type are types, not values.

tadelesh:
proposal: change `values` to `variants`

ArcturusZhang:
> proposal: change `values` to `variants`

I prefer `variantTypes` to emphasize that these are types instead of the values of the union

Issue #874, opened 2024-05-21
We have `any` in `SdkBuiltinKinds` for the built-in type `unknown` in typespec, why not `unknown`?


ArcturusZhang:
It is not a good practice that we change a name for a builtin type. Is it because the kind `unknown` is ambiguous from the type with a name of `unknown` and the case where there is an invalid kind which we do not know?

iscai-msft:
Yeah that was the original thinking, and because it aligned more closely with what we got from m4. Happy to change this though, it will just be a breaking change that will have to be driven across languages

tadelesh:
proposal: change `any` to `unknown`

PR #1463, opened 2024-09-03
Polish some API names in TCGC

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/874
Fixes https://github.com/Azure/typespec-azure/issues/933
Fixes https://github.com/Azure/typespec-azure/issues/1462
Fixes https://github.com/Azure/typespec-azure/issues/1466


.chronus/changes/change-some-bad-apis-2024-8-3-7-4-12.md
```diff
@@ -0,0 +1,12 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: breaking
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+1. The kind for `unknown` renamed from `any` to `unknown`.
+2. The `values` property in `SdkUnionType` renamed to `variantTypes`.
```
iscai-msft, line 9: `values` is different than `variantTypes` though, shouldn't it correspond to `variants`? Same for tuples

ArcturusZhang: I would like to have the `types` suffix because in other places we have `types` such as in `SdkArrayType` we have `valueType` - therefore I think maybe we should keep consistent

haolingdong-msft (approved):
overall looks good to me.

.chronus/changes/change-some-bad-apis-2024-8-3-7-4-12.md
ArcturusZhang: oh - and it is the type of the variant or in this way: in this context, the variant of the union is a type

tadelesh (approved):

.chronus/changes/change-some-bad-apis-2024-8-3-7-4-12.md
ArcturusZhang: I had an offline discussion with @tadelesh 
so our current naming convention is adding the `type` suffix everywhere, it would be weird if we do not do the same here, for example:
- the type names we have type suffix: `SdkModelType`, `SdkEnumType`, etc
- the properties inside a type if it is referring another type, we have the type suffix, such as the `keyType` `valueType` in `SdkDictionaryType`, `wireType` and `baseType` in `SdkDateTimeType`

So if we do not want the type suffix, we should do it consistently and thoroughly, we might have to remove all of them, which keep consistent with typespec apis: for type names we are adding a `Sdk` as prefix, such as `SdkModel`, `SdkClient`, `SdkEnum`, and those properties we do not need the `type` suffix such as we could have `wire` and `base` in `SdkDateTime`.

I think we could think this through, and if we decide to make this alignment, we could do it in next month's release

iscai-msft: I thought these `variantTypes` were referring to the `variant` values themselves, not the types of the variants. If it's the types of the `variant`s then `variantTypes` is appropriate. I'm not sure we need a total redesign if that's the case, do you agree?

iscai-msft (approved):

.chronus/changes/change-some-bad-apis-2024-8-3-7-4-12.md
ArcturusZhang: maybe we could put that as a future item to discuss.
At least now our types are consistent with each other on their names.
But whether we want to align with the style in typespec compiler (because it would be a big change), we could discuss that later.

ArcturusZhang:
> @ArcturusZhang I think the release just went out for the latest typespec, so this breaking change won't be part of it, which I think is good because a lot of people prefer having the breaking one a release after in tcgc, so we can better integrate it. Can you rebase this onto the release branch? [`main`/CONTRIBUTING.md#make-a-patch-release](https://github.com/Azure/typespec-azure/blob/main/CONTRIBUTING.md?rgh-link-date=2024-09-11T14%3A41%3A12Z#make-a-patch-release)

sure!

PR #1476, opened 2024-09-04
[TCGC] Fix client initialization for `subscriptionId`

msyyc:
- Before fix:
image
- After fix:
image

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -666,7 +666,7 @@ function addDefaultClientParameters<
       // if any sub operation groups have an subId param, the top level needs it as well
       subId = context.__clientToParameters
         .get(operationGroup.type)
-        ?.find((x) => x.isApiVersionParam);
+        ?.find((x) => isSubscriptionId(context, x));
       if (apiVersionParam) break;
     }
   }
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -660,7 +660,9 @@ function addDefaultClientParameters<
   if (apiVersionParam) {
     client.initialization.properties.push(apiVersionParam);
   }
-  let subId = context.__clientToParameters.get(client.__raw.type)?.find((x) => isSubscriptionId(context, x));
+  let subId = context.__clientToParameters
+    .get(client.__raw.type)
+    ?.find((x) => isSubscriptionId(context, x));
   if (!subId && context.arm) {
     for (const operationGroup of listOperationGroups(context, client.__raw)) {
       // if any sub operation groups have an subId param, the top level needs it as well
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -660,7 +660,7 @@ function addDefaultClientParameters<
   if (apiVersionParam) {
     client.initialization.properties.push(apiVersionParam);
   }
-  const subId = client.initialization.properties.find((x) => isSubscriptionId(context, x));
+  const subId = context.__clientToParameters.get(client.__raw.type)?.find((x) => isSubscriptionId(context, x));
   if (subId) {
     client.initialization.properties.push(subId);
   }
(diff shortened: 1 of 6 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -660,7 +660,16 @@ function addDefaultClientParameters<
...
+  let subId = context.__clientToParameters.get(client.__raw.type)?.find((x) => isSubscriptionId(context, x));
+  if (!subId && context.arm) {
+    for (const operationGroup of listOperationGroups(context, client.__raw)) {
+      // if any sub operation groups have an subId param, the top level needs it as well
+      subId = context.__clientToParameters
+        .get(operationGroup.type)
+        ?.find((x) => x.isApiVersionParam);
+      if (apiVersionParam) break;
+    }
+  }
```
msyyc, line 672: `subscriptionId` is special parameter in ARM SDK and we shall handle it with similar logic as `api-verison`.

msyyc: updated

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2176,6 +2178,41 @@ describe("typespec-client-generator-core: package", () => {
       strictEqual(apiVersionOpParam.correspondingMethodParams[0], apiVersionClientParam);
     });
 
+    it("global parameters", async () => {
```
weidongxu-microsoft, line 2181: nit, "subscriptionId client parameter" or "subscriptionId client property" ?

tadelesh (approved):

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -660,7 +660,18 @@ function addDefaultClientParameters<
   if (apiVersionParam) {
     client.initialization.properties.push(apiVersionParam);
   }
-  const subId = client.initialization.properties.find((x) => isSubscriptionId(context, x));
+  let subId = context.__clientToParameters
+    .get(client.__raw.type)
+    ?.find((x) => isSubscriptionId(context, x));
```
weidongxu-microsoft, line 665: I assume this part of the code checks whether "subscriptionId" is already added to client parameters?
Or, is it checks the "subscriptionId" on operations that does not be included in any operation group?

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/test/package.test.ts
msyyc: This case could cover all the client parameter of ARM sdk, including `credentail/endpoint/subscriptionId/apiVersion`. Maybe `client level signatures` is better.

packages/typespec-client-generator-core/src/package.ts
msyyc: Since we will update the logic in the future, I just keep same with api-verison: https://github.com/Azure/typespec-azure/blob/85d0f982f16cfb6ac44a85c989db17a7ae948064/packages/typespec-client-generator-core/src/package.ts#L647.

packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2176,6 +2178,41 @@ describe("typespec-client-generator-core: package", () => {
...
+        interface MyInterface {
+          get is ArmResourceRead<MyModel>;
+        }
+      `);
+
+      const sdkPackage = runnerWithArm.context.sdkPackage;
+      const client = sdkPackage.clients[0].methods.find((x) => x.kind === "clientaccessor")
+        ?.response as SdkClientType<SdkHttpOperation>;
+      for (const name of ["apiVersion", "subscriptionId", "endpoint", "credential"]) {
+        ok(client.initialization.properties.find((x) => x.name === name) !== undefined);
```
haolingdong-msft, line 2212: double confirm, those parameters would appear in `SdkHttpOperation.parameters` with `onClient === true`, right?

msyyc: Yeah

msyyc: Yeah, I added test for `onClient`

iscai-msft (approved):

iscai-msft:
thanks @msyyc!

Issue #1483, opened 2024-09-05
[Bug]: anonymous model in `HttpPart` template does not have a generated name


tadelesh:
### Describe the bug

See test here: https://github.com/Azure/typespec-azure/blob/85d0f982f16cfb6ac44a85c989db17a7ae948064/packages/typespec-client-generator-core/test/types/multipart-types.test.ts#L578
The model in the `HttpPart` template could not be found when generate naming.

In addition, in order to prevent any models without generated name, we should add a test in `afterEach` to check all the models in `context.models` and `context.enums` have names.

### Reproduction

As description.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1491, opened 2024-09-06
[TCGC] Fix naming logic for anonymous model wrapped by `HttpPart`

msyyc:
fix https://github.com/Azure/typespec-azure/issues/1483
The PR contains:
- Bug fix for empty name of anonymous model wrapped by `HttpPart`
- add `afterEach` in test to make sure the `sdkPackage.models/enums` must have non-empty name.


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -470,17 +470,28 @@ function getContextPath(
...
+      ) {
+        // handle array or dict
+        const dictOrArrayItemType: Type = currentType.indexer.value;
+        return dfsModelProperties(
+          expectedType,
+          dictOrArrayItemType,
+          pluralize.singular(displayName)
+        );
+      }
+
```
msyyc, line 494: The diff is actually not that large and I only add logic to peel off `HttpPart`

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
```diff
@@ -543,6 +554,7 @@ describe("typespec-client-generator-core: multipart types", () => {
     ) as SdkBodyModelPropertyType;
     ok(stringWithContentType);
     strictEqual(stringWithContentType.type.kind, "model");
+    strictEqual(stringWithContentType.type.name, "MultiPartRequestStringWithContentType");
```
haolingdong-msft, line 557: The name looks good to me.

haolingdong-msft (approved):

chunyu3 (approved):

packages/typespec-client-generator-core/test/types/array-types.test.ts
```diff
@@ -9,7 +9,16 @@ describe("typespec-client-generator-core: array types", () => {
   beforeEach(async () => {
     runner = await createSdkTestRunner({ emitterName: "@azure-tools/typespec-java" });
   });
-
+  afterEach(async () => {
+    for (const modelsOrEnums of [
+      runner.context.sdkPackage.models,
+      runner.context.sdkPackage.enums,
+    ]) {
+      for (const item of modelsOrEnums) {
```
iscai-msft, line 17: i don't think we should test it this way. Instead, we should add the test on each test and make sure that the generated name is what we expect it to be. We shouldn't just care that its generated name is not empty

msyyc: Agree that devs shall add check for generated name while devs may forget to add the check so this logic is just last Insurance to avoid similar issues.

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
```diff
@@ -563,6 +575,7 @@ describe("typespec-client-generator-core: multipart types", () => {
     ) as SdkBodyModelPropertyType;
     ok(bytesWithContentType);
     strictEqual(bytesWithContentType.type.kind, "model");
+    strictEqual(bytesWithContentType.type.name, "MultiPartRequestBytesWithContentType");
```
msyyc, line 578: test for anonymous model

packages/typespec-client-generator-core/test/types/array-types.test.ts
msyyc: It is here https://github.com/Azure/typespec-azure/pull/1491/files#r1756100153.

iscai-msft (approved):
Looks good! Do you want to release this before next month's release or can it wait until next month's release. If you want to release it before, we've started using a release branch https://github.com/Azure/typespec-azure/blob/main/CONTRIBUTING.md#make-a-patch-release

msyyc:
> Looks good! Do you want to release this before next month's release or can it wait until next month's release. If you want to release it before, we've started using a release branch https://github.com/Azure/typespec-azure/blob/main/CONTRIBUTING.md#make-a-patch-release

Other languages hope we could release a patch version asap so I will merge this into release branch

PR #1493, opened 2024-09-06
[tcgc] include core models if they're a property of a defined model

iscai-msft:
see [this](https://github.com/Azure/typespec-azure/pull/1493/files#diff-a7f44d943ba981ae14169029c19e0f3b982c3dbb366d080f60cc3861f0919f2fR832) test to learn more about the use case


tadelesh:
@iscai-msft i think we all agree not to do this in tcgc. i'll close it. if i miss anything, feel free to reopen.

Issue #1160, opened 2024-07-12
Adding missing common types in TypeSpec


allenjzhang:
- [ ] managedidentitywithdelegation.json
- [ ] mobo.json
- [ ] networksecurityperimeter.json

allenjzhang:
Should be done asap in the upcoming sprint.

timotheeguerin:
We have issues for the first two already, do you want to combine into this one?

Mobo issue https://github.com/Azure/typespec-azure/issues/249
managedidentitywithdelegation  issue https://github.com/Azure/typespec-azure/issues/944

markcowl:
est: 3

Issue #1358, opened 2024-08-12
[Bug]: missing-x-ms-identifiers rule produces false positives on 'extends' properties


darkoa-msft:
### Describe the bug

`missing-x-ms-identifiers` rule does not check base model for the existence of the identifier.

### Reproduction

```
model UpdateFile extends UpdateFileBase {
  @OpenAPI.extension("x-ms-identifiers", [])
  relatedFiles?: UpdateFileBase[];
}

model UpdateFileBase {
  fileName: string;
}

@OpenAPI.extension("x-ms-identifiers", ["fileName"])
  files?: UpdateFile[];
```

`@azure-tools/typespec-azure-resource-manager/missing-x-ms-identifiers: Property "fileName" is not found in "UpdateFile". Make sure value of x-ms-identifiers extension are valid property name of the array element.`

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1505, opened 2024-09-10
Add networksecurityperimeter and managedidentitywithdelegation common types 

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1160


packages/samples/common-types/openapi/v5/networksecurityperimeter.json
```diff
@@ -0,0 +1,518 @@
...
+          "description": "Address prefixes in the CIDR format for inbound rules",
+          "items": {
+            "type": "string"
+          }
+        },
+        "subscriptions": {
+          "type": "array",
+          "description": "Subscriptions for inbound rules",
+          "items": {
+            "$ref": "#/definitions/AccessRulePropertiesSubscriptionsItem"
```
timotheeguerin, line 67: This one I think used to be inline
image

maybe better if for now we can keep things structure(not introduce new definitions) as they were unless its not possible

packages/samples/common-types/openapi/v5/networksecurityperimeter.json
```diff
@@ -0,0 +1,518 @@
...
+      "description": "Information about a network security perimeter (NSP)",
+      "properties": {
+        "id": {
+          "type": "string",
+          "format": "arm-id",
+          "description": "Fully qualified Azure resource ID of the NSP resource",
+          "x-ms-arm-id-details": {
+            "allowedResources": [
+              {
+                "type": "Microsoft.Network/networkSecurityPerimeters"
```
timotheeguerin, line 190: There is a diff here 
image
good to investigate which one is correct

packages/samples/common-types/openapi/v5/networksecurityperimeter.json
```diff
@@ -0,0 +1,518 @@
...
+      "type": "object",
+      "description": "Result of a list NSP (network security perimeter) configurations request.",
+      "properties": {
+        "value": {
+          "type": "array",
+          "description": "Array of network security perimeter results.",
+          "items": {
+            "$ref": "#/definitions/NetworkSecurityPerimeterConfiguration"
+          },
+          "x-ms-identifiers": []
```
timotheeguerin, line 234: This wasn't there before
image

packages/samples/common-types/openapi/v5/networksecurityperimeter.json
AlitzelMendez: type is the correct one ([AutoRest Extensions for OpenAPI 2.0 | autorest (azure.github.io)](https://azure.github.io/autorest/extensions/#x-ms-arm-id-details)) thanks for your help!!!

packages/samples/common-types/openapi/v5/networksecurityperimeter.json
AlitzelMendez: I tried to address this one but I am unsure if it is a problem on my definition or when the generation of the common types is happening.

For what I saw, "x-ms-identifiers" is included when there is not an id for the array, in this case there is an id coming from resource: https://github.com/Azure/typespec-azure/blob/80f55767e0ff73699de4cba05b754529f8a5fc14/packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp#L170 (is not directly in there but it is pulled) even if I tried to add an id, I am not able to do so because it is already inherited from the other model, but when the common-type is getting generated it includes the x-ms-identifiers like it didn't have the id, am I missing something?

markcowl: I think this is related to this bug: https://github.com/Azure/typespec-azure/issues/1358  as the same logic is likely to be used

If so, we probably want to fix this before we merge this PR

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
```diff
@@ -0,0 +1,269 @@
...
+using TypeSpec.Versioning;
+using Azure.Core;
+using OpenAPI;
+
+@info({
+  title: "Network security perimeter common type definitions",
+  description: "Common types for network security perimeters based on a shared API specification. These common, versioned type definitions are intended for resource providers (RPs, except Network RP) to use, and reuse, for defining their own API versions that share a set of type definitions that is consistent across providers.",
+  contact: {},
+})
+@server(
```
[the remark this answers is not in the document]
markcowl: Do we need to add this to other common-types definitions for correct swagger geneeration?

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
```diff
@@ -0,0 +1,269 @@
...
+
+  /** Learning access mode - traffic to the resource is enabled for analysis but not blocked */
+  Learning: "Learning",
+
+  /** Audit access mode - traffic to the resource that fails access checks is logged but not blocked */
+  Audit: "Audit",
+}
+
+/** The type of identity that created the resource. */
+@added(Versions.v5)
```
markcowl, line 84: I don't see this in the common-types in the spec repo

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
```diff
@@ -0,0 +1,269 @@
...
+
+/** Network security configuration properties. */
+@added(Versions.v5)
+model NetworkSecurityPerimeterConfigurationProperties {
+  @visibility("read")
+  provisioningState?: NetworkSecurityPerimeterConfigurationProvisioningState;
+
+  /** List of provisioning issues, if any */
+  @visibility("read")
+  @OpenAPI.extension("x-ms-identifiers", [])
```
markcowl, line 176:
```suggestion
  @OpenAPI.extension("x-ms-identifiers", ["name"])
```

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
```diff
@@ -0,0 +1,269 @@
...
+
+  /** The link used to get the next page of results. */
+  nextLink?: url;
+}
+
+/**
+ * The default NetworkSecurityPerimeterConfigurationName.
+ */
+@added(Versions.v5)
+model NetworkSecurityPerimeterConfigurationNameParameter {
```
markcowl, line 263: We need to decide how to handle the segment name in this parameter model.  We might wanto to do this: https://github.com/Azure/typespec-azure/issues/1358

Or there may be a static value that is always be used.  This depends on usage inside specs

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
```diff
@@ -0,0 +1,269 @@
...
+  /** The name for a network security perimeter configuration */
+  @path
```
markcowl, line 265: Are the common-types decoration (-ref.tsp) files being added later?

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
AlitzelMendez: I think this one is empty: https://github.com/Azure/azure-rest-api-specs/blob/34f5146bc945549d087d38a8a593c0a5f475ad7f/specification/common-types/resource-management/v5/networksecurityperimeter.json#L107

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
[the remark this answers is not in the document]
AlitzelMendez: I don't think we need it, I added it to have the more similar generation to this: https://github.com/Azure/typespec-azure/blob/c469c8b695f2d5c412c2b9998ba5a14badac9e51/packages/samples/common-types/openapi/v5/networksecurityperimeter.json

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
AlitzelMendez: Mark, I think that is the link for the bug that you mentioned above, could you please share the link for the suggestion?

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
markcowl: Yeah, it probably should not be, but we should match

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
markcowl: Yes, I meant a template, like this: https://github.com/Azure/typespec-azure/blob/95cc6b95355aac89280f39bc4b5b9ae94f6ae16d/packages/typespec-azure-resource-manager/lib/common-types/private-links.tsp#L117-L127

markcowl (changes requested):

markcowl (approved):
Approved pending resolution of the two remaining issues

packages/typespec-azure-resource-manager/lib/common-types/network-security-perimeter.tsp
```diff
@@ -0,0 +1,260 @@
+using TypeSpec.Http;
+using TypeSpec.Versioning;
+using Azure.Core;
+using OpenAPI;
+
+@info({
```
timotheeguerin, line 6: should also remove info here, and augment it in the samples file instead

PR #1509, opened 2024-09-10
fix-model-properties-filter-metadata

qiaozha:


tadelesh:
it is by design not to remove metadata property in model because we believe it is either filtered by http lib or is specific defined with `@bodyRoot`.

markcowl:
closing as stale

Issue #1510, opened 2024-09-11
[Bug]: TCGC reports wrong diagnostics "conflicting-multipart-model-usage" for valid multipart cases


msyyc:
### Describe the bug

For the valid cases, TCGC reports wrong diagnostics "conflicting-multipart-model-usage"

### Reproduction

        @service({title: "Test Service"}) namespace TestService;

        model MultiPartRequest {
          id: string;
          profileImage: bytes;
        }
        
        @post op basic1(@header contentType: "multipart/form-data", @body body: MultiPartRequest): NoContentResponse;
        @put op basic2(@header contentType: "multipart/form-data", @body body: MultiPartRequest): NoContentResponse;

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

msyyc:
The root cause is that TCGC always set a model with `UsageFlags.Input`: https://github.com/Azure/typespec-azure/blob/a6365a1770d07b763f5ae9805374cda11085e180/packages/typespec-client-generator-core/src/types.ts#L1658 so it is not proper to distinguish whether a model is used both in common content-type and multipart content-type: https://github.com/Azure/typespec-azure/blob/a6365a1770d07b763f5ae9805374cda11085e180/packages/typespec-client-generator-core/src/types.ts#L1633-L1642.

There are 2 solutions:
- make usage `Input` and `MultipartFormData` excluded
- add new usage like `NonMultiPart`

msyyc:
> Personal opinion: Make `MultipartFormData` exclusive to `JSON`/`XML`, as all of these relates to the serialization of the model.

After discussion, we pretend to use this solution instead of adding new usage kind. Let me make a PR to fix it later.

PR #1511, opened 2024-09-11
[TCGC] Fix judge logic how to distinguish model is referred both in common content-type and multipart content-type

msyyc:
fix https://github.com/Azure/typespec-azure/issues/1510

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -1658,7 +1658,7 @@ function updateTypesFromOperation(
     const access = getAccessOverride(context, operation) ?? "public";
     diagnostics.pipe(updateUsageOrAccessOfModel(context, access, sdkType));
 
-    // after complete of usage calculation for httpBody, check whether it has
+    // after completion of usage calculation for httpBody, check whether it has
     // conflicting usage between multipart and regular body
     if (
       sdkType.kind === "model" &&
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -1331,8 +1331,8 @@ function updateMultiPartInfo(
         : undefined,
       contentType: httpOperationPart.body.contentTypeProperty
         ? diagnostics.pipe(
-          getSdkModelPropertyType(context, httpOperationPart.body.contentTypeProperty, operation),
-        )
+            getSdkModelPropertyType(context, httpOperationPart.body.contentTypeProperty, operation),
+          )
         : undefined,
       defaultContentTypes: httpOperationPart.body.contentTypes,
     };
@@ -1662,8 +1662,9 @@ function updateTypesFromOperation(
     // conflicting usage between multipart and regular body
     if (sdkType.kind === "model") {
       const isUsedInMultipart = (sdkType.usage & UsageFlags.MultipartFormData) > 0;
-      const isUsedInOthers = ((sdkType.usage & UsageFlags.Json) | (sdkType.usage & UsageFlags.Xml)) > 0;
-      if (!multipartOperation && isUsedInMultipart || multipartOperation && isUsedInOthers) {
+      const isUsedInOthers =
+        ((sdkType.usage & UsageFlags.Json) | (sdkType.usage & UsageFlags.Xml)) > 0;
+      if ((!multipartOperation && isUsedInMultipart) || (multipartOperation && isUsedInOthers)) {
         // This means we have a model that is used both for formdata input and for regular body input
         diagnostics.add(
           createDiagnostic({
@@ -1672,7 +1673,7 @@ function updateTypesFromOperation(
             format: {
               modelName: sdkType.name,
             },
-          })
+          }),
         );
       }
     }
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1677,6 +1657,27 @@ function updateTypesFromOperation(
...
     diagnostics.pipe(updateUsageOrAccessOfModel(context, access, sdkType));
+
+    // after complete of usage calculation for httpBody, check whether it has
+    // conflicting usage between multipart and regular body
+    if (
+      sdkType.kind === "model" &&
+      ((!multipartOperation && (sdkType.usage & UsageFlags.MultipartFormData) > 0) ||
+        (multipartOperation &&
+          (sdkType.usage & UsageFlags.MultipartFormData) > 0 &&
+          ((sdkType.usage & UsageFlags.Json) | (sdkType.usage & UsageFlags.Xml)) > 0))
```
msyyc, line 1668: As discussed in https://github.com/Azure/typespec-azure/issues/1510#issuecomment-2345259187, we shall usage existing usage `xml/Json` to distinguish whether a model has conflicting usage between multipart operation and regular operation.
And we shall move the logic position after complete of usage calculation.

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
```diff
@@ -61,6 +61,22 @@ describe("typespec-client-generator-core: multipart types", () => {
...
   });
+  it("multipart conflicting model usage for only multipart operations", async function () {
+    await runner.compile(
+      `
+        @service({title: "Test Service"}) namespace TestService;
+
+        model MultiPartRequest {
+          id: string;
+          profileImage: bytes;
+        }
```
weidongxu-microsoft, line 72: Should we add an `Address` model and verify usage of it is JSON and not MultipartFormData?

Also suggest we can verify the `MultiPartRequest` model be MultipartFormData and not JSON.

BTW, why to have a PUT and a POST in op?

msyyc: _Should we add an Address model and verify usage of it is JSON and not MultipartFormData?
Also suggest we can verify the MultiPartRequest model be MultipartFormData and not JSON._

-> Updated test. And I also add another test for mixed operation scenarios

msyyc: _BTW, why to have a PUT and a POST in op?_
-> With different method, there is no need to add route to distinguish them. Maybe it caused confusion, I have updated the test.

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
```diff
@@ -61,6 +61,68 @@ describe("typespec-client-generator-core: multipart types", () => {
...
+    deepEqual(multiPartRequest.usage, UsageFlags.MultipartFormData | UsageFlags.Input);
+  });
+  it("multipart conflicting model usage for mixed operations", async function () {
+    await runner.compile(
+      `
+        @service({title: "Test Service"}) namespace TestService;
+        model Address {city: string;}
+        model RegularRequest {
+          address: Address;
+        }
```
weidongxu-microsoft, line 100: nit, does Address still have usage=JSON, if there is no RegularReqeust?

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
```diff
@@ -61,6 +61,68 @@ describe("typespec-client-generator-core: multipart types", () => {
...
+        op basic1(@header contentType: "multipart/form-data", @body body: MultiPartRequest): NoContentResponse;
+        @post
+        @route("/basic2") 
+        op basic2(@header contentType: "multipart/form-data", @body body: MultiPartRequest): NoContentResponse;
+      `
+    );
+    deepEqual(runner.context.diagnostics.length, 0);
+    const address = runner.context.sdkPackage.models.find((x) => x.name === "Address");
+    ok(address);
+    deepEqual(address.usage, UsageFlags.Input);
```
msyyc, line 86: Since `usage.Json` is calculated recursively from regular body. If it is not used in regular body, the model usage will not be assigned to `Json`.

packages/typespec-client-generator-core/test/types/multipart-types.test.ts
msyyc: No. You can see https://github.com/Azure/typespec-azure/pull/1511/files#r1758270473

packages/typespec-client-generator-core/src/types.ts
iscai-msft: can you clean up this logic so it's easier to read? i.e. can you make vars like `isXml = sdkType.usage & UsageFlags.Xml > 0` because right now it's really hard to follow what the logic is trying to do here

tadelesh: @iscai-msft i've refined the logic, could you review again and help to merge?

iscai-msft: do we want this to merge into the release branch or into the main branch?

iscai-msft (approved):

packages/typespec-client-generator-core/src/types.ts
tadelesh: main is ok.

PR #1512, opened 2024-09-11
Replace `x-ms-skip-url-encoding` with `allowReserved`

live1206:
This is blocking a new RP [kubernetesruntime](https://github.com/Azure/azure-rest-api-specs/blob/f5321f9b29083f9ea4c028e7484504875e04a758/specification/kubernetesruntime/KubernetesRuntime.Management/storageclass.tsp#L273) generating SDK.

### DIFF
--- a/.chronus/changes/allow-reserved-2024-8-11-16-40-29.md
+++ b/.chronus/changes/allow-reserved-2024-8-11-16-40-29.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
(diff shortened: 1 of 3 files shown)


pshao25 (approved):
LGTM. I'm just not sure if we should discourage the usage of `x-ms-skip-url-encoding` since we have already `allowReserved`.

live1206:
> LGTM. I'm just not sure if we should discourage the usage of `x-ms-skip-url-encoding` since we have already `allowReserved`.

I read somewhere before, we would like to remove all `@extension` from resource-manager tempaltes.

PR #1515, opened 2024-09-11
feat(tcgc): add `SdkLroServiceMetadata`

### LINKED ISSUE #1236: [tcgc] lro info in tcgc interface
currently, if an operation is lro, then the service method's response in tcgc will be the logic result of the operation, and `resultPath` will show the `finalResultPath` of the lro metadata. from emitter's feedback, the envelop result is also usable in sdk code. so, we need to add more info to current tcgc interface.

@haolingdong-msft @archerzz @qiaozha could you help to list all the info you used from `getLroMetadata`? thanks

archerzz:
- add `SdkLroServiceMetadata` and corresponding LRO metadata
- update LRO metadata initialization lgoic
- add test cases

resolve #1236

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -627,6 +627,10 @@ export interface SdkPagingServiceMethod<TServiceOperation extends SdkServiceOper
 }
 
 interface SdkLroServiceMethodOptions {
+  /**
+   * @deprecated This property will be removed in future releases. Use `lroMetadata` for synthesized LRO metadata. If you still want to access primitive LRO info, use `lroMetadata.__raw`.
+   */
+  __raw_lro_metadata: LroMetadata;
   lroMetadata: SdkLroServiceMetadata;
 }
 
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -1,4 +1,8 @@
-import { LroMetadata, PagedResultMetadata } from "@azure-tools/typespec-azure-core";
+import {
+  FinalStateValue,
+  LroMetadata,
+  PagedResultMetadata,
+} from "@azure-tools/typespec-azure-core";
 import {
   DateTimeKnownEncoding,
   Diagnostic,
@@ -623,7 +627,56 @@ export interface SdkPagingServiceMethod<TServiceOperation extends SdkServiceOper
 }
 
 interface SdkLroServiceMethodOptions {
-  __raw_lro_metadata: LroMetadata;
+  lroMetadata: SdkLroServiceMetadata;
+}
+
+/**
+ * Long running operation metadata.
+ */
+export interface SdkLroServiceMetadata {
+  /** LRO metadata from TypeSpec core library */
+  __raw: LroMetadata;
+
+  /** Legacy `finalStateVia` value */
+  finalStateVia: FinalStateValue;
+  /** Polling step metadata */
+  pollingStep: SdkLroServicePollingStep;
+  /** Final step metadata */
+  finalStep?: SdkLroServiceFinalStep;
+  /** Synthesized final response metadata */
+  finalResponse?: SdkLroServiceFinalResponse;
+}
+
+/**
+ * Long running operation polling step metadata.
+ */
+export interface SdkLroServicePollingStep {
+  /** Response body type */
+  responseBody?: SdkModelType;
+}
+
+/**
+ * Long running operation final step metadata.
+ */
+export interface SdkLroServiceFinalStep {
+  /** Final step kind */
+  kind:
+    | "finalOperationLink"
+    | "finalOperationReference"
+    | "pollingSuccessProperty"
+    | "noPollingResult";
+}
+
+/**
+ * Synthesized long running operation response metadata.
+ */
+export interface SdkLroServiceFinalResponse {
+  /** Intact response type */
+  envelopeResult: SdkModelType;
+  /** Meaningful result type */
+  result: SdkModelType;
+  /** Property path to fetch {result} from {envelopeResult}. Note that this property is available only in some LRO patterns. */
+  resultPath?: string;
 }
 
 export interface SdkLroServiceMethod<TServiceOperation extends SdkServiceOperation>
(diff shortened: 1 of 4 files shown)


packages/typespec-client-generator-core/test/methods/lro.test.ts
```diff
@@ -0,0 +1,554 @@
...
+      );
+
+      const metadata = (method as SdkLroServiceMethod<SdkHttpOperation>).lroMetadata;
+      ok(metadata);
+      strictEqual(metadata.finalStateVia, FinalStateValue.azureAsyncOperation);
+      assert.isUndefined(metadata.finalStep);
+
+      // ARM LRO core types are different
+      // const pollingModel = runner.context.sdkPackage.models.find(
+      //   (m) => m.name === "ArmOperationStatusResourceProvisioningState"
```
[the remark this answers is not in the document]
iscai-msft: hmm ok thanks @archerzz can you open an issue for this?

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -154,35 +156,63 @@ function getSdkLroServiceMethod<TServiceOperation extends SdkServiceOperation>(
...
-  }
+  basicServiceMethod.response.type = metadata.finalResponse?.result;
 
-  basicServiceMethod.response.resultPath = metadata.finalResultPath;
+  basicServiceMethod.response.resultPath = metadata.finalResponse?.resultPath;
 
   return diagnostics.wrap({
     ...basicServiceMethod,
     kind: "lro",
-    __raw_lro_metadata: metadata,
```
weidongxu-microsoft, line 175: Should we deprecate this first, instead of removal (and I didn't see type change on this)?

This would relate to nightly build. If this is removed, we probably not able to get it pass till next TCGC release.

packages/typespec-client-generator-core/test/methods/lro.test.ts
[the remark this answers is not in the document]
archerzz: > hmm ok thanks @archerzz can you open an issue for this?

@iscai-msft sure, it's https://github.com/Azure/typespec-azure/pull/1515#discussion_r1754798220

packages/typespec-client-generator-core/src/package.ts
archerzz: added back `__raw_lro_metadata` with `@deprecated` in comments.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -623,7 +627,60 @@ export interface SdkPagingServiceMethod<TServiceOperation extends SdkServiceOper
...
+  /** Response body type */
+  responseBody?: SdkModelType;
+}
+
+/**
+ * Long running operation final step metadata.
+ */
+export interface SdkLroServiceFinalStep {
+  /** Final step kind */
+  kind:
```
haolingdong-msft, line 667: Would it be better to make it as enum instead of interface with `kind`? Because in TCGC's type convention, seems `kind` is used for describing a type, like `SdkType`, but I don't see any interface extends this `SdkLroServiceFinalStep`.

Also could we add some document to say what does each kind of final step mean, and what does corresponding `finalResponse` look like?

archerzz: I tend to keep it simple here because no one is really using its value and I'm not satified with original `LroMetadata` so I try to avoid copying its properties. I added `kind` just to make the struct not empty. `kind` is from weidong's suggestion. This is a copy of the core library `kind` which is a string: https://github.com/Azure/typespec-azure/blob/dfeee05c8fe3ec61ff28daeb0db59942ed08f8fe/packages/typespec-azure-core/src/lro-helpers.ts#L183

archerzz: We can gradually improve `SdkLroServiceMetadata`, per people's demand.

weidongxu-microsoft: It is actually me that suggested to make it a type (that can have more properties later), than an enum.

This is in case that we need more info in finalStep, in future.
But, yes, it probably would involve some break anyway (e.g. if we later make the type polymorphic).

packages/typespec-client-generator-core/test/methods/lro.test.ts
```diff
@@ -0,0 +1,554 @@
...
+        model Employee is TrackedResource<EmployeeProperties> {
+          ...ResourceNameParameter<Employee>;
+        }
+
+        model EmployeeProperties {
+          /** Age of employee */
+          age?: int32;
+        }
+
+        #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes"
```
weidongxu-microsoft, line 522: nit, does this suppress required when using ARM template?

packages/typespec-client-generator-core/test/methods/lro.test.ts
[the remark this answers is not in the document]
weidongxu-microsoft: At present, Java did not use the poll response from ARM.

One reason is the model be different between poling via "location" header and polling via "azure-async-operation" header. And almost no one know which pattern is used on any particular ARM LRO.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/test/methods/lro.test.ts
archerzz: I actually copied from here: https://github.com/Azure/typespec-azure/blob/dfeee05c8fe3ec61ff28daeb0db59942ed08f8fe/packages/samples/specs/resource-manager/operations/nobody-action/main.tsp#L85-L86

Anyway, I deleted all `#suppress` and the test cases are OK. Not sure if all those `#suppress` are unnecessary or the test framework has suppressed all warnings by default. @XiaofeiCao FYI

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: I would say if it's a type in the original definition instead of an enum, I lean towards keeping it a type

haolingdong-msft (approved):

Issue #875, opened 2024-05-21
`encode` in `SdkBuiltinType` should be optional or put sentinel values when an encoding is absent for the type


ArcturusZhang:
Potentially every built-in type could have an encoding (such as digits we would like to keep in payload for numbers), but it is not good practice that a type without encoding will always have a `encode` property with values.

iscai-msft:
@ArcturusZhang is this issue now fixed with your updates to sdk built in types?

tadelesh:
@iscai-msft are you agree to make encode optional in tcgc?

iscai-msft:
I think it should always be required based off of how we're currently utilizing `encode`: if there's no specifc rfc etc, we will use `string`.

We could do an overhaul where `encoding` is just a narrow list of encodings, which would be more accurate. However, I think this might be more pain than it's worth. Let me add this to the list to discuss for our next cross-language sync. If we all agree to have this breaking change, I think we can go ahead with it

PR #1539, opened 2024-09-14
[TCGC] Change `encode` to optional and no longer use `kind` as its fallback value if there is no encode

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/875


weidongxu-microsoft (approved):

haolingdong-msft (approved):
LGTM

Issue #1453, opened 2024-08-30
consider changing the type of `responses` from `Map<K,V>` to an array of responses


ArcturusZhang:
`JSON.stringify` does not support `Map<K, V>`, maybe because the key of a map may not be a string, but the key of JSON is required to be a string. This makes the object with `Map` property not persistable - especially for such language generators that need to pass information from emitter to the generator.

How about we change the type of `responses` from `Map<HttpStatusCodeRange | number, SdkHttpResponse>` to `SdkHttpResponse[]`, and put the `HttpStatusCodeRange | number` as a new property in `SdkHttpResponse`?

ArcturusZhang:
If it is OK maybe we could do it recently and make it in the next release.
This breaking change is only an API breaking, and its migration should be intuitive.

iscai-msft:
sounds good @ArcturusZhang can you drive this? Thank you!

ArcturusZhang:
has been fixed by #1540

PR #1540, opened 2024-09-14
[TCGC] Change the `responses` and `exceptions` to be arrays

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/1453

This is a breaking change in TCGC, but there are some reasons of making this change:
1. The `HttpOperation.responses` from typespec core library also defines as an array. 
2. The type `Map` is not persistable in typescript, which causes issues in some emitter implementation such as dotnet, because it needs to serialize everything in TCGC output and sends it to the generator written in other language (details [here](https://github.com/Azure/autorest.csharp/issues/5028))
3. In most cases, we have no more than 4 responses corresponding different status codes. In the majority of cases, we have 2 or 1. In such a small collections, maps do not have distinguishable performance improvement comparing with arrays with `find`, and arrays have the advantages above.

### DIFF
--- a/.chronus/changes/change-response-to-be-array-2024-8-14-11-27-13.md
+++ b/.chronus/changes/change-response-to-be-array-2024-8-14-11-27-13.md
@@ -4,6 +4,19 @@ packages:
   - "@azure-tools/typespec-client-generator-core"
 ---
 
-1. The type of `responses` and `exceptions` in `SdkHttpOperation` changed from `Map` to array.
-2. The type of `responses` in `SdkHttpOperationExample` changed from `Map` to array.
+1. The type of `responses` and `exceptions` in `SdkHttpOperation` changed from `Map<number | HttpStatusCodeRange | "*", SdkHttpResponse>` to `SdkHttpResponse[]`.
+2. The type of `responses` in `SdkHttpOperationExample` changed from `Map<number, SdkHttpResponseExampleValue>` to `SdkHttpResponseExampleValue[]`.
 3. `SdkHttpResponse` adds a new property `statusCodes` to store its corresponding status code or status code range.
+
+Migration hints:
+The type changed from map to array, and the key of the map is moved as a new property of the value type. For example, for code like this:
+```
+for (const [statusCodes, response] of operation.responses)
+```
+you could do the same in this way:
+```
+for (const response of operation.responses)
+{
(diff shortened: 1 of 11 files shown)


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -550,6 +550,7 @@ export interface SdkHttpResponse extends SdkServiceResponse {
   contentTypes?: string[];
   defaultContentType?: string;
   description?: string;
+  statusCodes: number | HttpStatusCodeRange | "*";
```
lirenhe, line 553: Is there any documentation update needed for the public interface change?

ArcturusZhang: updated

iscai-msft (approved):
thanks @ArcturusZhang

.chronus/changes/change-response-to-be-array-2024-8-14-11-27-13.md
```diff
@@ -0,0 +1,9 @@
+---
+changeKind: breaking
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+1. The type of `responses` and `exceptions` in `SdkHttpOperation` changed from `Map` to array.
+2. The type of `responses` in `SdkHttpOperationExample` changed from `Map` to array.
+3. `SdkHttpResponse` adds a new property `statusCodes` to store its corresponding status code or status code range.
```
tadelesh, line 9: nit: i don't know if chronus could handle markdown.

tadelesh (approved):
since responses are always small collection. i'm ok with this change as long as all languages have consensus.

lirenhe (approved):
It would be good to also provide the migration guide for emitters as this is a breaking change.

MaryGao (approved):
LGTM.

.chronus/changes/change-response-to-be-array-2024-8-14-11-27-13.md
ArcturusZhang: this works as pure text as well I believe

ArcturusZhang: but since the file is in `md`, it makes no sense if it cannot handle it properly - therefore I believe it should be able to handle it

PR #1541, opened 2024-09-14
[TCGC] update references for #1463

ArcturusZhang:
Update references after merged #1463

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -731,7 +731,7 @@ export type SdkExampleValue =
   | SdkUnionExampleValue
   | SdkModelExampleValue;
 
-export interface SdkExampleValueBase {
+interface SdkExampleValueBase {
   kind: string;
   type: SdkType;
   value: unknown;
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -731,7 +731,7 @@ export type SdkExampleValue =
   | SdkUnionExampleValue
   | SdkModelExampleValue;
 
-export interface SdkExampleValueBase {
+interface SdkExampleValueBase {
```
ArcturusZhang, line 734: this is not meant to be exported, others would not have been using, therefore despite this is a breaking but I think this should have no impact.

archerzz (approved):

packages/typespec-client-generator-core/src/interfaces.ts
weidongxu-microsoft: Unsure about this. For Java, the base class must be visible. But this is TS.

weidongxu-microsoft (approved):

msyyc (approved):

PR #1542, opened 2024-09-14
[TCGC] Fix `subscriptionId` for ARM SDK

msyyc:
fix lack of `subscription_id` of ARM SDK: https://github.com/Azure/azure-sdk-for-python/pull/37391/files#diff-6379cca2979bbcd25ca1a5e57303245925f770044d42a36754334a5e22f8e0e4

### DIFF
--- a/packages/typespec-client-generator-core/test/package.test.ts
+++ b/packages/typespec-client-generator-core/test/package.test.ts
@@ -2199,20 +2199,28 @@ describe("typespec-client-generator-core: package", () => {
           name: string;
         }
 
-        @armResourceOperations
-        interface MyInterface {
-          get is ArmResourceRead<MyModel>;
+        namespace MyClient {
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -2199,20 +2199,28 @@ describe("typespec-client-generator-core: package", () => {
...
-        @armResourceOperations
-        interface MyInterface {
-          get is ArmResourceRead<MyModel>;
+        namespace MyClient {
+          interface Operations extends Azure.ResourceManager.Operations {}
+
+          @armResourceOperations
+          interface MyInterface {
+            get is ArmResourceRead<MyModel>;
+          }
```
msyyc, line 2208: Updated test cases to add another operation group which doesn't have `subscriptionId` so that the bug could be catched.

archerzz (approved):

PR #1547, opened 2024-09-16
Default values supporting scalar and object

AlitzelMendez:
Issue: https://github.com/microsoft/typespec/issues/4238

### DIFF
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -1818,12 +1818,7 @@ export async function getOpenAPIForService(
         return defaultType.value.value ?? defaultType.value.name;
       case "ScalarValue":
         if (modelProperty) {
-          return serializeValueAsJson(
-            program,
-            defaultType,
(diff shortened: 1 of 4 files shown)


packages/typespec-autorest/src/openapi.ts
```diff
@@ -1801,7 +1802,7 @@ export async function getOpenAPIForService(
     return getSchemaForType(variant.type, schemaContext)!;
   }
 
-  function getDefaultValue(defaultType: Value): any {
+  function getDefaultValue(defaultType: Value, modelProperty?: ModelProperty): any {
```
timotheeguerin, line 1805: same as in openapi3, I think the whole function content can just be calling to serializeValueAsJson without anything else

AlitzelMendez: I don't think is the case, I replied with more information on the openapi3 comment, but I will reach you out so we can chat about this :)

timotheeguerin (approved):

PR #1560, opened 2024-09-18
[tcgc] Remove deprecated description details

### LINKED ISSUE #1386: [tcgc] remove deprecated `description` and `details`

We've switched to `doc` and `details` (pr [here](https://github.com/Azure/typespec-azure/pull/1379)). After merging this, we want to remove `doc` and `details` in our subsequent release

### Checklist

iscai-msft:
fixes #1386


.chronus/changes/remove_deprecated_description_details-2024-8-18-12-0-24.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: deprecation
```
iscai-msft, line 2: should this be marked deprecation or just bug fix since we deprecated it in a previous version

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -750,7 +710,7 @@ export enum UsageFlags {
 interface SdkExampleBase {
   kind: string;
   name: string;
-  description: string;
+  doc: string
```
iscai-msft, line 713: I made this breaking change so it's `.doc` everywhere, @tadelesh are you ok with that?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -553,7 +521,7 @@ export interface SdkHttpResponse extends SdkServiceResponse {
   kind: "http";
   contentTypes?: string[];
   defaultContentType?: string;
-  description?: string;
+  doc?: string;
```
iscai-msft, line 524: I made this breaking change so it's .doc everywhere, @tadelesh are you ok with that?

tadelesh: the reason i keep it `description` is it is not a direct type doc from typespec, but a property named description from http response in http lib. but i'm open for the change. @haolingdong-msft could you confirm if this change has any impact for java (you introduced it at first).

tadelesh (approved):

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: it's ok to change this.

packages/typespec-client-generator-core/src/interfaces.ts
haolingdong-msft: How does http lib name it? `description` or `doc`? I would prefer to be consistent with http lib's name.

tadelesh: description

iscai-msft: i did `doc` just for consistency in `tcgc`, but yeah I can keep it `description`. Not sure why it's different honestly between the libraries

iscai-msft: let me keep it `description` for consistency with `http` lib for now, we can always change later if tcgc users get too confused

PR #1627, opened 2024-09-28
Adding support for autorest for loading example files from nested sub-folders.

### LINKED ISSUE #1616: [Autorest] auto loading example files from sub-folders and maintains the folder structure in output

Currently example files are loaded only from single level. Need to maintain sub-folder structure. 

### Checklist

allenjzhang:
closes #1616


packages/typespec-autorest/src/openapi.ts
```diff
@@ -133,6 +134,7 @@ import {
   shouldInline,
 } from "@typespec/openapi";
 import { getVersionsForEnum } from "@typespec/versioning";
+import * as path from "path";
```
timotheeguerin, line 137: do not import things from node in emitters

timotheeguerin: all the path function should have equivalent in the compiler

packages/samples/specs/misc/nexted-examples/examples/read/read.json
chrisradek, line 1: Nit: Bit of a pain but I think you meant for these all to be `nested-examples` instead of `nexted-examples`

timotheeguerin: also with the name we do have `x-ms-examples` already as a folder, would be good to consolidate it or use consitent name for both

```
x-ms-examples/flat
x-ms-examples/nested
```

chrisradek (approved):

PR #1629, opened 2024-09-29
[tcgc] do not promote api version param to client if service is not versioned

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1391

cc: @haolingdong-msft @chunyu3


haolingdong-msft (approved):
LGTM

Issue #1608, opened 2024-09-25
[Bug]: fail to extract clientDefaultValue for apiVersion in server 


qiaozha:
### Describe the bug

In the below typespec, 
```ts
@server(
  "{host}.{subdomain}.{sufix}.com/{apiVersion}",
  "Confidential Ledger Service",
  {
    @path
    host?: string = "one",
    @path
    subdomain?: string = "two",
    @path
    sufix?: string = "three",
    @path
    apiVersion?: string = "v1"
  }
)

```
All of those templateArguments have defined default value for it, but only apiVersion failed to extract the default value.

<img width="886" alt="image" src="https://github.com/user-attachments/assets/89b2b21f-c8ae-439a-ad34-2c7c3440c4b8">

### Reproduction

N/A

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Issue #1595, opened 2024-09-24
[Bug]: should we have apiVersion lifted to client initialization if not all operation has this parameter?


qiaozha:
### Describe the bug

In this spec
```
      model ApiVersionParameter {
        @query
        "api-version": string;
      }
      @route("/test")
      op test(...ApiVersionParameter): string;
      @route("/test1")
      op test1(): string;
```

operation test1 doesn't have any parameter call apiVersion at all, but we still have an `api-version` parameter in the client.initialization.properties. with onClient: true and optional: false

![image](https://github.com/user-attachments/assets/b0d71e85-371d-4a56-9585-a2cb2dfd7a8d)

I wonder if this makes sense? 

### Reproduction

N/A 

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

tadelesh:
@iscai-msft could you take a look?

iscai-msft:
yep, i was meaning to take a look today, i'll bump this in priority. thanks!

msyyc:
I think:
-  if there is only one kind of api-version, then TCGC shall lift it into client initialization. It makes sense since  there is only one kind of api-version which will be convenient for SDK users as client level signatures. For those API that has no parameter api-version could just not use this client level signature. Just like `subscription_id` in MGMT SDK, not all API has this parameter however all language SDKs treat it as client level signatures.
- if there is multi kinds of api-version, then TCGC shall not lift it.

weidongxu-microsoft:
Kind of valid for me...

If the client has 99 API needs api-verison, 1 API does not need api-version, should we put it to client.initialization? If yes for 99:1 but no for 1:1, then where is the switch point...

msyyc:
[...]

PR #1630, opened 2024-09-30
[tcgc] do not override client default value for api version param in non-versioning service

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1608


packages/typespec-client-generator-core/test/packages/client.test.ts
```diff
@@ -388,6 +388,73 @@ describe("typespec-client-generator-core: package", () => {
...
+    strictEqual(endpointTemplateArg.name, "endpoint");
+    strictEqual(endpointTemplateArg.onClient, true);
+    strictEqual(endpointTemplateArg.optional, false);
+    strictEqual(endpointTemplateArg.kind, "path");
+    strictEqual(endpointTemplateArg.clientDefaultValue, "http://localhost:3000");
+
+    const apiVersionParam = templatedEndpoint.templateArguments[1];
+    strictEqual(apiVersionParam.clientDefaultValue, "v1");
+    strictEqual(apiVersionParam.name, "apiVersion");
+    strictEqual(apiVersionParam.onClient, true);
```
haolingdong-msft, line 453: I'm a bit confused. why is non-versioning service still have `apiVersionParam.onClient ===  true`?

weidongxu-microsoft: It seems valid in this way, when the `api-version` is defined in `@server`.

I'd be interest to see a case of `api-version` defined in op (without versioning in service), and that parameter should have `onClient=false`.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -536,7 +536,7 @@ function getEndpointTypeFromSingleServer<
       }
       const apiVersionInfo = updateWithApiVersionInformation(context, param, client.__raw.type);
       sdkParam.isApiVersionParam = apiVersionInfo.isApiVersionParam;
-      if (sdkParam.isApiVersionParam) {
+      if (sdkParam.isApiVersionParam && apiVersionInfo.clientDefaultValue) {
```
chunyu3, line 539: Is the `sdkParam.clientDefaultValue` always be override if apiVersionInfo.clientDefaultValue (if exist)? Is it possible that the sdkParam.clientDefaultValue has different value of  apiVersionInfo.clientDefaultValue, in that way , the sdkParam itself  default value is lost.

Considering the versioned service, and the `apiVersion` in the URL, with a default value (which may or may not the any enum value of the version enum)
Do we have UT to test this scenario?

tadelesh: for versioning service, we honor the api version emitter specified, not client default value.

packages/typespec-client-generator-core/test/packages/client.test.ts
haolingdong-msft: Could we add an unit test to test the case where api-version is not defined in `@service`, and `onClient` should be `false`?

tadelesh: we have an issue to track this: https://github.com/Azure/typespec-azure/issues/1595.

iscai-msft (approved):

Issue #1598, opened 2024-09-24
[Bug]: EnumMember type as a clientDefaultValue?


qiaozha:
### Describe the bug

for the given typespec, 
    @server(
      "{endpoint}/client/structure/{client}",
      "",
      {
        @doc("Need to be set as 'http://localhost:3000' in client.")
        endpoint: url,
    
        @doc("Need to be set as 'default', 'multi-client', 'renamed-operation', 'two-operation-group' in client.")
        client: ClientType = ClientType.Default,
      }
    )
the clientDefaultValue we get for client parameter is a type instead of a value, however if it's a constant, we will just get a value for clientDefaultValue. 
![image](https://github.com/user-attachments/assets/8e2e987a-1f2b-4d24-b619-b5d878a50602)

I feel like we should always extra the value for clientDefaultValue?

### Reproduction

N/A

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1631, opened 2024-09-30
[tcgc] support value type for client default value

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1598


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -569,3 +569,30 @@ export function isOnClient(
...
+      return value.value.value ?? value.value.name;
+    case "ObjectValue":
+      return Object.fromEntries(
+        [...value.properties.keys()].map((x) => [
+          x,
+          getValueTypeValue(value.properties.get(x)!.value),
+        ]),
+      );
+    case "ScalarValue":
+      // TODO: handle scalar value
```
haolingdong-msft, line 595: handle this TODO?

iscai-msft (changes requested):
the `value` in tsp purposefully is meant to describe service default values, not client default values, so we shouldn't be having this support. There are a lot of issues with roundtrippability with default values, so we should only have `clientDefaultValue` in the specific cases of endpoint values and client api versions. Where is the push for this support coming from?

tadelesh:
> the `value` in tsp purposefully is meant to describe service default values, not client default values, so we shouldn't be having this support. There are a lot of issues with roundtrippability with default values, so we should only have `clientDefaultValue` in the specific cases of endpoint values and client api versions. Where is the push for this support coming from?

see the issue linked in description. there are cases that endpoint default value use value types, the support is only for those cases.

iscai-msft:
ah ok, sorry for the confusion, let me remove the blocking pr

iscai-msft (approved):

Issue #1620, opened 2024-09-27
Azure init templates are not working out of the box


timotheeguerin:
Standalone arm for example has a dup key in the tspconfig.yaml
<img width="352" alt="image" src="https://github.com/user-attachments/assets/b3aeba0f-952a-44b8-859a-796976988a72">

and is missing a dependency on `@azure-tools/typespec-azure-rulesets`

we should add test like we did for the core init templates

PR #1643, opened 2024-10-03
Fix azure standalone init templates and add integration tests

timotheeguerin:
fix https://github.com/Azure/typespec-azure/issues/1620

Fixed the issue as well as setting up some integration and snapshot test same as we have in the compiler.


allenjzhang (approved):
It would be nice to have the scaffolding goes with each package installed and have compiler to recognize that so once a package is installed you can do "tsp init ARM" or "tsp init protobuf" with introduction of "tspTemplate" in the package.json. Let's open a tracking design issue for it.

PR #1659, opened 2024-10-10
[tcgc] remove projection for source model since typespec core has already fixed the issue

tadelesh:
because of this fix: https://github.com/microsoft/typespec/issues/4280, tcgc should not do projection again which will cause undefined issue when finding spread body model.


MaryGao (approved):
Approved and thanks for this fix!

PR #1664, opened 2024-10-10
[tcgc] add client initialization usage

### LINKED ISSUE #1518: Add ClientInitialization usage for models

Right now getAllModels does not include models which have usage as none.  For the client initialization model the usage is none since it isn't used in any of the service operations.  We should introduce a new usage flag and apply it to the client initialization model so that it is returned by getAllModels.

Emitters who don't do anything with this model can easily ignore it by looking for this usage flag.

### Checklist

iscai-msft:
fixes #1518, #1707


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -350,24 +350,26 @@ export interface SdkUnionType<TValueType extends SdkTypeBase = SdkType> extends
...
+  discriminatorProperty?: TPropertyType;
   baseModel?: SdkModelType;
   crossLanguageDefinitionId: string;
   apiVersions: string[];
 }
 
-export interface SdkInitializationType extends SdkModelType {
-  properties: SdkParameter[];
+export interface SdkInitializationType {
+  access: AccessFlags;
```
iscai-msft, line 371: i've realized we can't modify the `.access` on our existing `SdkInitializationType` to denote whether a client is publicly or internally initialized. There is both the context that the client options model is used, and the client options model itself. In this case, the context that the client options model is used has `access` that depends on whether it's a client or og. However, that doesn't change the actual `access` of the client options model itself.

This is a breaking change, so want to call this out and will call it out more in our sync

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -463,11 +424,18 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
...
-      for (const property of clientInitialization.properties) {
-        parameters.push(property);
+    if (tspClientInitialization) {
+      const clientInitialization = getSdkInitializationType(
+        context,
+        operationGroup,
+        tspClientInitialization,
+      );
+      for (const property of clientInitialization.model.properties) {
+        if (property.kind === "method") {
```
tadelesh, line 436: is there any case the initialization model property is not a parameter?

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -463,11 +424,18 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
...
+          parameters.push(property);
+        }
       }
     } else {
     }
```
tadelesh, line 441: if no customized client initialization, then we don't need to add parameter for client accessor?

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1778,6 +1839,10 @@ export function getAllModelsWithDiagnostics(
       diagnostics.pipe(updateTypesFromOperation(context, operation));
     }
     const ogs = listOperationGroups(context, client);
+    const clientInitModel = getClientInitialization(context, client.type);
+    if (clientInitModel) {
+      getSdkInitializationType(context, client, clientInitModel);
+    }
```
tadelesh, line 1845: is it a must that we need to get the client initialization model before generating sdk pacakge thing?
also, seems this logic forget to deal with operation group.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -781,6 +773,75 @@ function getSdkEnumValueType(
...
+  };
+}
+
+export function getSdkInitializationType(
+  context: TCGCContext,
+  client: SdkClient | SdkOperationGroup,
+  model: Model,
+): SdkInitializationType {
+  // convert to SDK type
+  const existingModel = context.modelsMap?.get(model) as SdkModelType<SdkParameter> | undefined;
```
tadelesh, line 806: do we allow to reuse model in operation as a client initialization model? if so, can we have a test? if not, we could return directly and shall have some lint.

tadelesh (commented):
left some comments, and i think for default access settings, we shall consider: https://github.com/Azure/typespec-azure/issues/1702.

packages/typespec-client-generator-core/src/package.ts
iscai-msft: no, i'm making them all method parameters because we also use that in the logic to pull them out of the method signature

packages/typespec-client-generator-core/src/package.ts
iscai-msft: we still add parameters for client initialization, the api version and endpoint parameters, we just don't iterate through a model and add them

packages/typespec-client-generator-core/src/types.ts
iscai-msft: If we don't have it here, I'm not able to pick up the model. I'll add logic for operation groups too

packages/typespec-client-generator-core/src/types.ts
iscai-msft: @m-nash since c# appears to be the main sdk using client initialization model generation, is there anything precluding us from having a model be used in both scenarios?

tadelesh:
close in favor of https://github.com/Azure/typespec-azure/pull/2027

PR #1676, opened 2024-10-14
add docs for discriminator

ArcturusZhang:


docs/howtos/Client Generation/06types.mdx
```diff
@@ -793,6 +793,160 @@ public final class Animal implements JsonSerializable<Animal> {
...
+        }
+      ],
+      "discriminatorProperty": {
+        // the same instance of the property in the properties list here
+      },
+      "discriminatedSubtype": {
+        "siamese": {
+          "kind": "model",
+          "name": "Siamese",
+          "properties": []
```
tadelesh, line 881: add `discriminatorValue` property

docs/howtos/Client Generation/06types.mdx
```diff
@@ -793,6 +793,197 @@ public final class Animal implements JsonSerializable<Animal> {
 </TabItem>
 </Tabs>
 
+### Discriminator
```
lirenhe, line 796: Could you also add some description/case of multiple-level discriminator?

ArcturusZhang: we kind of prefer that we do this in a separate PR.

tadelesh (approved):

PR #1700, opened 2024-10-16
Tcgc/keep core models

### LINKED ISSUE #1412: [Bug]: sdkPackage.models is missing Error type
### Describe the bug

Model `FailedLockToken` contains field `error` of type `Azure.Core.Foundations.Error` but it's missing in the array of model types. As a result, the type isn't emitted during codegen.

### Reproduction

Build https://github.com/Azure/azure-rest-api-specs/tree/main/specification/eventgrid/Azure.Messaging.EventGrid with tcgc version `0.44.3`

### Checklist

- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)

iscai-msft:
fixes #1412


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -329,7 +329,7 @@ export function intOrFloat(value: number): "int32" | "float32" {
  * @param t
  * @returns
  */
-export function isAzureCoreModel(t: Type): boolean {
+export function isAzureCoreTspModel(t: Type): boolean {
```
tadelesh, line 332: a concern for this helper is that for template model instance, it will also have the namespace of the template, shall we treat them as the azure core model? for example, each time i use `GetResourceOperationStatus` for the following tsp, i will get a concrete `ResourceOperationStatus` type that has `Azure.Core` namespace.
```
@resource("operations")
@parentResource(Resource)
model ResourceOperationStatus<
  Resource extends TypeSpec.Reflection.Model,
  StatusResult = Resource,
  StatusError = Foundations.Error
> is Azure.Core.Foundations.OperationStatus<StatusResult, StatusError>;

@Foundations.Private.ensureVerb("GetResourceOperationStatus", "GET")
  @readsResource(ResourceOperationStatus<Resource>)
  @Foundations.Private.ensureResourceType(Resource)
  GetResourceOperationStatus<
    Resource extends TypeSpec.Reflection.Model,
    StatusResult = Resource,
    StatusError = Foundations.Error,
    Traits extends TypeSpec.Reflection.Model = {}
  > is Foundations.ResourceOperation<
    ResourceOperationStatus<Resource, StatusResult, StatusError>,
    {},
    ResourceOperationStatusResponse<Resource, StatusResult, StatusError>,
    Traits & InterfaceTraits,
    ErrorResponse
  >;
```

iscai-msft: Since it's taking template namespace into account, wouldn't this mean that `isAzureCoreModel(ResourceOperationStatus) === false`? @timotheeguerin is there a good way to determine whether a model is originally defined in the `Azure.Core` namespace?

tadelesh: since typespec team confirm it is by design, let me approve and merge this pr. thanks.

tadelesh (approved):

PR #1705, opened 2024-10-17
[tcgc] add documentation for new namespace changes

iscai-msft:


tadelesh:
close as it is done in this pr: https://github.com/Azure/typespec-azure/pull/1786. we could refine it later.

PR #1712, opened 2024-10-21
[@azure-tools/azure-http-specs] backport cadl-ranch ARM test rename

XiaofeiCao:
- origin: https://github.com/Azure/cadl-ranch/pull/714
- Fixed singleton mockapi.ts. Location is not updatable. Tested locally.


XiaofeiCao:
@sarangan12 When doing backporting, I found that without `handler`, mockapi will check exact body equals:
`com.azure.core.management.exception.ManagementException: Status code 400, "{"message":"Body provided doesn't match expected body","expected":{"identity":{"type":"SystemAssigned"}},"actual":{"location":"eastus","tags":{"tagKey1":"tagValue1"},"properties":{},"identity":{"type":"SystemAssigned"}}}"
: Body provided doesn't match expected body`

https://github.com/Azure/typespec-azure/blob/1505570165cca5ab108961fb9ece65817130e556/packages/azure-http-specs/specs/azure/resource-manager/models/common-types/managed-identity/mockapi.ts#L89-L91

Here we just want the request body to contain an expected `identity` property. Do you know how to achieve that?

XiaofeiCao:
@sarangan12 My bad.. Seems there're some dependency issues on my local machine. Did a refresh install and build and everything works now!

weidongxu-microsoft (approved):
I think we can keep a few handlers, in the cases when we still need them (PATCH in ARM would be mixed behavior of normal PATCH and JSON-Merge-Patch). I assume this was what Sarangan did in cadl-ranch PR too.

(we can have follow up issue/PR to get rid of them, if we could find a way)

XiaofeiCao:
Seems we need https://github.com/microsoft/typespec/pull/4846 to be synced to submodule to fix the CI.

Issue #1552, opened 2024-09-17
`no-response-body` rule is triggering for some standard templates


timotheeguerin:
Some templates shown in the samles trigger that rule which is confusing

reason this wasn't caught before it is that rule is only triggered on namespace starting with `Azure.` (legacy behavior from before the linter system)

### Arm

```
  delete is ArmResourceDeleteWithoutOkAsync<Employee>;
  /** A sample HEAD operation to check resource existence */
  checkExistence is ArmResourceCheckExistence<Employee>;
```

[...]

PR #1740, opened 2024-10-24
no-response-body rule is triggering for some standard templates

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1552


docs/libraries/azure-resource-manager/rules/no-response-body.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+ARM response operations with status code 202 or 204 should not contain response body. Operations that are different for 202 and 204 should contain a response body.
+
+### For 202 and 204 status codes (response body should be empty)
+
+#### ❌ Incorrect
+
+```tsp
+model Test204BodyResponse {
```
markcowl, line 17: Since this is an ARM rule, I would change this to use our own constructs, rather than custom models with status code, for example:
```ArmNoContentResponse & {@body body: string}```

File: docs/libraries/azure-resource-manager/rules/no-response-body.md [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+```
+
+### For operations different than 202 and 204 status codes (response body should not be empty)
+
+#### ❌ Incorrect
+
+```tsp
+model Test201EmptyResponse {
```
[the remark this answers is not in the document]
markcowl: https://github.com/microsoft/typespec/blob/main/packages/http/lib/http.tsp#L58

packages/typespec-azure-resource-manager/src/rules/no-response-body.ts
```diff
@@ -1,33 +1,57 @@
...
 
 /**
- * verify an operation returns 202 should not contains response body.
+ * verify an operation returns 202 or 204 should not contain response body.
+ * verify that operations that are different for 202 and 204 should contain a response body.
  */
 export const noResponseBodyRule = createRule({
   name: "no-response-body",
+  description:
+    "Check that the body is empty for 202 and 204 responses, and not empty for non-204/non-202 responses.",
```
markcowl, line 13:
```suggestion
    "Check that the body is empty for 202 and 204 responses, and not empty for other success (2xx) responses.",
```

packages/typespec-azure-resource-manager/src/rules/no-response-body.ts
```diff
@@ -1,33 +1,57 @@
...
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-response-body",
   severity: "warning",
-  description: `The body of 202 response should be empty.`,
   messages: {
-    default: `The body of 202 response should be empty.`,
+    default: `The body of non-204/non-202 responses should not be empty.`,
```
markcowl, line 17:
```suggestion
    default: `The body of responses with success (2xx) status codes other than 202 and 204 should not be empty.`,
```

packages/typespec-azure-resource-manager/src/rules/utils.ts
```diff
@@ -28,6 +28,12 @@ export function isTemplatedInterfaceOperation(target: Operation) {
   );
 }
 
+export function isAzureSubNamespace(program: Program, ns: Namespace | undefined): boolean {
```
markcowl, line 31: This feels kind of hacky.  Does the core rule have a similar restriction?  If so, can we just borrow its mechanism for checking that this is not library code?

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
```diff
@@ -33,10 +35,46 @@ describe("typespec-azure-resource-manager: no response body rule", () => {
...
+      .toBeValid();
+  });
+
+  it("emit warnings if a 204 response has a body", async () => {
+    await tester
+      .expect(
+        `
+        namespace Azure.MyService;
+        
+        model TestAcceptedResponse {
```
markcowl, line 59: I would use the arm response templates in these tests

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
```diff
@@ -24,6 +24,8 @@ describe("typespec-azure-resource-manager: no response body rule", () => {
     await tester
       .expect(
         `
+        namespace Azure.MyService;
```
markcowl, line 27: shouldn't this cause the rule not to execute?

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
```diff
@@ -33,10 +35,46 @@ describe("typespec-azure-resource-manager: no response body rule", () => {
       .toBeValid();
   });
 
+  it("is valid if 204 responses have no body", async () => {
+    await tester
+      .expect(
+        `
+        namespace Azure.MyService;
```
markcowl, line 42: Doesn't this cause the rule not to execute?

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
markcowl: Not sure how this is emitting diagnostices if we are skipping namespaces starting with `Azure`

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
```diff
@@ -49,4 +87,22 @@ describe("typespec-azure-resource-manager: no response body rule", () => {
         message: `The body of 202 response should be empty.`,
       });
   });
+
+  it("emit warnings if 201 responses have no body", async () => {
+    await tester
+      .expect(
+        `
+        namespace Azure.MyService;
```
markcowl, line 95: same

markcowl (changes requested):

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
AlitzelMendez: we are skipping namespaces that doesn't start with `Azure`

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
AlitzelMendez: no, this is what makes the rule to execute, this rule is only executed on Azure namespaces, this is also the core rule for reference: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/src/rules/no-response-body.ts

packages/typespec-azure-resource-manager/src/rules/utils.ts
AlitzelMendez: yes, this was on core library, took that and deleted the changes on this file.

docs/libraries/azure-resource-manager/rules/no-response-body.md
```diff
@@ -0,0 +1,98 @@
...
+  @statusCode statusCode: 201;
+}
+op walk(): Test201EmptyResponse;
+```
+
+```json
+{
+  "responses": {
+    "201": {
+      "description": "The request has succeeded and a new resource has been created as a result."
```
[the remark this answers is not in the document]
AlitzelMendez: this text is in a doc decorator: https://github.com/microsoft/typespec/blob/fe8f3f1bbc47c193cb673f6637ae911118ad32ea/packages/http/src/decorators.ts#L325, do I need to reference it anywhere? :)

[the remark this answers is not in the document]
markcowl: oh, gotcha, this is the statusCode default description, makes sense.

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
markcowl: Most ARM namespace start with 'Microsoft.`  Our namespaces in typespec-azure, like Azure.Core and Azure.ResourceManager start with `Azure.`

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
```diff
@@ -33,6 +33,34 @@ describe("typespec-azure-resource-manager: no response body rule", () => {
       .toBeValid();
   });
 
+  it("is valid if 204 responses have no body", async () => {
```
markcowl, line 36: I think we need a test to verify that 200 responses with no body are valid in head, delete, and post operations

markcowl: If we already have these tests, then this looks good to merge

markcowl (changes requested):

packages/typespec-azure-resource-manager/src/rules/no-response-body.ts
```diff
@@ -1,33 +1,60 @@
-import { Operation, createRule } from "@typespec/compiler";
-import { getResponsesForOperation } from "@typespec/http";
-
+import { createRule, Operation } from "@typespec/compiler";
+import { getHttpOperation, getResponsesForOperation, HttpOperationResponse } from "@typespec/http";
 import { isTemplatedInterfaceOperation } from "./utils.js";
-
 /**
- * verify an operation returns 202 should not contains response body.
+ * verify an operation returns 202 or 204 should not contain response body.
```
markcowl, line 5:
```suggestion
 * verify that 202 or 204 responses do not contain a response body.
```

packages/typespec-azure-resource-manager/src/rules/no-response-body.ts
```diff
@@ -1,33 +1,60 @@
...
+ * verify that operations that are different for 202 and 204 should contain a response body.
```
markcowl, line 6:
```suggestion
 * verify that success (2XX) responses other than 202 and 204 contain a response body.
```

packages/typespec-azure-resource-manager/test/rules/no-response-body.test.ts
```diff
@@ -49,4 +77,87 @@ describe("typespec-azure-resource-manager: no response body rule", () => {
...
+          model Employee is TrackedResource<{}> {
+            @key("employeeName")
+            @segment("employeeName")
+            @path
+            name: string;
+          }
+          
+          @armResourceOperations
+          interface TestingOperations {
+            postEmployees is ArmProviderActionAsync<Response = Employee, Scope = SubscriptionActionScope>;
```
markcowl, line 157: Here I think we want `Response = void`  You could alternately use the ArmResourceActionSync template, with a void response type

markcowl (changes requested):

markcowl (approved):

PR #1744, opened 2024-10-25
Fix x-ms-identifiers rule to support inner properties

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1742


packages/typespec-azure-resource-manager/test/rules/missing-x-ms-identifiers.test.ts
```diff
@@ -130,4 +130,29 @@ describe("typespec-azure-core: no-enum rule", () => {
       )
       .toBeValid();
   });
+
+  it("allow array of x-ms-identifiers", async () => {
+    await tester
+      .expect(
+        `
+        model Pet {
+          @OpenAPI.extension("x-ms-identifiers", ["food/brand/name"])
```
timotheeguerin, line 139: The doc linked in the issue mentioned that those should start with / in the example.

Not really sure if that's accurate but maybe we should support it or confirm there isn't use case of one or the other

packages/typespec-azure-resource-manager/test/rules/missing-x-ms-identifiers.test.ts
```diff
@@ -130,4 +130,70 @@ describe("typespec-azure-core: no-enum rule", () => {
...
```
markcowl, line 139: I think we should also test an identifier that starts with `/` (i.e. /food/brand/name)

markcowl (changes requested):

PR #1773, opened 2024-10-31
Update migration doc

pshao25:
Update migration doc by adding `--fully-compatible` parameter.


docs/migrate-swagger/01-get-started.md
```diff
@@ -44,6 +44,14 @@ npm install -g @azure-tools/typespec-client-generator-cli
     tsp-client convert --swagger-readme [path to readme.md] --arm
     ```
 
+  - Convert a **control-plane** specification to fully compatible output:
```
allenjzhang, line 47: Will discuss in today's sync meeting

pshao25: We will have these features if turn on this flag:
1. Flatten related feature: `arm-resource-flattening` in tspconfig, `@@flattenProperty` on properties
2. `@operationId` on operation if not same as the generated operation id
3. Add some suppression for enum/union/model/parameter like no document, no enum for azure
4. If the operation response is different from the default response of the template, customize the response
5. If the body name is different from the default body name of the template, add clientName
6. Add suppression if the provisionningState in the resource model violates the linter

pshao25: We could reframe what it controls, but I think at least 4&5 we should expose to our customers.

Issue #1728, opened 2024-10-23
[Bug]: TCGC SDK method parameter is in-consistant with operation parameters with autoRouted constant path params


MaryGao:
### Describe the bug

SDK method parameter is in-consistant with operation parameters with constant path params.

With below typespec example

```
@doc("show example demo")
@autoRoute
op read(@path param: "foo/bar"): void;
```
SDK method would have `param` parameter as constand type.
![Image](https://github.com/user-attachments/assets/46ecde09-f04b-4ca6-918a-6912e5f21b79)

But relevant operation has 0 parameter. And there is no parameter in path `/foo/bar`.

![Image](https://github.com/user-attachments/assets/d77bd8a8-3201-441b-8d1a-2a667c949ff5)

After invetigation it is because the autoRoute feature do the replacement between the param and relevant value if it is string literal - https://github.com/microsoft/typespec/blob/main/packages/rest/src/rest.ts#L138-L143.

I think it's better to align the parameters between method and operation.

### Reproduction

NA

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

MaryGao:
I want to leverage this issue to discuss the design for method vs operation.
- it seems we have two different layers to duplicate things, why that?
- the in-consistancy would happen in parameters e.g spreading or above case
- the in-consistancy would happen in responses because of body or bodyRoot decorator
- what is the usage in getAllModels e.g `op test(@bodyRoot body: {@body body: Test}): void;`(https://github.com/microsoft/typespec/issues/4815)

Issue #1709, opened 2024-10-18
[tcgc] handle singleton key exclusion


iscai-msft:
Handle exclusion of singleton keys from operation parameters namespace [here](https://github.com/Azure/typespec-azure/blob/0f289160fadd84b6e0935d014740407f7e10354e/packages/typespec-azure-resource-manager/src/namespace.ts#L224) . Will require us to add logic to deal with `@typespec/http.setRouteOptionsForNamespace`

tadelesh:
a common solution: filter method parameter or model property that is decorated with `@path` but not in http operation path parameter list.

PR #1775, opened 2024-10-31
[tcgc] remove unused path parameter from method

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1709
fix: https://github.com/Azure/typespec-azure/issues/1728

### DIFF
--- a/packages/typespec-client-generator-core/test/package.test.ts
+++ b/packages/typespec-client-generator-core/test/package.test.ts
@@ -352,7 +352,7 @@ describe("typespec-client-generator-core: package", () => {
       strictEqual(method.parameters.length, 4);
       deepStrictEqual(
         method.parameters.map((x) => x.name),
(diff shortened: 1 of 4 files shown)


packages/typespec-client-generator-core/test/package.test.ts
```diff
@@ -349,7 +349,7 @@ describe("typespec-client-generator-core: package", () => {
       const method = getServiceMethodOfClient(sdkPackage);
       strictEqual(method.name, "create");
       strictEqual(method.kind, "basic");
-      strictEqual(method.parameters.length, 5);
+      strictEqual(method.parameters.length, 4);
```
MaryGao, line 352: why just length updated? but below deepStrictEqual for param name is not updated?

iscai-msft: fixed

iscai-msft (approved):

Issue #1596, opened 2024-09-24
[tcgc] support negation for scope in tcgc decorators


tadelesh:
Support scope like `!java` to represent all languages except java. Real use case could refer this: https://github.com/Azure/azure-rest-api-specs/blob/4883fa5dbf6f2c9093fac8ce334547e9dfac68fa/specification/batch/Azure.Batch/routes.tsp#L133.

live1206:
1. Use `!` for negation
2. When negate multiple scopes: `!(java, csharp)` or `!(java | csharp)`
3. Since negation means excluding the defined scopes from the rest, we don't want to fix the negation usage with normal usages. So, it is illegal to write `!java, csharp`
4. If agreed with 3, we will store the negation scope or the normal scope into state

tadelesh:
> 1. Use `!` for negation
> 2. When negate multiple scopes: `!(java, csharp)` or `!(java | csharp)`
> 3. Since negation means excluding the defined scopes from the rest, we don't want to mix the negation usage with normal usages. So, it is illegal to write `"!java, csharp"`
> 4. If agreed with 3, we will store the negation scopes or the normal scopes into state

agree with 3. only allow `!(java, chsarp)`

another situation is setting one decorator multiple times. e.g.:
```
@clientName("B", "!csharp")
@clientName("A", "java")
model T {}
```
the result should be the later override the former: T for csharp, B for else.

live1206:
Have we considered to avoid multiple usage of the same decorator? or we want to allow it for `@clientName`?

tadelesh:
> Have we considered to avoid multiple usage of the same decorator? or we want to allow it for `@clientName`?

see discussion here: https://github.com/Azure/typespec-azure/pull/1606#discussion_r1778791173

PR #1783, opened 2024-11-01
Implement scope negation for TCGC

live1206:
Resolves https://github.com/Azure/typespec-azure/issues/1596

There are 2 formats of scope negation:
```
@clientName("A", "!(python, java)")
Model foo
```
is equivalent to
```
@clientName("A", "!python, !java")
Model foo
```
We allow combination of normal scope and negation scope for different scopes
```
@clientName("A", "!python, csharp")
Model foo
```
Combination of same scope is also allowed
```
@clientName("A", "!python, python")
Model foo
```
is equivalent to
```
@clientName("A", "!python")
@clientName("A", "python")
Model foo
```
and equivalent to
```
@clientName("A")
Model foo
```

The rule for decorator override:
- for the same scope, the later decorator value wins regardless it's defined with normal scope or scope negation

Detailed override cases can be found in the tests of this PR.


packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3027,4 +3027,156 @@ describe("typespec-client-generator-core: decorators", () => {
...
+
+    it("disallow multiple separated negation scopes", async () => {
+      const runnerWithCSharp = await createSdkTestRunner({
+        emitterName: "@azure-tools/typespec-csharp",
+      });
+      const diagnostics = await runnerWithCSharp.diagnose(`
+        @service
+        @test namespace MyService {
+          @test
+          @clientName("TestRenamed", "!csharp, !java")
```
XiaofeiCao, line 3165: Why do we want to disallow this? 
There's a potential scenario where !csharp is added first, then java want the same. heuristically we'll append !java after it, though in practice we'll meet error..

live1206: For multiple negation scopes, user can simply write `!(csharp, java)`

XiaofeiCao: I don't know if it's just me.. I mean, if I saw `@clientName("TestRenamed", "!csharp")` in `client.tsp`, I'll definitely append `!java` after it...
On the other hand, `!csharp, !java` and `!(csharp, java)` look equivalent to me. What's the consideration to disallow the former?

live1206: So, we agreed that disallow `!csharp,  java`. I thought `!csharp, !java` would be confusing as well.
I am open for discussion on this.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -113,20 +124,53 @@ function setScopedDecoratorData(
...
+function getNegationScopes(scope?: LanguageScopes): string[] | undefined {
+  if (scope === undefined) {
+    return undefined;
+  }
+
+  const negationScopeRegex = new RegExp(/!\((.*?)\)/);
+  const negationScopeMatch = scope.match(negationScopeRegex);
+  if (negationScopeMatch) {
+    return negationScopeMatch[1].split(",").map((s) => s.trim());
+  } else if (!scope.includes(",") && scope.startsWith("!")) {
```
tadelesh, line 170: since we have two kinds of patten for scope, can we have a regex that could cover all cases and make sure all user's input are valid, else report diagnostic?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -113,20 +124,53 @@ function setScopedDecoratorData(
...
+    // if target doesn't exist in decorator map, create a new entry
+    if (!targetEntry) {
+      const newObject = Object.fromEntries(splitScopes.map((scope) => [scope, value]));
+      context.program.stateMap(key).set(target, newObject);
+      return;
+    }
+
+    // if target exists, overwrite existed value
+    const newObject = Object.fromEntries(splitScopes.map((scope) => [scope, value]));
+    context.program.stateMap(key).set(target, { ...targetEntry, ...newObject });
```
tadelesh, line 157: the main problem for the algorithm is it seems not support incremented set. for example:
```
@@doc(A, "else", "!java");
@@doc(A, "java", "java");
```

also, could you add some tests for that?

tadelesh (commented):
please also change the documentation for the scope parameter in `decorators.tsp` to explain the patten could be used in scope

packages/typespec-client-generator-core/test/decorators.test.ts
XiaofeiCao: Yeah, `!java, java` is definitely confusing. Though I don't see what harm does `!csharp, !java` do, and it's in fact heuristically more approachable.

Also, if we want to explicitly disallow this case, would be helpful to add more info to the diagnostic, as current one is confusing in this case: 
`Negation scope should not be combined with normal scope.` -> I'm not doing that..

live1206: That sounds reasonable to me, let me update the implementation.

live1206: updated

packages/typespec-client-generator-core/src/decorators.ts
live1206: Based on https://github.com/Azure/typespec-azure/pull/1783#discussion_r1825460771, we do have 2 syntax formats.
```
!(scope1, scope2, ...)
```
and
```
!scope1, !scope2, ...
```
I think it's clearer to handle them separately. Please take another look at the updated implementation.

packages/typespec-client-generator-core/src/decorators.ts
live1206: added

packages/typespec-client-generator-core/test/decorators.test.ts
XiaofeiCao: looks good, thanks!

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -113,20 +124,65 @@ function setScopedDecoratorData(
...
+    newObject[negationScopesKey] = negationScopes;
     context.program.stateMap(key).set(target, newObject);
     return;
+  } else if (scopes !== undefined && scopes.length > 0) {
+    // if scope specified, create or overwrite with the new value
+    const targetEntry = context.program.stateMap(key).get(target);
+
+    // if target doesn't exist in decorator map, create a new entry
+    if (!targetEntry) {
+      const newObject = Object.fromEntries(scopes.map((scope) => [scope, value]));
```
XiaofeiCao, line 139: nit, since we always need the newObject(whether targetEntry is empty or not), we could consider put it outside, and do `if(!targetEntry)` later. This way we could replace the `return` with `else`.

live1206: Right. It was like this in the first place, simplified.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -86,6 +87,16 @@ function getScopedDecoratorData(
...
+
+    // if the scope is negated, we should return undefined
+    const negationScopes = retval[negationScopesKey];
+    if (negationScopes !== undefined && negationScopes.includes(scope)) {
+      return undefined;
+    }
+    // if the scope is not negated, we should return the value for AllScopes
+    else {
+      return retval[AllScopes];
+    }
```
XiaofeiCao, line 99: do we need this else?

live1206: No, we don't. It can just fall back to the default return.

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3027,4 +3027,182 @@ describe("typespec-client-generator-core: decorators", () => {
...
+          @test
+          @access(Access.internal)
+          op func(
+            @body body: Test
+          ): void;
+        }
+      `);
+
+      const sdkPackage = runnerWithCSharp.context.sdkPackage;
+      const testModel = sdkPackage.models.find((x) => x.name === "Test");
```
tadelesh, line 3204: why this name is `Test`? i suppose it should be `TestRenamed`?

live1206:
```
          @clientName("TestRenamedAgain", "!csharp, !java")
          @clientName("TestRenamed", "csharp")
```
Since we define that later decorator overrides the former decorator, `@clientName("TestRenamedAgain", "!csharp, !java")` will take effect, right? Then this renaming will exclude csharp, so the model remains "Test".

tadelesh: the override is by single scope which mean you could set value for java, python, go with three separated decorators. then the problem is how should we handle the negation part.

live1206:
```
@clientName("A", "csharp")
@clientName("B", "java")
Model Foo
```
For normal scopes, we can just add them incrementally.

```
@clientName("A", "!csharp")
@clientName("B", "!java")
Model Foo
```
For the negation scope override, the value of other scopes changes. We can't just add the negated scopes to the original map.
Two options here:
- Simply override everything when a negation scope shows up including the others' value and the negation scopes
- Throw error if others' values are different with different negation scopes

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -6,6 +6,7 @@ namespace Azure.ClientGenerator.Core;
  * Changes the name of a method, parameter, property, or model generated in the client SDK
  * @param rename The rename you want applied to the object
  * @param scope The language scope you want this decorator to apply to. If not specified, will apply to all language emitters
+ * You can use "!" to specify negation such as "!(java, python)" or "!java, !python". Please note that, negation scope and normal scope should not be mixed.
```
XiaofeiCao, line 9: This would probably need update. Would recommend mentioning how we handle the overlapping edge case as well.

live1206: If you are talking about how to deal with decorator override. I think that should be a separate topic.
@tadelesh Do we have any documentation about how decorator override should work?

XiaofeiCao: I mean `Please note that, negation scope and normal scope should not be mixed.` this part, may need scroll to the right a bit.. 
Though user can definitely benefit from scope override documentation.

live1206: Repasting the description of the PR here
Combination of same scope is also allowed
```
@clientName("A", "!python, python")
Model foo
```
is equivalent to
```
@clientName("A", "!python")
@clientName("A", "python")
Model foo
```
and equivalent to
```
@clientName("A")
Model foo
```
For the case you mentioned, there is no override.

XiaofeiCao: I see. Then it's fine.

live1206: > I mean `Please note that, negation scope and normal scope should not be mixed.` this part, since we now allow mix of negation + normal scope. May need scroll to the right a bit.. Though user can definitely benefit from scope override documentation.

remoevd

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3027,4 +3027,311 @@ describe("typespec-client-generator-core: decorators", () => {
...
   });
+
+  describe("scope negation", () => {
+    it("single scope negation", async () => {
+      const runnerWithCSharp = await createSdkTestRunner({
+        emitterName: "@azure-tools/typespec-csharp",
+      });
+      await runnerWithCSharp.compile(`
+        @service
+        @test namespace MyService {
```
tadelesh, line 3038: `@test` is no need if you don't want to get the raw type. also could you remove the `@access`

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -3027,4 +3027,311 @@ describe("typespec-client-generator-core: decorators", () => {
...
+          }
+          @test
+          @access(Access.internal)
+          op func(
+            @body body: Test
+          ): void;
+        }
+      `);
+
+      const sdkPackage = runnerWithCSharp.context.sdkPackage;
```
tadelesh, line 3254: i suggest to use different runner with different language to test the result.

tadelesh (approved):
the algorithm looks good. please merge before you refine the test.

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: Added another runnerWithPython to verify the negation case in the same test.

packages/typespec-client-generator-core/test/decorators.test.ts
live1206: refined

PR #1793, opened 2024-11-02
Add tsp-client local spec development docs

catalinaperalta:
Adding docs to explain the local development scenario for folks who havent checked a spec in yet or who want to work on local changes and generate a client library based on those updates.


website/src/content/docs/docs/howtos/Generate with tsp-client/local_spec_quickstart.md
```diff
@@ -0,0 +1,42 @@
...
+title:  Quickstart - Local development
+---
+
+`tsp-client` allows you to develop your TypeSpec project locally before pushing changes upstream. These instructions can be used when you want to generate a client library from a local TypeSpec project that has the changes you wish to use.
+
+### Prerequisites
+
+- Install `tsp-client`:
+
+```pwsh
```
timotheeguerin, line 11:
```suggestion
```bash
```

timotheeguerin: same for the others

website/src/content/docs/docs/howtos/Generate with tsp-client/local_spec_quickstart.md
```diff
@@ -0,0 +1,42 @@
+---
+title:  Quickstart - Local development
```
timotheeguerin, line 2: Wondering if this name makes sense right next to `tsp-client usage`

should we name them differently?

catalinaperalta: I agree...Maybe it should be:
- `tsp-client command documentation`
- `tsp-client quickstart`

OR maybe just remove tsp-client and do:
- `Supported commands`
- `Quickstart (local development)`

catalinaperalta: I liked the second option more so updated to that, let me know if you have a better idea!

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: tsp-client usage
+title: Supported commands
```
timotheeguerin, line 2:
```suggestion
title: Commands
```

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
```diff
@@ -1,5 +1,5 @@
...
 ---
```
timotheeguerin, line 4: also wondering if you should hjave the install step as a separate page as you need it for both docs

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
timotheeguerin: actually thats not correct, it does more than show the commands, 
image

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
catalinaperalta: In my experience folks dont like to really read through multiple pages of docs, so I'd rather they have all the steps they need on the same page. We can look at this again in the future if there are more shared instructions between both pages.

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
```diff
@@ -1,5 +1,5 @@
...
-title: tsp-client usage
+title: Commands
```
timotheeguerin, line 2: Sorry reopening as my reply might have got lost, I think actually this name is not really accurate 
image

I think either we should split this file or maybe find a better name

website/src/content/docs/docs/howtos/Generate with tsp-client/local_spec_quickstart.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Quickstart (local development)
```
timotheeguerin, line 2: what would be the other type of developemnt?

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
catalinaperalta: Ok I dont want to do the refactoring for the other doc in this PR, but I agree I think we can find a way to split it. Maybe between usage & configuration (dev), commands and configurations, configuration (repo). 

For this PR I could revert the renaming we just did or update it to `Usage & configuration` which I think covers most of the content of the current documentation, what do you think?

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
catalinaperalta: Sorry I thought this comment was on the quickstart page...in my comment below I agree to splitting up this doc into separate sections

website/src/content/docs/docs/howtos/Generate with tsp-client/local_spec_quickstart.md
timotheeguerin: yeah I think it make sense as just QuickStart

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
timotheeguerin: however you feel, just depends I think on when the follow up will be done

timotheeguerin: `Usage & Configuration` dos also seems like a good name without needing a split

timotheeguerin:
```
tsp-client
  Quick Start
  Usage & Configuration
```

timotheeguerin (approved):
Looks good I think with the change discussed in the comments

website/src/content/docs/docs/howtos/Generate with tsp-client/tsp_client.md
catalinaperalta: Agree will update to those names for now, and I opened this tracking issue to look into splitting the docs: https://github.com/Azure/azure-sdk-tools/issues/9357

PR #1818, opened 2024-11-06
New custom component for rendering the client outputs

timotheeguerin:
New features
- Always show the typespec above
- sync the tabs
-  just defining the code blocks without the extra `TabItem`
image


website/src/content/docs/docs/howtos/Client Generation/02client.mdx
timotheeguerin, line 1: only changed this file for now if we like it I can easily bulk update it all

iscai-msft (approved):
beautiful

Issue #1827, opened 2024-11-07
[Bug]: propagation of usage flags makes base model type usage ambiguous


jhendrixMSFT:
### Describe the bug

Some change in `0.47.x` causes usage flags to propagate from derived type to base type. This means that it's no longer possible to identify base types that aren't explicitly annotated.

For Go, which does not have inheritance, we omit base types by default unless they're explicitly referenced (directly or transitively) via an operation.

For model-only builds we can only rely on usage flags to control the output. Now that these flags are propagated, the usage on base model types is ambiguous.

### Reproduction

For the following tsp, `BaseClassThatsPruned` has a usage of `output` even though it's not explicitly annotated as such.
```
import "@azure-tools/typespec-client-generator-core";

using Azure.ClientGenerator.Core;

@service({
  title: "ModelsOnly",
})
namespace ModelsOnly;

model BaseClassThatsPruned {
    id: int32;
}

model DerivedOne extends BaseClassThatsPruned {
    name: string;
}

model PreservedBaseClass {
    id: int32;
}

model DerivedTwo extends PreservedBaseClass {
    name: string;
}

@@access(ModelsOnly.PreservedBaseClass, Access.public);
@@access(ModelsOnly.DerivedOne, Access.public);
@@access(ModelsOnly.DerivedTwo, Access.public);
@@usage(ModelsOnly.PreservedBaseClass, Usage.output);
@@usage(ModelsOnly.DerivedOne, Usage.output);
@@usage(ModelsOnly.DerivedTwo, Usage.output);
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1834, opened 2024-11-11
[tcgc] add `disableUsageAccessPropagationToBase` flag and separate exception usage from output usage

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1827

add `disableUsageAccessPropagationToBase` flag: this flag aims to support language that does not generate base model, it will skip the base model's propagation but will not skip the property or subtype of the base model.

### DIFF
--- a/.chronus/changes/add_usage_helper-2024-10-11-17-1-36.md
+++ b/.chronus/changes/add_usage_helper-2024-10-11-17-1-36.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+add `disableUsageAccessPropagationToBase` to support language that does not generate base model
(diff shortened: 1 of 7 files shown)


iscai-msft (approved):

weidongxu-microsoft (approved):
What's the previous bahavior of model only used in `@error`? Does it has `Output`?

ArcturusZhang (commented):
what is the difference between `Error` usage and `Exception` usage? Why we need them both?

weidongxu-microsoft:
> > What's the previous behavior of model only used in `@error`? Does it has `Output`?
> 
> the model with `@error` or extends from model with `@error` will have `Error` usage. and model used in exception response will have usage `Output`. the problem here is `error` does not have propagation, and you could not filter out the model only used by that error model.

Thanks. Kind of same question as Dapeng's. Not sure if we just have an `Error` usage which covers all model related to `@error` model and its reference seems reasonable. Though if we do have some case that we must treat `Error` and `Exception` differently, I am fine with that.

Issue #1647, opened 2024-10-04
Autorest emitter silently emit no files when version option is used an doesn't match the versions of the service


timotheeguerin:
```
import "@typespec/versioning";

using TypeSpec.Versioning;


@versioned(Versions)
@service
namespace DemoService;

enum Versions {
  v1,
  v2,
}

model A {}
```
[Playground Link](https://azure.github.io/typespec-azure/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvdmVyc2lvbmluZyI7Cgp1c2luZyBUeXBlU3BlYy5WyR07CgoKQMcsZWQoxxlzKQpAc2VydmljZQpuYW1lc3BhY2UgRGVtb1PGFjsKCmVudW0gyDAgewogIHYxLMQGMiwKfQoKbW9kZWwgQSB7fQo%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22options%22%3A%7B%22%40azure-tools%2Ftypespec-autorest%22%3A%7B%22version%22%3A%22v13%22%7D%7D%7D)

set
```
options:
  "@azure-tools/typespec-autorest":
    version: v3

```

allenjzhang:
Need to introduce an error message on `no matching version found`.

PR #1838, opened 2024-11-12
No matching versions found

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1647


packages/typespec-autorest/src/emit.ts
```diff
@@ -128,6 +128,10 @@ export async function getAllServicesAtAllVersions(
       (v) => !options.version || options.version === v.version,
     );
 
+    if (versions.length === 0) {
+      reportDiagnostic(program, { code: "no-matching-version-found", target: NoTarget });
```
chrisradek, line 132: I'm not sure how common it is to define multiple services in a single TypeSpec project, but if we set the `target` to `service.type` it will highlight the namespace that doesn't have a matching version:
```suggestion
      reportDiagnostic(program, { code: "no-matching-version-found", target: service.type});
```

See [PR Playground Link](https://cadlplayground.z22.web.core.windows.net/typespec-azure/prs/1838/?c=aW1wb3J0ICJAdHlwZXNwZWMvdmVyc2lvbmluZyI7DQoNCnVzaW5nIFR5cGVTcGVjLlbJH8UeQMcvZWQoRGVtb1NlcnZpY2XIJnMpDQpAc8YUDQpuYW1lc3BhY2Ugyysgew0KICBlbnVtIMg1xhMgIHYxLMcJMsUJfcRxICBtb2RlbCBBIHt9DQrFE%2F8AhPcAhFRlc3T%2FAITzAIQgIHYz%2BQCN&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22options%22%3A%7B%22%40azure-tools%2Ftypespec-autorest%22%3A%7B%22version%22%3A%22v4%22%7D%7D%7D) for an example of 2 services failing for example.

chrisradek (approved):

Issue #1857, opened 2024-11-13
[tcgc] do not replace operation with `@override` when `listOperationsInOperationGroup`


tadelesh:
since `getHttpOperation` will rely on the operation signature, we should not replace it when list operations. instead, current logic has already handled the parameter signature for `@override` in service method processing logic.

PR #1858, opened 2024-11-13
[tcgc] do not replace operation with `@override` when `listOperationsInOperationGroup`

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1857


archerzz (approved):

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -586,7 +586,7 @@ export function listOperationsInOperationGroup(
     for (const op of current.operations.values()) {
       // Skip templated operations
       if (!isTemplateDeclarationOrInstance(op)) {
-        operations.push(getOverriddenClientMethod(context, op) ?? op);
+        operations.push(op);
```
iscai-msft, line 589: is this the decided way to proceed until every language can fully support overrides? I feel it would be better to still return these instead of the original, and we could tell emitter authors that it's overridden, and we can point back to the original operation as well

tadelesh: `listOperationsInOperationGroup` is a decorator helper, which return the raw typespec operation. if you return the override one, then http payload will be changed. it is a bug, not a design change.

.chronus/changes/fix_override-2024-10-13-15-32-53.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+do not replace operation with `@override` when `listOperationsInOperationGroup`
```
weidongxu-microsoft, line 7: Does it only affect `listOperationsInOperationGroup`?

Java emitter currently list `SdkServiceMethod` under `client.methods`.

From the context, this seems also should not be the `@override` operation -- it should have the signature of that in `SdkServiceMethod`, while having the signature of the original/overrided operation in `SdkServiceOperation`

tadelesh: it is just a bug for previous `@override` impl. for `SdkServiceMethod` it will use the `@override` operation, for `SdkServiceOperation` it will use the original operation, these two logic is right for the impl.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: ahh i see, approving

PR #1859, opened 2024-11-13
doc(howto): update csharp sample of `@encode(string)`

archerzz:
a follow up of https://github.com/Azure/autorest.csharp/issues/4919


website/src/content/docs/docs/howtos/Generate client libraries/06types.mdx
```diff
@@ -2581,7 +2581,8 @@ serialized_prop = json.dumps(prop, cls=SdkJSONEncoder, format="rfc3339")
 <ClientTabItem lang="csharp" >
 
 ```csharp
-TODO
+// Internal implementation
+utf8JsonWriter.WriteStringValue(Prop.ToString());
```
chunyu3, line 2585: there should has the `format` when call `WriteStringValue`

archerzz: > there should has the `format` when call `WriteStringValue`

@chunyu3 That's the next phase. Right now, we just rely on the `int.ToString()` e.g. the default formatting.

PR #1862, opened 2024-11-13
Tcgc/xml encoded name

iscai-msft:
fixes https://github.com/Azure/autorest.python/issues/2914

### DIFF
--- a/packages/typespec-client-generator-core/src/public-utils.ts
+++ b/packages/typespec-client-generator-core/src/public-utils.ts
@@ -41,6 +41,7 @@ import {
   getClientNamespaceStringHelper,
   getHttpBodySpreadModel,
   getHttpOperationResponseHeaders,
+  getTypeDecorators,
   isAzureCoreTspModel,
   isHttpBodySpread,
   parseEmitterName,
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -206,7 +207,14 @@ export function getLibraryName(
  * @returns
  */
 export function getWireName(context: TCGCContext, type: Type & { name: string }) {
-  // 1. Check if there's an encoded name
+  const xmlNameDecorator = ignoreDiagnostics(getTypeDecorators(context, type)).filter(
+    (x) => x.name === "TypeSpec.Xml.@name",
+  );
+  // 1. Check xml name encoding
+  if (xmlNameDecorator.length > 0) {
+    return xmlNameDecorator[0].arguments["name"];
```
catalinaperalta, line 215: So we're thinking that there really shouldnt be a case where encodedName is used with this decorator, right? We'll always return what's in @xml.name

tadelesh:
as i mentioned in the meeting, please hold this fix until i get the design for serialization refactor in tcgc, thanks.

tadelesh:
close in favor of https://github.com/Azure/typespec-azure/pull/2059

Issue #1710, opened 2024-10-18
`no-empty-model` rule shouldn't warn on record models


timotheeguerin:
```
model PortalExtensionCompileFileResult is Record<unknown>;
```
There is alreadt `no-arm-record` that will warn about the use of record so this rule just adds confusion to this case

[Playground Link](https://azure.github.io/typespec-azure/playground?c=bW9kZWwgUG9ydGFsRXh0ZW5zaW9uQ29tcGlsZUZpbGVSZXN1bHQgaXMgUmVjb3JkPHVua25vd24%2BOw%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fresource-manager%22%5D%7D%7D)

PR #1866, opened 2024-11-14
no-empty-model rule shouldn't warn on record models

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1710


packages/typespec-azure-resource-manager/src/rules/no-empty-model.ts
```diff
@@ -14,7 +14,7 @@ export const noEmptyModel = createRule({
   create(context) {
     return {
       model: (model: Model) => {
-        if (model.properties.size === 0) {
+        if (model.properties.size === 0 && model.sourceModel?.name !== "Record") {
```
markcowl, line 17: We should use the built-in compiler call for `isRecordModelType` as in this example: https://github.com/Azure/typespec-azure/blob/60f9e0682a8b3ff7e03fde33a2a199053db7d1d7/packages/typespec-autorest/src/openapi.ts#L1903

see import here: https://github.com/Azure/typespec-azure/blob/60f9e0682a8b3ff7e03fde33a2a199053db7d1d7/packages/typespec-autorest/src/openapi.ts#L87

markcowl (changes requested):

AlitzelMendez:
validation result: https://github.com/Azure/azure-rest-api-specs/actions/runs/11902850879

markcowl (approved):

PR #1886, opened 2024-11-19
Remove `Azure.Core.nextLink` in favor of  `TypeSpec.nextLink`

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/5048


iscai-msft (approved):

packages/typespec-azure-core/src/decorators.ts
```diff
@@ -268,15 +267,11 @@ export function getItems(program: Program, entity: Type): boolean | undefined {
...
-export const $nextLink: NextLinkDecorator = (context: DecoratorContext, entity: ModelProperty) => {
-  context.program.stateMap(AzureCoreStateKeys.nextLink).set(entity, true);
-};
-
 /**
  * Returns `true` if the property is marked with `@nextLink`.
  */
-export function getNextLink(program: Program, entity: Type): boolean | undefined {
-  return program.stateMap(AzureCoreStateKeys.nextLink).get(entity);
+export function getNextLink(program: Program, entity: ModelProperty): boolean {
```
markcowl, line 273: Why change the signature?  We would like to not break the behavior of this function, for those using it

packages/typespec-azure-core/src/decorators.ts
```diff
@@ -268,15 +267,11 @@ export function getItems(program: Program, entity: Type): boolean | undefined {
...
+  return program.stateSet(Symbol.for(`TypeSpec.nextLink`)).has(entity);
```
markcowl, line 274: Instead, can we call the accessor function in typespec core?

markcowl (changes requested):

markcowl (commented):
Also:  We should try this PR in the azure-rest-api-specs public repo, fix any issues and and link to that PR here

packages/typespec-azure-core/src/decorators.ts
AlitzelMendez: I tried that, but I got this error: https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/12146646232/job/33871047296#step:5:29355 
image I called this one: https://github.com/microsoft/typespec/blob/968312e3d0d456b74981bfe244a1ee0d01a9beda/packages/compiler/src/lib/decorators.ts#L124

is there another one?

AlitzelMendez:
@markcowl  sharing with you the other pull requests related to this change

### Related pull requestes
- E2E cadl-ranch tests: https://github.com/Azure/cadl-ranch/pull/778
- azure-rest-api-specs: This pull request is toward typespec-next: https://github.com/Azure/azure-rest-api-specs/pull/31741 this change was validated in https://github.com/Azure/azure-rest-api-specs/actions/runs/12149858838/job/33881414993?pr=31734
- azure-rest-api-specs-pr: This pull request is toward main as I don't thing typespec-next is up to date, but let me know if I should make the change to another branch: https://github.com/Azure/azure-rest-api-specs-pr/pull/20475 validated in: https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/12149946135/job/33881676248?pr=20475

tadelesh:
i'm wondering if we have plan also replace `@Azure.Core.items` with `@TypeSpec.pageItems`?

packages/typespec-azure-core/src/decorators.ts
markcowl: There is an `isNextLink` function, but it isn't exported.  I think we can stay with the current implementation, but mark the accessor as deprecated

AlitzelMendez: thanks Mark! that change needs to be done in a different pull request in typespec repository, right?

markcowl (approved):

Issue #1752, opened 2024-10-28
[TCGC] [LRO] `lroMetadata.pollingStep.responseBody.usage` is not correct when polling operation is defined in different namespace


haolingdong-msft:
## Issue

I believe this is an edge case. See below test case, `PollResponse` model in namespace `Cadl.ServiceNamespace` has usage === 0`, but it is actually used in polling operation.
```ts
import "@typespec/rest";
import "@azure-tools/typespec-client-generator-core";
import "@azure-tools/typespec-azure-core";


using TypeSpec.Rest;
using TypeSpec.Http;
using TypeSpec.Versioning;
using Azure.Core;
using Azure.Core.Traits;


@service({
  title: "ServiceNamespace",
})
@versioned(ApiVersions)
@route("/servicenamespace")
namespace Cadl.ServiceNamespace {
  enum ApiVersions {
    @useDependency(global.Azure.Core.Versions.v1_0_Preview_2)
    v2022_06_01_preview: "2022-06-01-preview",
  }

  model PollResponse {
    operationId: string;
    status: Azure.Core.Foundations.OperationState;
  }

  @pollingOperation(NonService.poll)
  @post
  @route("/post")
  op longRunning(): AcceptedResponse;

}

@useDependency(Azure.Core.Versions.v1_0_Preview_2, Cadl.ServiceNamespace.ApiVersions.v2022_06_01_preview)
namespace Cadl.NonService {
  @route("/poll")
  @get
  op poll(): Cadl.ServiceNamespace.PollResponse;
}

```
![Image](https://github.com/user-attachments/assets/6bbbcdab-185b-4fc4-8e84-d3d51af7f953)

## Solution
`PollResponse` should  be `public` access + `output` usage.

Issue #1530, opened 2024-09-12
[Bug][TCGC]: Arm LRO library models are not included in `sdkPackage.models` 


archerzz:
### Describe the bug

See the discussion: https://github.com/Azure/typespec-azure/pull/1515#discussion_r1754770163

Even we set `filter-out-core-models: false`, some Arm LRO library models are not included in `sdkPakcage.models`, like `ArmOperationStatusResourceProvisioningState`.

### Reproduction

Parse an LRO operation defined by ARM LRO template. Check the `sdkPakcage.models` with configuration `filter-out-core-models: false`. You can find an example in the link above.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

tadelesh:
the full lro logic should be considered when handling tcgc models. will be resolved when consolidating models and operations conversion logic.

Issue #311, opened 2024-02-26
Type changed for `envelopeResult` from `getLroMetadata`


tadelesh:
When calling `getLroMetadata` for `createOrUpdate` for the [template resource](https://github.com/Azure/autorest.python/blob/2f5b207a2ea1046521c086155c3711cadc830c79/packages/typespec-python/test/azure/specification/mgmt/sphere/product.tsp#L18), I got `logicResult` with right TypeSpec type `Product`, but `envelopResult` with a wrong calculated TypeSpec type `ProxyResource` as follow screenshot.
After some investigation, I found the type is changed after calling `getEffectiveModelType` [here](https://github.com/Azure/typespec-azure/blob/00f500c4ba7cf3fcae628bfbad651c87695fc0a5/packages/typespec-azure-core/src/lro-info.ts#L102). Since the resource model has a path parameter, so it will enter the logic of finding a naming model which is same to the resource model without this path parameter (filter logic). When it has no exactly same model, it will also find a model that is same with it with the same filter logic, then the final result will fall to the original model. But after the complicated process, the type has totally changed and the naming info lost, it make the emitter hard to distinguish if it is the same resource model. Either fix the `getEffectiveModelType` logic for templated model or change the logic of `getLroMetadata` coudl resolve it. I don't know whether it should be opened as a core compiler problem or a `getLroMetadata` problem, If wrong place, I could move it.
![image](https://github.com/Azure/typespec-azure/assets/1726438/7bb7420e-7a72-4f30-9ecb-e5faf6be032e)

markcowl:
I think this is by design, the finalEnvelopeResult (which matches the finalResult) is the value that *should be used by emitters.

tadelesh:
Let me explain more clearly. The main problem here is for the following TypeSpec, when we call `getEffectiveModelType`, we will get the result type with same structure with `Product`, but wrong name `ProxyResource`. I suppose it should return the type of `Product`.
```
model Product is ProxyResource<ProductProperties> {
  @doc("Name of product.")
  @pattern("^[\\w][\\w\\s]{1,48}[\\w]$|^\\.default$|^\\.unassigned$")
  @key("productName")
  @path
  @segment("products")
  name: string;
}
```

tadelesh:
[...]

Issue #1750, opened 2024-10-28
[TCGC][LRO] LRO initial response model should not be generated


haolingdong-msft:
Link https://github.com/Azure/typespec-azure/issues/1538#issuecomment-2367621254, create a seperate issue to make it more clear.
### Issue
For below example, `LroOperationStatus` model is initial response, and should not be generated. But TCGC will return the model with name `LroOperationStatusError` and 'public' access, thus after adopting TCGC's access, this model will be generated, which is incorrect. 

```ts
model LroOperationStatus<TStatusResult = never, TStatusError = Foundations.Error> {
    @key("operationId")
    @visibility("read")
    id: Azure.Core.uuid;

    @visibility("read")
    @lroStatus
    status: JobStatus;

    @visibility("read")
    createdDateTime?: utcDateTime;

    @visibility("read")
    expirationDateTime?: utcDateTime;

    @visibility("read")
    lastUpdateDateTime?: utcDateTime;

    error?: TStatusError;
    result?: TStatusResult;
  }

  op LroLongRunningRpcOperation<
    TParams extends TypeSpec.Reflection.Model,
    TResponse extends TypeSpec.Reflection.Model,
    Traits extends {} = {}
  > is Azure.Core.RpcOperation<
    TParams & RepeatabilityRequestHeaders,
    Foundations.AcceptedResponse<LroLongRunningStatusLocation<TResponse> &
      Foundations.RetryAfterHeader> &
      RepeatabilityResponseHeaders &
      LroOperationStatus,
    Traits
  >;

  @route("/jobs/{id}")
  op getJob is LroLongRunningPollOperation<JobResult>;

  @pollingOperation(LongRunning.getJob)
  @route("/jobs")
  op createJob is LroLongRunningRpcOperation<
    {
      @body body: JobData;
    },
    JobResult
  >;
```

### Solution
TCGC returns usage `0` to indicate the model should not be generated.

ArcturusZhang:
I personally do not think so.
TCGC should not make assumptions how a SDK generator would use those models related with an operation.
A generator could decide to generate the initial response as part of its API, and this model is really one of the possible return types of this operation.
Therefore from TCGC's perspective, this is by design and correct.

haolingdong-msft:
If TCGC keeps the current usage and access for LRO intial response, we should have a way to let emitter know when to update access and do not generate this model. But this will make TCGC's access less helpful as emitter still need to update it based on usage.
/cc @ArcturusZhang  @tadelesh

PR #1887, opened 2024-11-19
Introduce more usages to describe things related with the states of LRO

ArcturusZhang:
Remove this workaround since #311 is closed.
Fixes https://github.com/Azure/typespec-azure/issues/1752
Fixes https://github.com/Azure/typespec-azure/issues/1530

This PR adds three new usage flags: `LroInitial`, `LroPolling`, `LroFinalEnvelope`.
These new usage flags are previously represented by `Output`, now the `Output` flag no longer contain the semantics of these usages.
Therefore, if a model's usage is only `LroInitial`, this means it is only used as the initial response of an LRO.
if a model has both `LroInitial` and `Output`, this means it is used as the initial response of an LRO, and it also used as a normal output in other places.
The final result will not have above new usages unless it participates as those roles as well. It will have output usage.


packages/typespec-client-generator-core/test/methods/lro.test.ts
```diff
@@ -359,18 +369,48 @@ describe("typespec-client-generator-core: long running operation metadata", () =
...
         ok(pollingModel);
+        strictEqual(pollingModel.usage & UsageFlags.Input, 0); // polling model should not be input
+        notStrictEqual(pollingModel.usage & UsageFlags.Output, 0); // polling model should be output
         strictEqual(metadata.pollingStep.responseBody, pollingModel);
 
         assert.isUndefined(metadata.finalResponse);
       });
     });
+
+    it("LRO defined in different namespace", async () => {
```
tadelesh, line 387: for this [issue](https://github.com/Azure/typespec-azure/issues/1752), two namespaces do not have hierarchy. could you change your test to that scenario.

tadelesh (approved):

packages/typespec-client-generator-core/test/methods/lro.test.ts
ArcturusZhang: updated

iscai-msft:
@ArcturusZhang are you ok with this being released as part of the november release? in 2 weeks? If so, I would merge into main. If you'd like us to release this before the big release, can you merge into our release branch `release/november-2024` and let us know, then we'll do a tcgc release for you. Thanks for your contribution!

haolingdong-msft:
The changes in this pr looks good to me. Just want to bring up this related issue: https://github.com/Azure/typespec-azure/issues/1750 - If TCGC keeps the current usage and access for LRO intial response, we should have a way to let emitter know how to update access and do not generate this model.

ArcturusZhang:
> The changes in this pr looks good to me. Just want to bring up this related issue: #1750 - If TCGC keeps the current usage and access for LRO intial response, we should have a way to let emitter know when to update access and do not generate this model. But this will make TCGC's access less helpful as emitter still need to update it based on usage.

I think per discussion this morning, maybe it is better that we could introduce some new usages for these lro related models since they have different context to use.

haolingdong-msft:
> I think per discussion this morning, maybe it is better that we could introduce some new usages for these lro related models since they have different context to use.

Is there any issue to track? (my created issue is closed as planned)

tadelesh:
> > I think per discussion this morning, maybe it is better that we could introduce some new usages for these lro related models since they have different context to use.
> 
> Is there any issue to track? (my created issue is closed as planned, feel free to reopen it)

i think we could do the usage split in this pr. @ArcturusZhang thoughts?

ArcturusZhang:
> > > I think per discussion this morning, maybe it is better that we could introduce some new usages for these lro related models since they have different context to use.
> > 
> > 
> > Is there any issue to track? (my created issue is closed as planned, feel free to reopen it)
> 
> i think we could do the usage split in this pr. @ArcturusZhang thoughts?

yeah I will update those in this PR

weidongxu-microsoft (commented):
I don't have opinion on this in Java. You can go ahead if no objection from other languages.

I assume the usage is additive. Meaning we cannot know whether a model is "only used in LroInitial"?

ArcturusZhang:
> I don't have opinion on this in Java. You can go ahead if no objection from other languages.
> 
> I assume the usage is additive. Meaning we cannot know whether a model is "only used in LroInitial"?

not quite additive.
The new usage (for example `LroInitial`) is part of the old `Output`. New `Output` no longer contains `LroInitial`.
Therefore if you see one model only has `LroInitial`, this means this model is only used as the initial response of an lro. The output now means everything other than those have an explicit definitions such as the lro initial.

weidongxu-microsoft:
> not quite additive. The new usage (for example `LroInitial`) is part of the old `Output`. New `Output` no longer contains `LroInitial`. Therefore if you see one model only has `LroInitial`, this means this model is only used as the initial response of an lro. The output now means everything other than those have an explicit definitions such as the lro initial.

Got it. So only the "final result" would still be marked as usage=Output?

nit, maybe we can call it e.g. `OutputLroInitial` (kind of grouping the "output" family)?

ArcturusZhang:
> > not quite additive. The new usage (for example `LroInitial`) is part of the old `Output`. New `Output` no longer contains `LroInitial`. Therefore if you see one model only has `LroInitial`, this means this model is only used as the initial response of an lro. The output now means everything other than those have an explicit definitions such as the lro initial.
> 
> Got it. So only the "final result" would still be marked as usage=Output?
> 
> nit, maybe we can call it e.g. `OutputLroInitial` (kind of grouping the "output" family)?

I think it is a good idea to emphasize. what do you think on this suggestion? @tadelesh 
An issue would be there are already some usage is part of the `input` usage but not named in this way (for instance `jsonmergepatch`)

tadelesh:
> > > not quite additive. The new usage (for example `LroInitial`) is part of the old `Output`. New `Output` no longer contains `LroInitial`. Therefore if you see one model only has `LroInitial`, this means this model is only used as the initial response of an lro. The output now means everything other than those have an explicit definitions such as the lro initial.
> > 
> > 
> > Got it. So only the "final result" would still be marked as usage=Output?
> > nit, maybe we can call it e.g. `OutputLroInitial` (kind of grouping the "output" family)?
> 
> I think it is a good idea to emphasize. what do you think on this suggestion? @tadelesh An issue would be there are already some usage is part of the `input` usage but not named in this way (for instance `jsonmergepatch`)

i'm thinking of separating the input and output usage into a second layer, which means:
1. for first layer, we label them all as output
2. for second layer, we separate output to something like:
```
enum OutputUsage {
  Normal,
  LroInitial,
  Error,
  ...
}
```
That will be a breaking change, but much extensible and we should make sure each layer orthogonal.
Loop in @iscai-msft for thoughts.

tadelesh (approved):

tadelesh:
open this issue to track usage refactor: https://github.com/Azure/typespec-azure/issues/1941

Issue #1905, opened 2024-11-23
[Bug]: tcgc path parameters missing


gracewilcox:
### Describe the bug

Hi @tadelesh!

I'm generating Go code with typespec and running into an issue where path parameters are not generating at all. This is happening in the multiple Key Vault packages I'm trying to create. ([Link to typspec project](https://github.com/Azure/azure-rest-api-specs/tree/f89003f8c49a68051c3fb34bdccd7cbabc6ddd37/specification/keyvault/Security.KeyVault.Keys))

This issue happens with the Go emitter @azure-tools/typespec-go v0.3.2 and later, the previous version (v0.3.1) works.

I've I noticed when the go emitter calls [`tcgc.createSdkContext`](https://github.com/Azure/autorest.go/blob/7afad1d855757335be8b6429990a71e1210f7a47/packages/typespec-go/src/tcgcadapter/adapter.ts#L29) the returned `sdkContext.sdkPackage.clients[0].methods[0].parameters` do not contain any path parameters. Because that array is missing the path parameters, the go emitter is not generating the methods with the path parameters.

### Reproduction

in bug description

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1907, opened 2024-11-25
[tcgc] use the right path parameter name when filtering method parameter

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1905

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -412,12 +412,12 @@ function getSdkHttpResponseAndExceptions(
   context: TCGCContext,
   httpOperation: HttpOperation,
 ): [
-  {
-    responses: SdkHttpResponse[];
-    exceptions: SdkHttpErrorResponse[];
-  },
-  readonly Diagnostic[],
-] {
+    {
+      responses: SdkHttpResponse[];
+      exceptions: SdkHttpErrorResponse[];
+    },
+    readonly Diagnostic[],
+  ] {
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -655,7 +655,9 @@ function filterOutUselessPathParameters(
       param.__raw &&
       isPathParam(context.program, param.__raw) &&
       httpOperation.parameters.parameters.filter(
-        (p) => p.type === "path" && p.name === getWireName(context, param.__raw!),
+        (p) =>
+          p.type === "path" &&
+          p.name === (getPathParamName(context.program, param.__raw!) ?? param.name),
```
ArcturusZhang, line 660: we have to use `__raw`?

ArcturusZhang: does this mean we should filter by the parameter's name instead of its wire name, or the other way around?

ArcturusZhang: I think maybe we should check the part that never changes, here we should be checking if there wire name is the same, something like:
```
p.wireName === getWireName(context, param.__raw!)
```

tadelesh: `getWireName` does not support path/query parameter. the goal here is to filter out method parameter which is a path parameter but not used in operation parameter list. so, i believe use query param name is reasonable.

ArcturusZhang (approved):

PR #1919, opened 2024-11-27
[tcgc] support typespec common paging

tadelesh:
support typespec common paging.

currently, typspec does not support using pageable decorator in nested properties, related test is skipped.

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -595,7 +595,7 @@ export interface SdkBasicServiceMethod<TServiceOperation extends SdkServiceOpera
 }
 
 interface SdkPagingServiceMethodOptions {
-  __raw_paged_metadata: PagedResultMetadata;
+  __raw_paged_metadata?: PagedResultMetadata;
   nextLinkPath?: string; // off means fake paging
   nextLinkOperation?: SdkServiceOperation;
 }
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -54,6 +54,7 @@ export interface TCGCContext {
   decoratorsAllowList?: string[];
   previewStringRegex: RegExp;
   disableUsageAccessPropagationToBase: boolean;
+  __pagedResultSet: Set<SdkType>;
 }
 
 export interface SdkContext<
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -133,10 +133,7 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
   context: TCGCContext,
   operation: Operation,
   client: SdkClientType<TServiceOperation>,
-): [
-  SdkPagingServiceMethod<TServiceOperation>,
-  readonly Diagnostic[],
-] {
+): [SdkPagingServiceMethod<TServiceOperation>, readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
 
   const basic = diagnostics.pipe(
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -134,9 +134,9 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
   operation: Operation,
   client: SdkClientType<TServiceOperation>,
 ): [
-  SdkPagingServiceMethod<TServiceOperation> | SdkServiceMethod<TServiceOperation>,
-  readonly Diagnostic[],
-] {
+    SdkPagingServiceMethod<TServiceOperation> | SdkServiceMethod<TServiceOperation>,
+    readonly Diagnostic[],
+  ] {
   const diagnostics = createDiagnosticCollector();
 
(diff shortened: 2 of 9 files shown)


packages/typespec-client-generator-core/test/packages/paged-operation.test.ts
```diff
@@ -15,33 +15,112 @@ describe("typespec-client-generator-core: paged operation", () => {
...
+      @list
+      op test(): ListTestResult;
+      model ListTestResult {
+        results: {
+          @pageItems
+          values: Test[];
+        };
+        pagination: {
+          @TypeSpec.nextLink
+          nextLink: string;
```
weidongxu-microsoft, line 109: A thing not related to this PR.

What `nextLinkPath` be, if this property be
```
`next.link`: string;
```

tadelesh: good catch. i think export it as an array is more reasonable.

tadelesh: https://github.com/Azure/typespec-azure/issues/1920

weidongxu-microsoft: My guess is the design want to use JSON path. Though I am not sure how JSON path express that. And I guess not many language has built support for JSON path anyway...

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -167,7 +238,42 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
...
+    if (prop.type.kind === "Model") {
+      queue.push(prop);
+    }
+  }
+  while (queue.length > 0) {
+    const current = queue.shift()!;
+    segments.push(current);
+    for (const prop of (current.type as Model).properties.values()) {
+      if (judge(prop)) {
+        segments.push(prop);
```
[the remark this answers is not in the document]
tadelesh: rewrite the alg. previous one is wrong.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -595,7 +596,7 @@ export interface SdkBasicServiceMethod<TServiceOperation extends SdkServiceOpera
 }
 
 interface SdkPagingServiceMethodOptions {
-  __raw_paged_metadata: PagedResultMetadata;
+  __raw_paged_metadata?: PagedResultMetadata;
```
weidongxu-microsoft, line 599: Any reason why this become optional?

weidongxu-microsoft (approved):
LGTM

There is a bunch of code in the PR, so best to have another reviewer to approve.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: unbranded part does not have this metadata. since it is not recommended to use raw, i think such breaking is acceptable. looks like it only affect .net, i'll let them to change the usage.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -130,27 +133,95 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
   context: TCGCContext,
   operation: Operation,
   client: SdkClientType<TServiceOperation>,
-): [SdkPagingServiceMethod<TServiceOperation>, readonly Diagnostic[]] {
+): [
+  SdkPagingServiceMethod<TServiceOperation> | SdkServiceMethod<TServiceOperation>,
```
ArcturusZhang, line 137: this would be a bad api - a method to get paging method could return either a paging method or a normal method.
Could we do it in a different way?
In your following implementation, we only return basic when the paging cannot be resolved properly.
can we return this still in the format of paging method, but its paging information being empty?

tadelesh: make sense.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -167,7 +241,45 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
   });
 }
 
-function getPathFromSegment(context: TCGCContext, type: Model, segments?: string[]): string {
+function getPropertyPathFromModel(
+  context: TCGCContext,
+  model: Model,
+  judge: (property: ModelProperty) => boolean,
+): string | undefined {
```
ArcturusZhang, line 248: is it possible that we use tcgc type here?
In each generator, we will use this path in those types, therefore it would be more reasonable we get this path from the same source.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -167,7 +241,45 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
   });
 }
 
-function getPathFromSegment(context: TCGCContext, type: Model, segments?: string[]): string {
+function getPropertyPathFromModel(
+  context: TCGCContext,
+  model: Model,
+  judge: (property: ModelProperty) => boolean,
```
ArcturusZhang, line 247: is `judge` a conventional name in js? In .Net, we usually use `predicate` as the name here.

tadelesh:
added test for `getPropertyPathFromModel`

ArcturusZhang (approved):

Issue #1474, opened 2024-09-04
Can `x-ms-azure-resource ` be safely removed


timotheeguerin:
There is some spec still using it and seems like csharp autorest generator might make use of it.
1. Figure out if this is actually useful
2. If it is, why are spec using it not using the base resource types
3. If they have valid legacy reason then we need some deprecated decorator to get parity

markcowl:
- This is still used by older emitters for c-sharp
- We need an ARM decorator that will be consumed by autorest to add this extension
- The decorator should be called `@customAzureResource` and have an associated linting rule that discourages its usage (this can be excluded from legacy rulesets)

live1206:
> This is still used by older emitters for c-sharp

@markcowl Could you specify where the usage is in older emitter of csharp? I couldn't find the usage in https://github.com/Azure/autorest.csharp

markcowl:
There are actually multiple places where this or a similar mechanism (configuring paths as resource paths) are used in c-sharp

pshao25:
We only found ComminutyGallery fits this situation. Do you have more examples?

PR #1923, opened 2024-11-28
@@customeAzureResource decorator for resources that doesn't use the base resource types

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1474


packages/typespec-azure-resource-manager/lib/decorators.tsp
```diff
@@ -182,6 +182,12 @@ extern dec armCommonTypesVersion(
  */
 extern dec armVirtualResource(target: Model);
 
+/**
+ * This decorator is used on resources that do not satisfy the definition of a resource
+ * but need to be identified as such.
+ */
+extern dec customAzureResource(target: Model);
```
markcowl, line 189: I think we want this in the namespace `Azure.ResourceManager.Legacy`

packages/typespec-azure-resource-manager/src/lib.ts
```diff
@@ -36,6 +36,12 @@ export const $lib = createTypeSpecLibrary({
         default: "Resource types must include a string property called 'name'.",
       },
     },
+    "arm-custom-resource-usage-discourage": {
+      severity: "warning",
+      messages: {
+        default: "Avoid using this decorator as it does not provide validation for ARM resources.",
```
markcowl, line 42:
```suggestion
        default: "Avoid using this decorator except when converting old APIs to TypeSpec, as it does not provide validation for ARM resources.",
```

packages/typespec-azure-resource-manager/src/rules/utils.ts
```diff
@@ -93,6 +93,14 @@ export function isInternalTypeSpec(
   );
 }
 
+export function isLegacyTypeSpec(
```
markcowl, line 96: I am not sure where this is used, but I would not expect Service TypeSpec to use `.Legacy` in their namespaces.

I think we should instead add a new linting rule that throws if folks use customAzureResource, and think about having a new 'legacy' ruleset that does not include this rule.  But I don't think the new ruleset needs to be in this PR

packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -92,6 +94,19 @@ export const $armVirtualResource: ArmVirtualResourceDecorator = (
   }
 };
 
+export const $customAzureResource: CustomAzureResourceDecorator = (
+  context: DecoratorContext,
+  entity: Model,
+) => {
+  const { program } = context;
+  if (isTemplateDeclaration(entity)) return;
+  if (!isLegacyTypeSpec(program, entity)) {
```
markcowl, line 103: I think we can just report this diagnostic.  Ideally, we would move this check into a linting rule, so we could eventually exclude it from a future 'arm brownfield'  ruleset

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/decorators.tsp
AlitzelMendez: moved

packages/typespec-azure-resource-manager/src/rules/utils.ts
AlitzelMendez: removed this check

packages/typespec-azure-resource-manager/src/resource.ts
AlitzelMendez: created the linting rule :)

markcowl (approved):

PR #1929, opened 2024-11-29
azure-http-specs, add clientNamespace

weidongxu-microsoft:
fix https://github.com/Azure/cadl-ranch/issues/758
move from https://github.com/Azure/cadl-ranch/pull/775

This scenario mainly test the case when `@clientNamespace` applied to TypeSpec Namespace/Interface/Model/Enum.

So far it does not test corner cases, like client/subclient in different namespace, name conflict for model in different namespace, etc.

Current cases include
1. clients in different clientNamespace
2. client and model in different clientNamespace
3. model and the model of its property in different clientNamespace


packages/azure-http-specs/specs/client/namespace/client.tsp
```diff
@@ -0,0 +1,31 @@
...
+using Client.ClientNamespace;
+
+@route("/client/client-namespace")
+namespace ClientNameSpaceClient;
+
+@client({
+  name: "ClientNamespaceFirstClient",
+  service: Client.ClientNamespace,
+})
+@clientNamespace("client.clientnamespace", "java")
```
tadelesh, line 16: is it possible to remove the scope? i believe every language needs to test ns customization.

weidongxu-microsoft: Every language would need to use slight different SDK namespace. E.g. csharp would use maybe `Client.ClientNamespace`.

So I think whenever we had `@clientNamespace`, it would be one line for one language.

tadelesh: @msyyc could you help to add python part?

msyyc: updated for python.

tadelesh (approved):

packages/azure-http-specs/specs/client/namespace/client.tsp
```diff
@@ -0,0 +1,36 @@
...
+@clientNamespace("client.clientnamespace", "python")
```
MaryGao, line 17: @v-jiaodi Could you help add js part?

v-jiaodi: pr link: https://github.com/Azure/typespec-azure/pull/2224

packages/azure-http-specs/specs/client/namespace/client.tsp
JoshLove-msft: I don't think we should have different customization like this per language. The emitters need to be able to use the correct casing. This isn't testing that namespace adheres to Azure standards for each language.

PR #1945, opened 2024-12-03
add client initialization docs

joheredi:


website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
```diff
@@ -555,3 +555,75 @@ petActionClient.pet();
...
+  blobName: string;
+}
+
+@@clientInitialization(Storage, StorageMockClientOptions);
+```
+
+```python
+from storage import StorageClient
+
+client = StorageClient(blob_name="myBlobName", endpoint="<my-endpoint>", ...)
```
msyyc, line 593: @iscai-msft I understand `blob_name` is keyword-only signature?

website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
```diff
@@ -555,3 +555,75 @@ petActionClient.pet();
...
+
+using Azure.ClientGenerator.Core;
+
+namespace Customizations;
+
+model StorageClientOptions {
+  blobName: string;
+}
+
+@@clientInitialization(Storage, StorageMockClientOptions);
```
weidongxu-microsoft, line 587: `StorageMockClientOptions` is wrong.

I've fixed in the PR with Java doc https://github.com/Azure/typespec-azure/pull/1949/files

website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
```diff
@@ -555,3 +555,75 @@ petActionClient.pet();
...
+With `@clientInitialization`, you can pass in additional parameters you would like your client to have, by passing in an input model. All properties of the input model will be appended to the current default list of client initialization parameters. Additionally, these client parameters will no longer appear on service methods that previously had them as part of the method signature. The generated code will automatically pass in the inputted value from the client init to the service.
+
+<ClientTabs>
+
+```typespec title="main.tsp"
+@service
+namespace Storage;
+
+op upload(blobName: string): void;
+op download(blobName: string): void;
```
weidongxu-microsoft, line 572: Can we use `@path` as an example (e.g. `@path blobName: string)`)?

There is currently limitation in Java emitter when this "option to clientInitialization" is a property within a request body (instead of a normal parameter like path/query/header).

iscai-msft: ok I'll update the docs

website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
iscai-msft: since i'm passing all by name order doesn't matter, but you're right it is kind of confusing, let me update the docs

lmazuel (approved):

website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
```diff
@@ -555,3 +555,77 @@ petActionClient.pet();
 ```
 
 </ClientTabs>
+
+### Adding Client Initialization Parameters
+
+By default, we only generate our clients with initialization parameters for `endpoint`, `credential`, and `apiVersion`, whenever any of these are applicable. There are cases where spec authors would like their clients to have additional input parameters. This is where the `@clientInitialization` decorator comes in.
```
RickWinter, line 561: Why are implicit and not explicit? Should we make TSP authors be explicit and declare the client initiziation takes these three and any others?

weidongxu-microsoft (approved):

website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
iscai-msft: it's because these three parameters are required and can't be removed, and we already know what they are. We don't want to make users have to re-enter these 3 parameters. If we required users to input them in, we would raise a warning if the users don't input it, but to me it seems better to just include them automatically, since you can't remove them.

iscai-msft: Going to resolve so we can merge this docs pr, @RickWinter if you still feel this is an issue, can you open a separate one? thanks!

RickWinter: Is the magic that the tsp compiler adds them? I wouldn't want to rely on emitters knowing to do this magic

weidongxu-microsoft: These are usually from different decorators in typespec source. For now, TCGC does not require dev explicitly write them in typespec clientInitialization (if it does, basically every single service need to add these bunch of parameters).

For example, `credential` be via `@useAuth`, e.g. https://github.com/weidongxu-microsoft/typespec/blob/java_migrate-specs/packages/http-specs/specs/authentication/api-key/main.tsp#L9
`endpoint` via `@server`, e.g. https://github.com/weidongxu-microsoft/typespec/blob/java_migrate-specs/packages/http-specs/specs/server/path/single/main.tsp#L9-L16
`api-version` is typically `Azure`, and from `@versioned`, e.g. https://github.com/weidongxu-microsoft/typespec/blob/java_migrate-specs/packages/http-specs/specs/versioning/added/main.tsp#L13

PR #1976, opened 2024-12-11
`typespec-azure` repository - Upload Manifest file to the new storage account

sarangan12:
In the `cadl-ranch` repository, during the publish process, the CI publishes the manifest file to the `azuresdkcadlranch` storage account. You can find the code [here](https://github.com/Azure/cadl-ranch/blob/main/.azure-pipelines/publish.yml#L35)

With the new Azure Spec Dashboard, the `typespec-azure` repository CI has to publish the manifest file to the `typespec` storage account. I have tested this code in CI in my branch. You can find the link [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4403105&view=logs&j=eab7c258-2ac5-5cd8-1109-0a296aef509c&t=99d957b2-3f06-5bc3-a8e6-5126b230307b).

Please review and approve the PR. Thanks


packages/azure-spec-dashboard/src/apis.ts
```diff
@@ -58,6 +59,32 @@ export function getManifestClient() {
...
+  } else {
+    if (
+      (generatorReports["azure"] as any)[key][0]["scenariosMetadata"].packageName ===
+      "@azure-tools/azure-http-specs"
+    ) {
+      return (generatorReports["azure"] as any)[key][0];
+    } else {
+      return (generatorReports["azure"] as any)[key][1];
+    }
+  }
```
chrisradek, line 85: I can't tell from looking at this why you're accessing different indices based on these conditions. Can you add some comments so it's clearer what is being returned in each case?

sarangan12: @chrisradek Refactored the code and added comment

chrisradek (approved):

PR #1978, opened 2024-12-12
[tcgc] add `@isApiVersion` decorator

### LINKED ISSUE #1706: [tcgc] Add a decorator to signal non-standard service api version patterns

For brownfield services with non-standard api version patterns we need a way to signal that the parameter or header is specifying the api version for the service in the generated code. We typically default api version parameters to the latest version in generated code, I would like a tcgc decorator that will signal when client libraries should treat a parameter or header like this.

For example, the storage blob service specifies and `x-ms-version` header on all requests for api versions. 

```
/** Specifies the version of the operation to use for this request. */
  @apiVersion
  @header("x-ms-version")
  version: string,
```

cc @iscai-msft 

### Checklist

iscai-msft:
fixes #1706

cc @jhendrixMSFT 

also, fix some typo for decorator docs.


packages/typespec-client-generator-core/test/decorators/is-api-version.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      `);
+      const sdkPackage = runner.context.sdkPackage;
+      // there will be no api version param on client, bc the service isn't versioned
+      const apiVersionClientParam = sdkPackage.clients[0].initialization.properties.find(
+        (x) => x.name === "version",
+      );
+      ok(!apiVersionClientParam);
+      const method = sdkPackage.clients[0].methods[0];
+      strictEqual(method.kind, "basic");
+      const apiVersionParam = method.parameters.find((x) => x.name === "version");
```
chunyu3, line 31: Isn't the apiVersion parameter bumped to the client level?

tadelesh: good catch. i think some logic is wrong for api version elevation. let me try to fix it tomorrow.

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
```diff
@@ -511,18 +511,19 @@ export type ParamAliasDecorator = (
...
  * @clientNamespace("ContosoClient")
  * namespace Contoso;
  * ```
  * @example
  * ```typespec
- * @clientName("ContosoJava", "java")
- * @clientName("ContosoPython", "python")
- * @clientName("ContosoCSharp", "csharp")
- * @clientName("ContosoJavascript", "javascript")
+ * @clientNamespace("ContosoJava", "java")
```
lmazuel, line 523: Why is this PR including things about clientNamespace?

tadelesh: i just noticed this nit for the `@clientNamespace` doc and fixed it.

tadelesh: @lmazuel if no concern, could you also help to approve this pr?

lmazuel (approved):
Approve on the principal reading doc and content (didn't read the actual code)

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
lmazuel: @tadelesh approved!

weidongxu-microsoft (approved):

PR #1991, opened 2024-12-17
Update labels.ts to include CLI/PSH area label

allenjzhang:


mikeharder (approved):

eng/config/labels.ts
```diff
@@ -32,6 +32,10 @@ export const AreaLabels = defineLabels({
     color: "846da1",
     description: "Issues for Azure specific ide features",
   },
+  "cli/psh": {
```
markcowl, line 35: Also need to add a record correlating the label with a path (or no path) in the AreaPaths structure on line 44 below

markcowl: Probably no path, in this case

PR #1998, opened 2024-12-18
Add `@scope` decorator

### LINKED ISSUE #964: [TCGC] Need a way to remove operation from TCGC

For .NET SDK, we have requirement to remove certain operations, such as `Operations.list` should always be removed as duplicated within resource manager.
Right now, the best approach we have is below, [details](https://azure.github.io/typespec-azure/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3yvIK3Jlc291cmNlLW1hbmFnZXIiOwoKdXNpbmcgVHlwZVNwZWMuSHR0cDvQFVJlc3TRFVbpAKnIG0HEZS5Db3JlzhJSx3dNxnbPL2xpZW50R2VuZXJhdG9yxz8KLyoqIENvbnRvc28gyEAgUHJvdmlkZXIg5gDAbWVudCBBUEkuICovCkBhcm3IIE5hbWVzcGFjZQpAc2VydmljZSh7CiAgdGl0bGU6ICLHWMguSHVi5gCFIiwKfSkKQOcBgmVkKOcA43MpCm7IUiBNaWNyb3NvZnQu0kfvALhBUEkgx01z5ACjZW51bSDIVCDkAJLELjIwMjEtMTAtMDEtcHJldmlld8g1xDQgIEB1c2VEZXBlbmRlbmN5KPUBU%2BgBjXMudjFfMF9QxkhfMSnEQGFybUNvbW1vbuQBu3PHKtdIyynLVDXESGDyAKlgLAp95gDzQegA9esBDiDoAm%2FkAMhtb2RlbCBFbXBsb3llZSBpcyBUcmFja2Vk6ACCPMgcUHJvcGVydGllcz7lASQuLukApuQB2FBhcmFtZXRlcskxPjvoAIbJX3DJRNJ8ymDpAYNBZ2Ugb2YgZcg%2F5QF4YWdlPzogaW50MzI7CscpQ2l0edIqY2l0eT86IHN0cuUDDccsUHJvZmls01lAZW5jb2RlKCJiYXNlNjR1cmwi5AFgcMYwPzogYnl0ZXPJSFRoZSBzdGF0dXPES3RoZSBsYXN0IOQAxWF0aW9u5QL%2BICBAdmlzaWJpbGl0eSgicmVhZMhVxBjlA6dTdGF0xF%2FlAaPMFOkBQMRrzDLFeOUAwmHpAc3Fb0Bscm%2FEO3VzCnVu5AKq0VTlAVzmARIs7ADKyEcgY3JlxCdyZXF1ZXN0IGhhcyBiZWVuIGFjY2VwdGVkxGcgIEHHDjogIsgLItZQacRA5AC06QDByETsAJw6ICLMD9pMdXBkYXTET8VDVccOOiAiyAvKO%2BkEn%2BkAxOYA3GTnAZpTdWNjZWVk5QDFyQzTP8U25AFNZmFpbMk%2BRsUNOiAixgncOHdhcyBjYW5jZco%2BQ8cP5AT2xwv%2FAUAgZGVsZekBgETEDeYA%2BcgL5AUm7wNvbW926gHK6QNxTW92ZVLHFegDasRzbW92xGhmcm9tIGxvY%2BYAvMVuxBPxA0jLM3RvzzF0b8ov9wCVc3BvbnPrBIPmAJbHFuwAl%2B4DZMU%2BxWTGfOYC7s1uaW50ZXJm5AYUT%2BgDjHMgZXh0ZW5kc%2FYG%2Bi7LKXt9CuUGtMgjyhsKQEDkAwVzcyjkAvNzcynLaugA3uYEvGdldOQBuEHKRlJlYWTsBRkg5wKMT3LlArjlAefLL0PHHVJlcGxhY2VBc3luY84%2F5QMI0DdQYXRjaFPMLCzzBV%2FFQOYCTs9A5QJLZVdpdGhvdXRPa9N3bGlzdEJ5yDBHcm91cM9ETMUiUGFyZW501DxTdWJzY3JpcOUCLcY7xjPMGcw56AYMIHNhbXBs6wMaYWPFRHRoYXTmAhfpBc90byBkaWZmZeQAhO8CmsUp7gCyQcVI7wE36wMlyA3mApzzAJJIRUFE6gX4xH5jaGVja%2BoAqmV4aXN0ZW7mB3Igxh5FyRTvAfvNHe4HTQ%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fresource-manager%22%5D%7D%7D)
```
@armResourceOperations
interface Employees {
  get is ArmResourceRead<Employee>;
}

@@access(Employees.get, Access.internal, "csharp");
```
But this operation still exists, .NET SDK need to handle it specifically.
Is it possible to remove this operation from TCGC result? for instance, 
```
@@access(Employees.get, Access.never, "csharp");
```
Or can we have a new decorator to remove an operation in TCGC?

### Checklist

live1206:
Resolves #964

The scope parameter of `@scope` decorator is not working as the normal decorator.
For a normal decorator, we need to get the decorator value from state map based on the scope and the current emitter name. If the value hit from the state map with current emitter name, it's a positive case, if not hit, it's a negative case.
```
@clientName("Renamed", csharp)
model A;
```
With the normal decorator scope handling, if the decorator is absent it is a negative case. For the case below, we don't need to handle `@clientname` decorator, it is a negative case.
```
model A;
```

But for `@scope` decorator, when it is absent, it means `AllScopes`, which is also a positive case. For below two cases, python emitter should include this operation, both of them are positive case.
```
op func: void
```
and 
```
@scope("!csharp")
op func: void
```

For incremental applying, the mechanism should be similar to the normal decorator scope handling.
- for the same scope, the later decorator value wins regardless it's defined with normal scope or scope negation

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -22,6 +22,7 @@ import {
   getClientNameOverride,
   getClientNamespace,
   getOverriddenClientMethod,
+  IsInScope,
   listClients,
   listOperationGroups,
   listOperationsInOperationGroup,
@@ -606,9 +607,11 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
   const diagnostics = createDiagnosticCollector();
   const retval: SdkMethod<TServiceOperation>[] = [];
   for (const operation of listOperationsInOperationGroup(context, client)) {
-    retval.push(
-      diagnostics.pipe(getSdkServiceMethod<TServiceOperation>(context, operation, sdkClientType)),
-    );
+    if (IsInScope(context, operation)) {
+      retval.push(
+        diagnostics.pipe(getSdkServiceMethod<TServiceOperation>(context, operation, sdkClientType)),
+      );
+    }
   }
   for (const operationGroup of listOperationGroups(context, client)) {
     // We create a client accessor for each operation group
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -22,7 +22,6 @@ import {
   getClientNameOverride,
   getClientNamespace,
   getOverriddenClientMethod,
-  IsInScope,
   listClients,
   listOperationGroups,
   listOperationsInOperationGroup,
@@ -607,11 +606,9 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
   const diagnostics = createDiagnosticCollector();
   const retval: SdkMethod<TServiceOperation>[] = [];
   for (const operation of listOperationsInOperationGroup(context, client)) {
-    if (IsInScope(context, operation)) {
-      retval.push(
-        diagnostics.pipe(getSdkServiceMethod<TServiceOperation>(context, operation, sdkClientType)),
-      );
-    }
+    retval.push(
+      diagnostics.pipe(getSdkServiceMethod<TServiceOperation>(context, operation, sdkClientType)),
+    );
   }
   for (const operationGroup of listOperationGroups(context, client)) {
     // We create a client accessor for each operation group
--- a/packages/typespec-client-generator-core/test/decorators.test.ts
+++ b/packages/typespec-client-generator-core/test/decorators.test.ts
@@ -2971,7 +2971,30 @@ describe("typespec-client-generator-core: decorators", () => {
   });
 
   describe("scope decorator", () => {
-    it("remove operation from csharp client", async () => {
+    it("include operation from csharp client", async () => {
+      const runnerWithCSharp = await createSdkTestRunner({
+        emitterName: "@azure-tools/typespec-csharp",
+      });
+      await runnerWithCSharp.compile(`
+        @service
+        namespace MyService {
+          @clientName("TestRenamed", "csharp")
+          model Test {
+            prop: string;
+          }
+          @scope("csharp")
+          op func(
+            @body body: Test
+          ): void;
+        }
+      `);
+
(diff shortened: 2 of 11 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -606,9 +607,11 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
   const diagnostics = createDiagnosticCollector();
   const retval: SdkMethod<TServiceOperation>[] = [];
   for (const operation of listOperationsInOperationGroup(context, client)) {
```
tadelesh, line 609: i think you should change `listOperationsInOperationGroup`, instead of filter operation in `pacakge.ts` and `types.ts`.

packages/typespec-client-generator-core/test/decorators.test.ts
```diff
@@ -2969,4 +2969,134 @@ describe("typespec-client-generator-core: decorators", () => {
       ok(testModel);
     });
   });
+
+  describe("scope decorator", () => {
```
tadelesh, line 2973: you may also test if the models used in operation have the correct usage.

live1206: Actually, the model does not exist in the output package, test updated.

tadelesh (approved):

Issue #1985, opened 2024-12-13
[Bug]: tcgc result metadata is not being extracted correctly


qiaozha:
### Describe the bug

For this tsp
```
@error
model Error {
    code: int32;
    message: string;
}
 
@pagedResult
model Bar {
    @items
    lists: string[];
    @TypeSpec.nextLink
    nextLink: string;
}
 
model Child extends Bar {
    message: string
}
 
@post
op test(): Error | Child; 
```

the itemsSegments and nextLinkSegments are correct, but the method.nextLinkPath and method.response.resultPath is empty string. 
![Image](https://github.com/user-attachments/assets/ffc840d0-b716-4e15-b04e-490573bd5abe)

### Reproduction

N/A

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

tadelesh:
it is a bug for tcgc to handling page result with inheritance.

PR #2004, opened 2024-12-19
[tcgc] get correct pageable info for azure pageable operation with inheritance return model

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1985

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -291,6 +291,9 @@ function getPropertyPathFromSegment(
   for (const segment of segments) {
     const property = current.properties.get(segment);
     if (!property) {
+      if (current.baseModel) {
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -291,6 +291,9 @@ function getPropertyPathFromSegment(
   for (const segment of segments) {
     const property = current.properties.get(segment);
     if (!property) {
+      if (current.baseModel) {
+        return getPropertyPathFromSegment(context, current.baseModel, segments);
+      }
```
weidongxu-microsoft, line 296: Should we do a `while` (if baseModel has another baseModel)?

tadelesh: if the base does not have the property we want and also has base, it will recursive call `getPropertyPathFromSegment`.

weidongxu-microsoft (approved):

Issue #1441, opened 2024-08-28
[TCGC] Could TCGC provide an easier way to map from `SdkMethodParameter` to  `SdkPathParameter | SdkQueryParameter | SdkHeaderParameter | SdkBodyParameter`?


haolingdong-msft:
Example:

TSP:
```ts
  @route("upload/{name}")
  @post
  uploadFile(
    @path name: string,
    @header contentType: "multipart/form-data",
    file_data: bytes,

    // this field should not be serialized to form-data
    @visibility("read") readOnly: string,

    constant: "constant",
  ): OkResponse;
```

Generated java sdk:
```java

    @Generated
    @ServiceMethod(returns = ReturnType.SINGLE)
    Response<Void> uploadFileWithResponse(String name, BinaryData uploadFileRequest, RequestOptions requestOptions) {
        // Protocol API requires serialization of parts with content-disposition and data, as operation 'uploadFile' is
        // 'multipart/form-data'
        return this.serviceClient.uploadFileWithResponse(name, uploadFileRequest, requestOptions);
    }

    @Generated
    @ServiceMethod(returns = ReturnType.SINGLE)
    public void uploadFile(String name, FileDataFileDetails fileData) {
        // Generated convenience method for uploadFileWithResponse
        RequestOptions requestOptions = new RequestOptions();
        UploadFileRequest uploadFileRequestObj = new UploadFileRequest(fileData);
        BinaryData uploadFileRequest = new MultipartFormDataHelper(requestOptions)
            .serializeFileField("file_data", uploadFileRequestObj.getFileData().getContent(),
                uploadFileRequestObj.getFileData().getContentType(), uploadFileRequestObj.getFileData().getFilename())
            .serializeTextField("constant", uploadFileRequestObj.getConstant())
            .end()
            .getRequestBody();
        uploadFileWithResponse(name, uploadFileRequest, requestOptions).getValue();
    }
```

code-model:
<img width="608" alt="image" src="https://github.com/user-attachments/assets/6a334dc2-4b84-4819-9a5b-d38fdd06b889">

TCGC response:
`SdkServiceMethod`'s `file_data` parameter

<img width="451" alt="image" src="https://github.com/user-attachments/assets/dd692c5b-cf8f-4c77-91da-e8b6714a931a">

`SdkHttpOperation`'s `bodyParameter` parameter
[...]

PR #2010, opened 2024-12-23
[tcgc] add `getHttpOperationParameter` helper function

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1441

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -523,6 +523,7 @@ export function getCorrespondingMethodParams(
 ): [SdkModelPropertyType[], readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
 
+  // 1. To see if the service parameter is a client parameter.
   const operationLocation = getLocationOfOperation(operation)!;
   let clientParams = context.__clientToParameters.get(operationLocation);
   if (!clientParams) {
@@ -539,6 +540,7 @@ export function getCorrespondingMethodParams(
     return diagnostics.wrap(correspondingClientParams);
   }
 
+  // 2. To see if the service parameter is api version parameter that has been elevated to client.
   if (serviceParam.isApiVersionParam && serviceParam.onClient) {
     const existingApiVersion = clientParams?.find((x) => isApiVersion(context, x));
     if (!existingApiVersion) {
@@ -556,6 +558,8 @@ export function getCorrespondingMethodParams(
     }
     return diagnostics.wrap(existingApiVersion ? [existingApiVersion] : []);
   }
+
+  // 3. To see if the service parameter is subscription parameter that has been elevated to client (only for arm service).
   if (isSubscriptionId(context, serviceParam)) {
     const subId = clientParams.find((x) => isSubscriptionId(context, x));
     if (!subId) {
@@ -574,13 +578,13 @@ export function getCorrespondingMethodParams(
     return diagnostics.wrap(subId ? [subId] : []);
   }
 
-  // to see if the service parameter is a method parameter or a property of a method parameter
+  // 4. To see if the service parameter is a method parameter or a property of a method parameter.
   const directMapping = findMapping(methodParameters, serviceParam);
   if (directMapping) {
     return diagnostics.wrap([directMapping]);
   }
 
-  // to see if all the property of service parameter could be mapped to a method parameter or a property of a method parameter
+  // 5. To see if all the property of the service parameter could be mapped to a method parameter or a property of a method parameter.
   if (serviceParam.kind === "body" && serviceParam.type.kind === "model") {
     const retVal = [];
     for (const serviceParamProp of serviceParam.type.properties) {
@@ -594,6 +598,7 @@ export function getCorrespondingMethodParams(
     }
   }
 
+  // If mapping could not be found,  TCGC will report error since we can't generate the client code without this mapping.
   diagnostics.add(
     createDiagnostic({
       code: "no-corresponding-method-param",
@@ -607,6 +612,12 @@ export function getCorrespondingMethodParams(
   return diagnostics.wrap([]);
 }
 
+/**
+ * Try to find the mapping of a service paramete or a property of a service parameter to a method parameter or a property of a method parameter.
+ * @param methodParameters
+ * @param serviceParam
+ * @returns
+ */
 function findMapping(
   methodParameters: SdkModelPropertyType[],
   serviceParam: SdkHttpParameter | SdkModelPropertyType,
@@ -615,7 +626,7 @@ function findMapping(
   const visited: Set<SdkModelType> = new Set();
   while (queue.length > 0) {
     const methodParam = queue.shift()!;
-    // http operation parameter/body parameter/property of body parameter could either from an operation parameter directly or from a property of an operation parameter
+    // HTTP operation parameter/body parameter/property of body parameter could either from an operation parameter directly or from a property of an operation parameter.
     if (
       methodParam.__raw &&
       serviceParam.__raw &&
@@ -623,7 +634,7 @@ function findMapping(
     ) {
       return methodParam;
     }
-    // this following two hard code mapping is for the case that TCGC help to add content type and accept header is not exist
+    // Two following two hard coded mapping is for the case that TCGC help to add content type and accept header when not exists.
     if (
       serviceParam.kind === "header" &&
       serviceParam.serializedName === "Content-Type" &&
@@ -638,6 +649,7 @@ function findMapping(
     ) {
       return methodParam;
     }
+    // BFS to find the mapping.
     if (methodParam.type.kind === "model" && !visited.has(methodParam.type)) {
       visited.add(methodParam.type);
       let current: SdkModelType | undefined = methodParam.type;
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -535,7 +535,9 @@ export function getCorrespondingMethodParams(
       twoParamsEquivalent(context, x.__raw, serviceParam.__raw) ||
       (x.__raw?.kind === "ModelProperty" && getParamAlias(context, x.__raw) === serviceParam.name),
   );
-  if (correspondingClientParams.length > 0) return diagnostics.wrap(correspondingClientParams);
+  if (correspondingClientParams.length > 0) {
+    return diagnostics.wrap(correspondingClientParams);
+  }
 
   if (serviceParam.isApiVersionParam && serviceParam.onClient) {
     const existingApiVersion = clientParams?.find((x) => isApiVersion(context, x));
@@ -552,7 +554,7 @@ export function getCorrespondingMethodParams(
       );
       return diagnostics.wrap([]);
     }
-    return diagnostics.wrap(clientParams.filter((x) => isApiVersion(context, x)));
+    return diagnostics.wrap(existingApiVersion ? [existingApiVersion] : []);
   }
   if (isSubscriptionId(context, serviceParam)) {
     const subId = clientParams.find((x) => isSubscriptionId(context, x));
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -55,16 +55,6 @@ export interface TCGCContext {
   previewStringRegex: RegExp;
   disableUsageAccessPropagationToBase: boolean;
   __pagedResultSet: Set<SdkType>;
-  __paramMapping: Map<
-    SdkMethodParameter,
-    (
-      | SdkPathParameter
-      | SdkQueryParameter
-      | SdkHeaderParameter
-      | SdkCookieParameter
-      | SdkBodyParameter
-    )[]
-  >;
 }
 
 export interface SdkContext<
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -55,6 +55,7 @@ export interface TCGCContext {
   previewStringRegex: RegExp;
   disableUsageAccessPropagationToBase: boolean;
   __pagedResultSet: Set<SdkType>;
+  __paramMapping: Map<SdkMethodParameter, (SdkPathParameter | SdkQueryParameter | SdkHeaderParameter | SdkCookieParameter | SdkBodyParameter)[]>;
 }
 
 export interface SdkContext<
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -55,7 +55,16 @@ export interface TCGCContext {
   previewStringRegex: RegExp;
   disableUsageAccessPropagationToBase: boolean;
   __pagedResultSet: Set<SdkType>;
-  __paramMapping: Map<SdkMethodParameter, (SdkPathParameter | SdkQueryParameter | SdkHeaderParameter | SdkCookieParameter | SdkBodyParameter)[]>;
+  __paramMapping: Map<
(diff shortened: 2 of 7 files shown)


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +708,59 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
+      }
+    }
+    if (operation.bodyParam) {
+      for (const cp of operation.bodyParam.correspondingMethodParams) {
+        if (cp === param) {
+          result.push(operation.bodyParam);
+        }
+      }
+    }
+    if (result.length === 0 && param?.type.kind === "model" && !visited.has(param.type)) {
```
archerzz, line 754: Is this a case of spread?

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +708,59 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
+  while (result.length === 0 && queue.length > 0) {
+    const param = queue.pop();
+    for (const p of operation.parameters) {
+      for (const cp of p.correspondingMethodParams) {
+        if (cp === param) {
+          result.push(p);
+        }
+      }
+    }
+    if (operation.bodyParam) {
```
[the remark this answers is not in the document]
tadelesh: yes. tcgc separate the body parameter with other http parameters.

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +708,59 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
+  const result: (
+    | SdkPathParameter
+    | SdkQueryParameter
+    | SdkHeaderParameter
+    | SdkCookieParameter
+    | SdkBodyParameter
+  )[] = [];
+  const queue: SdkModelPropertyType[] = [param];
+  const visited: Set<SdkModelType> = new Set();
+  while (result.length === 0 && queue.length > 0) {
```
weidongxu-microsoft, line 738: I didn't quite get the while condition. In which case it would get into 2nd loop?

Better to add a comment on it.

packages/typespec-client-generator-core/src/public-utils.ts
weidongxu-microsoft: Seems grouping? Again, better to add comment.

tadelesh: it is for `@bodyRoot`, for example:
```
import "@typespec/http";
using Http;
model TestRequest {
  @header
  h: string;

  @query
  q: string;

  prop1: string;
  prop2: string;
}
op test(@bodyRoot request: TestRequest): void;
```

packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts
```diff
@@ -0,0 +1,420 @@
...
+      model Shelf {
+        name: string;
+        theme?: string;
+      }
+      model CreateShelfRequest {
+        @body
+        body: Shelf;
+      }
+      op createShelf(...CreateShelfRequest): Shelf;
+    `);
```
weidongxu-microsoft, line 145: There is no `@bodyRoot` in this test?

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: added a test with `@override`.

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: comment added.

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +707,61 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
+  const queue: SdkModelPropertyType[] = [param];
+  const visited: Set<SdkModelType> = new Set();
+  // BFS to find the corresponding http parameter.
+  // An http parameter will be mapped to a method/client parameter, several method/client parameters (body spread case), or one property of a method property (metadata on property case).
+  // So, when we try to find which http parameter a method parameter corresponds to, we compare the `correspondingMethodParams` list directly.
+  // When we try to find which http parameter a property corresponds to, we need to consider the following cases:
+  // 1. The property itself is an http parameter.
+  // 2. The property is a model, and the properties of the model are http parameters, recursively.
+  // 3. The property is a model, and the base model of the model has http parameters, recursively.
+  // When we find one, we return directly since a method/client parameter or property could only be used in one http parameter.
```
weidongxu-microsoft, line 738: Is this assert valid for below case?

```
      model Params {
        @query foo: string;
        @query bar: string;
      }
      op func(...Params): void;
      `,
      `
      namespace MyCustomizations;
      op func(params: MyService.Params): void;
      @@override(MyService.func, MyCustomizations.func);
```

tadelesh: i think so. for `params`, its corresponding http parameters is `undefined`. you need to get them with `foo` and `bar` property. see my last test.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -55,6 +55,16 @@ export interface TCGCContext {
   previewStringRegex: RegExp;
   disableUsageAccessPropagationToBase: boolean;
   __pagedResultSet: Set<SdkType>;
+  __paramMapping: Map<
```
weidongxu-microsoft, line 58: I didn't see it used anywhere?

packages/typespec-client-generator-core/src/public-utils.ts
weidongxu-microsoft: Ok, so a method parameter would map to at most 1 parameter, and strictly an HTTP parameter (body param or query param or header param or `undefined`, but not to property in body).

It is somewhat unexpected that one can go to the properties inside the model and call `getHttpOperationParameter` on it. Let's see what other reviewers think.

tadelesh: a little bit force and back for this. just want to make it simple. current spec does not contain such mix usage at least.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: removed.

packages/typespec-client-generator-core/src/public-utils.ts
weidongxu-microsoft: One concern is, if I call the method, and find the parameter maps to a kind=body, I seems won't be able to tell at this state, whether the parameter is the body, or it actually be a property within that body?

archerzz (approved):

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: currently, it will tell you this parameter is used as body. but it could not tell you the property of this parameter is also used as a query patameter.

weidongxu-microsoft: What I mean is,
```
model Model { key: string }

op myOp(@header h: string, @query q: string, @path p: string, @body body: Model): void;
```
call on "body" method param would map to `body` body param of the `Model`.

And
```
op myOp(@header h: string, @query q: string, @path p: string, ...Model): void;
```
call on "key" method param would also map to `body` body param of the `Model`.
```

Maybe it is not that bad, maybe we can use the mapping from the other side to determine...

Comment (tadelesh) [reply packages/typespec-client-generator-core/src/public-utils.ts:738]: i got your point. you mean there is no way to tell how to map the exact value. then you are right, tcgc suggest to map from http parameter to method parameter.

**Review by weidongxu-microsoft — APPROVED** [review]
We can go with it, if no one objects.

Though I'd expect refinements or breaks, when dev start using it.

**Comment by tadelesh** [comment]
@qiaozha please help to review. current solution is kind of a tradeoff which should meet all language's requirement, but not so convenience for exact value mapping.

File: packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts [hunk 7]
```diff
@@ -0,0 +1,511 @@
...
+
+    strictEqual(parameters.length, 3);
+
+    for (const param of parameters) {
+      if (param.name === "body") {
+        const httpParam = getHttpOperationParameter(method, param);
+        ok(httpParam);
+        strictEqual(httpParam.kind, "body");
+        strictEqual(httpParam.serializedName, "body");
+      } else if (param.name === "contentType") {
```
qiaozha, line 110: I remmebr contentType is generated by tcgc in this case, is the result suppose to be undefined ?

packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts
```diff
@@ -0,0 +1,511 @@
...
+        const httpParam = getHttpOperationParameter(method, param);
+        ok(httpParam);
+        strictEqual(httpParam.kind, "header");
+        strictEqual(httpParam.serializedName, "Content-Type");
+      } else if (param.name === "accept") {
```
qiaozha, line 115: the same concern for accept parameter

packages/typespec-client-generator-core/src/public-utils.ts
qiaozha: I tend to agree with weidong that we should provide such information whether a parameter is from the body property or from the body itself otherwise I don't know how to handle the serialization logic for spread body model with only one property inside it ?

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +707,61 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
+      }
+    }
+    if (param?.kind === "property" && param?.type.kind === "model" && !visited.has(param.type)) {
+      visited.add(param.type);
+      let current: SdkModelType | undefined = param.type;
+      while (current) {
+        for (const prop of param.type.properties) {
+          queue.push(prop);
+        }
+        current = current.baseModel;
```
qiaozha, line 762: just a little curious, what would we get if A extends B and we are spreading A, and we are querying the properties from B ?

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +707,61 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
 }
+
+/**
+ * Find corresponding http parameter list for a client initialization or service method parameter or the property of that parameter.
+ *
+ * @param method
+ * @param param
+ * @returns
+ */
+export function getHttpOperationParameter(
```
qiaozha, line 718: wonder if the performance would be a concern here?

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -552,8 +556,10 @@ export function getCorrespondingMethodParams(
       );
       return diagnostics.wrap([]);
     }
-    return diagnostics.wrap(clientParams.filter((x) => isApiVersion(context, x)));
+    return diagnostics.wrap(existingApiVersion ? [existingApiVersion] : []);
   }
+
+  // 3. To see if the service parameter is subscription parameter that has been elevated to client (only for arm service).
```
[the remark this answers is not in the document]
tadelesh: you could also pass the client initialization parameter to this function. but you also need to pass a method, since we need to find the corresponding http operation parameter. see the `client parameter` test.

packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts
tadelesh: tcgc will also add `content-type` and `accept` header parameter. so the mapping could work.

packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts
tadelesh: see above reply.

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: then maybe i need to give you a segment array?

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: i think it's ok. since we already have the corresponding map, it may take some time but not too much.

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: current logic is for parameter type. when you spread A, the parameter will be all the properties of A and B, then all parameters will get separated mapping.

packages/typespec-client-generator-core/src/public-utils.ts
qiaozha: I think that works. And it will also works if we want to support group a few parameters into one parameter like @weidongxu-microsoft proposed before?

tadelesh: @qiaozha @weidongxu-microsoft i changed the impl to return http body parameter or the property of http body parameter. then it could handle the spread case.

weidongxu-microsoft: LGTM.

So now in spread case, we will find the mapped type is a ModelProperty, and its model would always be `operation.bodyParam`.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: correct.

packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts
qiaozha: how can we know if the http content-type and accept header is added by tcgc or not? do we have isGeneratedName in header parameter?

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -698,3 +707,48 @@ export function isAzureCoreModel(t: SdkType): boolean {
...
+    for (const cp of p.correspondingMethodParams) {
+      if (cp === param) {
+        return p;
+      }
+    }
+  }
+  if (operation.bodyParam) {
+    for (const cp of operation.bodyParam.correspondingMethodParams) {
+      if (cp === param) {
+        if (operation.bodyParam.type.kind === "model" && operation.bodyParam.type !== param.type) {
```
qiaozha, line 744: just a little curious, what would happen if this body parameter is spread of record?

packages/typespec-client-generator-core/src/public-utils.ts
qiaozha: it doesn't seem to work for the group cases? should we consider that in the future?

tadelesh: it is by design. current helper deal with the 1-1 mapping of parameter/property. if a parameter is a grouping, then you need to call with the parameter property instead of parameter.

packages/typespec-client-generator-core/test/public-utils/get-http-operation-parameter.test.ts
tadelesh: yes

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: whatever kind of spread, the body parameter will always be a model. if spread, the type will be different.

qiaozha (approved):

Issue #1696, opened 2024-10-16
[tcgc] have `@clientInitialization` decorator take in optional `access` input


iscai-msft:
### Clear and concise description of the problem

Currently, the `@clientInitialization` decorator solely takes in a required model input. However, as we are trying to move away from the `@operationGroup` decorator, we want to use the `@clientInitialization` decorator to also specify whether a client is publicly or internally instantiable. We are currently using client hierarchy and the `@operationGroup` decorator to determine whether the initialization on a client is public or not. Instead, we should almost do the opposite. Get rid of the `@operationGroup` decorator, based off of client hierarchy, we can default over whether a client is publicly or internally instantiable, but allow users to specify the access level of initialization.

// PROPOSED CHANGES TO @CLIENTINITIALIZATION DECORATOR

// Client with public initialization -> maps to current def of client
// Client with internal initialization -> maps to current def of operation group

// current design for @clientInitialization(client, model, scope?)
// proposed designs for 

// 1. @clientInitialization(client, model?, scope?)
// 2. @clientInitialization(client, access?, scope?)
// 3. @clientInitialization(client, model?, access?, scope?)

// @clientInitialization(client, {model?: Model, access?: AccessFlags}, scope?: string)

// @@operationGroup(MyService) -> @@clientInitialization(MyService, {access: "internal"})
// @@clientInitialization(MyService, {model: MyClientInitialization})

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

weidongxu-microsoft:
What would you write for an Interface as subclient (with clientInitialization access=internal, as usual) without `@operationGroup`?

weidongxu-microsoft:
Also, I assume there is a complementary concept in SdkClientAccessor

```
export interface SdkClientAccessor<TServiceOperation extends SdkServiceOperation>
  extends SdkMethodBase {
  kind: "clientaccessor";
  response: SdkClientType<TServiceOperation>;
}
```

[...]

Issue #1715, opened 2024-10-21
[tcgc] add client accessor support to `client.tsp`


iscai-msft:
Currently, TCGC doesn't support public client accessors from parent to subclient. 

The canonical use case is something like `storage`, where there is the access from parent `BlobServiceClient` to child `ContainerClient`, with `BlobServiceClient.getContainerClient(containerName: string)` as a client accessor between the two.

By default, with parents and subclients, unless an explicit client accessor method is defined, there is no generated public accessor method from parent to subclient. If a user wants a public accessor method from parent to subclient, they will define it in `client.tsp` like so

```tsp
// client.tsp

op getContainerClient(containerName: string): Azure.Storage.Blob.Container;
```

If you want to define a client accessor and have `access` still be `internal`, the `@access` decorator can still be applied. This use case would be if there are special internal parameters that you have to pass to the subclient that we don't want users to deal with

## TODO

1. We need to make sure that clients / namespaces / interfaces can be listed as a return type in an operation. This might require working with the typespec team.

iscai-msft:
I still feel that having an explicit definition is better, instead of going down a logic map way of initialization and accessor. I guess the root of the issue is, besides storage and some other individual examples, we don't have a good feel if a public client has a public subclient, whether we have to add a separate ctor.

The thing with explicitly defining is we don't have to tie ourselves to logic. I don't think we're 100% blocked by typespec support actually, we can return namespaces and interfaces from operations (ex [here](https://typespec.io/playground/?c=CkBzZXJ2aWNlCm5hbWVzcGFjZSBBIHsKCiAgb3Agc3ViTsgXQ2xpZW50KCk6IEI7yR5JbnRlcmbNHkM7xD3KTkIgewogIH3EFWnILiBDxxV9Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)). Finally, I also think we want to give support to spec authors to explicitly state what parameters should get passed from parent to child etc

weidongxu-microsoft:
First, wondering whether we have different logic for accessor without parameter and accessor with parameter. My following comment is on assumption that we didn't distinguish the 2. Both are just client accessor.

Now, wondering for how many languages, this statement is currently true?
> By default, with parents and subclients, unless an explicit client accessor method is defined, there is no generated public accessor method from parent to subclient.

In this default mode, then how the subclient get initialized?

[...]

Issue #1707, opened 2024-10-18
[tcgc] @clientInitialization not properly handling parameters


catalinaperalta:
### Clear and concise description of the problem

Discussed this with @iscai-msft offline, seems that when we specify a parameter that should be initialized on the client like in the example below the generated code does add it to the client, however the parameter isn't removed from the operation itself. 

Client initialization example:
```
// client.tsp

model StorageClientOptions {
  /** Specifies the version of the operation to use for this request. */
  version: string;
}

@@clientInitialization(Storage.Blob, StorageClientOptions);
```

Example output in python:

Client looks good: 
![Image](https://github.com/user-attachments/assets/5d2192d3-3e2c-4bee-bad3-bba08abf39dd)

Operations **shouldnt** have the version parameter anymore:
![Image](https://github.com/user-attachments/assets/23d6ad27-4ff5-4765-a6e0-3b176fbba882)

TCGC should remove the version references from operations.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #1702, opened 2024-10-17
[tcgc] design for Java emitter to influence the default access property of SdkInitializationType and SdkClientAccessor


weidongxu-microsoft:
Upon https://github.com/Azure/typespec-azure/issues/1696

## Context

In TCGC, I take
- `SdkInitializationType.access` means whether the client/subclient is directly initializable (from builder or client class ctor).
- `SdkClientAccessor.access` means whether the subclient can be accessed by this client access method (from parent client of this subclient).
- `SdkInitializationType.model.access` means whether the model of the `SdkInitializationType` is public or not.

In Java, there is currently 2 pattern when accessing a subclient.
1. "client builder pattern" as
```java
// client
Client client = new ClientBuilder().buildClient();

// subclient via builder (equivalent to client initialization in other languages)
SubClient subclient = new ClientBuilder()
  .parameter(<param>)
  .buildSubClient();
```

I assume it maps to TCGC
```
// subclient initialization type
subclient.initialization.access=public;

// client access method from (parent) client
client.<clientaccessor_method>.access=internal;
```

2. "client accessor pattern" as
```java
// client
Client client = new ClientBuilder().buildClient();

// subclient via client accessor
SubClient subclient = client.getSubClient(<param>);
```

I assume it maps to TCGC
```
// subclient initialization type
subclient.initialization.access=internal;

// client access method from (parent) client
client.<clientaccessor_method>.access=public;
```

At present, DPG and unbranded by default uses "client builder pattern". And we also need certain configure to switch to "client accessor pattern" for some modules (e.g. Azure Face API, unbranded OpenAI)

And currently the TCGC value of `SdkInitializationType.access` and `SdkClientAccessor.access` is not what Java expected. Therefore, Java emitter does not use the TCGC value for now.
It would infer the client/subclient from whether the client has parent client.

PS: Java emitter would still treat the `SdkInitializationType.model.access` as "internal" or "none", as Java uses its properties directly and not the model (meaning there is currently no `##ClientOptions` class for Java).

## With "access" in `@clientInitialization`

Afterward, Java emitter must always use the TCGC value, as it cannot tell whether service has overridden the default `@clientInitialization`'s "access" value.

## Problem to solve

Java emitter would still want to affect TCGC to provide the expect default value (when service did not set the "access"). And this default is different from other languages.

[...]

PR #2027, opened 2024-12-26
[tcgc] client initialization and accessor refactor

tadelesh:
1. Change `@clientInitialization` decorator's `options` parameter to `ClientInitializationOptions` type. The options now could set the client initialization method's access and client accessor's access. This is a behavior breaking change for this decorator. All specs that use this decorator should change from `@clientInitialization(CustomizedOption)` to `@clientInitialization({parameters: CustomizedOption})
2. `getClientInitialization` now return `ClientInitializationOptions` types. It includes the info of the client initialization parameter list, the client initialization access value and the client accessor access value.
3. Add `clientInitialization` property to `SdkClientType`. It indicate the initialization method's signature, including the parameter list and access, etc.

Related design could be seen in this gist: https://gist.github.com/tadelesh/bb1b04501d25dd1f19bc5b164fb79810.

Also resolve: #1696, #1518, #1702, #1707, #1715


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -524,45 +523,63 @@ function getClientDefaultApiVersion(
...
+  const name = `${client.kind === "SdkClient" ? client.name : getLibraryName(context, client.type)}Options`;
+  const initializationModel: SdkInitializationType = {
+    __raw: client.service,
+    doc: "Initialization class for the client",
+    kind: "model",
+    properties: [],
+    name,
+    isGeneratedName: true,
+    access: initializationOptions?.access ?? (client.kind === "SdkClient" ? "public" : "internal"),
+    usage: UsageFlags.Input,
```
weidongxu-microsoft, line 543: How can emitter filter out this model?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -698,6 +708,10 @@ export interface SdkClientAccessor<TServiceOperation extends SdkServiceOperation
   response: SdkClientType<TServiceOperation>;
 }
 
+export interface SdkClientInitializationMethod extends SdkMethodBase {
+  kind: "clientinitialization";
+}
```
weidongxu-microsoft, line 713: Another method for Java emitter to filter out.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -95,7 +95,11 @@ export interface SdkClientType<TServiceOperation extends SdkServiceOperation>
   clientNamespace: string; // fully qualified namespace
   doc?: string;
   summary?: string;
+  /**
+   * @deprecated Use `clientInitialization` instead.
+   */
   initialization: SdkInitializationType;
+  clientInitialization: SdkClientInitializationMethod;
```
weidongxu-microsoft, line 102: It is a bit strange that `SdkClientInitializationMethod` is here, where `SdkClientAccessor<>` appears in `methods`?

packages/typespec-client-generator-core/src/package.ts
tadelesh: it is not included in the `sdkPackage.models` and is deprecated now.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: it is not in the `methods` list. you could use as you will.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: i also want to move client assessor method out of the method list. just tbd.

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -449,17 +449,37 @@ extern dec useSystemTextJsonConverter(target: Model, scope?: valueof string);
...
+ * Client initialization options.
+ */
+model ClientInitializationOptions {
+  /**
+   * Redefine the client initialization parameters.
+   */
+  parameters?: Model;
+
+  /**
+   * Determines the accessibility of the client initialization method.
```
msyyc, line 473:
```suggestion
   * Determines the accessibility of the client initialization method. "public" means the client could be initialized separately.
```

msyyc: Additional annotation could make it easier to understand.

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -449,17 +449,37 @@ extern dec useSystemTextJsonConverter(target: Model, scope?: valueof string);
...
+   */
+  access?: EnumMember;
+
+  /**
+   * Determines the accessibility of the client accessor method.
```
msyyc, line 478:
```suggestion
   * Determines the accessibility of the client accessor method. "public" means client could be got from parent client.
```

tadelesh:
two things to be discussed:
1. shall we put initialization method into methods list or move client accessor method out of methods list to keep consistency?
2. if client access is public, shall we put it to be the top-level client?

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -449,17 +449,37 @@ extern dec useSystemTextJsonConverter(target: Model, scope?: valueof string);
...
+
+  /**
+   * Determines the accessibility of the client initialization method. `Access.public` means the client could be initialized separately.
+   */
+  access?: EnumMember;
+
+  /**
+   * Determines the accessibility of the client accessor method. `Access.public` means client could be got from parent client.
+   */
+  accessorAccess?: EnumMember;
```
iscai-msft, line 480: i don't think this name is very clear, but I don't have a great suggestion. Accessible from parent? I'm just not sure that a spec author will have an understanding of "accessor" like us tcgc authors do

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1029,36 +1033,56 @@ const clientInitializationKey = createStateSymbol("clientInitialization");
...
-    properties: initializationProps,
+    parameters: model
+      ? model.properties.map((property: SdkModelPropertyType): SdkMethodParameter => {
+          property.onClient = true;
+          property.kind = "method";
+          return property as SdkMethodParameter;
+        })
+      : undefined,
+    access: options.properties.get("access")
+      ? options.properties.get("access").type.value
```
iscai-msft, line 1081: nit: I think you can do `options.properties.get('access')?.type?.value || undefined`, or maybe you don't even need `|| undefined`

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: are we doing the breaking change for `initialization` -> `clientInitialization` so we are the same as the decorator? I'm not 100% sure we need to make this change, but it's not a hill I'm going to die on

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -525,45 +524,63 @@ function getClientDefaultApiVersion(
...
+    let clientParams = context.__clientToParameters.get(client.type);
+    if (!clientParams) {
+      clientParams = [];
+      context.__clientToParameters.set(client.type, clientParams);
+    }
+    for (const param of initializationOptions.parameters) {
+      clientParams.push(param);
+    }
+  }
+  const name = `init${client.kind === "SdkClient" ? client.name : getLibraryName(context, client.type)}`;
```
iscai-msft, line 571: since the name doesn't super matter, should we just call it "init"? I think in terms of conflicts, since each one will be for a separate client, it should be fine

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -590,34 +607,38 @@ function getSdkMethods<TServiceOperation extends SdkServiceOperation>(
...
-      response: operationGroupClient,
-      apiVersions: getAvailableApiVersions(context, operationGroup.type, client.type),
-      crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, operationGroup.type)}.${name}`,
-      decorators: [],
-    });
   }
   return diagnostics.wrap(retval);
 }
 
+function createClientAccessorMethod<TServiceOperation extends SdkServiceOperation>(
```
iscai-msft, line 617: just to clarify, we still have client accessor methods, denoting the relationship between a parent and child client. At the same time, instead of treating the initialization information in tcgc as a pure model, we are looking at it more like another operation. I like this way of thinking, I'm just trying to make sure I'm reading this correctly

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,379 @@
...
+There is no way to indicate that a sub client could not be initialized from the parent client.
+3. `@operationGroup` is an Azure concept, which means a sub client that could not be initialized directly. 
+It is better to consolidate `@client` and `@operationGroup`.
+
+## New proposal
+
+1. Change `@clientInitialization` decorator and add `clientInitialization` property to `SdkClientType`
+- Change `@clientInitialization` decorator's `options` parameter to `ClientInitializationOptions` type to accept `access` and `accessorAccess` setting.
+- Add `clientInitialization` property to `SdkClientType`. It indicate the initialization method's signature, including the parameter list and access.
+- When `access` set to `Access.public`, it means this client could be initialized separately, while `Access.internal` could not. Default to `Access.public` for client and `Access.internal` for sub client.
```
iscai-msft, line 218: can you make clear that this is the `access` for the initialization property?

With this design, there seem to be 3 accesses associated with a `client`.
1. The access of the `client` itself. This can't be toggled rn, it's always `public`
2. The access of the client's init. If `public`, can directly init, else you can not
3. The access of the client's accessor. If `public`, you can create this client from its parent. Else, not

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,379 @@
...
+
+  @@clientInitialization(InnerGroup, {parameters: InnerGroupClientOptions, access: Access.public, accessorAccess: Access.internal});
+}
+```
+The above tsp gets client `MyServiceClient` and sub client `InnerGroup`. 
+The `MyServiceClient` has a client accessor method whose response is `InnerGroup`. The method has a parameter named `blobName`. The method is `internal`.
+The `InnerGroup`'s `initialization` model's properties contains a property named `blobName`. The method is `public`.
+The method `upload` no longer has `blobName` parameter, its corresponding operation's parameter `blobName` is mapped to the `blobName` property of `InnerGroup`'s `initialization` model.
+
+2. Move client accessor method out of the normal methods list of client and add a new `clientAccessor` property to put it.
```
iscai-msft, line 301: will the `clientAccessor` be on the parent client or the child client? Off the top of my head, we want to know a client's subclients, so I think it would be better to have this on the parent

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,379 @@
...
+- The idea is from the feedback that the client access way is too complicated.
+
+3. Consolidate `@client` and `@operationGroup`
+- Deprecate decorator `@operationGroup` and `SdkOperationGroup` type.
+- Current explicitly `@operationGroup` could be migrated to `@client`. If `@client` is nested, then it is a sub-client, will follow previous operation group default logic.
```
iscai-msft, line 306: when we move `@operationGroup` to `@client`, we will move them to a client with init access `internal` and accessor access `public`, right?

tadelesh: correct.

packages/typespec-client-generator-core/design-docs/client.md
tadelesh: it is on the parent's `getXXX` method.

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,379 @@
...
+sub_client = TestSubClient(endpoint="endpoint", credential=AzureKeyCredential("key"))
+result = sub_client.do_something()
+```
+- Sub client has additional parameter than parent client
+```python
+client = TestClient(endpoint="endpoint", credential=AzureKeyCredential("key"))
+sub_client = TestSubClient(endpoint="endpoint", credential=AzureKeyCredential("key"), sub_name="sub")
+result = sub_client.do_something()
+```
+4. Combine 3 and 4. The sub client could either be get from parent client or be initialized separately.
```
srnagar, line 36: This is 4. Did you mean to combine 2 and 3?

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,379 @@
...
+```python
+client = TestClient(endpoint="endpoint", credential=AzureKeyCredential("key"))
+sub_client = client.sub_client(sub_name="sub")
+result = sub_client.do_something()
+```
+3. Client with sub client and sub client could only be initialized separately
+- Sub client does not have additional parameter than parent client (ARM case)
+```python
+client = TestClient(endpoint="endpoint", credential=AzureKeyCredential("key"))
+sub_client = TestSubClient(endpoint="endpoint", credential=AzureKeyCredential("key"))
```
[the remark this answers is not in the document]
tadelesh: yes. for such case, it is not a sub client. it is a little bit tricky here.
if the client hierarchy in tcgc follow how the client is initialized, then tcgc should elevate this sub client to top level. also, its parent client's sub clients list should not have this sub client.
but for another case: a sub client that could not only initialized by parent but also initialized individually, should tcgc do the elevation? does this sub client could be both in top level clients list and its parent client's sub clients list?

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: i think we could keep the current model style since it at least has name that could be used in emitter.

weidongxu-microsoft:
Note that for Java, we again want to know whether the e.g. `initializedBy` is provided by TCGC or by TypeSpec. The case in Java is that all subclient is by default "initialized by" ClientBuilder of root client, unless specified different by dev (when specified, emitter would follow that).

It maybe supported by 2 ways.
1. `initializedBy==undefined` if not specified in TypeSpec -- this may not be easy to understand
2. emitter can provide a "defaultSubClientInitializedBy" to TCGC. In Java's case, it could be "ClientBuilder"

I am not sure whether JavaScript have similar ask.

packages/typespec-client-generator-core/design-docs/client.md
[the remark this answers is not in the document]
tadelesh: i tend to keep it simple. client hierarchy will follow the spec define structure. emitter see initializedBy to decide how to generate the client ctor.

packages/typespec-client-generator-core/lib/decorators.tsp
tadelesh: will change to use `initailizedBy`

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: i tend to keep it simple. client hierarchy will follow the spec define structure. emitter see initializedBy to decide how to generate the client ctor.

packages/typespec-client-generator-core/src/package.ts
tadelesh: after design review, we decided to keep the original model impl. also, client accessor will be deprecated and switch to `subClients` property.

packages/typespec-client-generator-core/src/package.ts
tadelesh: init method way is droped.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1029,26 +1035,71 @@ const clientInitializationKey = createStateSymbol("clientInitialization");
...
   context: TCGCContext,
   entity: Namespace | Interface,
 ): SdkInitializationType | undefined {
-  const model = getScopedDecoratorData(context, clientInitializationKey, entity);
-  if (!model) return model;
-  const sdkModel = getSdkModel(context, model);
+  let options = getScopedDecoratorData(context, clientInitializationKey, entity);
+  if (options === undefined) return undefined;
+  // backward compatibility
+  if (options.properties.get("initializedBy")) return undefined;
```
weidongxu-microsoft, line 1100: why `return undefined` when properties have `initializedBy`?

I assume it should still get to the value of `parameters`, if available (kind of ignore `initializedBy`)?

Or, it should be `undefined` only if `initializedBy=parent`?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1062,6 +1113,49 @@ export function getClientInitialization(
...
+  if (options === undefined) return undefined;
+
+  // backward compatibility
+  if (
+    options.properties.get("initializedBy") === undefined &&
+    options.properties.get("parameters") === undefined
+  ) {
+    return {
+      parameters: options,
+    };
```
weidongxu-microsoft, line 1137: No `initializedBy` value? I assume you need to provide one as well (either parent or individual, based on whether this is root client)

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -127,6 +169,31 @@ export interface DecoratorInfo {
...
+  kind: "client";
+  name: string;
+  clientNamespace: string; // fully qualified namespace
+  doc?: string;
+  summary?: string;
+  /**
+   * @deprecated Use `init.paramters` instead.
+   */
+  initialization: SdkInitializationType;
+  init: SdkClientInitializationType;
```
weidongxu-microsoft, line 184: nit, personally I'd like the full word `initialization`. But appears here we name it `init` because we need to handle the deprecate procedure.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -114,6 +99,63 @@ export interface SdkOperationGroup {
...
+}
+
+/**
+ * Options used to indicate how to initialize a client.
+ * `parameters` is a model that used to .
+ * `initializedBy` is a flag that indicates how the client is initialized.
+ */
+export interface ClientInitializationOptions {
+  parameters?: Model;
+  initializedBy?: InitializedByFlags;
```
weidongxu-microsoft, line 154: Should we make `initializedBy` "required"?

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -565,6 +561,86 @@ function getSdkInitializationType(
...
+      );
+    } else if (
+      client.kind === "SdkOperationGroup" &&
+      initializationOptions.initializedBy === InitializedByFlags.Individually
+    ) {
+      diagnostics.add(
+        createDiagnostic({
+          code: "invalid-initialized-by",
+          target: client.type,
+          format: { message: "Sub client could not only be initialized individually." },
```
weidongxu-microsoft, line 622: Would it be clearer if message say "Sub client must have 'parent' specified in 'initializedBy'." (we better tell user what need to be done, instead of what cannot) ?

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: the default logic is in `package.ts`, not included here.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: i'm not 100% like current name. but not have more idea. @iscai-msft do you have any suggestion. this pr will not break anything, but just deprecate things, so we could not use previous property name.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: this is for decorator, not for type graph in tcgc.

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: fixed.

weidongxu-microsoft:
> i think tcgc could give you a flag to not set default.

Current design (not finalized) in Java would be an emitter option to control what emitter handles the subclients.
By default, "initializedBy = parent" in Azure would be mapped to "initializedBy = parentBuilder".

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: we could change the whole thing to "constructor" instead of "initialization"? Don't know if it's that much work though, I'm ok with `init` if we want to keep it that way

tadelesh:
@srnagar @m-nash could you help to review?

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,564 @@
+# TCGC client related type design
+
+We have many client related types, decorators and concepts in TCGC and some of them are implicit, which makes author hard to know how to generate a needed client. This doc aims to revisit all client related things in TCGC and try to have a consolidated design output.
+
+## User scenario
+
+1. Single client
+
+```Python
```
srnagar, line 9: It would be great to add one strongly-typed language sample as well in the doc.

packages/typespec-client-generator-core/design-docs/client.md
```diff
@@ -0,0 +1,564 @@
...
+- Add check for `initializedBy`, root clients could only have `individually` value.
+
+2. Deprecate client accessor method. Add `subClients` property to `SdkClientType` and put all sub clients in this list.
+
+3. Consolidate `@client` and `@operationGroup`
+
+- Deprecate decorator `@operationGroup` and `SdkOperationGroup` type.
+- Current explicitly `@operationGroup` could be migrated to `@client`. If `@client` is nested, then it is a sub client, will follow previous operation group default logic.
+- Add `subClients`, `clientPath` properties to the `SdkClient` type to keep backward compatible for metadata type.
+- Add `getClientPath` helper to provide similar function for TCGC `SdkClientType` type.
```
srnagar, line 564: Thanks for a very good write-up on this.

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -449,17 +464,33 @@ extern dec useSystemTextJsonConverter(target: Model, scope?: valueof string);
...
+  /**
+   * Redefine the client initialization parameters you would like to add to the client.
+   * By default, we apply endpoint, credential, and api-version parameters. If you specify parameters model, we will append the properties of the model to the parameters list of the client initialization.
+   */
+  parameters?: Model;
+
+  /**
+   * Determines how the client could be initialized. Use `InitializedBy` enum to set the value. The value could be `InitializedBy.individually`, `InitializedBy.parent` or `InitializedBy.individually | InitializedBy.parent`.
+   */
+  initializedBy?: EnumMember | Union;
```
srnagar, line 491: Why is this optional? What is the default when this is not defined? Will this assume `individually` or `parent` or both? We should update the docs here to clarify.

packages/typespec-client-generator-core/src/interfaces.ts
srnagar: We could use `clientInitialization`

srnagar (approved):

packages/typespec-client-generator-core/src/interfaces.ts
iscai-msft: i switched it to `clientInitialization`

packages/typespec-client-generator-core/design-docs/client.md
iscai-msft: i think this is up to separate languages to add

packages/typespec-client-generator-core/lib/decorators.tsp
weidongxu-microsoft: This part is in decorator, which mean the dev does not have to specify it.

I think Chenjie have this be "required" in TCGC API, so TCGC always gives a value.

Issue #2031, opened 2025-01-03
[tcgc] `crossLanguageDefinitionId` has redundant `.`


tadelesh:
for anonymous case, it seems tcgc always add a redundant `.`, e.g., `crossLanguageDefinitionId: Contoso.WidgetManager..anonymous.clientRequestId`.

PR #2034, opened 2025-01-03
[tcgc] refine cross language definition id logic

tadelesh:
1. refine all generated type's cross language definition id
2. for parameters, use the operation name as the prefix of cross language definition id
3. for anonymous model, add namespace for first section of the cross language definition id

fix: https://github.com/Azure/typespec-azure/issues/2031

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -11,11 +11,9 @@ import {
   Model,
   ModelProperty,
   Operation,
-  Type,
 } from "@typespec/compiler";
 import { getServers, HttpServer } from "@typespec/http";
 import { resolveVersions } from "@typespec/versioning";
-import { camelCase } from "change-case";
 import {
   getAccess,
   getClientInitialization,
@@ -390,7 +388,7 @@ function getSdkMethodResponse(
       name: createGeneratedName(context, operation, "UnionResponse"),
       isGeneratedName: true,
       clientNamespace: client.clientNamespace,
-      crossLanguageDefinitionId: getCrossLanguageDefinitionId(context, operation),
+      crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, operation)}.UnionResponse`,
       decorators: [],
     };
   } else if (responseTypes.size === 1) {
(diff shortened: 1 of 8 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -755,7 +731,7 @@ function getSdkEndpointParameter<TServiceOperation extends SdkServiceOperation =
       variantTypes: types,
       name: createGeneratedName(context, rawClient.service, "Endpoint"),
       isGeneratedName: true,
-      crossLanguageDefinitionId: getCrossLanguageDefinitionId(context, rawClient.service),
+      crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, rawClient.service)}.Endpoint`,
```
weidongxu-microsoft, line 734: nit, in the test case, the value is `Contoso.WidgetManager.endpoint`

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -390,7 +388,7 @@ function getSdkMethodResponse(
       name: createGeneratedName(context, operation, "UnionResponse"),
       isGeneratedName: true,
       clientNamespace: client.clientNamespace,
-      crossLanguageDefinitionId: getCrossLanguageDefinitionId(context, operation),
+      crossLanguageDefinitionId: `${getCrossLanguageDefinitionId(context, operation)}.UnionResponse`,
```
weidongxu-microsoft, line 391: So, for these kind of changes, we are making a `crossLanguageDefinitionId` that being cross language in SDK, but does not exist in TypeSpec itself?

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/package.ts
tadelesh: for the parameter, it is `endpoint`, for the union type, it is `Endpoint`.

packages/typespec-client-generator-core/src/package.ts
tadelesh: yes. for these tcgc generated types, we create the definition id. maybe in efv2, we will try to create native typespec type, then we could do the mapping.

Issue #1729, opened 2024-10-23
[tcgc] `getWireName` should honor content type


tadelesh:
xml support in typespec have `@encodedName` to rename the xml tags name, however current tcgc does not honor content type when `getWireName`, but hard code to `application/json`. we need to fix that first. but an operation may have multiple content types, thus will have different wire name for different content type, since tcgc currently does not support multiple content type, we may do that when we add multi-content type support.

tadelesh:
Add new interfaces to represent detail of the serialization info, also will move multipart related options into it:
```typescript
export interface SerializationOptions {
  json?: JsonSerializationOptions;
  xml?: XmlSerializationOptions;
  multipart?: MultipartOptions;
}

export interface JsonSerializationOptions {
  name?: string;
}

export interface XmlSerializationOptions {
  name?: string;
  attribute?: boolean;
  ns?: string | EnumMember;
  prefix?: string;
  unwrapped?: boolean;
}
```
Deprecate current `serializedName` in `SdkEndpointParameter`, `SdkBodyModelPropertyType`, `SdkHeaderParameter`, `SdkQueryParameter`, `SdkPathParameter`, `SdkBodyParameter`, `SdkServiceResponseHeader`.
Add new `serializationOptions?: SerializationOptions;` property to the following types as well as `SdkModelType`:
```typescript
export interface SdkQueryParameter extends SdkModelPropertyTypeBase {
  kind: "query";
  collectionFormat?: CollectionFormat;
  /**
   * @deprecated This property is deprecated. Use `allowReserved` instead.
   * @param serviceParam
   */
  serializedName: string;
  serializationOptions?: SerializationOptions;
  correspondingMethodParams: SdkModelPropertyType[];
  explode: boolean;
}
```
This change could cover all possible serialization logic with extensibility.

PR #2059, opened 2025-01-09
[tcgc] new serialization options for model and model property

tadelesh:
1. Add `serializationOptions` property to `SdkModelType` and `SdkBodyModelPropertyType`. Its type is `SerializationOptions` which contains the info of how to serialize to Json/Xml/Multipart value.
2. Deprecate `serializedName` property in `SdkBodyModelPropertyType`, use `serializationOptions.xxx.name` instead.
3. Deprecate `multipartOptions` in `SdkBodyModelPropertyType`, use `serializationOptions.multipart` instead.

resolve: https://github.com/Azure/typespec-azure/issues/1729

Add new interfaces to represent detail of the serialization info, also will move multipart related options into it:
```typescript
export interface SerializationOptions {
  json?: JsonSerializationOptions;
  xml?: XmlSerializationOptions;
  multipart?: MultipartOptions;
}

/**
 * For Json serialization.
 * The name will come from explicit setting of `@encodedName("application/json", "NAME")`.
 */
export interface JsonSerializationOptions {
  name: string;
}

/**
 * For Xml serialization.
 * The `name`/`itemsName` will come from explicit setting of `@encodedName("application/xml", "NAME")` or `@xml.Name("NAME")`.
 * Other properties come from `@xml.attribute`, `@xml.ns`, `@xml.unwrapped`.
 * The `itemsName` and `itemsNs` are used for array items.
 */
export interface XmlSerializationOptions {
  name?: string;
  attribute?: boolean;
  ns?: {
    namespace: string;
    prefix: string;
  };
  unwrapped?: boolean;

  itemsName?: string;
  itemsNs?: {
    namespace: string;
    prefix: string;
  };
}
```
The serialization info will be added only when it is used in request body or response body with corresponding content type.
The serialization options update logic could follow usage/access update logic.
This change could cover all possible serialization logic with extensibility.


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -408,11 +409,46 @@ export interface SdkModelPropertyTypeBase extends DecoratedType {
...
+  xml?: XmlSerializationOptions;
+  multipart?: MultipartOptions;
+}
+
+/**
+ * For Json serialization.
+ * The name will come from explicit setting of `@encodedName("application/json", "NAME")`.
+ */
+export interface JsonSerializationOptions {
+  name: string;
```
iscai-msft, line 423: Will this also include the default value if there is no `@encodedName` property? Meaning if I have a model like below used in json (or for that matter xml, multipart) operation call, will the `.name` value be `original`?

```tsp
model InputModel {
  original: string
}
```

iscai-msft (commented):
overall really like this design @tadelesh! can you confirm this design will also work on response body? thanks!

tadelesh:
> overall really like this design @tadelesh! can you confirm this design will also work on response body? thanks!

Yes.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: Yes. `resolveEncodedName` will fall back to the original name if no `@encodedName`.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -408,11 +409,46 @@ export interface SdkModelPropertyTypeBase extends DecoratedType {
...
+}
+
+/**
+ * For Xml serialization.
+ * The `name`/`itemsName` will come from explicit setting of `@encodedName("application/xml", "NAME")` or `@xml.Name("NAME")`.
+ * Other properties come from `@xml.attribute`, `@xml.ns`, `@xml.unwrapped`.
+ * The `itemsName` and `itemsNs` are used for array items.
+ */
+export interface XmlSerializationOptions {
+  name?: string;
```
weidongxu-microsoft, line 433: In what case, name would be undefined?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -434,6 +470,7 @@ export type SdkModelPropertyType =
   | SdkCookieParameter;
 
 export interface MultipartOptions {
+  name: string;
```
weidongxu-microsoft, line 473: I remember from @msyyc that the `name` (field name) of a part could be undefined, in some cases (multipart/mixed ?).

Are we able to know in which case, the `name` is provided by TCGC, instead of a real serialized name on wire?

tadelesh: the name is either from the property name or from the http part name. let me double check with @msyyc.

msyyc: Each field has a name. As @tadelesh said, it is either from property name or http part name. 
image

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -449,11 +486,18 @@ export interface MultipartOptions {
 export interface SdkBodyModelPropertyType extends SdkModelPropertyTypeBase {
   kind: "property";
   discriminator: boolean;
+  /**
+   * @deprecated This property is deprecated. Use `serializationOptions.xxx.name` instead.
+   */
   serializedName: string;
```
msyyc, line 492: Is it necessary to keep considering we already have `serializationOptions`?

packages/typespec-client-generator-core/src/interfaces.ts
weidongxu-microsoft: Is it doc for multipart/form-data or multipart/mixed?

image

I am fine TCGC only support form-data. In such case, I expect it output diagnostic error on mixed.

msyyc: I see. I make an issue to track it https://github.com/Azure/typespec-azure/issues/2065

weidongxu-microsoft (commented):
I assume the serializedName from method parameter is not changed, right?

Also, I assume the `lroMetadata.finalResponse.resultPath` is referring the `serializationOptions.json.name`?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -408,11 +409,46 @@ export interface SdkModelPropertyTypeBase extends DecoratedType {
...
+  name: string;
+}
+
+/**
+ * For Xml serialization.
+ * The `name`/`itemsName` will come from explicit setting of `@encodedName("application/xml", "NAME")` or `@xml.Name("NAME")`.
+ * Other properties come from `@xml.attribute`, `@xml.ns`, `@xml.unwrapped`.
+ * The `itemsName` and `itemsNs` are used for array items.
+ */
+export interface XmlSerializationOptions {
```
weidongxu-microsoft, line 432: For boolean like `attribute` and `unwrapped`, what would it mean if it is undefined? Is it an "attribute" or not an "attribute"? I'd hope we had some clarity in TCGC.

tadelesh: for xml serialization, only property could have this flag. so for model, set it to false is meaningless for model.

weidongxu-microsoft: So, for property, it would always have a true/false on `attribute`?
And what about `unwrapped`?

If possible, have it in document.

weidongxu-microsoft: > for xml serialization, only property could have this flag. so for model, set it to false is meaningless for model.

BTW, why a model has a `name` in XML? I don't think model has a `name` in JSON. Is XML something different?

tadelesh:
> I assume the serializedName from method parameter is not changed, right?
yes.

> Also, I assume the `lroMetadata.finalResponse.resultPath` is referring the `serializationOptions.json.name`?
no. it refers the client name, not wire name.

tadelesh:
@iscai-msft @weidongxu-microsoft if no further concern, could you approve? i'd like to make it in the 0.50.0.

weidongxu-microsoft:
> > I assume the serializedName from method parameter is not changed, right?
> 
> yes.
> 
> > Also, I assume the `lroMetadata.finalResponse.resultPath` is referring the `serializationOptions.json.name`?
> 
> no. it refers the client name, not wire name.

I thought client name can be different per language (`@clientName`). So this value be different if run in different emitter?

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: xml will use model name as the tag name. `Container` `EnumerationResults` all from the model name. refer the spec: https://github.com/Azure/azure-rest-api-specs/blob/a26af9acb30387724eff64229ba7f3259600681a/specification/storage/data-plane/Microsoft.BlobStorage/stable/2025-01-05/blob.json#L10507.

tadelesh:
> > > Also, I assume the `lroMetadata.finalResponse.resultPath` is referring the `serializationOptions.json.name`?
> > 
> > 
> > no. it refers the client name, not wire name.
> 
> I thought client name can be different per language (`@clientName`). So this value be different if run in different emitter?
> 
> image
> 
> ```
>     @clientName("pollResult")
>     @encodedName("application/json", "poll_result")
>     result?: TStatusResult;
> ```
> 
> This thing seems not right. I will raise another issue, as it not exactly relate to this PR.

seems a bug. yes. for different emitter, it could follow the name that emitter get from tcgc.

weidongxu-microsoft (approved):
LGTM

Though prefer that the user of the XML to take a look...

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -408,11 +409,46 @@ export interface SdkModelPropertyTypeBase extends DecoratedType {
   crossLanguageDefinitionId: string;
 }
 
+export interface SerializationOptions {
```
jhendrixMSFT, line 412: This needs some doc comments. Can multiple be set? None be set and if so, what does that mean?

iscai-msft (approved):

tadelesh:
@jhendrixMSFT i've refined the doc. could you help to confirm the xml part is good for you?

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -408,11 +409,52 @@ export interface SdkModelPropertyTypeBase extends DecoratedType {
   crossLanguageDefinitionId: string;
 }
 
+/**
+ * Options to show how to serialize a type.
+ * A type that is used in multiple operations with different wire format could have multiple options set. For example, a type could be serialized as JSON in one operation and as XML in another operation.
+ * A type that has no special serialization logic will have no options set. For example, a type that is used in a HTTP query parameter will have no serialization options set.
```
jhendrixMSFT, line 415: I assume that binary payloads will also have no options set?

tadelesh: yes. refined.

jhendrixMSFT (approved):

PR #1983, opened 2024-12-12
x-ms-identifiers alternatives

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1473

PR #2069, opened 2025-01-13
Revert "x-ms-identifiers alternatives (#1983)"

AlitzelMendez:
This reverts commit 426b7399df9aa02bc998a4b7cd4297fa49f576d1.
reverts: https://github.com/Azure/typespec-azure/commit/426b7399df9aa02bc998a4b7cd4297fa49f576d1

I need to handle azure-rest-api-specs and it is significant, also I need to consider Azure.Core.Page as a scenario


chrisradek (approved):
Thanks! I think you might need 1 more changelog so that publishing the dev version works properly (currently this removes 1 and adds 1, so the version published would be the same version as the most recent one)

markcowl (approved):

AlitzelMendez:
> Thanks! I think you might need 1 more changelog so that publishing the dev version works properly (currently this removes 1 and adds 1, so the version published would be the same version as the most recent one)
@chrisradek, 
Github is having some issues: https://www.githubstatus.com/ so I can't push the new change log right now, but as soon as I am able to, I will merge this

PR #2080, opened 2025-01-15
[doc] update namespace related docs

tadelesh:
1. add sub namespace example for clients
2. add sub namespace example for models
3. rearrange the sections

### DIFF
--- a/website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
+++ b/website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
@@ -4,7 +4,7 @@ title: Clients
 
 import { ClientTabs, ClientTabItem } from "@components/client-tabs";
 
-This page documents the default client hierarchy behavior as well as how to customize clients. For an overview of the setup, please visit the previous page.
+This page documents the client behavior as well as how to customize clients. For an overview of the setup, please visit the previous page.
 
 JS RLC is not in the business of customization. it will ignore client.tsp and the follow scenarios will not have impact on the JS RLC user experiences. In this context, TypeScript part means JS Modular Emitter.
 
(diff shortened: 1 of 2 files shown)


weidongxu-microsoft (commented):
You seem missing language tab item element like
```
<ClientTabItem lang="java" >
...
</ClientTabItem>
```
in some of the tabs.

website/src/content/docs/docs/howtos/Generate client libraries/06types.mdx
```diff
@@ -6,6 +6,166 @@ import { ClientTabs, ClientTabItem } from "@components/client-tabs";
...
+// TODO
+```
+
+</ClientTabItem>
+<ClientTabItem lang="java" >
+
+As `flavor=azure`, Java emitter generates the model classes under `models` package.
+
+```java
+import myservice.models.Foo;
```
weidongxu-microsoft, line 56: cspell fails on this `myservice`.

Maybe just call the namespace `Service` in source?

website/src/content/docs/docs/howtos/Generate client libraries/02client.mdx
```diff
@@ -67,6 +67,74 @@ client.pet();
...
+
+<ClientTabs>
+
+```typespec title="main.tsp"
+namespace PetStore;
+
+@service({
+  title: "Dog Store",
+})
+namespace Dogs {
```
lmazuel, line 80: This is interesting, because your example is two servies in one TypeSpec. Which I didn't even think we supported, so if we do we should keep it for sure :)

I was thinking more about sub-namespaces, like in todo app:
https://github.com/allenjzhang/typespec-e2e-demo/blob/main/todoApp/spec/main.tsp

Meaning, one service, and sub-namespaces are operation groups (or subclient)

TLDR: If we support two services, great let's keep that. But then add also example with sub-namespaces and operation groups / sub client.

tadelesh: weidong has reminded me about multiple service limitation. tcgc used to report warning for this, but could give the correct type graph. let me check with all languages to see if it works.
the doc already has the case of operation group with sub namespace. but since sub client from operation group could only get from parent, current language usage code could not show the difference for sub client in sub namespace. do you think we should add the generated code to the doc to show such feature?

PR #2087, opened 2025-01-16
[tcgc] Add @apiVersion scenarios and mock apis

catalinaperalta:
Add scenarios and mock apis for the tcgc `@apiVersion` decorator


packages/azure-http-specs/specs/azure/client-generator-core/api-version/main.tsp
```diff
@@ -0,0 +1,36 @@
...
+import "@typespec/spector";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+using Spector;
+
+@doc("Test for api-version decorator.")
+@scenarioService("/azure/client-generator-core/api-version")
+namespace _Specs_.Azure.ClientGenerator.Core.ApiVersion;
```
weidongxu-microsoft, line 11: I assume if we are testing azure api-version, the service itself should be versioned (instead of just value "1234" in mockapi)?
E.g. https://github.com/Azure/typespec-azure/blob/020ca109d6fad08b8de8fcd6ea264ad6ef43d277/packages/azure-http-specs/specs/azure/core/basic/main.tsp#L16-L31

Also, unsure how client works when 3 different scheme of api-version appears in a same client. Need to test.

tadelesh (commented):
i think the intension for this pr is to test the new `@apiVersion` decorator. but i did not see it in the spec. am i miss anything?

packages/azure-http-specs/specs/azure/client-generator-core/api-version/main.tsp
catalinaperalta: Separated out the scenarios into different specs. Also, ci seems to be fixed now

catalinaperalta:
> i think the intension for this pr is to test the new `@apiVersion` decorator. but i did not see it in the spec. am i miss anything?

I only noticed that one of the examples didnt have the apiVersion decorator over it, but it's fixed now

weidongxu-microsoft:
1. the `@scenarioService` already do `@server` and `@service`
2. the `@scenarioService` must be on the namespace that contains operations (e.g. in your case, the namespace in "main.tsp")

Also, I kind of think the origin folder is better, e.g. "azure/client-generator-core/api-version/header` etc.

If I write the test, it would be

azure/client-generator-core/api-version/header/main.tsp (`@scenarioService` in this tsp)

```
/**
 * DO NOT GENERATE FROM THIS FILE USE client.tsp
 * This is just to simulate a service entrypoint
 */
import "./service.tsp";

using TypeSpec.Http;
using TypeSpec.Versioning;
using global.Azure.ClientGenerator.Core;
using Spector;

@scenarioService(
  "/azure/client-generator-core/api-version/header",
  {
    versioned: ApiVersions,
  }
)
namespace _Specs_.Azure.ClientGenerator.Core.ApiVersion.Header;

model HeaderApiVersionParam {
  @header("x-ms-version")
  version: string;
}

@scenario
@scenarioDoc("Set a header as the  service api version")
@doc("Header api version parameter.")
@route("/header")
@post
op headerApiVersion(...HeaderApiVersionParam): OkResponse;
```

azure/client-generator-core/api-version/service.tsp

```
import "@typespec/http";
import "@typespec/versioning";
import "@typespec/spector";

using TypeSpec.Http;
using TypeSpec.Versioning;
using Azure.ClientGenerator.Core;
using Spector;

@doc("Supported api versions.")
enum ApiVersions {
  @doc("Api version 2025-01-01.")
  @useDependency(Azure.Core.Versions.v1_0_Preview_2)
  v2025_01_01: "2025-01-01",
}
```

There is now not much in service.tsp, so it may be directly inserted into e.g. "header.tsp"

catalinaperalta:
> 1. the `@scenarioService` already do `@server` and `@service`
> 2. the `@scenarioService` must be on the namespace that contains operations (e.g. in your case, the namespace in "main.tsp")
> 3. generally the TypeSpec namespace be different in different main.tsp (as emitter typically use the TypeSpec namespace for SDK package), so we'd like to have different suffix like `.Header`, `.Path` in different main.tsp.
> 
> Also, I kind of think the origin folder is better, e.g. "azure/client-generator-core/api-version/header` etc.
> 
> If I write the test, it would be
> 
> azure/client-generator-core/api-version/header/main.tsp (`@scenarioService` in this tsp)
> 
> ```
> /**
>  * DO NOT GENERATE FROM THIS FILE USE client.tsp
>  * This is just to simulate a service entrypoint
>  */
> import "./service.tsp";
> 
> using TypeSpec.Http;
> using TypeSpec.Versioning;
> using global.Azure.ClientGenerator.Core;
> using Spector;
> 
> @scenarioService(
>   "/azure/client-generator-core/api-version/header",
>   {
>     versioned: ApiVersions,
>   }
> )
> namespace _Specs_.Azure.ClientGenerator.Core.ApiVersion.Header;
> 
> model HeaderApiVersionParam {
>   @header("x-ms-version")
>   version: string;
> }
> 
> @scenario
> @scenarioDoc("Set a header as the  service api version")
> @doc("Header api version parameter.")
> @route("/header")
> @post
> op headerApiVersion(...HeaderApiVersionParam): OkResponse;
> ```
> 
> azure/client-generator-core/api-version/service.tsp
> 
> ```
> import "@typespec/http";
> import "@typespec/versioning";
> import "@azure-tools/typespec-azure-core";
> import "@typespec/spector";
> import "@azure-tools/typespec-client-generator-core";
> 
> using TypeSpec.Http;
> using TypeSpec.Versioning;
> using Azure.ClientGenerator.Core;
> using Spector;
> 
> @doc("Supported api versions.")
> enum ApiVersions {
>   @doc("Api version 2025-01-01.")
>   @useDependency(Azure.Core.Versions.v1_0_Preview_2)
>   v2025_01_01: "2025-01-01",
> }
> ```
> 
> There is now not much in service.tsp, so it may be directly inserted into e.g. "header.tsp"

Done. It would be great if we could have the docs mention this info. For example, not needing to specify `@server` and `@service`. I noticed some of the scenarios included that so I had added it here as well.

packages/azure-http-specs/specs/azure/client-generator-core/api-version/header/main.tsp
```diff
@@ -0,0 +1,40 @@
...
+using Azure.ClientGenerator.Core;
+using Spector;
+
+@scenarioService(
+  "/azure/client-generator-core/api-version/header",
+  {
+    versioned: ApiVersions,
+  }
+)
+namespace Client.AlternateApiVersion.Service.Header;
```
weidongxu-microsoft, line 22: nit, personally I'd prefer namespace with Azure.ClientGenerator.Core prefix like `_Specs_.Azure.ClientGenerator.Core` (the _Specs_ here is mostly for disambiguate from e.g. Azure.Core).

Not anything major, Java can use the `@clientNamespace` to specify any SDK namespace we want.
So the only effect would be on the summary doc.

PS: please run `pnpm generate-scenarios-summary` before merge.
@sarangan12 The CI seems still not fixed (about the verification of summary matches the lib)

weidongxu-microsoft (approved):
Please run `pnpm generate-scenarios-summary` before merge.
@sarangan12 The CI seems still not fixed (about the verification of summary matches the lib)

packages/azure-http-specs/specs/azure/client-generator-core/api-version/header/main.tsp
catalinaperalta: I still got conflicts after updating the namespace to the one you suggested. Here is an example of the error message:
```
error: ✘ Namespace _Specs_.Azure doesn't have member Core: C:/Users/caperal/repos/typespec-azure/packages/azure-http-specs/specs/azure/client-generator-core/api-version/path/main.tsp:27:18
```

I'll leave the namespace as is for now, but if you feel strongly I can update it in a follow up PR.

weidongxu-microsoft: When we define our namespace as `_Specs_.Azure.ClientGenerator.Core` or `.Azure.ClientGenerator.Core`, the `Azure.ClientGenerator.Core` (any namespace under Azure from other libs) would need to be written as `global.Azure.ClientGenerator.Core` to avoid the ambiguity.

E.g. `using Azure.ClientGenerator.Core;` -> `using global.Azure.ClientGenerator.Core;`, etc.

catalinaperalta: Got it, will do in a separate PR and ping you. Thanks for the info

PR #2092, opened 2025-01-16
[common-types] Fix difference on privateEndpointConnectionName for common-types

AlitzelMendez:
Closing final gaps on common types.

Fixing this current difference 
image

### DIFF
--- a/packages/samples/common-types/openapi/v3/privatelinks.json
+++ b/packages/samples/common-types/openapi/v3/privatelinks.json
@@ -211,8 +211,8 @@
     }
   },
   "parameters": {
-    "PrivateEndpointConnectionParameter": {
-      "name": "name",
+    "PrivateEndpointConnectionName": {
+      "name": "privateEndpointConnectionName",
       "in": "path",
       "description": "The name of the private endpoint connection associated with the Azure resource.",
       "required": true,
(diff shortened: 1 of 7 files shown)


packages/typespec-azure-resource-manager/lib/backcompat.tsp
```diff
@@ -16,7 +16,7 @@ alias PrivateEndpointConnectionProperties = CommonTypes.PrivateEndpointConnectio
 alias PrivateLinkServiceConnectionState = CommonTypes.PrivateLinkServiceConnectionState;
 alias PrivateEndpointConnectionProvisioningState = CommonTypes.PrivateEndpointConnectionProvisioningState;
 alias PrivateEndpointServiceConnectionStatus = CommonTypes.PrivateEndpointServiceConnectionStatus;
-alias PrivateEndpointConnectionParameter = CommonTypes.PrivateEndpointConnectionParameter;
+alias PrivateEndpointConnectionName = CommonTypes.PrivateEndpointConnectionName;
```
allenjzhang, line 19: I am not sure about this breaking change fix. Can't we not apply `@encodedName()` to correct this only common-type generation script to get the matching swagger and have the -ref to point to the swagger name?

AlitzelMendez: I have tried this approach, and I am not getting the desired results, encodedName is not impacting at all the name, I will investigate further

allenjzhang (approved):

PR #2094, opened 2025-01-16
x-ms-identifiers alternatives 

AlitzelMendez:
Reintroduce changes in: https://github.com/Azure/typespec-azure/pull/1983


AlitzelMendez:
validation:
- [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/actions/runs/13143374213)

.chronus/changes/x-ms-identifiers-alternatives-2024-11-12-14-15-30.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Use the @identifiers decorator to identify and utilize identifiers for x-ms-identifiers. Additionally, use the @key decorator to identify identifiers.
```
timotheeguerin, line 8:
```suggestion
Use the `@identifiers` decorator to identify and utilize identifiers for x-ms-identifiers. Additionally, use the `@key` decorator to identify identifiers.
```

packages/typespec-azure-resource-manager/lib/decorators.tsp
```diff
@@ -191,3 +191,19 @@ extern dec resourceBaseType(
...
+ *
+ * @example
+ * ```typespec
+ * model Pet {
+ *  @identifiers(["size"])
+ *  dog: Dog;
+ * }
+ * ```
+ */
+extern dec identifiers(entity: ModelProperty, properties: string[]);
```
timotheeguerin, line 209: make properties be `valueof string[]`

alternatively could we not have this be using the actual model property reference?

timotheeguerin:
```
@identifiers(Foo.id, Foo.secondId)
```

AlitzelMendez: if I remember correctly there was a problem with that approach for more complex scenarios, for example 
`@identifiers(Foo.id, Foo.secondId.otherId)`  wasn't able to resolve the `otherId` and this is supported for x-ms-identifiers, but I can check again this

timotheeguerin:
```
Foo.secondId::type.otherId
```
would technically be the right syntax but yeah wwe don't have a way to get the trace up.

But would change to `valueof` at least, it will simplify the js code and will not allow you to pass `string[]`

packages/typespec-autorest/test/openapi-output.test.ts
```diff
@@ -847,6 +847,228 @@ describe("typespec-autorest: extension decorator", () => {
...
+      }
+      model PetList {
+        value: Pet[]
+      }
+      @route("/Pets")
+      @get op list(): PetList;
+      `,
+    );
+    ok(oapi.paths["/Pets"].get);
+    deepStrictEqual(oapi.definitions.PetList.properties.value["x-ms-identifiers"], []);
```
markcowl, line 930: In this case, it should inject no extension at all, not `x-ms-identifiers: []`

markcowl (changes requested):
Looks like there's at least one typespec diagnostic in a test, and you need to run `pnpm format` as well

packages/typespec-autorest/test/openapi-output.test.ts
AlitzelMendez: I think this can be a little bit more impactful as it is the default behavior right now ([playgroun example](https://azure.github.io/typespec-azure/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3JlIjsKCnVzaW5nIFR5cGVTcGVjLkh0dHA70BVSZXN00RVWyXLIG0HEWS5Db3Jl0hIuVHJhaXRzOwoKCiAgbW9kZWwgUGV0IHsKICAgIG5hbWU6IHN0csVNxBJhZ2U6IGludDMyxBB9zDRMaXPIOHZhbHVlOsQVW13HJ0Byb3V0ZSgiL1BldHMiKcQSZ2V0IG9wIGxpc3QoKcUuxEM7&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fdata-plane%22%5D%7D%7D)) is that ok?

markcowl: Yeah, I didn't realize we were doing this for data plane specs.  If there is no behavior change then there's no reason not to do this

markcowl (approved):

AlitzelMendez:
### Validation 
- [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/actions/runs/13209457938/job/36880060229?pr=32521)
- [azure-rest-api-specs-pr](https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/13209462335/job/36880071537?pr=21234)

Issue #234, opened 2024-02-07
[ARM]:  Support resource with multiple paths


markcowl:
- Create a canonical pattern for an ARM resource that can be hosted under multiple paths
  - Resources at each of these paths have the same resource schema and may have the same or different operations
  - example: PolicyDefinition (subscription, resource group)
  - Resources may also have recursive parentage (ManagementGroups)

pshao25:
We have several (maybe we can say many) RPs in which a same resource model defined is used in different clients. For example, [this client](https://github.com/Azure/azure-rest-api-specs/blob/508e46868618ea226265c3ff8ffcda3a341cfa96/specification/web/resource-manager/Microsoft.Web/stable/2023-12-01/WebApps.json#L107) and [this client](https://github.com/Azure/azure-rest-api-specs/blob/508e46868618ea226265c3ff8ffcda3a341cfa96/specification/web/resource-manager/Microsoft.Web/stable/2023-12-01/WebApps.json#L9061) share the same resource model `Site`. 

> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}
> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}/slots/{slot}

[...]

PR #2095, opened 2025-01-17
Add legacy operations for multi-path scenario

pshao25:
To proceed https://github.com/Azure/typespec-azure/issues/234, add the legacy operation template into the library according to this [design](https://gist.github.com/markcowl/825f8827d1e280a1c5ffeb57eb968677). It's almost the same as @markcowl 's design, except:
1. Change the body name and doc for CustomPatchSync/CustomPatchAsync to align with those of ArmCustomPatchSync/ArmCustomPatchAsync.
2. Change the DeleteAsync to DeleteWithoutOkAsync because its corresponding ArmResourceDeleteAsync is deprecated.

### DIFF
--- a/packages/typespec-azure-resource-manager/lib/Legacy/operations.tsp
+++ b/packages/typespec-azure-resource-manager/lib/Legacy/operations.tsp
@@ -27,10 +27,7 @@ namespace Azure.ResourceManager.Legacy {
     @autoRoute
     @doc("Create a {name}", Resource)
     @armResourceCreateOrUpdate(Resource)
-    @Azure.Core.Foundations.Private.defaultFinalStateVia(#[
-      "location",
-      "azure-async-operation"
-    ])
+    @Azure.Core.Foundations.Private.defaultFinalStateVia(#["location", "azure-async-operation"])
     @put
     CreateOrUpdateAsync<
       Resource extends Azure.ResourceManager.CommonTypes.Resource,
@@ -145,11 +142,7 @@ namespace Azure.ResourceManager.Legacy {
         Azure.Core.Foundations.RetryAfterHeader,
       Parameters extends {} = {},
       Response extends {} = ArmDeletedResponse | ArmDeleteAcceptedLroResponse<LroHeaders> | ArmDeletedNoContentResponse
-    >(
-      ...ParentParameters,
-      ...ResourceTypeParameter,
-      ...Parameters,
-    ): Response | ErrorType;
+    >(...ParentParameters, ...ResourceTypeParameter, ...Parameters): Response | ErrorType;
 
     /**
      * @dev Delete a resource synchronously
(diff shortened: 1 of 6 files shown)


markcowl (commented):
I wonder if we want to export the types here as experimental until we have used them on a few more specs.

pshao25:
> I wonder if we want to export the types here as experimental until we have used them on a few more specs.

Isn't the design ready? What else we need to design more? If we don't get this in, every time when customer asks, we need to give them a copy of a legacy operation.

AlitzelMendez:
> I wonder if we want to export the types here as experimental until we have used them on a few more specs.

do you mean like having an `Azure.ResourceManager.Experimental`?

packages/typespec-azure-resource-manager/lib/Legacy/operations.tsp
```diff
@@ -0,0 +1,272 @@
...
+      Response extends TypeSpec.Reflection.Model | void,
+      Parameters extends {} = {}
+    >(
+      ...ParentParameters,
+      ...ResourceTypeParameter,
+      ...Parameters,
+
+      @doc("The content of the action request")
+      @bodyRoot
+      body: Request,
```
markcowl, line 223: I think we want to add two add an additional parameter for every one of these operations that has a request body, like:

```tsp
RequestBody extends {} = {
  @doc("The content of the action request")
      @bodyRoot
      body: Request
},
```

This would allow optional request bodies, which are not standard, but occur in legacy apis

packages/typespec-azure-resource-manager/lib/Legacy/operations.tsp
```diff
@@ -0,0 +1,272 @@
...
+      Response extends {} = ArmResponse<Resource>
+    >(
+      ...ParentParameters,
+      ...ResourceTypeParameter,
+      ...Parameters,
+      @doc("The resource properties to be updated.") @bodyRoot properties: PatchModel,
+    ): Response | ErrorType;
+
+    /**
+     * @dev Delete a resource asynchronously
```
markcowl, line 127: I'd like us to be explicit that these APIs are subject to change, and should only be used in converting legacy APIs to typespec.

If we could, I would consider an experimental types export like this: https://github.com/microsoft/typespec/blob/main/packages/events/package.json#L28

But we would have to validate that this works with imports and standard tsp compilation.  The best current solution might be docs and either a decorator or rule that marks these types as volatile, and issues a warning if you use them (perhaps through conbining a decorator and linting rule, so that it could be globally suppressed in config).

markcowl (changes requested):

packages/typespec-azure-resource-manager/lib/Legacy/operations.tsp
markcowl: It also allows *renaming* the body parameter without decorators.

markcowl:
closing in favor of this PR: https://github.com/Azure/typespec-azure/pull/2653

Issue #2088, opened 2025-01-16
[Bug]: tcgc remove nullable union name which cause problems for us


qiaozha:
### Describe the bug

For the given TypeSpec, tcgc will remove the nullable union name and provide a generated name for property error type in `FullBackupOperation` as `FullBackupOperationError`
```tsp
model FullBackupOperation {
  error?: Error;
}

union Error {
  null,
  {
    code?: string,
    message?: string,
    innerError?: Error,
  },
}
```
This is problematic for us because JS will generate the type inline for anonymous models. however, in this case Error has circle reference which cause us are not able to generate it in inline way. 
```ts
model FullBackupOperation {
   error?: nullable | {
       code? string;
       message?: string;
       innerError?: nullable | {
          ....// recursively inline definition.
       }
   }
}
```

But the original typespec, it won't be a problem for us, because we can generate type alias for Error, and there's no problem to have circle reference there. 
```ts
type Error = null | {
    code?: string;
    message?: string;
    innerError: Error;
}
```
Could tcgc reserve the original nullable union name? @tadelesh 

### Reproduction

N/A

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

maorleger:
I'd love to be able to follow along or have an ETA for when this will be fixed as it is a blocker for KeyVault's migration to TypeSpec in JS. @qiaozha / @tadelesh any workarounds I can apply in the meanwhile?

tadelesh:
i'll try to make it before the mid of next week and release a hotfix version of tcgc.

qiaozha:
I don't think we have any work arounds here because the original information already lost.

maorleger:
I noticed that `@azure-tools/typespec-client-generator-core": "0.50.1"` has been released and I _believe_ it includes this fix is that right?

If so, I tried generating code using the above version (after bumping the version and running `tsp-client generate-lock-file` and am still running into the same error:

```

[...]

PR #2097, opened 2025-01-20
[tcgc] support `name` and `isGeneratedName` for nullable type

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2088

### DIFF
--- a/.chronus/changes/add_nullable_name-2025-0-20-17-36-45.md
+++ b/.chronus/changes/add_nullable_name-2025-0-20-17-36-45.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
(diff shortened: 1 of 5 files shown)


lirenhe (approved):

weidongxu-microsoft (commented):
In which case the nullable would have a non-generated name? `union NullableModel { Model, null }` ?

What's the rule for generated name on nullable? The name of non-null type, or something else?

tadelesh:
> In which case the nullable would have a non-generated name? `union NullableModel { Model, null }` ?

yes. only named union with `null`.

> What's the rule for generated name on nullable? The name of non-null type, or something else?

it follows the same rule of anonymous model/union. something like `ModelProperty` for `model Model { property: string | null}`.

weidongxu-microsoft (approved):

Issue #2072, opened 2025-01-14
[Bug]: TCGC, lroMetadata.finalResponse.resultPath is neither clientName nor serializedName


weidongxu-microsoft:
### Describe the bug

TCGC 0.49.1

`lroMetadata.finalResponse.resultPath` is neither clientName nor serializedName

### Reproduction

[...]

PR #2102, opened 2025-01-21
Add a property `resultProperties` in `lroMetadata` to show the downstream how to get the final result of a lro

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/2072

The `resultPath` in `LroMetadata` uses `string` as its type.
A `string` type could mean many things, in this case, for a property, it could have three names (at least): its name in typespec, its  `clientName`, its `encodedName`.
This PR adds a new property in that model to replace this string type property to avoid such confusion to reference the corresponding property, instead use one specific name of the property.

We could have future work items of changing pageable result in similar ways.


iscai-msft (commented):
with `.resultPath` we were trying to point users to how to get to the actual result (which is shown in the method's `result`). How does `.resultProperties` tell us how to get to the result, instead of what the result is. I hear what you're saying though about confusion over whether the property name is the client name / encoded name etc.

ArcturusZhang:
> with `.resultPath` we were trying to point users to how to get to the actual result (which is shown in the method's `result`). How does `.resultProperties` tell us how to get to the result, instead of what the result is. I hear what you're saying though about confusion over whether the property name is the client name / encoded name etc.

this `resultProperties` tells the consumer on exactly which property to use to get to the result.
We make it an array to be defensive in case in the future lro could have multiple layers like paging info
why this `resultProperties` is better than `resultPath` because this array contains the exact same reference of those properties the customer need to use. If they want to use client name, they could get it via `.name`, if they want to use serialized name, they could get it via `.serializationOptions.json.name`.

To make your statement more precise:
> with `.resultPath` we were trying to point users to how to get to the actual result (which is shown in the method's `result`)
it should be:
with `.resultProperties` we were trying to point users to the **properties** they want to follow on the model `envelopeResult` to get to the **property** that holds the actual result (which is shown in the method's `result` **property**)

iscai-msft:
oh I see, I got the wrong idea from your PR. Yes this makes sense, so it's a `resultPath` that instead of a string is a path through the properties. We should discuss this tonight in our office hours. Thanks @ArcturusZhang!

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -529,6 +529,11 @@ export interface SdkMethodResponse {
   kind: "method";
   type?: SdkType;
   resultPath?: string; // if exists, tells you how to get from the service response to the method response.
+  /**
+   * An array of properties to fetch {result} from the {response} model. Note that this property is available only in some LRO patterns.
+   * Temporarily this is not enabled for paging now.
+   */
+  resultProperties?: SdkModelPropertyType[];
```
ArcturusZhang, line 536: suggestions: `resultSegments`

tadelesh (approved):

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -364,6 +358,41 @@ function getServiceMethodLroMetadata(
...
+    // find the property inside the envelope result using the final result path
+    let sdkProperty: SdkBodyModelPropertyType | undefined = undefined;
+    for (const property of envelopeResult.properties) {
+      if (property.__raw === undefined || property.kind !== "property") {
+        continue;
+      }
+      if (property.__raw?.name === resultPath) {
+        sdkProperty = property;
+        break;
+      }
```
weidongxu-microsoft, line 386: nit, personally I'd avoid `continue` in loop if possible. E.g. here we only need to reverse the condition of the `if`.

weidongxu-microsoft (approved):

PR #2148, opened 2025-01-28
[WIP] x-ms-identifiers validating core approach

AlitzelMendez:


AlitzelMendez:
talked with Mark and this is very azure-resource-manager specific, closing

PR #2156, opened 2025-01-30
Adds typespec-azure-telemetry package

chrisradek:


.chronus/changes/add-azure-telemetry-take-3-2025-0-30-10-37-58.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
```
timotheeguerin, line 2:
```suggestion
changeKind: internal
```

Just as it will show up in the following version otherwise

packages/typespec-azure-telemetry/src/http-overrides/fetch.ts
```diff
@@ -0,0 +1,31 @@
+import { ExtensionConfig } from "../reporters/types.js";
+
+export function getFetchHttpOverride(): ExtensionConfig["httpXHROverride"] {
```
timotheeguerin, line 3: whats the reason for those overrides, can add a comment?

packages/typespec-azure-telemetry/src/http-overrides/node-http.ts
```diff
@@ -0,0 +1,53 @@
+import type { IncomingHttpHeaders } from "http";
+import * as https from "https";
+import { ExtensionConfig } from "../reporters/types.js";
+
+export function getNodeHttpOverride(): ExtensionConfig["httpXHROverride"] {
```
timotheeguerin, line 5: same here

packages/typespec-azure-telemetry/src/reporters/telemetry-reporter.ts
```diff
@@ -0,0 +1,70 @@
...
+export function createTelemetryReporter(props: CreateTelemetryReporterProps): TelemetryReporter {
+  try {
+    // Some quick checking to return no-op reporter
+    if (!props || !props.instrumentationKey || isTelemetryDisabled()) {
+      return createNoOpTelemetryReporter();
+    }
+
+    const appInsightsClient = getAppInsightsClient({
+      instrumentationKey: props.instrumentationKey,
+      httpXHROverride: supportsFetch() ? getFetchHttpOverride() : getNodeHttpOverride(),
```
timotheeguerin, line 21: don't we always have fetch? compiler depends on it too

packages/typespec-azure-telemetry/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+```bash
+npm install @azure-tools/typespec-azure-telemetry
+```
+
+## Telemetry Configuration
+
+Telemetry collection is on by default.
+
+To opt out, set the environment variable `DISABLE_TYPESPEC_AZURE_TELEMETRY` to a truthy value.
```
timotheeguerin, line 15: can add basic docs on how to use this package?

packages/typespec-azure-telemetry/src/utils/common-properties.browser.ts
```diff
@@ -0,0 +1,18 @@
...
+
+function getRandomUUID() {
+  try {
+    if (typeof crypto !== "undefined" && typeof crypto.randomUUID === "function") {
+      return crypto.randomUUID();
+    }
+  } catch {
+    // ignore error
+  }
+  return;
```
timotheeguerin, line 17: what happens if it errors out? no session id or empty id?

chrisradek: No sessionId in that case. It will still show up in app insights, but each event will be treated as being part of their own session.

That said - this file is mostly inconsequential for now since we'll use the NoOpTelemetryReporter in browser builds.

PR #2161, opened 2025-01-31
[tcgc] add `namespace` flag

iscai-msft:


packages/typespec-client-generator-core/src/context.ts
```diff
@@ -79,6 +79,7 @@ export async function createSdkContext<
     decoratorsAllowList: [...defaultDecoratorsAllowList, ...(options?.additionalDecorators ?? [])],
     previewStringRegex: options?.versioning?.previewStringRegex || tcgcContext.previewStringRegex,
     disableUsageAccessPropagationToBase: options?.disableUsageAccessPropagationToBase ?? false,
+    namespace: context.options["namespace"],
```
m-nash, line 82: I might be missing it but where are we blocking / ignoring this outside of azure flavor?  @srnagar if I recall correctly that is what we were doing is azure only since thats the only use case that actually needs this?

iscai-msft: My understanding is we weren't going to block unbranded from using it, but we are really mainly adding it for Azure compatibility.

The general use case I have for this is:
1. there's a `main.tsp` that defines the service under one main namespace (say `namespace DocumentIntelligence`)
2. there's a `client.tsp` with client-customizations under one main namespace (say `namespace DocumentIntelligenceCustomizations`)
3. we use `namespace` flag in tspconfig.yaml to give the azure name, so `namespace: Azure.DocumentIntelligence`

iscai-msft: I personally feel like it's cleaner to keep it as is, and when we split up tcgc, we can keep `namespace` in the `tcgc` library. I think I'd almost prefer if we only want azure to use it, to call it something like `azureNamespace`. I don't feel very strongly about this though

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -13,6 +13,7 @@ const EmitterOptionsSchema: JSONSchemaType<SdkEmitterOptions> = {
     "examples-directory": { type: "string", nullable: true },
     "examples-dir": { type: "string", nullable: true },
     "emitter-name": { type: "string", nullable: true },
+    namespace: { type: "string", nullable: true },
```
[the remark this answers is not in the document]
iscai-msft: we don't actually since there isn't any weird characters like a `-`. I originally had it with quotes, but I think formatting got rid of it for me

packages/typespec-client-generator-core/src/context.ts
srnagar: Yeah, I agree with just keeping it as is.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -593,3 +595,25 @@ export function getValueTypeValue(
...
+
+  // We explicitly want namespaces that are used to define a client, not namespaces that define the service of a client
+  const clientNamespaces = listClients(context).map((x) =>
+    x.type.kind === "Namespace" ? x.type : x.type.namespace,
+  );
+  const globalNamespaces = rootNamespaces.filter((x) => clientNamespaces.includes(x));
+  // if we override with namespace flag, we should override the global namespace to the namespace flag
+  if (globalNamespaces.length !== 1) {
+    throw new Error(
+      "You can only use the `--namespace` flag in conjunction with a single namespace.",
```
srnagar, line 615: Can we provide a bit more details in this error message? What does "in conjunction with a single namespace" mean in this context?

srnagar (approved):

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: another concern: for service uses azure core or azure arm template will have models with `Azure` root namespace which is another root namespaces different with service's root namespace. will this also be baned?

iscai-msft: that isn't an issue because we go based off of where clients are defined

iscai-msft: switched to using a diagnostic. This is slightly breaking, bc we are technically breaking some public method signatures. However, since idt any emitters actually call these functions, I think it's ok. Let me know your feelings

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -593,3 +595,45 @@ export function getValueTypeValue(
...
+      segments.unshift(currNamespace.name);
+      if (rootNamespaces.includes(currNamespace)) {
+        globalNamespaces.push(segments.join("."));
+        break;
+      }
+      currNamespace = currNamespace.namespace;
+    }
+  }
+  // if we override with namespace flag, we should override the global namespace to the namespace flag
+  if (globalNamespaces.length !== 1) {
```
weidongxu-microsoft, line 628: I still not sure whether I understand what the logic for.

If we have 2 client, one on TypeSpec namespace `Service.Client1`, anther on `Service.Client2`, the logic here would anyway produce a `globalNamespaces` with 2 items (`rootNamespaces` would contain 1 `Service`)?

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -593,3 +595,45 @@ export function getValueTypeValue(
...
+/**
+ * Get all of the root user-defined namespaces in a TypeSpec definition if they're tied to a client
+ * @param context
+ * @returns
+ */
+export function getRootUserDefinedNamespaceName(
+  context: TCGCContext,
+): [string, readonly Diagnostic[]] {
+  const diagnostics = createDiagnosticCollector();
+  const rootNamespaces = [...context.program.getGlobalNamespaceType().namespaces.values()];
```
tadelesh, line 608: a little bit confusing with the logic. what will `getGlobalNamespaceType` return?

iscai-msft: global namespace returns the step above all of the namespaces. For example, if you just have an empty typespec with imports to `TypeSpec` libraries, it will return an unnamed namespace at the root. It will then have `.namespaces = [TypeSpecNamespace]`. If you have something more robust, i.e. what we will see in the usages of the namespace flag,

```tsp
...
using TypeSpec.Http;
using Azure.ClientGenerator.Core;

namespace Communication {
 ...
}
```
it will return an unnamed namespace with `.namespaces = [TypeSpecNamespace, AzureNamespace, CommunicationNamespace]`.

What i'm trying to find here is the root namespace defined by the user

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1209,19 +1212,30 @@ export const $clientNamespace: ClientNamespaceDecorator = (
...
-): string {
+): [string, readonly Diagnostic[]] {
+  const diagnostics = createDiagnosticCollector();
   const override = getScopedDecoratorData(context, clientNamespaceKey, entity);
-  if (override) return override;
+  if (override) {
+    const rootNamespaceName = diagnostics.pipe(getRootUserDefinedNamespaceName(context));
+    // if you've defined a namespace flag, we replace the root name with the flag
+    return diagnostics.wrap(
+      override.replace(rootNamespaceName, context.namespace || rootNamespaceName),
```
tadelesh, line 1222: better to judge undefined of `context.namespace` to prevent always find `rootNamespaceName` part.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -593,3 +595,45 @@ export function getValueTypeValue(
...
+  // if we override with namespace flag, we should override the global namespace to the namespace flag
+  if (globalNamespaces.length !== 1) {
+    diagnostics.add(
+      createDiagnostic({
+        code: "unclear-namespace-overriding",
+        messageId: "default",
+        format: {},
+        target: context.program.getGlobalNamespaceType(),
+      }),
+    );
```
tadelesh, line 636: can we just report there is multiple user defined ns? i don't prefer couple this function with namespace override config.

tadelesh (commented):
overall looks good. some impl optimization comments left.

weidongxu-microsoft:
The current implementation seems doing (in the case that `namespece` defined in emitter option -- `<EmitterOptionNamespace>`)
1. find a dev defined base TypeSpec namespace, as `<RootUserDefinedNamespace>`
2. error if there is more than one candidate
3. for all TypeSpec namespace encountered, replace the `<RootUserDefinedNamespace>.Suffix` to `<EmitterOptionNamespace>.Suffix`

---

If this is the implementation, there may be a few problem

First, in customization of client.tsp, it is possible to write clients in more than one namespace (and the Service itself would be in yet another namespace)
e.g.
```
// main.tsp
namespace Service;
```
```
// client.tsp
namespace Customization;

namespace FirstClient {
  interface OperationGroup1;
}
namespace SecondClient {
  interface OperationGroup2;
}
```
In such case, it is pretty difficult to resolve the `<RootUserDefinedNamespace>` (there is `Customization.`FirstClient`, `Customization.SecondClient`, and `Service` and maybe more)

Second, there is some service that defines more than one namespace (and models in a completely different namespace)
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/liftrneon/Neon.Postgres.Management/client.tsp#L6-L14
`Neon.Postgres` is the namespace for client, and `LiftrBase*` has some models.

Also, some `TypeSpec.*` or `Azure.*` models could complicate the problem. SDK would like them generated to `<EmitterOptionNamespace>`

---

In my opinion, it seems easier to just map any TypeSpec namespace to `<EmitterOptionNamespace>` (meaning, `Serivce`, `Service.Model`, `TypeSpec.Rest.Resource`, etc. all be `<EmitterOptionNamespace>` in SDK namespace), if no `@clientNamespace` specified.
If `@clientNamespace` is specified to a TypeSpec namespace, all client/model/enum under this TypeSpec namespace would use this overridden SDK namespace.
If `@clientNamespace` is specified to a TypeSpec model, the model would use this overridden SDK namespace.

tadelesh:
for first problem, current pr will find `Customization` and replace it with user's config.
for second problem, we should consider it carefully. for models not belong to the client's ns, current pr does not change the ns's prefix, shall we put them in the original ns?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1224,6 +1227,9 @@ export function getClientNamespace(
 }
 
 function getNamespaceFullNameWithOverride(context: TCGCContext, namespace: Namespace): string {
+  if (context.namespace) {
+    return context.namespace;
+  }
   const segments = [];
   let current: Namespace | undefined = namespace;
   while (current && current.name !== "") {
```
weidongxu-microsoft, line 1235: I assume the `@clientNamespace` (if specified on TypeSpec Namespace or Model) is having higher priority than `namespace` in emitter option.

@srnagar for confirmation.

tadelesh:
@iscai-msft just fyi, this [pr](https://github.com/Azure/typespec-azure/pull/2219) may conflict with your pr. it's better to switch this pr to merge to release branch.

Issue #2112, opened 2025-01-22
[TCGC] introduce cache for model properties


ArcturusZhang:
Unlike other types, we have cache in TCGC, such as `SdkModelType`, etc.
This ensures us if we put the same instance of `Model` from compiler, we always get back the same instance of `SdkModelType`.

We should introduce cache for model properties as well, now every time we give the same `ModelProperty` into `getSdkModelPropertyType`, it gives back a new instance every time.

PR #2167, opened 2025-02-05
add cache for `SdkModelPropertyType`

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/2112

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -1304,41 +1304,46 @@ export function getSdkModelPropertyType(
 ): [SdkModelPropertyType, readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
 
-  const clientParams = operation
-    ? context.__clientToParameters.get(getLocationOfOperation(operation))
-    : undefined;
-  const correspondingClientParams = clientParams?.find((x) =>
-    twoParamsEquivalent(context, x.__raw, type),
-  );
-  if (correspondingClientParams) return diagnostics.wrap(correspondingClientParams);
-  const base = diagnostics.pipe(getSdkModelPropertyTypeBase(context, type, operation));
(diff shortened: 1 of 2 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1304,41 +1304,46 @@ export function getSdkModelPropertyType(
...
 
-  const clientParams = operation
-    ? context.__clientToParameters.get(getLocationOfOperation(operation))
-    : undefined;
-  const correspondingClientParams = clientParams?.find((x) =>
-    twoParamsEquivalent(context, x.__raw, type),
-  );
-  if (correspondingClientParams) return diagnostics.wrap(correspondingClientParams);
-  const base = diagnostics.pipe(getSdkModelPropertyTypeBase(context, type, operation));
+  let property = context.referencedPropertyMap?.get(type);
```
archerzz, line 1307: Maybe we should initialize `context.referencePropertyMap` the first time we access it, instead of initializing it when we want to put the property in the cache in `updateReferencedPropertyMap()`?

archerzz (approved):

packages/typespec-client-generator-core/src/types.ts
ArcturusZhang: well I am mimicking the same way we did for types `referencedTypeMap` therefore I think this should be output the scope of this PR.
JS does not have as convenient mechanism as we have in dotnet that we could let a property to initialize the first time we use it. To do this we have to construct a `class`, but here we only have those interfaces, therefore I think what we have now is our best compromise

tadelesh:
could you add at least one test to see if two call can get one property type?
also, there is some places that will create property, such as discriminated property. with intuition, i believe it is irrelevant with your pr. but please confirm.

tadelesh (approved):

iscai-msft (approved):

PR #2172, opened 2025-02-05
[tcgc] switch to `getLifecycleVisibilityEnum` call from deprecated `getVisibility`

### LINKED ISSUE #2171: Replace deprecate `getVisibility` usage in TCGC
`getVisibility` is [deprecated in core](https://github.com/microsoft/typespec/blob/main/packages/compiler/src/core/visibility/core.ts#L276). TCGC uses `getVisibility` in [types.ts](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/types.ts#L28). 

This method should be replaced by `getVisibilityForClass`.

iscai-msft:
fixes #2171

### DIFF
--- a/packages/typespec-client-generator-core/src/public-utils.ts
+++ b/packages/typespec-client-generator-core/src/public-utils.ts
@@ -12,9 +12,9 @@ import {
   createDiagnosticCollector,
   getEffectiveModelType,
   getFriendlyName,
+  getLifecycleVisibilityEnum,
   getNamespaceFullName,
   getProjectedName,
-  getVisibility,
   ignoreDiagnostics,
   listServices,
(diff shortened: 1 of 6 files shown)


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -12,9 +12,9 @@ import {
   createDiagnosticCollector,
   getEffectiveModelType,
   getFriendlyName,
+  getLifecycleVisibilityEnum,
```
srnagar, line 15: this import can be removed as it's not used here.

srnagar (approved):
Thanks @iscai-msft for the quick fix.

PR #2173, opened 2025-02-05
Uptake versioning mutators

timotheeguerin:
fix https://github.com/microsoft/typespec/pull/4931


packages/typespec-autorest/src/openapi.ts
```diff
@@ -458,6 +458,7 @@ export async function getOpenAPIForService(
         prop.type.kind === "Scalar" &&
         ignoreDiagnostics(
           program.checker.isTypeAssignableTo(
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
```
markcowl, line 461: What will projectionBase have at this point, shouldn't this just be prop.type?

packages/typespec-autorest/src/openapi.ts
```diff
@@ -750,6 +751,7 @@ export async function getOpenAPIForService(
   }
 
   function isBytes(type: Type) {
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
```
markcowl, line 754: same question

markcowl (commented):
A couple of questions.  ALso, will want to look at how this impacts typespec-providerhub-controller - I *think* since this emitter is using canonical versioning and does not make use of the version map or projections, that it likely needs no change, but it would be good to check.

packages/typespec-autorest/src/openapi.ts
timotheeguerin: Projection base was to make it work with std scalars but with mutators I preventing mutating the compiler types so it's unnecessary

chrisradek (approved):
Changes look good to me - I'm not sure how to verify the provider hub changes Mark mentioned though.

Issue #2175, opened 2025-02-06
[TCGC] In some cases, the properties in a model may not have `serializationOptions`


ArcturusZhang:
Spec: https://github.com/Azure/azure-rest-api-specs/blob/9df71d5a717e4ed5e6728e7e6ba2fead60f62243/specification/batch/Azure.Batch/routes.tsp#L114

In the above operation, all the properties of the body type `BatchPoolCreateContent` do not have `serializationOptions` value:
![Image](https://github.com/user-attachments/assets/9da00b1d-04b4-4f04-88af-b4654aa39f67)

which leads to example value loss.

This should be a regression when we introduced `serializationOptions`.

tadelesh:
root cause is the json/xml judge in tcgc is not robust. with this content-type: https://github.com/Azure/azure-rest-api-specs/blob/9df71d5a717e4ed5e6728e7e6ba2fead60f62243/specification/batch/Azure.Batch/common.tsp#L132, tcgc failed to identify the request model as a json payload.

Issue #1999, opened 2024-12-18
[tcgc] should tcgc unify the content type logic for bytes format inference


qiaozha:
for this case, the contentType is a union of application/json or text, but tcgc returns `bytes` as the default encoding 

```
  @get
  getSchemaByVersion is Azure.Core.Foundations.Operation<
    Azure.Core.Foundations.ItemKeysOf<Version>,
    OkResponse & SchemaContent & SchemaContentType
  >;

@doc("The content type for given schema. Each schema type has an associated content-type.")
model SchemaContentType {
  @doc("The content type for given schema.")
  @header("Content-Type")
  contentType: SchemaContentTypeValues;
}

@doc("Content of the schema.")
model SchemaContent {
  @body
  @doc("String representation (UTF-8) of the schema.")
  content: bytes;
}

union SchemaContentTypeValues {
  avro: "application/json; serialization=Avro",
  @added(ServiceApiVersions.V2022_10)
  json: "application/json; serialization=json",
  @added(ServiceApiVersions.V2022_10)
  custom: "text/plain; charset=utf-8",
  @added(ServiceApiVersions.V2023_07_01)
  protobuf: "text/vnd.ms.protobuf",
}
```
After some investigation with @tadelesh offline, tcgc is using strict equal to "application/json" to set the default encoding as base64 
https://github.com/Azure/typespec-azure/blob/5506281742813438a2f2607cc152e08be35a16a1/packages/typespec-client-generator-core/src/types.ts#L184

But, previously, we have an ask from client emitter side that suggests us to leverage the logic from dotnet https://github.com/Azure/autorest.typescript/issues/2200 and fixed in https://github.com/Azure/autorest.typescript/pull/2258/files#diff-7dd8f9407e7839a21c48e98c9da493fa7da52885017e1df4dbf18a346b9933c8 

I wonder if tcgc could also leverage this ?

PR #2176, opened 2025-02-06
[tcgc] refine content type judgement

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1999
fix: https://github.com/Azure/typespec-azure/issues/2175


packages/typespec-client-generator-core/src/media-types.ts
```diff
@@ -0,0 +1,105 @@
...
+        subtype: parsed[2],
+        suffix: parsed[3],
+        parameter: parsed[4],
+      };
+    }
+  }
+  return undefined;
+}
+
+export function knownMediaType(mediaType: string) {
```
ArcturusZhang, line 46: should this be `getKnownMediaType` or something? this method name is missing a verb

ArcturusZhang: also I think we should put the explicit return type here.

packages/typespec-client-generator-core/src/media-types.ts
ArcturusZhang, line 1: so if these methods are not designed to be called by those consumers of TCGC (such as our emitters), we should simplify these methods.
for instance, `knownMediaType` method calls `parseMediaType` which calculates a big regex is not quite efficient.
So are `isMediaTypeJson` and `isMediaTypeXml`.

We could change these methods to accept that parsed struct returned by `parseMediaType` to avoid this.

For instance, I notice in the other file in this PR, we check if the media type is json, if so do something, and then we check if it is xml, if so do something else. During this process, that regex is calculated twice, which is unnecessary.

packages/typespec-client-generator-core/src/media-types.ts
tadelesh: code is copied from autorest [impl](https://github.com/Azure/autorest/blob/faf5c1168232ba8a1e8fe02fbc28667c00db8c96/packages/libs/codegen/src/media-types.ts#L53). let me remove the useless code for now.

iscai-msft (approved):

packages/typespec-client-generator-core/src/media-types.ts
tadelesh: tcgc has many places that could be optimized. but the optimization may come with coupling (need to pass context to reduce recalculation). i prefer to keep the simplicity for now.

weidongxu-microsoft (approved):

PR #2179, opened 2025-02-06
[tcgc] allow models-only packages

### LINKED ISSUE #1755: [tcgc] explore what it would mean to have a `@client(modelsOnly)`
Right now, our clients are exclusively tied to `@service` decorators. If there's no `@service` decorator, we don't have clients. However, for libraries like EventGrid, which are models only, we aren't able to generate anything. In cases like this, we would still like to return a list of models.

Would this mean explicitly adding to the `@client` decorator that I am a models-only library, so I don't need a service?

cc @timotheeguerin @tadelesh @swathipil

iscai-msft:
fixes #1755

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -25,6 +25,7 @@ import {
   getEncode,
   getKnownValues,
   getLifecycleVisibilityEnum,
+  getLocationContext,
   getSummary,
   getVisibilityForClass,
   ignoreDiagnostics,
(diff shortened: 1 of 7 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1867,8 +1868,11 @@ export function handleAllTypes(context: TCGCContext): [void, readonly Diagnostic
     }
   }
   // update for orphan models/enums/unions
-  for (const client of listClients(context)) {
-    const namespaces = [client.service];
+  const allNamespaces = [...context.program.getGlobalNamespaceType().namespaces.values()].filter(
```
tadelesh, line 1871: when generating name for anonymous model, there is a similar logic. see here: https://github.com/Azure/typespec-azure/blob/70dab25c415321e29f251a4ee42427ef13264ce8/packages/typespec-client-generator-core/src/public-utils.ts#L349.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1867,8 +1868,11 @@ export function handleAllTypes(context: TCGCContext): [void, readonly Diagnostic
...
+    (x) => getLocationContext(context.program, x).type === "project",
```
ArcturusZhang, line 1872: this apparently is causing a regression when we are bumping the new version of typespec and TCGC.
Now all orphan models are somehow removed.

PR #2180, opened 2025-02-07
Point to tsp-client usage docs in intro page

catalinaperalta:
Addressing feedback provided by @heaths that tsp-client usage docs should be more visible and easier to find from the intro page for folks working on Azure specs and client libraries.


website/src/content/docs/docs/intro.md
```diff
@@ -55,6 +55,7 @@ for specific scenarios:
 
 Some additional sources:
 
+- The `tsp-client` CLI provides several commands to help during the TypeSpec development process, such as converting a swagger specification to a TypeSpec project, providing a simple interface to generate client libraries within the `azure-sdk-for-<language>` repositories, and more. For more information on the CLI tool, see the [tsp-client usage](https://aka.ms/azsdk/tsp-client) documentation
```
heaths, line 58: <https://aka.ms/azsdk/tsp-client> 404s. But wherever it leads, will that include installation instructions? It wasn't so much usage as much as how we acquire it. It would be better in the Installation page on this site.

heaths (changes requested):

website/src/content/docs/docs/intro.md
catalinaperalta: Yes the tsp-client usage docs contain installation instructions. The url to the page that was being targeted before just changed with the recent docs updates. I'd give it a few minutes before testing the link again

heaths (approved):

website/src/content/docs/docs/intro.md
catalinaperalta: Moved to the installation page @heaths

PR #2187, opened 2025-02-08
azure-http-specs, config java namespace, avoid _Specs_ appear in SDK namespace

weidongxu-microsoft:
context: Java does not want to see the default `_specs_` in its namespace https://github.com/Azure/autorest.java/tree/main/typespec-tests/src/main/java/_specs_

### DIFF
--- a/.chronus/changes/azure-http-specs_config-java-namespace-2025-1-8-11-59-45.md
+++ b/.chronus/changes/azure-http-specs_config-java-namespace-2025-1-8-11-59-45.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@azure-tools/azure-http-specs"
+---
+
+Update SDK namespace for Java, avoid _Specs_ appears in SDK namespace.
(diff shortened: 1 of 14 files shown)


iscai-msft (approved):
i wonder if all languages will be moving towards this, since python has the same issue. I think this is the best way to do it now without breaking everyone. We could gradually move over this way based off of language availability. Once everyone's moved over / agrees to be broken, and the namespace flag is in we can switch over to using that instead

weidongxu-microsoft:
> i wonder if all languages will be moving towards this, since python has the same issue. I think this is the best way to do it now without breaking everyone. We could gradually move over this way based off of language availability. Once everyone's moved over / agrees to be broken, and the namespace flag is in we can switch over to using that instead

Yeah. I could ask around whether other language also want to remove the `_Specs_`. Unsure in which case we had to use the `_Specs_` (we likely used it for dis-ambiguate, but for [mgmt cases](https://github.com/Azure/typespec-azure/blob/9cbd963b1050826c44892a5c42a4a288f17205e3/packages/azure-http-specs/specs/azure/resource-manager/resources/main.tsp#L24) it seems normal `Azure.*` namespace can work).

Issue #2133, opened 2025-01-27
[Bug]: tcgc tagged a wrong name to anonymous union


qiaozha:
### Describe the bug

this is the tsp.
```
union Prompt {
  string,
  string[],
  TokenArray,
  TokenArrayArray,
  null,
}
```
the Prompt type should be a nullable, but Prompt.type's type should be an anonymous non-nullable union, however, what tcgc provide is still a named non-nullable union Prompt

![Image](https://github.com/user-attachments/assets/7262a0d6-f137-41e0-8afc-da396cd1064a)

### Reproduction

N/A

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Issue #2132, opened 2025-01-26
[Bug]: tcgc crashed when handling this union


qiaozha:
### Describe the bug

this is the tsp.
```
union A {
  null,
  A
}
```
And this is the error message 

![Image](https://github.com/user-attachments/assets/970bb98a-153e-42c9-85e8-58a67c083093)

### Reproduction

N/A

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2196, opened 2025-02-11
[tcgc] refine union logic and fix known issues

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2132
fix: https://github.com/Azure/typespec-azure/issues/2133


iscai-msft (approved):

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -49,6 +49,12 @@ export const $lib = createTypeSpecLibrary({
         default: "Cannot have a union containing only null types.",
       },
     },
+    "union-circular": {
+      severity: "error",
+      messages: {
+        default: "Cannot have a union containing self.",
+      },
+    },
```
ArcturusZhang, line 57: I think we should open an issue in microsoft/typespec because typespec as a language should never allow this - a type defined in this way is impossible to exist.

tadelesh: issue here: https://github.com/microsoft/typespec/issues/6106

PR #2202, opened 2025-02-11
add design doc

iscai-msft:


design-doc.md
```diff
@@ -0,0 +1,137 @@
+# Design Document: Adding Support for Next Link Paging with Re-injected Query Parameters
+
+## Overview
+
+Currently, the client treats the `nextLink` in paging as opaque and does not inject query parameters that were part of the initial request when following the next link. This proposal aims to enhance the client by adding capability to reinject query parameters into next requests for paging operations.
+
+## Problem
+
+This is an existing issue in 1-2 brownfield Azure services: Keyvault and potentially AppConfig.
```
lmazuel, line 9: I guarantee it's in AppConfig :). And in Storage. Storage has two problems: continuation token and query parameters.

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Keyvault
+
+KV only has one instance in `listCertificates`, there is an optional query parameter `includePending`, which dictates whether to include pending certificates in the returned list of certificates. The service does not automatically reinject the value of `includePending` into the next link, so for all subsequent calls during paging, you will have to reinject `includePending=True` if you would like to continue seeing pending certificates in subsequent pages.
+
+### AppConfig
+
+I can't get in touch with anyone on the client-side team working with AppConfig since they're all no longer with us.
```
lmazuel, line 17: Take any list operations with optional parameters, and we can check the Python implementation or do a call, it's quite easy to check.

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+
+The examples used throughout the rest of this doc will reference the `includePending` issue from KV.
+
+## Requirements
+
+1. Solves the issue for our 3 Azure services: KeyVault, Storage, AppConfig.
+2. Open question of whether this support is required for 3p. For Azure services, requiring clients to reinject parameters is against guidelines. However, should we restrict 3p authors from describing this in their specs, even if it might not be best practice?
```
lmazuel, line 24: This is required for 3P. The whole discussion started about including a resolution of this for the GA of TypeSpec, which is a 3P event.

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+3. Work with templates
```
lmazuel, line 25: Good point

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Pros
+
+1. Definitely easiest path forward, since all of the code is pretty much here for us. All we have to do is fix [this](https://github.com/Azure/typespec-azure/issues/1880) issue
+2. No new decorators or syntax introduced
+
+#### Cons
+
+1. Ugly definition
+2. We are gradually moving away from paging from `azure-core`, and unclear if there would be a counterpart in unbranded paging, so this won't be a good idea for 3p support or non-brownfield service support
```
lmazuel, line 49: Are you implying that `nextPageOperation` doesn't exist in 3P? If that's the case, I feel it should be done.

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+3. No 3p support
+
+#### Todos
+
+1. Fix [this](https://github.com/Azure/typespec-azure/issues/1880) github issue
+2. Ensure support for `@nextPageOperation` in all language emitters
```
lmazuel, line 55: I would expect TCGC to abstract that

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Cons
+
+- Adds another decorator
+- Definition isn't the most clear
+- Do we need to add a new decorator if we can polish up `@nextPageOperation` to work and persist that decorator?
+
+#### Todos
+
+1. Figure out where to put the `@reinject` decorator
```
lmazuel, line 129: We can start TCGC at first and change it later

design-doc.md
```diff
@@ -0,0 +1,137 @@
...
+2. Implement the `@reinject` decorator
+3. Get emitters to understand the `@reinject` decorator
+
+## Things to Decide
+
+- [ ] design to choose
+- [ ] Should this be for azure brownfield services only? Thoughts about this being 3p behavior?
```
lmazuel, line 136: This shold include 3P, at least long term, so we should have a vision of how we move to 3P.

iscai-msft: Sounds good. The reason I have it open is after discussing with some people on the TypeSpec team, they're not convinced it should be supported for 3p, so I feel that point is still up for debate to some, so we should definitely keep talking about it

design-doc.md
iscai-msft: yeah, tcgc would have to design how to abstract that, I'm thinking through a templated url value. Let me update the docs with this suggestion actually

design-doc.md
iscai-msft: Yeah, `nextPageOperation` is a decorator in `Azure.Core`, and there aren't current plans to move it to TypeSpec core

design-doc.md
iscai-msft: Let me check this myself. I'll follow up with storage people offline, because I was told explicitly that this wasn't an issue

design-doc.md
```diff
@@ -0,0 +1,150 @@
+# Design Document: Adding Support for Next Link Paging with Re-injected Query Parameters
+
+## Overview
+
+Currently, we treat the `nextLink` in paging as opaque and does not inject query parameters that were part of the initial request when following the next link. This proposal aims to enhance our `nextLink` handling by adding capability to reinject query parameters into next requests for paging operations.
```
srnagar, line 5: Are we limiting this to reinjecting only query params or should we include request headers too?

design-doc.md
```diff
@@ -0,0 +1,150 @@
...
+
+#### Todos
+
+1. Fix [this](https://github.com/Azure/typespec-azure/issues/1880) github issue
+2. Ensure support for `@nextPageOperation` in all language emitters
+
+### Adding an incomplete next link scalar type
+
+```tsp
+scalar incompleteNextLink<ParametersToReinject[]> extends url;
```
srnagar, line 71: We should consider naming this scalar something different - `nextLinkWithAdditionalParams<>`

In the KV scenario, the next link is not *incomplete* as the next link returned by the service still works. The only issue is without `includePending` appended to the next link, the results of the next page don't include the pending certs.

design-doc.md
```diff
@@ -0,0 +1,150 @@
...
+
+model ListCertificateOptions {
+    includePending?: string;
+}
+
+model Page {
+    @pagedItems items: Certificate[];
+    @nextLink nextLink: incompleteNextLink<[ListCertificateOptions.includePending]>;
```
srnagar, line 79: Is there a simpler way to define that all query params of the original request (1st page) should be added to the nextLink?

design-doc.md
```diff
@@ -0,0 +1,150 @@
...
+If we found more use cases, we could elevate the definition of this scalar into either the `Azure.Core` library, or even more generic.
+
+#### Pros
+
+1. Flexible level of support. We could start with the scalar just being a local definition, but if we found other services needing this, we could elevate it up further in the library-chain
+2. Kind of fits in with the paging design that is currently in tsp.
+
+#### Cons
+
+1. Slightly more involved than getting `@nextPageOperation` to work, but not by much.
```
srnagar, line 94: For the spec author, discoverability of the scalar option is a challenge.

design-doc.md
```diff
@@ -0,0 +1,150 @@
...
+- [ ] Usage with azure core templates? Usage with paging templates?
+
+#### Pros
+
+- Simple implementation, which is good considering our need to GA, and the fact that this definitely is more of a corner case.
+
+#### Cons
+
+- Adds another decorator
+- Definition isn't the most clear
```
srnagar, line 137: The name `@reinject` isn't very obvious. If we name this `@nextLinkQuery`, it would be easy to understand that this is a query param for the next link.



```tsp
list is Azure.Core.ResourceOperations<ServiceTraits>.ResourceList<
    Certificate,
    {
        @nextLinkQuery @query includePending?: string;
    },
>;
```

design-doc.md
```diff
@@ -0,0 +1,150 @@
...
+- Do we need to add a new decorator if we can polish up `@nextPageOperation` to work and persist that decorator?
```
srnagar, line 138: `nextPageOperation` is only available for Azure. If we need this to be supported for 3p, we will have to move this up as well.

design-doc.md
iscai-msft: we could allow the template to take in a model . I do know there is the `RequestParameter` concept. I'll bring this up in the meeting and see if mark has any ideas

design-doc.md
iscai-msft: i believe we should limit this to only query params. Are there instances / is this something we want to consider? I'll bring this up in the meeting as a question to answer as well

PR #2203, opened 2025-02-12
fix url for extension resource in mockapi

v-hongli1:
### Context

- Fix URL for Extension Resource.


packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -5,6 +5,10 @@ export const Scenarios: Record<string, ScenarioMockApi> = {};
 const SUBSCRIPTION_ID_EXPECTED = "00000000-0000-0000-0000-000000000000";
 const RESOURCE_GROUP_EXPECTED = "test-rg";
 const LOCATION_EXPECTED = "eastus";
+const SUBSCRIPTION_SCOPE_URI = `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}`;
+const RESOURCE_GROUP_SCOPE_URI = `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/resourceGroups/${RESOURCE_GROUP_EXPECTED}`;
+const RESOURCE_SCOPE_URI = `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/resourceGroups/${RESOURCE_GROUP_EXPECTED}/providers/Azure.ResourceManager.Resources/topLevelTrackedResources/top`;
+const TENANT_SCOPE_RUI = "";
```
tadelesh, line 11: RUI or URI?

v-hongli1: Fixed by [3d7d0ca](https://github.com/Azure/typespec-azure/pull/2203/commits/3d7d0cab433727b91ede08dc31826b3516d68466)

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -202,14 +206,12 @@ Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = passOnSucces
     kind: "MockApiDefinition",
   },
   {
-    uri: "/subscriptions/:subscriptionId/resourceGroups/:resourceGroup/providers/Azure.ResourceManager.Resources/topLevelTrackedResources/:topLevelResourceName/providers/Azure.ResourceManager.Resources/extensionsResources/:extensionName",
+    uri: "/:resourceUri/providers/Azure.ResourceManager.Resources/extensionsResources/:extensionName",
```
XiaofeiCao, line 209: Does this work for four identical `uri`s?
Also, does it allow both single `/` and double `//`?

v-hongli1: Fixed by [9658831](https://github.com/Azure/typespec-azure/pull/2203/commits/9658831fac4fdc139b493a39e100fa115d01cc17)

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -240,115 +239,59 @@ Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_createOrUpdate = p
...
+  createExtensionCreateMockApiDefinition(
+    RESOURCE_GROUP_SCOPE_URI,
+    EXTENSION_RESOURCE_NAME,
+    validResourceGroupExtensionsResource,
+  ),
+  createExtensionCreateMockApiDefinition(
+    SPLIT_RESOURCE_GROUP_SCOPE_URI,
+    EXTENSION_RESOURCE_NAME,
+    validResourceGroupExtensionsResource,
+  ),
```
XiaofeiCao, line 257: I suspect this kind of definition would require client to write test for **BOTH** RESOURCE_GROUP_SCOPE_URI and SPLIT_RESOURCE_GROUP_SCOPE_URI. 
We only need client to write **EITHER** one of them to pass the test.

v-hongli1: I think both resource URIs of the formats `RESOURCE_GROUP_SCOPE_URI` and `SPLIT_RESOURCE_GROUP_SCOPE_URI` should be included in the test scope.

XiaofeiCao: Yes, for mockapi, we need to support both endpoints.
Though for clients that calls API, they only need to call either one(either  `/` or `//`). 
Like the test you write here: https://github.com/microsoft/typespec/pull/5929/files
e.g. We only need to call with resourceUri `EXTENSION_RESOURCE_SUBSCRIPTION_URL`, no need for `SPLIT_EXTENSION_RESOURCE_SUBSCRIPTION_URL`.

The `passOnSuccess` in mockapi may require all the scenarios be called successfully by client to pass this test, which is not what we want.

XiaofeiCao: Would you try [`withServiceKeys`](https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/payload/pageable/mockapi.ts#L38)?

You'll need:
1. Define 4 serviceKeys, `Tenant`, `Subscription`, `ResourceGroup`, `Resource`.
2. Use custom handler
3. In customer hander, return same `pass: {key}` for the same scope(either `/` or `//`), e.g. `pass: Tenant` for tenant level resource.

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -516,13 +922,169 @@ Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_delete = passOnSuc
...
+    handler: (req: MockRequest) =>
+      requestHandler(req, "list", RESOURCE_SCOPE_URI, validResourceExtensionsResource),
     kind: "MockApiDefinition",
   },
 ]);
 
 Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_listByScope = passOnSuccess([
   {
-    uri: "/subscriptions/:subscriptionId/resourceGroups/:resourceGroup/providers/Azure.ResourceManager.Resources/extensionsResources",
+    uri: "//subscriptions/:subscriptionId/resourceGroups/:resourceGroup/providers/Azure.ResourceManager.Resources/extensionsResources",
```
XiaofeiCao, line 1087: Why the change here? IMHO since we are using `withServiceKeys`, there should be no `passOnSuccess` anymore.

v-hongli1: Removed redundant parts by [6d3a56e](https://github.com/Azure/typespec-azure/pull/2203/commits/6d3a56edf3f754deec6eec21e4d52d8d97974e94)

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -151,16 +163,113 @@ const validResourceExtensionsResource = {
...
+    case RESOURCE_SCOPE_URI:
+      return "Resource" as const;
+    default:
+      throw new ValidationError(
+        "Invalid resource uri",
+        `${RESOURCE_GROUP_SCOPE_URI} | ${SUBSCRIPTION_SCOPE_URI} | ${TENANT_SCOPE_URI} | ${RESOURCE_SCOPE_URI}`,
+        resourceUri,
+      );
+  }
+}
```
XiaofeiCao, line 183: Should we check for both "/" and "//"?

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -151,16 +163,113 @@ const validResourceExtensionsResource = {
...
 // extension tracked resource
-Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = passOnSuccess([
+Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = withServiceKeys([
+  "ResourceGroup",
+  "Subscription",
+  "Tenant",
+  "Resource",
+]).pass([
+  {
+    uri: "/:resourceUri/providers/Azure.ResourceManager.Resources/extensionsResources/:extensionName",
```
XiaofeiCao, line 247: or without "/" prefix?

v-hongli1: This version [bd42d17](https://github.com/Azure/typespec-azure/pull/2203/commits/bd42d17208014816dc40104885eff013266f1177) can check both "/" and "//".

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -151,17 +163,99 @@ const validResourceExtensionsResource = {
...
-Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = passOnSuccess([
+Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = withServiceKeys([
+  "ResourceGroup",
+  "Subscription",
+  "Tenant",
+  "Resource",
+]).pass([
   {
-    uri: "/subscriptions/:subscriptionId/resourceGroups/:resourceGroup/providers/Azure.ResourceManager.Resources/extensionsResources/:extensionName",
+    uri: getUri(RESOURCE_GROUP_SCOPE_URI, EXTENSION_RESOURCE_NAME, "get"),
```
XiaofeiCao, line 256: Why does it look like it's still checking only "//"? 
```typescript
const RESOURCE_GROUP_SCOPE_URI = `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/resourceGroups/${RESOURCE_GROUP_EXPECTED}`;

function getUri(resourceUri: string, extensionName: string, requestMethod: string): string {
  switch (requestMethod) {
    case "list":
      return `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`;
    default:
      return `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources/${extensionName}`;
  }
}
```

Did you have generated java code? Did the test pass for both cases?

v-hongli1: This function can generate URIs containing "//" or "/", so both conditions can be check. which has been verified by the test case.

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -248,9 +254,12 @@ function getUri(resourceUri: string, extensionName: string, requestMethod: strin
 // extension tracked resource
 Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = withServiceKeys([
   "ResourceGroup",
+  "SplitResourceGroup",
   "Subscription",
+  "SplitSubscription",
   "Tenant",
   "Resource",
+  "SplitResource",
```
XiaofeiCao, line 262: With this change, user still have to pass both `SplitXX` and `XX` tests. This is the same as `passOnSuccess`.

v-hongli1: Rollback to [bd42d17](https://github.com/Azure/typespec-azure/pull/2203/commits/bd42d17208014816dc40104885eff013266f1177)

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -170,16 +264,30 @@ Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_get = passOnSucces
       status: 200,
       body: json(validResourceGroupExtensionsResource),
     },
+    handler: (req: MockRequest) =>
+      requestHandler(req, "get", RESOURCE_GROUP_SCOPE_URI, validResourceGroupExtensionsResource),
     kind: "MockApiDefinition",
   },
   {
-    uri: "/subscriptions/:subscriptionId/providers/Azure.ResourceManager.Resources/extensionsResources/:extensionName",
+    uri: getUri(RESOURCE_GROUP_SCOPE_URI.substring(1), EXTENSION_RESOURCE_NAME, "get"),
```
XiaofeiCao, line 272: Do we have to always duplicate two operations? 
If `uri: /${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`, what will `resourceUri` be when the request is `//subscriptions/${SUBSCRIPTION_ID_EXPECTED}/providers/Azure.ResourceManager.Resources/extensionsResources`?

v-hongli1: 1. Although the operations are repeated, the· `resourceUri` in the Uri is different. For example, the `resourceUri` of the Subscription scope is: `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}` and `subscriptions/${SUBSCRIPTION_ID_EXPECTED}`
2. The resourceUri is `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}` be when the request is `//subscriptions/${SUBSCRIPTION_ID_EXPECTED}/providers/Azure.ResourceManager.Resources/extensionsResources`.

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -151,17 +163,137 @@ const validResourceExtensionsResource = {
...
+    default:
+      throw new ValidationError(
+        "Invalid request method",
+        `"get" | "put" | "patch" | "delete" | "list"`,
+        requestMethod,
+      );
+  }
+}
+
+function getUri(resourceUri: string, extensionName: string, requestMethod: string): string {
```
XiaofeiCao, line 239: Is `extensionName` always `extension`? Do we need this parameter here?

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -151,17 +163,137 @@ const validResourceExtensionsResource = {
...
+  switch (requestMethod) {
+    case "list":
+      return `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`;
+    default:
+      return `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources/${extensionName}`;
+  }
+}
+
+function getUriForTenant(
```
XiaofeiCao, line 248: Why a separate method for tenant? 
Will `getUri("/", extensionName, requestMethod)` + `getUri("", extensionName, requestMethod)` work for tenant resource?

v-hongli1: Fixed by the [commit#22089c1](https://github.com/Azure/typespec-azure/pull/2203/commits/22089c1d3e99d0d62f04e5f3f7560e034608c143)

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -151,17 +163,121 @@ const validResourceExtensionsResource = {
...
+  }
+}
+
+function getUri(resourceUri: string, requestMethod: string): string {
+  switch (requestMethod) {
+    case "list":
+      if (resourceUri === "/" || resourceUri === "") {
+        return `${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`;
+      }
+      return `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`;
```
XiaofeiCao, line 245: Isn't this equivalent to just
```typescript
return `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`;
```
?

v-hongli1: This is not the case. When resourceUri is `/` and the URI format is `${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`, the returned URI is `//providers/Azure.ResourceManager.Resources/extensionsResources`, but if the URI format is `/${resourceUri}/providers/Azure.ResourceManager.Resources/extensionsResources`, the returned URI is `///providers/Azure.ResourceManager.Resources/extensionsResources`, which is not what the request requires.

XiaofeiCao: I see. Nit, better to use `"/" + TENANT_SCOPE_URI` and `TENANT_SCOPE_URI` here.

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -443,92 +782,160 @@ Scenarios.Azure_ResourceManager_Resources_ExtensionsResources_update = passOnSuc
...
       },
+    },
+    response: {
+      status: 204,
+    },
+    handler: (req: MockRequest) => requestHandler(req, "delete", TENANT_SCOPE_URI, null),
+    kind: "MockApiDefinition",
+  },
+  {
+    uri: getUri("", "delete"),
```
XiaofeiCao, line 868: Please replace `""` with `TENANT_SCOPE_URI`, also `"/"` with `TENANT_SCOPE_URI + "/"`. 
There are multiple places.

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -98,8 +110,8 @@ const validResourceGroupExtensionsResource = {
 };
 
 const validSubscriptionExtensionsResource = {
-  id: `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/providers/Azure.ResourceManager.Resources/extensionsResources/extension`,
-  name: "extension",
+  id: `${SUBSCRIPTION_SCOPE_URI}/providers/Azure.ResourceManager.Resources/extensionsResources/extension`,
+  name: `${EXTENSION_RESOURCE_NAME}`,
```
weidongxu-microsoft, line 114: Any difference from this?
```suggestion
  name: EXTENSION_RESOURCE_NAME,
```

v-hongli1: Fixed by the [commit#e90e90c](https://github.com/Azure/typespec-azure/pull/2203/commits/e90e90cf288973e618b4d5cd3029856bc0155cf5)

weidongxu-microsoft (approved):

PR #2216, opened 2025-02-13
[tcgc] refine flags

iscai-msft:
refine all tcgc flags and add docs.


antkmsft (approved):

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+
+Currently, emitters don't really use it, but it would be great to move language emitters to use this flag. Java should be excluded from this list because it doesn't apply to their naming structure.
+
+Should remove any logic that derives namespace names from the value of `package-name`.
+
+Default value is `none`.
+
+## 4. `flatten-union-as-enum`
+
+Whether you want TCGC to return unions of strings as enums. Default value is `true`. I feel this should not be a `tspconfig.yaml` flag, but an option you can pass to TCGC when calling `createSdkContext`
```
lmazuel, line 23: How do we configure it if it's not a flag?

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+
+When set to `true`, the emitter will generate low-level protocol methods for each service operation if `@protocolAPI` is not set for an operation. Default value is `true`.
+
+## 2. `generate-convenience-methods`
+
+When set to `true`, the emitter will generate low-level protocol methods for each service operation if `@convenientAPI` is not set for an operation. Default value is `true`.
+
+## 3. `package-name`
+
+The `package-name` override you would like for your generated library. By default individual language emitters should calculate the package name for their library from the `SdkPackage.namespaces` that `tcgc` returns to them.
```
lmazuel, line 13: Feels like we may want to kill it?

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+
+Whether you want TCGC to return unions of strings as enums. Default value is `true`. I feel this should not be a `tspconfig.yaml` flag, but an option you can pass to TCGC when calling `createSdkContext`
+
+## 5. `examples-dir`
+
+Specifies the directory where the emitter will look for example files. If the flag isn’t set, the emitter defaults to using an `examples` directory located at the project root.
+
+## 6. `emitter-name`
+
+The name of the emitter you are calling TCGC from. TCGC uses this to keep track of what scope you belong to, and what information to set for you. I feel this should not be a `tspconfig.yaml` flag, but an option you can pass to TCGC when calling `createSdkContext`
```
lmazuel, line 31: +1

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+
+## 1. `generate-protocol-methods`
+
+When set to `true`, the emitter will generate low-level protocol methods for each service operation if `@protocolAPI` is not set for an operation. Default value is `true`.
+
+## 2. `generate-convenience-methods`
+
+When set to `true`, the emitter will generate low-level protocol methods for each service operation if `@convenientAPI` is not set for an operation. Default value is `true`.
+
+## 3. `package-name`
```
weidongxu-microsoft, line 11: I wasn't aware we had a `package-name` in TCGC, when I asks for https://github.com/microsoft/typespec/issues/5758

Yeah, we can use this, if we agree there would be no need to setting a version.

Though unsure whether there would be another `package-description` though.

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+
+The `package-name` override you would like for your generated library. By default individual language emitters should calculate the package name for their library from the `SdkPackage.namespaces` that `tcgc` returns to them.
+
+Currently, emitters don't really use it, but it would be great to move language emitters to use this flag. Java should be excluded from this list because it doesn't apply to their naming structure.
+
+Should remove any logic that derives namespace names from the value of `package-name`.
+
+Default value is `none`.
```
weidongxu-microsoft, line 19: Eh, default `none`? Why not `undefined`?

weidongxu-microsoft (commented):
I think Java emitter uses all of them except `package-name`

`flatten-union-as-enum` and `emitter-name` would certainly affect TCGC behavior but I agree they seems not an emitter option.

packages/typespec-client-generator-core/design-docs/flags.md
msyyc: Python is fine to add `package-version/package-description` or not since they are not necessary configurations for a python package.
But I think maybe we could add these configs in case users want to set them when calling emitter.

nit: 
Python could consume `package-verison` at [here](https://github.com/microsoft/typespec/blob/ba241e8e5561e3b99d4a7b6195aa193c35cbae63/packages/http-client-python/generator/pygen/codegen/templates/version.py.jinja2#L4) and `package-description` at [here](https://github.com/microsoft/typespec/blob/ba241e8e5561e3b99d4a7b6195aa193c35cbae63/packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/setup.py.jinja2#L27).

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
+# TCGC `tspconfig.yaml` flags
```
MaryGao, line 1: I'd like to clarify the context of tcgc flags here, does that mean 1) tcgc would use these flags or 2) the common flags cross all emitters?

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+
```
MaryGao, line 2: missing `api-version`? is the `flavor` in-scoped?

weidongxu-microsoft: Yeah, `api-version`

`flavor` probably will go away from emitter, as we should have 1 emitter for unbranded, and 1 emitter for azure?

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,35 @@
...
+The `package-name` override you would like for your generated library. By default individual language emitters should calculate the package name for their library from the `SdkPackage.namespaces` that `tcgc` returns to them.
+
+Currently, emitters don't really use it, but it would be great to move language emitters to use this flag. Java should be excluded from this list because it doesn't apply to their naming structure.
+
+Should remove any logic that derives namespace names from the value of `package-name`.
+
+Default value is `none`.
+
+## 4. `flatten-union-as-enum`
```
MaryGao, line 21: another point is even we could expose this flag to customers that doesn't mean the customer's config would always take effective. it depends on how emitter handle them and is allowed to override or not.

packages/typespec-client-generator-core/design-docs/flags.md
MaryGao: For now package name, version and description would be useful for JS. And the package name is required but others are optional in Azure scope.

packages/typespec-client-generator-core/design-docs/flags.md
iscai-msft: we have been talking about coalescing to `package-name` across the js, csharp, and python libraries (java falls out of this). Would `package-name` live in tcgc or would each language emitter have their own flag called `package-name`? Now that I'm writing it out, I think it is the latter.

packages/typespec-client-generator-core/design-docs/flags.md
iscai-msft: Yeah, I messed up on the exact semantics, let me update the docs to `undefined`

packages/typespec-client-generator-core/design-docs/flags.md
iscai-msft: oh yep! I can't believe I missed `api-version`, thank you for the catch!

packages/typespec-client-generator-core/design-docs/flags.md
iscai-msft: it can be passed in as an option to `createSdkContext` which is called by the individual emitters. So emitters would look like `createSdkContext(..., {flattenUnionAsEnum: false})`

lmazuel: That doesnt' reply to my question :). We made it a flag so users can decide the behavior they want, how to user decide what behavior they want, if it's not through a flag?

packages/typespec-client-generator-core/design-docs/flags.md
```diff
@@ -0,0 +1,39 @@
...
+
+Whether you want TCGC to return unions of strings as enums. Default value is `true`. I feel this should not be a `tspconfig.yaml` flag, but an option you can pass to TCGC when calling `createSdkContext`
+
+## 5. `examples-dir`
+
+Specifies the directory where the emitter will look for example files. If the flag isn’t set, the emitter defaults to using an `examples` directory located at the project root.
+
+## 6. `emitter-name`
+
+The name of the emitter you are calling TCGC from. TCGC uses this to keep track of what scope you belong to, and what information to set for you. I feel we should remove this from `tspconfig.yaml`. It is currently the second optional argument to pass to `createSdkContext`, which I think is the only place you should be able to pass it in
```
lmazuel, line 31: Agree, shouldn't be a flag

packages/typespec-client-generator-core/design-docs/flags.md
tadelesh: it is not a flag for end user. it is an option for emitter. let me refine the doc.

packages/typespec-client-generator-core/design-docs/flags.md
tadelesh: the flag tcgc will use and get from emitter's config in `tspconfig.yaml`.

packages/typespec-client-generator-core/design-docs/flags.md
tadelesh: the `package-name` will be soon deprecated in this pr: https://github.com/Azure/typespec-azure/pull/2161

packages/typespec-client-generator-core/design-docs/flags.md
tadelesh: it should not configured by user. it is only used by emitter. i moved it from flag to options.

packages/typespec-client-generator-core/design-docs/flags.md
tadelesh: this flag has been removed.

tadelesh (approved):

Issue #2188, opened 2025-02-10
[Bug]: lots of same console output from TCGC


weidongxu-microsoft:
### Describe the bug

https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/decorators.ts#L911-L914

![Image](https://github.com/user-attachments/assets/2d281341-b1f5-4401-8d6b-fc87de4f2e6a)

### Reproduction

run some emitter on tsp with `@override`, e.g.
https://github.com/microsoft/typespec/blob/e43995f7defff3c9b062af6c769b63eb35744d2a/packages/http-client-java/generator/http-client-generator-test/tsp/method-override.tsp

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2217, opened 2025-02-14
[tcgc] remove useless typing and console log

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2188
fix: https://github.com/Azure/typespec-azure/issues/2218

### DIFF
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -909,11 +909,6 @@ function collectParams(
           params.push(value);
         } else if (!sourceProp.model) {
           params.push(value);
-        } else {
-          // eslint-disable-next-line no-console
-          console.log(
-            `We are not counting "${sourceProp.name}" as part of a method parameter because it's been added by Azure.Core templates`,
-          );
         }
       }
     }
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -60,8 +60,8 @@ import {
   AllScopes,
   clientNameKey,
   clientNamespaceKey,
+  findRootSourceProperty,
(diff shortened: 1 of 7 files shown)


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -909,11 +909,6 @@ function collectParams(
           params.push(value);
         } else if (!sourceProp.model) {
           params.push(value);
-        } else {
-          // eslint-disable-next-line no-console
-          console.log(
-            `We are not counting "${sourceProp.name}" as part of a method parameter because it's been added by Azure.Core templates`,
-          );
```
ArcturusZhang, line 916: do we need to do something or report something when we ignore it?

ArcturusZhang (approved):

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: i removed the core model filtering logic. i believe it will cause method param mapping issues. also, we should not specialize the logic for azure core model. cc: @iscai-msft

iscai-msft: we did this because we didn't want users to have to deal with the injected values from core templates. is there a way to check if a property was added by a library template instead of the user?

tadelesh: the filtering will cause `SdkServiceMethod` level missing params. thus i removed such logic. @catalinaperalta do you have real case here?

tadelesh: @catalinaperalta i merged this pr and will release a new tcgc version soon. if any unexpected behavior happens, feel free to ping me directly.

PR #2222, opened 2025-02-14
[tcgc] refine logic to iterate over all user-defined namespaces

iscai-msft:


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -591,3 +592,27 @@ export function hasNoneVisibility(context: TCGCContext, type: ModelProperty): bo
   const visibility = getVisibilityForClass(context.program, type, lifecycle);
   return visibility.size === 0;
 }
+
+export function getAllUserDefinedNamespaces(
+  context: TCGCContext,
+  namespace?: Namespace,
+  retval?: Namespace[],
+): Namespace[] {
+  if (!retval) {
```
joheredi, line 601: Is there any way we can use this typekit to filter non user defined namespaces?

https://github.com/microsoft/typespec/blob/main/packages%2Fcompiler%2Fsrc%2Fexperimental%2Ftypekit%2Fkits%2Ftype.ts#L127

iscai-msft: ooh that's a great idea, let me try that and see what happens

tadelesh:
could you please also update the code here: https://github.com/Azure/typespec-azure/blob/6bf27328400c186664518d85e1d43b4136d0aa75/packages/typespec-client-generator-core/src/public-utils.ts#L348. it is missing in the original pr.

iscai-msft:
> could you please also update the code here:
> 
> https://github.com/Azure/typespec-azure/blob/6bf27328400c186664518d85e1d43b4136d0aa75/packages/typespec-client-generator-core/src/public-utils.ts#L348
> 
> 
> it is missing in the original pr.

the reason for this pr is to remove the need to iterate over `listClients`. The root issue was I used to stop once I saw a namespace that wasn't user defined (which would stop at just `Azure.` and not continue recursing). With the new code, I keep recursing, so I will find the necessary clients

tadelesh:
> > could you please also update the code here:
> > https://github.com/Azure/typespec-azure/blob/6bf27328400c186664518d85e1d43b4136d0aa75/packages/typespec-client-generator-core/src/public-utils.ts#L348
> > 
> > it is missing in the original pr.
> 
> the reason for this pr is to remove the need to iterate over `listClients`. The root issue was I used to stop once I saw a namespace that wasn't user defined (which would stop at just `Azure.` and not continue recursing). With the new code, I keep recursing, so I will find the necessary clients

yes, i understand this pr's work. but my ask is actually from the original pr's comment: https://github.com/Azure/typespec-azure/pull/2179#discussion_r1945855708. there is something miss in that pr.

tadelesh:
and i'm think of if current impl could support versioning?

tadelesh (approved):

tadelesh:
@weidongxu-microsoft could you help to test this version to see if all orphan model is calculated?

PR #2226, opened 2025-02-18
[doc][multipart] remove old format for multipart

msyyc:
Old multipart has been deprecated so we shall remove related doc.


website/src/content/docs/docs/howtos/Generate client libraries/12multipart.mdx
```diff
@@ -176,7 +176,7 @@ def upload(
 ```
 
 ```csharp
-
+# TODO
```
msyyc, line 179: @ArcturusZhang If .NET doesn't support this feature, maybe you could add the linking issue as JS does.

weidongxu-microsoft (approved):

PR #2236, opened 2025-02-20
updates `@OpenAPI.extension` usage to use value kinds

chrisradek:
Depends on: https://github.com/microsoft/typespec/pull/6078


packages/samples/common-types/src/mobo.tsp
```diff
@@ -6,4 +6,4 @@ using OpenAPI;
 
 namespace Azure.ResourceManager.CommonTypes;
 
-@@extension(ManagedOnBehalfOfConfiguration.moboBrokerResources, "x-ms-identifiers", ["id"]);
+@@extension(ManagedOnBehalfOfConfiguration.moboBrokerResources, "x-ms-identifiers", #["id"]);
```
timotheeguerin, line 9:
```suggestion
@@identifiers(ManagedOnBehalfOfConfiguration.moboBrokerResources, #["id"]);
```

I think would be the correct new way

chrisradek: Talked offline - there's still some work to do to get `@identifiers` working everywhere. Using it here causes some 'duplicate' errors once ARM is imported.

timotheeguerin (approved):

timotheeguerin (approved):

timotheeguerin (approved):

PR #2237, opened 2025-02-20
Update the client generation document to provide valid typespec samples

lirenhe:
Update the below 2 documents to validate all existing typespec samples
1. Remove the pattern that we do not support.
2. Remove the TypeSpec code that would cause lint error.
3. Update the code to include model to make a valid sample
4. Update the related generated code.

### DIFF
--- a/website/src/content/docs/docs/howtos/Generate client libraries/06types.mdx
+++ b/website/src/content/docs/docs/howtos/Generate client libraries/06types.mdx
@@ -707,14 +707,6 @@ public final class Animal implements JsonSerializable<Animal> {
 model Animal {
   name: string;
   kind: string;
-  ...Record<string | int32>;
-}
-```
-
-```typespec
-model Animal {
-  name: string;
-  kind: string;
   ...Record<string>;
   ...Record<int32>;
 }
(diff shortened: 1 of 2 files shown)


website/src/content/docs/docs/howtos/Generate client libraries/06types.mdx
```diff
@@ -703,14 +703,6 @@ public final class Animal implements JsonSerializable<Animal> {
...
 <ClientTabItem lang="typespec" >
 
-```typespec
-model Animal {
-  name: string;
-  kind: string;
-  ...Record<string | int32>;
-}
-```
-
```
tadelesh, line 713: this is to demonstrate there are two ways to write additional properties with union types. do we need to delete it?

website/src/content/docs/docs/howtos/Generate client libraries/06types.mdx
```diff
@@ -1005,20 +997,6 @@ model Ragdoll extends Cat {
...
-  Ragdoll,
-}
-
-model Siamese {}
-
-model Ragdoll {}
-```
-
 </ClientTabItem>
 <ClientTabItem lang="tcgc">
```
tadelesh, line 1002: the next line could be removed.

tadelesh (approved):
i think we could also remove the tcgc representation in this doc since it is for spec author/sdk user, not for emitter dev.

lirenhe:
> i think we could also remove the tcgc representation in this doc since it is for spec author/sdk user, not for emitter dev.

Good point.  We need to continue updating https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/guideline/.   Merge this PR and let's update TCGC part in a separate effort.

PR #2250, opened 2025-02-23
Fix common-types specification differences with azure-rest-api-specs

AlitzelMendez:
Feeback provided in: https://github.com/Azure/azure-rest-api-specs/pull/32562


AlitzelMendez:
## Validatation
Currently under performance, still need to do some fixes 
- https://github.com/Azure/azure-rest-api-specs/actions/runs/13504230854/job/37729914282?pr=32781
- https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/13504484534/job/37730758657?pr=21514

packages/typespec-azure-resource-manager/lib/backcompat.tsp
```diff
@@ -6,8 +6,8 @@ namespace Azure.ResourceManager;
 alias InfrastructureEncryption = CommonTypes.InfrastructureEncryption;
 alias KeyEncryptionIdentity = CommonTypes.KeyEncryptionKeyIdentity;
 alias KeyEncryptionKeyIdentity = CommonTypes.KeyEncryptionKeyIdentity;
-alias CustomerManagedKeyEncryption = CommonTypes.CustomerManagedKeyEncryption;
-alias EncryptionConfiguration = CommonTypes.Encryption;
+alias CustomerManagedKeyEncryption = CommonTypes.customerManagedKeyEncryption;
```
markcowl, line 9: Just reiterating the comment above.  This is perpetuating a bug that does not impact the wire API

packages/samples/common-types/openapi/v5/customermanagedkeys.json
```diff
@@ -6,34 +6,6 @@
   },
   "paths": {},
   "definitions": {
-    "CustomerManagedKeyEncryption": {
```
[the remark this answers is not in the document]
AlitzelMendez: I reverted the change in `customermanagedkeys` and kept it only in places where it is necessary.

 It is necessary because there are places where some post-processing occurs: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/autorest.md#tag-package-common-type 

image

packages/samples/common-types/openapi/v4/types.json
```diff
@@ -506,7 +471,7 @@
           "readOnly": true
         },
         "systemData": {
-          "$ref": "#/definitions/SystemData",
+          "$ref": "#/definitions/systemData",
```
AlitzelMendez, line 474: verify if this change doesn't impact SDK team, Wei Hu

AlitzelMendez: in this case typespec is changed, is this ok? or do I need to create a special case in where this is changed on post processing and in tsp we keep upper case

live1206: If we are generating SDK from swagger, this will cause issue for SDK generation for all languages.
Because you changed the definition id here, and the other services referencing to this(with upper case) will not be able to find it. Modular 4 will not be able to compile the swagger, the SDK generation will fail in the first step.

AlitzelMendez: yes, we are keeping this to lower case (same as original swagger)  https://github.com/Azure/azure-rest-api-specs/blob/e256596e6701c96989581b97909a773f2520074f/specification/common-types/resource-management/v4/types.json#L381

my question is related to TypeSpec side, does it matter the casing of TypeSpec?

allenjzhang (approved):
nit: 
- remove `LocationResourceParameter`
- double check `x-ms-parameter-location` on `SubscriptionId`, `ApiVersion` with SDK

packages/samples/common-types/openapi/v4/types.json
live1206: For TypeSpec input, I think the first letter should be fine, all language emitters should have taken care of them already. 
We should not change the casing within the word as it would cause word separation changes.
I can double confirm with the team.

pshao25 (changes requested):
@AlitzelMendez @allenjzhang We talked before, the naming case issue will cause no services could update their specs.

AlitzelMendez:
> @AlitzelMendez @allenjzhang We talked before, the naming case issue will cause no services could update their specs.

we are fixing the naming case issue on this pull request, or do you still see a problem on this pull request?

packages/samples/common-types/openapi/v3/types.json
```diff
@@ -769,6 +775,7 @@
       "description": "The ID of the target subscription. The value must be an UUID.",
       "required": true,
       "type": "string",
+      "minLength": 1,
       "x-ms-parameter-location": "client"
```
AlitzelMendez, line 779: @live1206 we want to double check with you if adding this is ok (in relation with the original swagger) as this line `"x-ms-parameter-location": "client"` is not present in there: https://github.com/Azure/azure-rest-api-specs/blob/ad550d2f7adc2ca8f586080113eb220c3a6f3515/specification/common-types/resource-management/v3/types.json#L658

packages/samples/common-types/openapi/v3/types.json
```diff
@@ -721,6 +655,78 @@
...
+        },
+        "lastModifiedAt": {
+          "type": "string",
+          "format": "date-time",
+          "description": "The timestamp of resource last modification (UTC)"
+        }
+      }
     }
   },
   "parameters": {
```
AlitzelMendez, line 732: same in ApiVersionParameter, we are adding 
```json
"x-ms-parameter-location": "client",
"x-ms-client-name": "apiVersion"
```

which is not part of : https://github.com/Azure/azure-rest-api-specs/blob/ad550d2f7adc2ca8f586080113eb220c3a6f3515/specification/common-types/resource-management/v3/types.json#L622

is this change ok from sdk perspective?

packages/samples/common-types/openapi/v4/types.json
live1206: Confirmed with all language owners, the SDK generation will not be affected.
But we are wondering why we are changing the model names directly, can we keep the names upper case and use `@encdoeName` to change the generated swagger to lower case?

packages/samples/common-types/openapi/v3/types.json
live1206: I think default location is client.

packages/samples/common-types/openapi/v3/types.json
live1206: These are not needed, but should not change the SDK generation anyway.

packages/samples/common-types/openapi/v5/customermanagedkeys.json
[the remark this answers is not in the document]
markcowl: I would argue that this is a reason to update those queries when the change is made, rather than perpetuate a bad naming schema

Issue #2251, opened 2025-02-24
[Bug]: tcgc union isGeneratedName is problematic


qiaozha:
### Describe the bug

See https://github.com/Azure/autorest.typescript/pull/2753/files#r1965549152 

The union has a name in typespec definition, but tcgc still returns isGeneratedName as true.

### Reproduction

N/A

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2252, opened 2025-02-24
[tcgc] fix regression of 0.51.2

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2251
`isGeneratedName` always set to true only when it is the inner type of nullable union.
Remove parameter cache to retain HTTP metadata info for model property.

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -338,88 +338,84 @@ export function getSdkHttpParameter(
   location?: "path" | "query" | "header" | "body" | "cookie",
 ): [SdkHttpParameter, readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
-  let parameter = context.__httpParameterCache?.get(param);
-
-  if (!parameter) {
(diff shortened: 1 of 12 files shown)


weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/http.ts
```diff
@@ -338,88 +338,84 @@ export function getSdkHttpParameter(
...
-      parameter = {
-        ...base,
-        optional: param.optional,
-        collectionFormat: getCollectionFormat(context, param),
-        correspondingMethodParams,
-        kind: "header",
-        serializedName: getHeaderFieldName(program, param) ?? base.name,
-      };
-    }
-    context.__httpParameterCache.set(param, parameter);
```
qiaozha, line 420: just curious why not just remove this line? Feel like the other changes are not quite necessary?

qiaozha (approved):

packages/typespec-client-generator-core/src/http.ts
tadelesh: just rollback to previous version.

Issue #2053, opened 2025-01-08
TCGC should map the streaming response into binary type


lirenhe:
### Clear and concise description of the problem

The support for SSE/Streaming is still under design, to quickly workaround the spec that contains streaming response, TCGC should treat streaming with binary to enable each emitter to still generate the runnable code with this feature.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2256, opened 2025-02-24
[tcgc] map all stream request and response type to bytes

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/2053


weidongxu-microsoft (approved):
I assume emitter also need to add stream lib?

packages/typespec-client-generator-core/test/packages/streams.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      strictEqual(method.operation.responses.length, 1);
+      strictEqual(method.operation.responses[0].type?.kind, "bytes");
+      strictEqual(method.operation.responses[0].type?.encode, "bytes");
+    });
+
+    it("intersection stream response", async () => {
+      await runner.compileWithBuiltInService(
+        `
+        @route("/")
+        op get(): JsonlStream<Thing> & { @statusCode statusCode: 204; };
```
qiaozha, line 175: what would the tcgc behavior be if it's ssestream?

tadelesh: same behavior as json stream. let me add tests for sse.

iscai-msft (approved):

packages/typespec-client-generator-core/test/packages/streams.test.ts
tadelesh: done

qiaozha (approved):

PR #2257, opened 2025-02-24
update `@header` usage to value syntax

chrisradek:
Updates to address https://github.com/microsoft/typespec/issues/4114
Depends on: https://github.com/microsoft/typespec/pull/6130


iscai-msft (approved):
looks good, i don't think i would label this a breaking change in the changeset since we're still supporting the old way. thanks @chrisradek

timotheeguerin (approved):

Issue #2215, opened 2025-02-13
[tcgc] remove calls to deprecated projection functions


iscai-msft:
Swtich from `buildVersionProjections` -> `getVersioningMutators`, `projectProgram` -> `unsafe_mutateSubgraphWithNamespace`.

See [this](https://github.com/microsoft/typespec/pull/4931) pr for inspiration

PR #2276, opened 2025-02-27
[tcgc] try to migrate to mutator

tadelesh:
find two issues:
~1. when doing mutator, it seems the decorator data will be set again: https://github.com/microsoft/typespec/issues/6180~
2. for spread case, the model property will not be consistent: https://github.com/microsoft/typespec/issues/6181

resolve: https://github.com/Azure/typespec-azure/issues/2215


tadelesh:
find another problem: https://github.com/microsoft/typespec/issues/6198

iscai-msft:
closing in favor of https://github.com/Azure/typespec-azure/pull/2305

PR #2297, opened 2025-03-04
Fix using `@identifiers` on array of model with id not respected

### LINKED ISSUE #2296: `@identifiers` drops config if the target has an id property
```tsp
import "@azure-tools/typespec-azure-resource-manager";

using Azure.ResourceManager;

model Foo {
  @identifiers(#[])
  ids: Bar[];
}

model Bar {
  id: string;
}
```

[Playground Link](https://azure.github.io/typespec-azure/playground/?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtxhVyZXNvdXJjZS1tYW5hZ2VyIjsKCnVzaW5nIEHEIC5SxyBNxh87Cgptb2RlbCBGb28gewogIEBpZGVudGlmaWVycygjW10pCiAgaWRzOiBCYXJbXTsKfcgxQmFyxTFpZDogc3RyaW5nxB0%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%5D%7D%7D)

in this case the user explicitly said there was no key but the precense of `id` take over and make it drop the property

timotheeguerin:
fix  #2296

### DIFF
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -1855,13 +1855,17 @@ export async function getOpenAPIForService(
     }
   }
 
-  function ifArrayItemContainsIdentifier(program: Program, array: ArrayModelType) {
+  function ifArrayItemContainsIdentifier(
+    program: Program,
+    array: ArrayModelType,
+    armIdentifiers: string[],
+  ) {
     if (array.indexer.value?.kind !== "Model") {
       return true;
     }
     return (
       getExtensions(program, array).has("x-ms-identifiers") ||
-      getProperty(array.indexer.value, "id")
+      (getProperty(array.indexer.value, "id") && armIdentifiers.includes("id"))
     );
   }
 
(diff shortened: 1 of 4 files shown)


AlitzelMendez (approved):

packages/typespec-autorest/src/openapi.ts
timotheeguerin, line 1: Seems to have more issues where it gets injected when it shouldn't 

https://azure.github.io/typespec-azure/playground/?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtxhVyZXNvdXJjZS1tYW5hZ2VyIjsKCnVzaW5nIEHEIC5SxyBNxh87Cgptb2RlbCBGb28gewogIC8vQGlkZW50aWZpZXJzKCNbXSkKICBpZHM6IEJhcltdOwp9yDNCYXLFM0BrZXkgb3RoZXI6IHN0cmluZ8Ql&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%5D%7D%7D

timotheeguerin: make sense, i'll file a follow up issue then to decide if we want to fix that or not in the future

markcowl (approved):

Issue #2294, opened 2025-03-04
[tcgc] add paging response info into `pagingMetadata`


tadelesh:
we do that for lro. to be consistency, we should also put response info into paging metadata.

Issue #2292, opened 2025-03-04
SerializedName should not be set to `body`


JoshLove-msft:
For root body parameters, why is the `serializedName` property set to `body`? I think this is intended to convey that the parameter is the body of the request, but this is confusing as a non-root parameter could also have a serialized name of "body". I would have expected this to be empty instead. Ctrl + F for "serializedName: body"

This leads to emitter code like https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/emitter/src/lib/operation-converter.ts#L176

[...]

Issue #2291, opened 2025-03-03
Expose the ItemPropertyName in SdkPagingServiceMetadata


JoshLove-msft:
This is on the PagedResultMetadata, which _might_ be accessible via `__raw`, but it should be a top level property. __raw is typed as `PagedResultMetadata | PagingOperation` so it isn't clear if this will always be accessible.

JoshLove-msft:
After testing, __raw is actually PagingOperation, so I needed to do this:
```ts
ItemsPropertyName: (method.pagingMetadata.__raw as PagingOperation).output!.pageItems.property
```

JoshLove-msft:
Related https://github.com/Azure/typespec-azure/issues/2294

PR #2298, opened 2025-03-04
[tcgc] improvement for paging and body parameter

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/2294
resolve: https://github.com/Azure/typespec-azure/issues/2292
resolve: https://github.com/Azure/typespec-azure/issues/2291

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -185,7 +185,7 @@ function getSdkHttpParameters(
         kind: "body",
         name,
         isGeneratedName: true,
-        serializedName: "body",
+        serializedName: "",
         doc: getDoc(context.program, tspBody.type),
         summary: getSummary(context.program, tspBody.type),
         onClient: false,
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -814,6 +814,8 @@ export interface SdkPagingServiceMetadata<TServiceOperation extends SdkServiceOp
   continuationTokenParameterSegments?: SdkModelPropertyType[];
   /** Segments to indicate how to get continuation token value from response. */
   continuationTokenResponseSegments?: SdkModelPropertyType[];
+  /** Segments to indicate how to get page items from response. */
+  pageItemsSegments?: SdkModelPropertyType[];
 }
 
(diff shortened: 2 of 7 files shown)


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -185,7 +185,7 @@ function getSdkHttpParameters(
         kind: "body",
         name,
         isGeneratedName: true,
-        serializedName: "body",
```
ArcturusZhang, line 188: well if we would like to keep this property, we should fallback to its body

iscai-msft:
@tadelesh I'm going to try to rebase this onto `main`, since they are about to do their release

packages/typespec-client-generator-core/src/http.ts
weidongxu-microsoft: Should we just remove the `serializedName` property at all?

tadelesh: we could not do that. for explicitly defined body, it might have different client name. e.g., `op test(@body request: string): void`. in previous swagger example, we will set the request example with `{"request": "test"}`.

tadelesh: yes.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -814,6 +814,8 @@ export interface SdkPagingServiceMetadata<TServiceOperation extends SdkServiceOp
   continuationTokenParameterSegments?: SdkModelPropertyType[];
   /** Segments to indicate how to get continuation token value from response. */
   continuationTokenResponseSegments?: SdkModelPropertyType[];
+  /** Segments to indicate how to get page items from response. */
+  pageItemsSegments?: SdkModelPropertyType[];
```
weidongxu-microsoft, line 818: Should it be `required`?

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: for azure, it might be undefined. sad. https://github.com/Azure/typespec-azure/blob/405af28b4b9b3c16aa7de4bcbbb704bd12408249/packages/typespec-azure-core/src/decorators.ts#L101

Issue #2285, opened 2025-03-03
[azure-http-specs] ARM case for Exception


XiaofeiCao:
1. Normal case, ErrorResponse https://github.com/Azure/typespec-azure/blob/9355615e7ba500932baa523fc1e7defadac539be/packages/typespec-azure-resource-manager/lib/operations.tsp#L27
2. Custom error. 

server error response content, e.g. 
> Status code 404, "{"error":{"code":"ResourceNotFound","message":"The Resource 'Microsoft.Compute/virtualMachines/myvm' under resource group 'javacsmrg48243' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix"}}"

PR #2300, opened 2025-03-04
azure-http-specs, arm test case for Error

XiaofeiCao:
- fix https://github.com/Azure/typespec-azure/issues/2285

### DIFF
--- a/packages/azure-http-specs/spec-summary.md
+++ b/packages/azure-http-specs/spec-summary.md
@@ -800,6 +800,60 @@ maxpagesize=3
 }
 ```
 
+### Azure_ResourceManager_CommonProperties_Error_createOrReplace
+
+- Endpoint: `put https://management.azure.com`
+
+Resource PUT operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.CommonProperties/confidentialResources/confidential",
+Expected query parameter: api-version=2023-12-01-preview
+Expected request body:
+
+```json
+{
+  "location": <any string>,
+  "properties": {
+    "username": "myusername",
+    "password": "mypassword"
+  }
+}
+```
+
+Expected response status code: 403
+Expected response body:
+
+```json
+{
+  "error": {
+    "code": "AuthorizationFailed",
+    "message": "Your account or service principal doesn't have sufficient access to complete the operation.",
+    "innerError": {
+      "exceptiontype": "general"
+    }
+  }
+}
+```
+
+### Azure_ResourceManager_CommonProperties_Error_get
+
+- Endpoint: `get https://management.azure.com`
+
+Resource GET operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.CommonProperties/confidentialResources/confidential",
+Expected query parameter: api-version=2023-12-01-preview
+
+Expected response status code: 404
+Expected response body:
+
+```json
+{
+  "error": {
+    "code": "ResourceNotFound",
+    "message": "The Resource 'Azure.ResourceManager.CommonProperties/confidentialResources/confidential' under resource group 'test-rg' was not found."
+  }
+}
+```
+
 ### Azure_ResourceManager_CommonProperties_ManagedIdentity_createWithSystemAssigned
 
 - Endpoint: `put https://management.azure.com`
--- a/packages/azure-http-specs/specs/azure/resource-manager/common-properties/error.tsp
+++ b/packages/azure-http-specs/specs/azure/resource-manager/common-properties/error.tsp
@@ -1,0 +1,161 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/versioning";
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-azure-resource-manager";
+import "@typespec/spector";
+
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using TypeSpec.Versioning;
+using Azure.Core;
+using Azure.ResourceManager;
+using TypeSpec.OpenAPI;
+using Spector;
+
+namespace Azure.ResourceManager.CommonProperties;
+
+@resource("confidentialResources")
+model ConfidentialResource is TrackedResource<ConfidentialResourceProperties> {
+  ...ResourceNameParameter<ConfidentialResource>;
+}
+
+@doc("Confidential Resource Properties.")
+model ConfidentialResourceProperties {
+  @visibility(Lifecycle.Read)
+  @doc("The status of the last operation.")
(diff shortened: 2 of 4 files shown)


packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
```diff
@@ -126,3 +126,60 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_updateWithUserA
...
+  uri: "/subscriptions/:subscriptionId/resourceGroups/:resourceGroup/providers/Azure.ResourceManager.CommonProperties/confidentialResources/:resourceName",
+  method: "get",
+  request: {
+    params: {
+      subscriptionId: SUBSCRIPTION_ID_EXPECTED,
+      resourceGroup: RESOURCE_GROUP_EXPECTED,
+      resourceName: "confidential",
+      "api-version": "2023-12-01-preview",
+    },
+    status: 404,
```
anuchandy, line 140: Is it intentional that we define a field `request.status` with an expected response status code?

anuchandy (changes requested):

packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
anuchandy, line 140: Is it intentional that we define a field `request.status` with an expected response status code?

packages/azure-http-specs/spec-summary.md
```diff
@@ -800,6 +800,60 @@ maxpagesize=3
...
+Expected response body:
+
+```json
+{
+  "error": {
+    "code": "AuthorizationFailed",
+    "message": "Your account or service principal doesn't have sufficient access to complete the operation.",
+    "innerError": {
+      "exceptiontype": "general"
+    }
```
weidongxu-microsoft, line 832: this part seems not ARM?
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v6/types.json#L83-L135

packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
XiaofeiCao: Yes. It's for matching the 404 `response.status`, to pass Spector e2e test.
Spector code: https://github.com/microsoft/typespec/blob/3f7204c0a4e05aa04126ac736f640b2c0047b3b4/packages/spector/src/actions/server-test.ts#L24-L29

Existing usage: https://github.com/microsoft/typespec/blob/657d4a0f1c86acb2a0d31b51aa1d128d2a4d6571/packages/http-specs/specs/authentication/http/custom/mockapi.ts#L26

XiaofeiCao: @sarangan12 May need your opinion here. I thought `response.status` should be enough, to match `passOnCode(status)`. Why is `request.status` still in need?

packages/azure-http-specs/spec-summary.md
XiaofeiCao: No, `CloudError` is to test custom defined error. 
e.g. Compute: https://github.com/Azure/azure-rest-api-specs/pull/32748/files#diff-bd60c6106ada2ea78b87a4cf98040f192149df43b988f22ec3c34c7a50729bf6R29

Normal ARM exception is tested in `GET` operation.

XiaofeiCao: Sure. 400 should be enough.

packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
anuchandy: Thanks for clarifying

weidongxu-microsoft (approved):
update summary

packages/azure-http-specs/specs/azure/resource-manager/common-properties/error.tsp
```diff
@@ -0,0 +1,159 @@
...
+    ```json
+    {
+      "error": {
+        "code": "ResourceNotFound",
+        "message": "The Resource 'Azure.ResourceManager.CommonProperties/confidentialResources/confidential' under resource group 'test-rg' was not found."
+      }
+    }
+    ```
+    """)
+  get is ArmResourceRead<ConfidentialResource>;
```
MaryGao, line 127: this get can't set error model?

MaryGao (approved):
LGTM and left confirmation comments.

packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
```diff
@@ -126,3 +126,58 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_updateWithUserA
...
   });
+
+Scenarios.Azure_ResourceManager_CommonProperties_Error_get = passOnCode(404, {
+  uri: "/subscriptions/:subscriptionId/resourceGroups/:resourceGroup/providers/Azure.ResourceManager.CommonProperties/confidentialResources/:resourceName",
+  method: "get",
+  request: {
+    params: {
+      subscriptionId: SUBSCRIPTION_ID_EXPECTED,
+      resourceGroup: RESOURCE_GROUP_EXPECTED,
+      resourceName: "confidential",
```
MaryGao, line 137: what does the name mean? do we leverage name to differentiate the cases?

XiaofeiCao: > do we leverage name to differentiate the cases?

Yes. We don't want resources across scenarios be of the same name, e.g. `Employee`. It'll be a pain importing them when writing tests.
Here I named it originally for the 403 case. Now that the 403 is replaced with 400 case, it does make little sense here. Open to other names, maybe just `ErrorResource`?

packages/azure-http-specs/specs/azure/resource-manager/common-properties/error.tsp
XiaofeiCao: Yes, it can. Here it's for testing predefined `Error` model, in contrast with user-defined `CloudError`.

packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
tadelesh: since the mock always return build-in error. it's better to rename the op.

packages/azure-http-specs/specs/azure/resource-manager/common-properties/error.tsp
tadelesh: since the mock always return build-in error. it's better to rename the op.

tadelesh (approved):

PR #2305, opened 2025-03-04
[tcgc] move to mutators

iscai-msft:
fixes https://github.com/Azure/typespec-azure/issues/2215


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -595,3 +607,80 @@ export function findRootSourceProperty(property: ModelProperty): ModelProperty {
...
+ *
+ * See https://github.com/microsoft/typespec/issues/6247
+ *
+ * @param context
+ */
+export function listAllServiceNamespaces(context: TCGCContext): Namespace[] {
+  const serviceNamespaces: Namespace[] = [];
+  let globalNamespace = context.globalNamespace;
+  if (!globalNamespace) {
+    globalNamespace = handleVersioningMutationForGlobalNamespace(context);
```
anuchandy, line 678: Should we store the mutated global namespace to `context` or it happens somewhere else? just reading the doc of `TCGCContext.globalNamespace`

```ts
globalNamespace?: Namespace; // the root of all tsp namespaces for this instance. Starting point for traversal, so we don't call mutation multiple times
```

anuchandy: in case TCGC is supposed to store it in the context, I wonder why we not store the resolved `serviceNamespaces` array in `context` instead of `globalNamespace`? Computing `serviceNamespaces ` seems to involve some work, but I'm unsure if it always computes the same result.

.chronus/changes/tcgc-migrateToMutator-2025-2-4-12-45-51.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
```
timotheeguerin, line 2:
```suggestion
changeKind: feature
```

probably more accurate

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -595,3 +607,80 @@ export function findRootSourceProperty(property: ModelProperty): ModelProperty {
...
+  if (!retval) {
+    retval = [];
+  }
+  if (retval.includes(namespace)) return retval;
+  if (getLocationContext(context.program, namespace).type === "project") {
+    retval.push(namespace);
+  }
+  for (const ns of namespace.namespaces.values()) {
+    listAllUserDefinedNamespaces(context, ns, retval);
+  }
```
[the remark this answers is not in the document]
iscai-msft: yes, one common example is something like `Azure.My.Service`. `Azure` can show up as a pre-defined namespace (depending on import order). however, we keep recursing until we reach the end still. So we don't immediately return when we see a pre-defined namespace, we just don't add them to the returned list

packages/typespec-client-generator-core/src/internal-utils.ts
iscai-msft: yep `.globalNamespace` is the mutated global namespace. maybe I can change the name

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -619,3 +631,59 @@ export function findRootSourceProperty(property: ModelProperty): ModelProperty {
...
+}
+
+export function handleVersioningMutationForGlobalNamespace(context: TCGCContext): Namespace {
+  const globalNamespace = context.program.getGlobalNamespaceType();
+  const service = listServices(context.program)[0];
+  if (!service) return globalNamespace;
+  const allApiVersions = getVersions(context.program, service.type)[1]
+    ?.getVersions()
+    .map((x) => x.value);
+  if (!allApiVersions) return globalNamespace;
```
tadelesh, line 661: can we just return the `globalNamespace` for `all` version? the type graph without mutation seems should represent `all` just as previous we do with projection.

packages/typespec-client-generator-core/test/types/body-model-property-types.test.ts
```diff
@@ -11,17 +11,6 @@ describe("typespec-client-generator-core: body model property types", () => {
...
-    for (const modelsOrEnums of [
-      runner.context.sdkPackage.models,
-      runner.context.sdkPackage.enums,
-    ]) {
-      for (const item of modelsOrEnums) {
-        ok(item.name !== "");
-      }
-    }
-  });
-
```
tadelesh, line 24: why we remove such post tests? it is used to check if tcgc could provide all types with name.

packages/typespec-client-generator-core/test/types/enum-types.test.ts
```diff
@@ -12,16 +12,6 @@ describe("typespec-client-generator-core: enum types", () => {
...
-  afterEach(async () => {
-    for (const modelsOrEnums of [
-      runner.context.sdkPackage.models,
-      runner.context.sdkPackage.enums,
-    ]) {
-      for (const item of modelsOrEnums) {
-        ok(item.name !== "");
-      }
-    }
-  });
```
tadelesh, line 24: same concern.

tadelesh (commented):
i'm wondering if current impl could cover the real service case: https://github.com/Azure/azure-rest-api-specs/blob/13d840307175afdc951f74ffded36d5cdda2da91/specification/ai/Face/client.tsp#L59. face use another namespace to do customization. but customization namespace could only depend on one version. i'm wondering the type under that namespace could successfully mutate to old version (`v1_1_preview_1`).
we don't have such tests in tcgc,  i believe we should add one.

packages/typespec-client-generator-core/test/types/body-model-property-types.test.ts
iscai-msft: i removed them bc they do make debugging a lot messier. I'll add them back bc I don't feel too strongly

packages/typespec-client-generator-core/src/internal-utils.ts
timotheeguerin: oh I guess if you mean that all is the unmutated code then I think its fine too, no? what I think will be a nightmare right now at least is mutating every single version and try to merge that

iscai-msft: ooh ok, yeah let me give it a try then

PR #2306, opened 2025-03-04
Openapi extension emit schema

chrisradek:
Depends on https://github.com/microsoft/typespec/pull/6229
Fixes https://github.com/microsoft/typespec/issues/6077

This PR updates `@azure-tools/typespec-autorest` to support emitting schemas for Types passed into the `@OpenAPI.extension` decorator.

### DIFF
--- a/.chronus/changes/openapi-extension-emit-schema-2025-2-4-10-35-45.md
+++ b/.chronus/changes/openapi-extension-emit-schema-2025-2-4-10-35-45.md
@@ -1,0 +1,24 @@
+---
+changeKind: breaking
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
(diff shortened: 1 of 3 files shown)


.chronus/changes/openapi-extension-emit-schema-2025-2-4-10-35-45.md
timotheeguerin, line 1: I am a tiny bit worried about this change for the value it brings. Wondering if we might be better to just cut of the type conversion for 1.0 so at least using a type is exposed as broken with teh codefix to the user and maybe we add the feature parity in a few month.

Here I am worried that a lot of 3p customers might not pulled the last version and this would be a bigger break(quite confusing I assume as well) with no codefix

PR #2307, opened 2025-03-04
[typespec-autorest] Fix logic around `canShare` and `@invisible`

witemple-msft:
This PR fixes a bug with the handling of `@invisible(Lifecycle)` in typespec-autorest. Currently, it considers a property shareable if it either has default visibility (all modifiers), or if it does not contain any modifiers that cannot be expressed through x-ms-mutability. This was almost correct, except for the case where the property is _completely invisible_, in which case the property cannot be shared. This behavior was unique to typespec-autorest due to its logic around handling x-ms-mutability and the openapi3 emitter was not affected because it only attempts to share read-only properties.

In practice, `@invisible(Lifecycle)` is extremely rarely used because it means that the property would not be present in any operation payloads.

This PR adds a test against regressions ensuring that `@invisible(Lifecycle)` correctly hides properties.

### DIFF
--- a/.chronus/changes/witemple-msft-autorest-invisible-fix-2025-2-4-16-21-51.md
+++ b/.chronus/changes/witemple-msft-autorest-invisible-fix-2025-2-4-16-21-51.md
@@ -4,4 +4,4 @@ packages:
   - "@azure-tools/typespec-autorest"
 ---
 
-Fixed a bug causing properties marked `@invisble(Lifecycle)` to remain included in payloads."
+Fixed a bug causing properties marked `@invisible(Lifecycle)` to remain included in payloads."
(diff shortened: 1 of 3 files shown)


timotheeguerin (approved):
Do we need a hotfix for this?

.chronus/changes/witemple-msft-autorest-invisible-fix-2025-2-4-16-21-51.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+Fixed a bug causing properties marked `@invisble(Lifecycle)` to remain included in payloads."
```
witemple-msft, line 7:
```suggestion
Fixed a bug causing properties marked `@invisible(Lifecycle)` to remain included in payloads."
```

witemple-msft:
@timotheeguerin `@invisble(Lifecycle)` is so little-used that I don't know if it matters very much, but we could. Across several thousand tests here and in core as well as hundreds of REST API specs we only found one thing that tipped me off to this behavior. @allenjzhang what do you think?

PR #2316, opened 2025-03-06
Remove c-sharp service emitter

markcowl:


core
timotheeguerin, line 1: Probably not expected change?

timotheeguerin (commented):
might also be reference in the top level tsconfig.json

pnpm-lock.yaml
timotheeguerin, line 1: might be an issue with the core submodule here too

core
markcowl: Oh, yeah, my bad

markcowl:
> might also be reference in the top level tsconfig.json

Looks like it isn't

timotheeguerin (approved):

PR #2318, opened 2025-03-06
azure-http-specs, validate orphan model serializable

weidongxu-microsoft:
As discussed, we want to verify that an orphan model is by default serializable as JSON.

Since there were an existing test on orphan model, I just enhanced it and added an operation to verify the serialization.

### DIFF
--- a/packages/azure-http-specs/specs/azure/client-generator-core/usage/main.tsp
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/usage/main.tsp
@@ -13,8 +13,9 @@ namespace _Specs_.Azure.ClientGenerator.Core.Usage;
 
 @scenario
 @scenarioDoc("""
-  This scenario contains two public operations. Both should be generated and exported.
-  The models are override to roundtrip, so they should be generated and exported as well.
+  This scenario contains 4 public operations. All should be generated and exported.
+  'OrphanModel' is not used but specified as 'public' and 'input', so it should be generated in SDK. The 'orphanModelSerializable' operation verifies that the model can be serialized to JSON.
+  The other models are override to roundtrip, so they should be generated and exported as well.
   """)
 @global.Azure.ClientGenerator.Core.clientNamespace("azure.clientgenerator.core.usage", "java")
 namespace ModelInOperation {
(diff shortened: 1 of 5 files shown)


packages/azure-http-specs/specs/azure/client-generator-core/usage/main.tsp
```diff
@@ -90,11 +91,31 @@ namespace ModelInOperation {
...
+    {
+      "name": "name",
+      "desc": "desc"
+    }
+    ```
+    """)
+  @global.Azure.ClientGenerator.Core.convenientAPI(false)
+  @route("/orphanModelSerializable")
+  @put
+  op orphanModelSerializable(@body body: unknown): NoContentResponse;
```
weidongxu-microsoft, line 109: I used the `unknown` to avoid ref the `OrphanModel`.

Let me know any language that having difficulty to serialize a `OrphanModel` model into this op.

packages/azure-http-specs/specs/azure/client-generator-core/usage/main.tsp
```diff
@@ -90,11 +91,31 @@ namespace ModelInOperation {
...
 @doc("Not used anywhere, but access is override to public so still need to be generated and exported with serialization.")
 @global.Azure.ClientGenerator.Core.usage(global.Azure.ClientGenerator.Core.Usage.input)
 @global.Azure.ClientGenerator.Core.access(global.Azure.ClientGenerator.Core.Access.public)
 model OrphanModel {
+  @global.Azure.ClientGenerator.Core.clientName("modelName")
   name: string;
+
+  @encodedName("application/json", "desc")
+  description: string;
 }
```
weidongxu-microsoft, line 121: I've updated the model a little, to mix typespec name, client name, encoded name. And we expect SDK find the correct name for serialization.

packages/azure-http-specs/specs/azure/client-generator-core/usage/main.tsp
```diff
@@ -90,11 +91,31 @@ namespace ModelInOperation {
...
+  @route("/orphanModelSerializable")
+  @put
+  op orphanModelSerializable(@body body: unknown): NoContentResponse;
 }
 
 @doc("Not used anywhere, but access is override to public so still need to be generated and exported with serialization.")
 @global.Azure.ClientGenerator.Core.usage(global.Azure.ClientGenerator.Core.Usage.input)
 @global.Azure.ClientGenerator.Core.access(global.Azure.ClientGenerator.Core.Access.public)
 model OrphanModel {
+  @global.Azure.ClientGenerator.Core.clientName("modelName")
```
tadelesh, line 116: i think we do need to have a way to indicate how to serialize an orphan model. it seems weird that i have an orphan model that one property has a client name.

weidongxu-microsoft: `@clientName` appears expected for me. This case maybe a bit uncommon, but I assume it be pretty common that one language want one name, while other language want another.

And the eventgrid lib does exact this with `@clientName`
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/propertyNameOverrideCsharp.tsp
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/propertyNameOverrideJava.tsp

But yes, we may need some way to say default to JSON, either emitter option, option to TCGC, or option to `@service`

XiaofeiCao (approved):

tadelesh (approved):

PR #2326, opened 2025-03-06
[tcgc] add namespace flag

iscai-msft:


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1135,7 +1136,18 @@ export function getClientNamespace(
   entity: Namespace | Interface | Model | Enum | Union,
 ): string {
   const override = getScopedDecoratorData(context, clientNamespaceKey, entity);
-  if (override) return override;
+  if (override) {
```
srnagar, line 1139: It would be good to add a comment here (or on the method doc) to document the intent with an example.

srnagar (approved):
LGTM

packages/typespec-client-generator-core/src/context.ts
```diff
@@ -103,6 +103,7 @@ export async function createSdkContext<
     decoratorsAllowList: [...defaultDecoratorsAllowList, ...(options?.additionalDecorators ?? [])],
     previewStringRegex: options?.versioning?.previewStringRegex || tcgcContext.previewStringRegex,
     disableUsageAccessPropagationToBase: options?.disableUsageAccessPropagationToBase ?? false,
+    namespaceFlag: context.options["namespace"],
```
msyyc, line 106: Not sure other languages, while Python doesn't need this flag.

msyyc (commented):
Overall looks good to me.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1130,12 +1131,36 @@ export const $clientNamespace: ClientNamespaceDecorator = (
   setScopedDecoratorData(context, $clientNamespace, clientNamespaceKey, entity, value, scope);
 };
 
+/**
+ * Returns the client namespace for a given entity. The order of operations is as follows:
+ *
+ * 1. if `@clientNamespace` is applied to the entity, this wins out.
+ *    a. If the `--namespace` flag is passed in during generation, we will replace the root of the client namespace with the flag.
```
weidongxu-microsoft, line 1138: For (1), I assume it be "`@clientNamespace` is applied to the entity, or applied to the Namespace contains the entity"

I am not sure about what (1a) means. If `@clientNamespace` is provided to the entity or its Namespace, the `--namespace` options seems just won't have effect?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1130,12 +1131,36 @@ export const $clientNamespace: ClientNamespaceDecorator = (
...
+ */
 export function getClientNamespace(
   context: TCGCContext,
   entity: Namespace | Interface | Model | Enum | Union,
 ): string {
   const override = getScopedDecoratorData(context, clientNamespaceKey, entity);
-  if (override) return override;
+  if (override) {
+    // if @clientNamespace is applied to the entity, this wins out
+    const userDefinedNamespace = listAllUserDefinedNamespaces(context).find((x) =>
```
weidongxu-microsoft, line 1152: Feel some difficulty to understand what account as UserDefinedNamespace.

Is `Service.Models` a UserDefinedNamespace?
`TypeSpce.Rest.Resources`?
`Azure.Core`?
`LiftrBase` in https://www.npmjs.com/package/@azure-tools/typespec-liftr-base ?

tadelesh: `isUserDefined` typekit seems judge if a type is defined in project scope, which means `LiftrBase`'s ns is not user defined.

tadelesh, line 1152: can this find the shortest user defined namespace? i believe the replacement should for the shortest user defined one?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1146,6 +1171,9 @@ export function getClientNamespace(
 }
 
 function getNamespaceFullNameWithOverride(context: TCGCContext, namespace: Namespace): string {
+  if (context.namespaceFlag) {
+    return context.namespaceFlag;
+  }
```
tadelesh, line 1176: duplicate with line 1161

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -1150,8 +1150,8 @@ export function getSdkPackage<TServiceOperation extends SdkServiceOperation>(
   const crossLanguagePackageId = diagnostics.pipe(getCrossLanguagePackageId(context));
   const allReferencedTypes = getAllReferencedTypes(context);
   const sdkPackage: SdkPackage<TServiceOperation> = {
-    name: getClientNamespaceString(context)!,
-    rootNamespace: getClientNamespaceString(context)!,
+    name: getClientNamespaceString(context)!, // eslint-disable-line @typescript-eslint/no-deprecated
```
tadelesh, line 1153: since we do not deprecate this name, we need to change to a non-deprecate way.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1130,12 +1131,36 @@ export const $clientNamespace: ClientNamespaceDecorator = (
...
+ * @returns
+ */
 export function getClientNamespace(
   context: TCGCContext,
   entity: Namespace | Interface | Model | Enum | Union,
 ): string {
   const override = getScopedDecoratorData(context, clientNamespaceKey, entity);
-  if (override) return override;
+  if (override) {
+    // if @clientNamespace is applied to the entity, this wins out
```
tadelesh, line 1151: nit:
```suggestion
    // if `@clientNamespace` is applied to the entity, this wins out
```

packages/typespec-client-generator-core/src/decorators.ts
weidongxu-microsoft: Got it. So far the logic should be fine if `LiftrBase` is not UserDefinedNamespace. We want `namespace` option to override that anyway.

packages/typespec-client-generator-core/src/context.ts
iscai-msft: this is the flag we're adding to tcgc as an override to the `namespace is namespace` mantra we have

packages/typespec-client-generator-core/src/decorators.ts
srnagar: Yes, `@clientNamespace` applies to whatever entity or parent it is applied to and will take precedence over `--namespace` flag in tspconfig.

PR #2327, opened 2025-03-06
Update samples with expected date-time format for http header

AlitzelMendez:
follow up of pr: https://github.com/microsoft/typespec/pull/6289


packages/samples/test/output/azure/core/data-plane/api-path-parameter/@typespec/openapi3/openapi.2022-08-31.yaml
```diff
@@ -679,7 +679,7 @@ components:
       description: The request should only proceed if the entity was modified after this time.
       schema:
         type: string
-        format: date-time
+        format: http-date
     Azure.Core.ConditionalRequestHeaders.ifNoneMatch:
```
AlitzelMendez, line 683: :)

PR #2346, opened 2025-03-10
Pending Azure Lib update for Http library cleanup

allenjzhang:


packages/samples/specs/misc/appconfig/api.json
```diff
@@ -185,8 +185,7 @@
                 "locked",
                 "etag"
               ]
-            },
-            "collectionFormat": "csv"
+            }
```
timotheeguerin, line 188: for autorest we might want to keep how it was generating before

allenjzhang: To minimize spec impact?

packages/samples/test/output/azure/core/data-plane/api-path-parameter/@azure-tools/typespec-autorest/2022-08-31/openapi.json
```diff
@@ -1353,7 +1353,6 @@
       "items": {
         "type": "string"
       },
-      "collectionFormat": "multi",
```
timotheeguerin, line 1356: seems like this one is unexpected at least

PR #2349, opened 2025-03-11
Update Azure Libs with latest HTTP Lib

allenjzhang:


packages/typespec-azure-core/src/rules/prefer-csv-collection-format.ts
timotheeguerin, line 1: you can remove this rule there is already a [no-query-explode](https://azure.github.io/typespec-azure/docs/libraries/azure-core/rules/no-query-explode/)

packages/typespec-autorest/src/openapi.ts
```diff
@@ -1482,12 +1470,8 @@ export async function getOpenAPIForService(
...
-    let collectionFormat = headerOptions.format;
-    if (
-      !collectionFormat &&
-      (typeof headerOptions.explode === "boolean" || $.array.is(prop.type))
-    ) {
+    let collectionFormat;
+    if (typeof headerOptions.explode === "boolean" || $.array.is(prop.type)) {
       collectionFormat = headerOptions.explode ? "multi" : "csv";
     }
     if (collectionFormat && !["csv", "ssv", "tsv", "pipes"].includes(collectionFormat)) {
```
timotheeguerin, line 1477: I think this technically cann't happen now? should the error also be removed?

Might technically have to check for the encode for `ssv` and `pipes` though I don't think anyone is using it in spec repo so we could not support them for now

timotheeguerin (approved):

packages/typespec-autorest/src/openapi.ts
allenjzhang: Lets leave it for now as it won't get hit. If TCGC is updated with consolidated `@encode` logic, we can added it back top and this check could be live again.

tadelesh: i've added `ssv` and `pipes` support for tcgc and autorest emitter.

PR #2353, opened 2025-03-11
[tcgc] add license config support

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/2273


packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -13,6 +13,10 @@ const EmitterOptionsSchema: JSONSchemaType<SdkEmitterOptions> = {
     "examples-directory": { type: "string", nullable: true },
     "examples-dir": { type: "string", nullable: true },
     "emitter-name": { type: "string", nullable: true },
+    "license-name": { type: "string", nullable: true },
```
iscai-msft, line 16: would it be better to have a flag structure more like this?
```yaml
license?:
    name: str
    company: str
    link?: str
    description?: str
```

This is just my guess for what would be required in a license property and what wouldn't be

iscai-msft (approved):

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -46,6 +46,10 @@ export interface TCGCContext {
   flattenUnionAsEnum?: boolean;
   apiVersion?: string;
   examplesDir?: string;
+  licenseName?: string;
+  licenseLink?: string;
+  licenseCompany?: string;
+  licenseDescription?: string;
```
msyyc, line 52: Better to add some annotation.

packages/typespec-client-generator-core/src/license.ts
```diff
@@ -0,0 +1,139 @@
...
+\n
+You should have received a copy of the GNU Lesser General Public License\n
+along with this program.  If not, see <https://www.gnu.org/licenses/>.\n`,
+  },
+};
+
+export function getLicenseInfo(context: TCGCContext): LicenseInfo | undefined {
+  if (!context.licenseName || context.licenseName.trim() === "") {
+    return undefined;
+  }
```
msyyc, line 120: It seems licenseName is necessary however users may not know it. So before return undefined, maybe we could report a warning diagnostics like "licenseName is necessary"

tadelesh: if there is no `licenseName` configured, we will leave the license part as empty.

packages/typespec-client-generator-core/src/lib.ts
weidongxu-microsoft: Personally, I like the structured form too.

One thing to know, previously we are not able to use the command line to set nested property.
e.g. we can do `--option "@typespec/http-client-java.license-name=..."`, but not `--option "@typespec/http-client-java.license.name=..."`
I am not sure whether tsp supports it now.

packages/typespec-client-generator-core/src/license.ts
```diff
@@ -0,0 +1,139 @@
...
+The above copyright notice and this permission notice shall be included in\n
+all copies or substantial portions of the Software.\n
+\n
+THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\n
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\n
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE\n
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER\n
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,\n
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN\n
+THE SOFTWARE.\n`,
```
weidongxu-microsoft, line 26: Does any SDK use the full form?

Java uses a short form
```
Copyright (c) Microsoft Corporation. All rights reserved.
Licensed under the MIT License.
```

tadelesh: python use the full form in `LICENSE` file. do you think i need to add a short form into the type? or let emitter to do it with `license.name` and `license.company`?

packages/typespec-client-generator-core/src/lib.ts
tadelesh: changed to form.

packages/typespec-client-generator-core/src/license.ts
tadelesh: added `header`.

ArcturusZhang (approved):
LGTM!

.chronus/changes/license-2025-2-11-20-38-31.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Add `licenseInfo` property of `LicenseInfo` type to `SdkPackage`, which is used to indicate the license for the package and could be configured by `license.name`, `license.company`, `license.link`, `license.header` and `license-description` in `tspconfig.yaml`.
```
weidongxu-microsoft, line 7: nit `license-description` -> `license.description`

weidongxu-microsoft (approved):
Some license would ask for a `<year>` (GPL?), but I guess we can have what we have now.

qiaozha (approved):

iscai-msft (approved):
looks great, thanks @tadelesh

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -88,8 +96,14 @@ export interface SdkEmitterOptions {
   "flatten-union-as-enum"?: boolean;
   "api-version"?: string;
   "examples-dir"?: string;
-  "emitter-name"?: string;
   namespace?: string;
+  license?: {
```
JoshLove-msft, line 100: Should there be a displayString property that gives the default display?

tadelesh: if there is not `license` config, emitter should not generate any license header or license file.

tadelesh: oh yes. emitter could use `sdkPackage.license` info. if it is not undefined, emitter could use `sdkPackage.license.header` for the license header comment and use `sdkPackage.license.description` for the separated license file content directly.

JoshLove-msft: Understood, but would it make sense to have an API for emitters to use? e.g.
```function getLicenseDisplayString(license: LicenseInfo) : string```

tadelesh: for current design, all the config for license should be transparent for emitter, i.e. emitter only needs to use `context.sdkPackage.license` to get the display string, no need to care about the raw config in `tspconfig.yaml`.

tadelesh: i'll merge this pr to catch the coming release. we could discuss further if more helper function is needed.

PR #2356, opened 2025-03-11
[tcgc] maintain superset of api versions for types

### LINKED ISSUE #2330: [bug][TCGC] wrong `apiVersions` for language emitters
For https://github.com/Azure/azure-rest-api-specs/tree/main/specification/deviceregistry/DeviceRegistry.Management, TCGC provide wrong apiVersions of `endpoint`:

Image

Expected behavior:
`apiVersions` of `endpoint` shall also contain 3 api-verisons as `client` instead of 2.

iscai-msft:
fixes #2330

### DIFF
--- a/packages/typespec-client-generator-core/.vscode/launch.json
+++ b/packages/typespec-client-generator-core/.vscode/launch.json
@@ -1,25 +1,24 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-    "version": "0.2.0",
-    "configurations": [
-        
-      {
-        "name": "Generate typespec",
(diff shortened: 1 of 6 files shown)


packages/typespec-client-generator-core/.vscode/launch.json
timotheeguerin, line 1: did you meant to commit this one?

iscai-msft: yeah, just bc we didn't have one before, so if we want to debug in the future it'd be good to have. Let me know if this goes against repo etiquette

timotheeguerin: I thikn ideally we'd reuise the top level compile scratch but also think this is ok, just not a big fan of having things here that need to open the sub window (and miss from the top level config then)

PR #2369, opened 2025-03-12
Validate and update sample output for media type hint change

witemple-msft:
This is the Azure companion of https://github.com/microsoft/typespec/pull/6357

This PR updates the recorded sample output to account for changes to the default content types of scalars.

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -68,7 +68,7 @@ import {
   twoParamsEquivalent,
 } from "./internal-utils.js";
 import { createDiagnostic } from "./lib.js";
-import { isMediaTypeJson } from "./media-types.js";
+import { isMediaTypeJson, isMediaTypeOctetStream, isMediaTypeTextPlain } from "./media-types.js";
 import {
   getCrossLanguageDefinitionId,
   getEffectivePayloadType,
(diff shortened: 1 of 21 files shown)


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -277,7 +277,10 @@ function createContentTypeOrAcceptHeader(
 ): Omit<SdkMethodParameter, "kind"> {
   const name = bodyObject.kind === "body" ? "contentType" : "accept";
   let type: SdkType = getTypeSpecBuiltInType(context, "string");
-  // for contentType, we treat it as a constant IFF there's one value and it's application/json.
+  // for contentType, we treat it as a constant IFF there's one value and it's one of:
+  //  - application/json
+  //  - text/plain
+  //  - application/octet-stream
```
tadelesh, line 283: after revisit the code logic, i believe this function should always return constant. this function will only be called when this is no explicitly defined `content-type` header with body request or `accept` header with body response. then the content type will always come from the guess of http lib, and should always be one value. so we either choose to make it a parameter with client default value or constant. since for the case of user defined only one content type explicitly, tcgc do not transfer it to a string parameter, i think we should keep consistency here. if the content is not what tsp author need, then an explicit content type header should be added. cc: @iscai-msft for confirmation.

iscai-msft: yes, I am favoring switching things over to returning a string with a `clientDefaultValue` going forward though

tadelesh: but it's ok to merge current impl. we could refine it in another tcgc pr.

tadelesh (approved):
approve for tcgc part

timotheeguerin (approved):

packages/typespec-client-generator-core/src/http.ts
witemple-msft: Yeah, let's merge this to keep the status quo, and then I'll rely on folks with a better grasp of TCGC to handle the rework here if you'd like to change these to all be strings instead of constants.

timotheeguerin (approved):

PR #2378, opened 2025-03-13
[tcgc] output auth related info in the yaml of code model

tadelesh:
more convenient for emitter to know the exact info of auth.


packages/typespec-client-generator-core/src/context.ts
```diff
@@ -154,7 +154,8 @@ async function exportTCGCOutput(context: SdkContext) {
           return undefined; // skip keys starting with "__" from the output
         }
         if (k === "scheme") {
-          return undefined; // remove credential schema
+          const { model, ...rest } = v;
+          return rest; // remove credential schema's model property
```
msyyc, line 158: It is better to add some test cases

iscai-msft (approved):

Issue #1901, opened 2024-11-22
Support "x-ms-format": "duration-constant"


pshao25:
Some swagger will have `"x-ms-format": "duration-constant"` on their model property for example. We may need a new scalar for it. What's more, TCGC needs to pass this value to emitter. 

BTW I have some trials just FYI
1. 
```
model Test {
  @format("duration-constant")
  input: duration;
}
```
This gives error `Cannot apply @format decorator to type it is not a string`

2. 
```
model Test {
  @format("duration-constant")
  input: string;
}
```
This gives warning `'string' format 'duration-constant' is not supported in Autorest. It will not be emitted`. The emitted swagger does not emit this either.

3.
```
model Test {
  @encode("duration-constant")
  input: duration;
}
```
This emits
```
"input": {
  "type": "string",
  "format": "duration-constant"
}
```
SDK has this error:
```
fatal   | System.InvalidOperationException: The MaximumDuration property of VirtualMachineInstallPatchesParameters ends with "Duration" but does not use the duration format to be generated
   as TimeSpan type. Add "format": "duration" with directive in autorest.md for the property if it's ISO 8601 format like P1DT2H59M59S. Add "x-ms-format": "duration-constant" if it's the con 
  stant format like 1.2:59:59.5000000. If the property does not conform to a TimeSpan format, please use "x-ms-client-name" to rename the property for the client.
```
In other words, `"x-ms-format": "duration-constant"` is a must in the generated swagger.

pshao25:
Summary for Option3:

- Use encode to express duration-constant in TypeSpec
- C# can check this string and do whatever is needed
- Other emitters must not crash, and either ignore (and return string) or decide to handle it like C#
- We need to document this somewhere in a list of acceptable encoding
- Swagger compat should be done with directives in autorest.md

weidongxu-microsoft:
For
```
model Test {
  @encode("duration-constant")
  input: duration;
}
```

TCGC gives
```
                                  type:
                                    kind: duration
                                    decorators: []
                                    name: duration
                                    encode: duration-constant
                                    wireType:
                                      kind: string
                                      decorators: []
                                      name: string
                                      doc: A sequence of textual characters.
                                      crossLanguageDefinitionId: TypeSpec.string
                                    doc: A duration/time period. e.g 5s, 10h
                                    crossLanguageDefinitionId: TypeSpec.duration
```

[...]

PR #2379, opened 2025-03-13
Add test for duration-constant

pshao25:
Add tests for duration-constant as the first step to fix https://github.com/Azure/typespec-azure/issues/1901


packages/azure-http-specs/specs/azure/payload/encoding/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+  Expected request body:
+  ```json
+  {
+    "input": "1.02:59:59.5000000"
+  }
+  ```
+  """)
+@put
+@route("/duration-constant")
+op durationConstant(@body body: DurationModel): NoContentResponse;
```
iscai-msft, line 33: i'm confused: based off of the name, it seems that duration should be a constant, however it isn't specified as one in the spec

weidongxu-microsoft (commented):
Please follow similar folder/namespace pattern as http-specs
https://github.com/microsoft/typespec/tree/main/packages/http-specs/specs/encode/bytes

"azure/encode/bytes" maybe

packages/azure-http-specs/specs/azure/payload/encoding/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+using global.Azure.ClientGenerator.Core;
+using Spector;
+
+@doc("Test for azure related encode decorator.")
+@scenarioService("/azure/payload/encoding")
+@scenario
+@scenarioDoc("""
+  Test case for azure specific encoding. SDK should generate correct serialization format according to the set encoding.
+  """)
+namespace _Specs_.Azure.Payload.Encoding;
```
weidongxu-microsoft, line 15: Do we have to use the `_Specs_.` in `_Specs_.Azure.Payload.Encoding`? Would `.Azure.### work?

pshao25: Yes, as long as it starts with `Azure`, we should add some prefix. Any namespace starts with `Azure` treats differently at least for .net.

weidongxu-microsoft: https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/resource-manager/resources/main.tsp#L24

pshao25: Hmmm, not sure if arm is different. Do you have any problem with this prefix? I saw others under azure folder have this prefix.

packages/azure-http-specs/specs/azure/payload/encoding/main.tsp
pshao25: Actually it isn't though it's name is constant. In swagger we could have "x-ms-format": "duration-constant" which means the serialized payload will be like `1.2:59:59.5000000`. We have discussion with TypeSpec team and the conclusion is that swagger is like this in TypeSpec
```
model Test {
  @encode("duration-constant")
  input: duration;
}
```

packages/azure-http-specs/specs/azure/payload/encoding/main.tsp
weidongxu-microsoft: Yes, I have.

If you insist on this, use `@clientNamespace` to rename for Java
e.g. https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/core/page/main.tsp#L22

pshao25: Added.

iscai-msft (approved):

packages/azure-http-specs/specs/azure/encode/duration/main.tsp
```diff
@@ -0,0 +1,35 @@
...
+using TypeSpec.Http;
+using global.Azure.ClientGenerator.Core;
+using Spector;
+
+@doc("Test for azure related encode decorator.")
+@scenarioService("/azure/encode/duration")
+@scenario
+@scenarioDoc("""
+  Test case for azure specific encoding. SDK should generate correct serialization format according to the set encoding.
+  """)
```
weidongxu-microsoft, line 14: Should you put the scenario/scenarioDoc on the operation, instead of the whole file?

weidongxu-microsoft (approved):

PR #2381, opened 2025-03-13
Uptake cleanup of http library

timotheeguerin:
Uptake change required by this pr https://github.com/microsoft/typespec/pull/6433

There is a few spec that use the includePayload decorator but they are all marked as a pattern you shoul;dn't do already using internal apis so we'll just add `Private.` in front and we'll have to figure out how to solve this so we can completely remove the decorator


markcowl (approved):
I think we need a PR against the specs repo *soon*

PR #2396, opened 2025-03-17
[core] add internal `parameterizedNextLink` scalar

iscai-msft:
Following discussion [here](https://github.com/Azure/typespec-azure/blob/05a94ba2b3e8f23b6f99cba0c02785f434cbc06e/design-doc.md), we decided to use an internal scalar to support parameterized next link paging. This PR is the first step, adding an internal scalar type that supports this.

The next step would be doing a TCGC pr to take the information exposed here and pass it to emitters


.chronus/changes/core-addParameterizedNextLink-2025-2-17-14-28-54.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-core"
+---
+
+add internal scalar type `parameterizedNextLink` to handle next links that don't meet guidelines
```
timotheeguerin, line 7: is the type supposed to be internal?

iscai-msft: would it be ok if we made it internal? you're right, right now the decorator is internal but the scalar is public. Can people use internal types in their specs?

timotheeguerin: it warn if they use from the `Private` namespace. But we also have a `Legacy` namespace for bad patterns, maybe it belongs there

iscai-msft: yeah I think legacy makes sense, let me move it into there

iscai-msft: hmm not finding `Legacy`. I'm seeing `obsolete`, everything there's deprecated so not sure it's the one to use

markcowl: Legacy exists in Azure.ResourceManager, but not here.  I would suggest adding a Azure.Core.Legacy namespace specifically for this, I am sure it won't be the last thing we put in this namespace.

We should likely have a similar linting rule that emits a warning when components from this namespace are used.

packages/typespec-azure-core/lib/legacy.tsp
```diff
@@ -0,0 +1,7 @@
+using TypeSpec.Reflection;
+
+namespace Azure.Core.Legacy;
+
+@Azure.Core.Foundations.Private.parameterizedNextLinkConfig(TypeSpec.Reflection.ModelProperty[])
+scalar parameterizedNextLink<ParameterizedParams extends TypeSpec.Reflection.ModelProperty[]>
```
timotheeguerin, line 6: add doc comment?

timotheeguerin (approved):

PR #2402, opened 2025-03-18
[tcgc] ignore visibility when finding response type if it is anonymous model

tadelesh:
with previous logic, http library will return a created anonymous model without `Create` visibility, and tcgc could not find the original model, but use this anonymous model as the response type (`GetResponse`). see [playground](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=QHNlcnZpY2UNCm5hbWVzcGFjZSBUZXN0Q2xpZW50IHsNCiAgbW9kZWzFFsYQICBAdmlzaWJpbGl0eShMaWZlY3ljbGUuQ3JlYXRlKcYjY8UNOiBzdHJpbmc7DQrGF3JlYWTLFSAgfcYab3AgZ2V0KCk6xWg7DQp9DQo%3D&e=%40azure-tools%2Ftypespec-client-generator-core&options=%7B%7D), line 59.
with current logic, tcgc could find the original model and will use that model as the response type (`Test`). see [playground](https://cadlplayground.z22.web.core.windows.net/typespec-azure/prs/2402/?c=QHNlcnZpY2UNCm5hbWVzcGFjZSBUZXN0Q2xpZW50IHsNCiAgbW9kZWzFFsYQICBAdmlzaWJpbGl0eShMaWZlY3ljbGUuQ3JlYXRlKcYjY8UNOiBzdHJpbmc7DQrGF3JlYWTLFSAgfcYab3AgZ2V0KCk6xWg7DQp9DQo%3D&e=%40azure-tools%2Ftypespec-client-generator-core&options=%7B%7D), line59.

### DIFF
--- a/packages/typespec-client-generator-core/src/public-utils.ts
+++ b/packages/typespec-client-generator-core/src/public-utils.ts
@@ -119,9 +119,10 @@ export function getClientNamespaceString(context: TCGCContext): string | undefin
 }
 
 /**
- * If the given type is an anonymous model and all of its properties excluding
- * header/query/path/status-code are sourced from a named model, returns that original named model.
- * Otherwise the given type is returned unchanged.
+ * If the given type is an anonymous model, returns a named model with same shape.
+ * The finding logic will ignore all the properties of header/query/path/status-code metadata,
+ * as well as the properties that are not visible in the given visibility if provided.
+ * If the model found is also anonymous, the input type is returned unchanged.
  *
  * @param context
  * @param type
--- a/packages/typespec-client-generator-core/src/public-utils.ts
+++ b/packages/typespec-client-generator-core/src/public-utils.ts
@@ -17,7 +17,14 @@ import {
   isService,
   resolveEncodedName,
 } from "@typespec/compiler";
-import { HttpOperation, getHttpOperation, getHttpPart, isMetadata } from "@typespec/http";
+import {
+  HttpOperation,
+  Visibility,
+  getHttpOperation,
+  getHttpPart,
+  isMetadata,
+  isVisible,
+} from "@typespec/http";
 import { Version, getVersions } from "@typespec/versioning";
 import { pascalCase } from "change-case";
 import pluralize from "pluralize";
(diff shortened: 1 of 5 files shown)


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -120,7 +127,11 @@ export function getClientNamespaceString(context: TCGCContext): string | undefin
  * @param type
  * @returns
  */
-export function getEffectivePayloadType(context: TCGCContext, type: Model): Model {
+export function getEffectivePayloadType(
+  context: TCGCContext,
+  type: Model,
+  visibility?: Visibility,
+): Model {
```
weidongxu-microsoft, line 134: Update the docstring?

I am not sure whether `Visibility.Read` mean here. In the test, `Visibility.Create` would be not affect the effective type. Then what kind of property would affect? `@visibility(Visibility.Read) prop`?

tadelesh: doc updated. the visibility here means only take the properties with the visibility provided into account when finding the potential named model.

weidongxu-microsoft: So, the test case works, because the response from http lib is
```
{
  read: string
}
```
and via this `getEffectivePayloadType(..., Visibility.READ)`, it would map to `Test` model?

weidongxu-microsoft (approved):
LGTM

Not sure how much its impact on test. Hope real SDK won't be affected (I think it unlikely).

msyyc (approved):

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh: correct.

tadelesh:
> LGTM
> 
> Not sure how much its impact on test. Hope real SDK won't be affected (I think it unlikely).

i doubt at least batch will be impacted?

weidongxu-microsoft:
> i doubt at least batch will be impacted?

Do you mean "specification/batch/Azure.Batch"? I seem not finding Visibility there.

PR #2403, opened 2025-03-18
Add non-resource arm tests

pshao25:
In many cases, when a model has id, name, type, customers still want it to be a non resource. To realize it, we ask them 
1. This model should not extend from Resource model form arm libraries.
2. All the operations on that model should not be marked with any arm related decorators and the interface too.
3. They should write these operations in low level way.

Add this test to ensure our client can correctly handle this case.

### DIFF
--- a/packages/azure-http-specs/specs/azure/resource-manager/non-resource/mockapi.ts
+++ b/packages/azure-http-specs/specs/azure/resource-manager/non-resource/mockapi.ts
@@ -1,0 +1,30 @@
+import { json, passOnSuccess, ScenarioMockApi } from "@typespec/spec-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+const SUBSCRIPTION_ID_EXPECTED = "00000000-0000-0000-0000-000000000000";
+const LOCATION_EXPECTED = "eastus";
+
+const nonResource = {
+  id: "id",
+  name: "hello",
+  type: "nonResource"
+};
+
+Scenarios.Azure_ResourceManager_NonResource_NonResourceOperations_get = passOnSuccess({
+  uri: "/subscriptions/:subscriptionId/providers/Microsoft.NonResource/locations/:location/otherParameters/:parameter",
+  method: "get",
+  request: {
+    params: {
+      subscriptionId: SUBSCRIPTION_ID_EXPECTED,
+      location: LOCATION_EXPECTED,
+      parameter: "hello",
+      "api-version": "2023-12-01-preview",
+    },
+  },
+  response: {
+    status: 200,
+    body: json(nonResource),
+  },
+  kind: "MockApiDefinition",
+});
--- a/packages/azure-http-specs/specs/azure/resource-manager/non-resource/non-resource.tsp
+++ b/packages/azure-http-specs/specs/azure/resource-manager/non-resource/non-resource.tsp
@@ -1,0 +1,71 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-azure-resource-manager";
+import "@typespec/spector";
+
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using Azure.Core;
+using Azure.ResourceManager;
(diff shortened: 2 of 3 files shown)


packages/azure-http-specs/specs/azure/resource-manager/non-resource/mockapi.ts
```diff
@@ -0,0 +1,30 @@
...
+const LOCATION_EXPECTED = "eastus";
+
+const nonResource = {
+  id: "id",
+  name: "hello",
+  type: "nonResource"
+};
+
+Scenarios.Azure_ResourceManager_NonResource_NonResourceOperations_get = passOnSuccess({
+  uri: "/subscriptions/:subscriptionId/providers/Microsoft.NonResource/locations/:location/otherParameters/:parameter",
```
XiaofeiCao, line 15: Why does the URL look exactly like a resource? Does this reflect the real use case?

XiaofeiCao: This test is only for validating that SDK can generate without reporting error, and we don't have requirements on how SDK looks, right?

pshao25: I'm not sure I follow you. What's the problem it looks like a resource url? As long as they define it as a non resource operation, we should take it as a non resource operation, no matter how it looks like. 

Same as the model, no matter how it looks like a real resource. As long as user defines it as non resource model, we should take it as a non resource model.

As yes, it is a real case.

pshao25: How SDK looks like is each language's own decision. I can only speak to how C# looks like. You decide how Java looks like, etc...

pshao25: To give you more insights, TypeSpec's advantages over swagger is that it gives clear information about what is a resource, what is not a resource, instead of the guessing logic in swagger. We definitely should honor such information to unveil such advantages. So I add these tests.

XiaofeiCao: Thanks. Currently Java hasn't taken advantage of the TypeSpec information.

XiaofeiCao (approved):
LGTM, thanks!

XiaofeiCao:
Is `PUT` non resource possible(My guess is no)? 
image

pshao25:
> Is `PUT` non resource possible(My guess is no)? image

Why not? I assume they can do whatever they want.

XiaofeiCao:
> Why not? I assume they can do whatever they want.

OK, would you help add this case? Java is currently having trouble identifying it as a non-resource.

packages/azure-http-specs/specs/azure/resource-manager/non-resource/non-resource.tsp
```diff
@@ -0,0 +1,122 @@
...
+   */
+  id?: string;
+
+  /**
+   * The name of the result
+   */
+  name?: string;
+
+  /**
+   * The result resource type
```
weidongxu-microsoft, line 31: nit, doc "resource type" while you are saying this is not resource in test case

packages/azure-http-specs/specs/azure/resource-manager/non-resource/non-resource.tsp
```diff
@@ -0,0 +1,122 @@
...
+  type?: string;
+}
+
+/**
+ * Operations on non resource model should not be marked as `@armResourceOperations`.
+ */
+interface NonResourceOperations {
+  @scenario
+  @scenarioDoc("""
+    It's non resource get operation operating on non resource model, though the model has `id`, `name`, `type` properties.
```
weidongxu-microsoft, line 42: non-resource ?

"non" is not a word -- maybe let AI proof read the doc?

weidongxu-microsoft (approved):
I am not sure whether ARM allows a PUT of non-resource.

Typically, we see GET does this, as well as some POST. So far I could not remember PUT or PATCH.

But anyway fine as a test.

PR #2435, opened 2025-03-24
Change OperationStatusResult.resourceId to armResourceIdentifier

pshao25:
Change `OperationStatusResult.resourceId` to `armResourceIdentifier` as what we did for `OperationStatusResult.id`.

Justification:
Though semantically it is the same as before, in the client we have the logic "if the string is armResourceIdentifier from Azure.Core, then we use the ResourceId type from our core library". If `OperationStatusResult.resourceId` is giving us a `string` type, we don't know it is actually from `Azure.Core`, so we won't do the replacement to the core type.


packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -218,9 +218,8 @@ model OperationStatusResult {
 
   /** Fully qualified ID of the resource against which the original async operation was started. */
   @visibility(Lifecycle.Read)
-  @format("arm-id")
   @added(Versions.v5)
-  resourceId?: string;
```
markcowl, line 223: My question is how this impacts our common types output in typespec-autorest.  Adding @AlitzelMendez  as a reviewer for comment

pshao25: It changes nothing, we do common type replacement.
`test: Azure.ResourceManager.CommonTypes.OperationStatusResult;` produces 
```
"test": {
          "$ref": "../../../common-types/resource-management/v5/types.json#/definitions/OperationStatusResult"
        }
```

AlitzelMendez: thanks Pan!! I will just double check and if everything looks good I will approve

AlitzelMendez (approved):

PR #2440, opened 2025-03-25
azure-http-specs, fix mockapi in flattenProperties and ARM cases

weidongxu-microsoft:
- fix simple error in flattenProperties and ARM PUT
- fix content-type for ARM PATCH -- it should be "application/json" (this does not affect test, though)
- add handler func for ARM PATCH, that only validate the property that matters (PATCH allows unchanged property either not sent, or sent with old value)

### DIFF
--- a/packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
+++ b/packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
@@ -1,4 +1,11 @@
-import { json, passOnCode, passOnSuccess, ScenarioMockApi } from "@typespec/spec-api";
+import { deepEquals } from "@typespec/compiler/utils";
+import {
+  json,
+  passOnCode,
+  passOnSuccess,
+  ScenarioMockApi,
+  ValidationError,
+} from "@typespec/spec-api";
 
 export const Scenarios: Record<string, ScenarioMockApi> = {};
 
@@ -88,6 +95,11 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_createWithSyste
     method: "put",
     request: {
       body: json({
+        location: "eastus",
+        tags: {
+          tagKey1: "tagValue1",
+        },
+        properties: {},
         identity: createExpectedIdentity,
       }),
       pathParams: {
@@ -115,7 +127,7 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_updateWithUserA
         identity: updateExpectedIdentity,
       }),
       headers: {
-        "Content-Type": "application/merge-patch+json",
+        "Content-Type": "application/json",
       },
       pathParams: {
         subscriptionId: SUBSCRIPTION_ID_EXPECTED,
@@ -131,6 +143,19 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_updateWithUserA
       body: json(validUserAssignedAndSystemAssignedManagedIdentityResource),
     },
     kind: "MockApiDefinition",
+    handler: (req) => {
+      if (!deepEquals(req.body["identity"], updateExpectedIdentity)) {
+        throw new ValidationError(
+          "Body should contain 'identity' property",
+          updateExpectedIdentity,
+          req.body,
+        );
+      }
+      return {
+        status: 200,
+        body: json(validUserAssignedAndSystemAssignedManagedIdentityResource),
+      };
+    },
   });
 
 Scenarios.Azure_ResourceManager_CommonProperties_Error_getForPredefinedError = passOnCode(404, {
--- a/packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
+++ b/packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
@@ -165,6 +165,7 @@ Scenarios.Azure_ResourceManager_CommonProperties_Error_createForUserDefinedError
   method: "put",
   request: {
     body: json({
+      location: "eastus",
       properties: {
         username: "00",
       },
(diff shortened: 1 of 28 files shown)


packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
```diff
@@ -115,7 +127,7 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_updateWithUserA
         identity: updateExpectedIdentity,
       }),
       headers: {
-        "Content-Type": "application/merge-patch+json",
```
weidongxu-microsoft, line 118: ARM should be "application/patch"

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -791,7 +791,7 @@ Scenarios.Azure_ResourceManager_Resources_Singleton_update = passOnSuccess({
       },
     }),
     headers: {
-      "Content-Type": "application/merge-patch+json",
```
weidongxu-microsoft, line 794: PATCH in ARM is "application/json" (though part of it behavior like a json-merge-patch)
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/standbypool/resource-manager/Microsoft.StandbyPool/stable/2025-03-01/standbypool.json#L334

pshao25: We wrapped the body with `json`. Do we need to add the header once more?

weidongxu-microsoft: Emm, seems not. Other API does not have this `content-type` header.
Let me try remove it.

XiaofeiCao (approved):

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -916,7 +917,7 @@ Scenarios.Azure_ResourceManager_Resources_TopLevel_update = passOnSuccess({
           description: "valid2",
         },
       },
-      "application/merge-patch+json",
+      "application/json",
```
pshao25, line 920: Do we need this?

packages/azure-http-specs/specs/azure/resource-manager/resources/mockapi.ts
```diff
@@ -1051,7 +1071,7 @@ Scenarios.Azure_ResourceManager_Resources_Nested_update = passOnSuccess({
...
```
pshao25, line 1074: And this?

pnpm-workspace.yaml
```diff
@@ -7,3 +7,5 @@ packages:
   - "!core/packages/http-client-csharp/**"
   - "!core/packages/http-client-java/**"
   - "!core/packages/http-client-python/**"
+  - "!core/packages/typespec-vs/**"
+  - "!core/packages/typespec-vscode/**"
```
timotheeguerin, line 11: why is this added here? we do want to be able to build them here too, they are part of the core workspace

weidongxu-microsoft: Sorry, this file should not be included, reverted (I typically do this in local, as build of these plugins required specific version of .NET lib) -- appreciate any suggestion that can do e.g. "only build essential lib on pnpm build".

timotheeguerin: Hhm, they should skip if you don't have the right requirements locally.

Can you file an issue with what you see.

I think we might want to find a more generalized way to skip things if you don't have the right items

packages/azure-http-specs/specs/azure/resource-manager/common-properties/mockapi.ts
```diff
@@ -88,6 +95,11 @@ Scenarios.Azure_ResourceManager_CommonProperties_ManagedIdentity_createWithSyste
     method: "put",
     request: {
       body: json({
+        location: "eastus",
+        tags: {
+          tagKey1: "tagValue1",
+        },
+        properties: {},
```
pshao25, line 102: The empty `{}` is what Java needs?

weidongxu-microsoft: I think Java sends the empty object in `properties` property.

The resource has a model for `properties`. Just it only contain read-only property.
https://github.com/Azure/typespec-azure/blob/a1ab10e20e53d4738f196d0df539bba01d842083/packages/azure-http-specs/specs/azure/resource-manager/common-properties/managed-identity.tsp#L18-L36

pshao25: If user don't set the properties, shouldn't we send nothing?

pshao25: I tried to find out a workaround in .net to send the `{}`, but failed.

weidongxu-microsoft: Sending a `{}` and sending nothing is not exactly same...

Would Dapeng know about how this be sent? I assume ARM resource would send a `properties` in almost all cases. Just in this case there is no writable property in this model.

pshao25: Our logic is if you don't touch the properties, then send nothing. Or maybe you can have a value set for `Properties.ProvisioningState` so that we can touch it?

weidongxu-microsoft: Yeah, I can add a writable property within `properties`. But it would require test change in emitters (as they now need to set this).

So here I just add a `handler` to again only verify the `identity` whether this test is for.

pshao25 (approved):

antkmsft:
@weidongxu-microsoft, now the test started failing for us, see #2491

weidongxu-microsoft:
> @weidongxu-microsoft, now the test started failing for us, see #2491

I've answered in your issue.

My guess is that your spector/spec-api lib is outdated (all of these libs need to upgrade together).

PR #2441, opened 2025-03-25
[tcgc] add new decorator emptyStringAsNull

haiyuazhang:


packages/typespec-client-generator-core/test/decorators/general-list.test.ts
```diff
@@ -338,11 +338,17 @@ describe("csharp only decorator", () => {
   it("@useSystemTextJsonConverter", async function () {
     runner = await createSdkTestRunner(
       {},
-      { additionalDecorators: ["Azure\\.ClientGenerator\\.Core\\.@useSystemTextJsonConverter"] },
+      {
```
live1206, line 341: The test description is for another decorator.
We should separate it to a isolated test if we want to make sure the e2e works.

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -608,3 +608,18 @@ extern dec scope(target: Operation, scope?: valueof string);
  * ```
  */
 extern dec apiVersion(target: ModelProperty, value?: valueof boolean, scope?: valueof string);
+
+/**
+ * Whether a model needs to treat empty string as null, this is only used for backward compatibility for csharp.
```
live1206, line 613: This is not only for backward compatibility.
And this might be used for other language scope as well.

haiyuazhang: removed the inaccurate description.

packages/typespec-client-generator-core/test/decorators/general-list.test.ts
haiyuazhang: refined the tests.

live1206 (approved):
LGTM
@tadelesh for final approval

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -608,3 +608,18 @@ extern dec scope(target: Operation, scope?: valueof string);
...
+ *
+ * @example
+ * ```typespec
+ * @emptyStringAsNull
+ * model MyModel {
+ *   prop: string;
+ * }
+ * ```
+ */
+extern dec emptyStringAsNull(target: Model, scope?: valueof string);
```
ArcturusZhang, line 625: From feature's perspective, it makes more sense if this decorator applies to the property.
I think we made it targets the model because our old configuration is using the model name as its value, but that is actually a bad decision - because apparently, in most cases, not all the string properties in that model want to treat empty string as null.
So here I would say we should change the target to `Property`.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1251,3 +1252,9 @@ function IsInScope(context: TCGCContext, entity: Operation): boolean {
   }
   return true;
 }
+
+export const $emptyStringAsNull: EmptyStringAsNullDecorator = (
+  context: DecoratorContext,
+  target: Model,
+  scope?: LanguageScopes,
+) => {};
```
ArcturusZhang, line 1260: I think we need some validation in this decorator - for instance, if we insist it to apply to a model, we need to validate the model has a property whose type is string.
And if we change the target to property, we should validate it should only apply to a string-typed property.

packages/typespec-client-generator-core/lib/decorators.tsp
live1206: As offline discussed, it does look clearer to target on property instead of model.
But autorest.csharp emitter puts this config on model now. When we need to migrate from OpenAPI to TypeSpec, we need to add the decorator on property instead of model. We can handle this during tsp conversion.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1251,3 +1252,9 @@ function IsInScope(context: TCGCContext, entity: Operation): boolean {
   }
   return true;
 }
+
+export const $emptyStringAsNull: EmptyStringAsNullDecorator = (
+  context: DecoratorContext,
+  target: Model,
```
tadelesh, line 1258: why the target is model? i suppose it should apply to model property.

packages/typespec-client-generator-core/lib/decorators.tsp
haiyuazhang: change the decorator's target to model property.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: validation added.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: change decorator's target to model property.

packages/typespec-client-generator-core/test/decorators/deserialize-empty-string-as-null.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        model A {
+          @deserializeEmptyStringAsNull
+          prop1: string[];
+        }
+    `);
+
+    expectDiagnostics(diagnostics, {
+      code: "@azure-tools/typespec-client-generator-core/invalid-deserializeEmptyStringAsNull-target-type",
+    });
+  });
```
ArcturusZhang, line 80: maybe in the future we could have a requirement that this could apply to array or dictionary when its value type is a string-like scalar, but it is fine for now.

packages/typespec-client-generator-core/test/decorators/deserialize-empty-string-as-null.test.ts
```diff
@@ -0,0 +1,118 @@
...
+            model A {
+              @deserializeEmptyStringAsNull
+              prop1: int64like;
+            }
+        `);
+
+    expectDiagnostics(diagnostics, {
+      code: "@azure-tools/typespec-client-generator-core/invalid-deserializeEmptyStringAsNull-target-type",
+    });
+  });
```
ArcturusZhang, line 117: could we add a new test case to validate it will error out when we apply this to a property with a model as its type?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1251,3 +1252,30 @@ function IsInScope(context: TCGCContext, entity: Operation): boolean {
   }
   return true;
 }
+
+export const $deserializeEmptyStringAsNull: DeserializeEmptyStringAsNullDecorator = (
+  context: DecoratorContext,
+  target: ModelProperty,
+  scope?: LanguageScopes,
+) => {
+  if (target.kind !== "ModelProperty" || target.type.kind !== "Scalar") {
```
tadelesh, line 1261:
```suggestion
  if (target.type.kind !== "Scalar") {
```

ArcturusZhang (approved):

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -293,6 +293,13 @@ export const $lib = createTypeSpecLibrary({
         default: paramMessage`Invalid 'initializedBy' value. ${"message"}`,
       },
     },
+    "invalid-deserializeEmptyStringAsNull-target-type": {
+      severity: "error",
+      messages: {
+        default:
+          "@deserializeEmptyStringAsNull can only be applied to ModelProperty of type 'string' or a Scalar type derived from 'string'.",
```
tadelesh, line 300:
```suggestion
          "@deserializeEmptyStringAsNull can only be applied to `ModelProperty` of type 'string' or a `Scalar` derived from 'string'.",
```

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1251,3 +1252,30 @@ function IsInScope(context: TCGCContext, entity: Operation): boolean {
...
+      code: "invalid-deserializeEmptyStringAsNull-target-type",
+      format: {},
+      target: target,
+    });
+    return;
+  }
+
+  if (target.type.name !== "string") {
+    const scalarType = target.type as Scalar;
+    if (scalarType.baseScalar === undefined || scalarType.baseScalar.name !== "string") {
```
tadelesh, line 1272: this could not handle multi-layer extends for scalar, you need to find the root base scalar.
```
scalar armid extends uuid;
scalar uuid extends string;
```

tadelesh (approved):

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: the logic of finding base Scalar type has been refined.

tadelesh (approved):

packages/typespec-client-generator-core/test/decorators/deserialize-empty-string-as-null.test.ts
```diff
@@ -0,0 +1,189 @@
...
+      {},
+      { additionalDecorators: ["Azure\\.ClientGenerator\\.Core\\.@deserializeEmptyStringAsNull"] },
+    );
+
+    await runner.compileWithBuiltInService(`
+        scalar l1 extends string;
+        scalar l2 extends l1;
+
+        model A {
+          @deserializeEmptyStringAsNull("csharp")
```
MaryGao, line 57: What if there is no scope specified like in spector case defined? Will this misleading customers if some languages don't support this? 

https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/client-generator-core/deserialize-empty-string-as-null/main.tsp#L18-L21

```
@doc("This is a Model contains a string-like property of type url.")
model ResponseModel {
  @deserializeEmptyStringAsNull
  sampleUrl: url;
}
```

PR #2447, opened 2025-03-25
[tcgc] emitter options split

iscai-msft:
BREAKING: emitter options will now be individually exposed and split into `BrandedSdkEmitterOptions` and `UnbrandedSdkEmitterOptions`. This change allows emitters to choose which flags to support, BUT it means that you will have to explicitly choose the flags to support. Example code snippet in an emitter:

```ts
import { BrandedSdkEmitterOptions, UnbrandedSdkEmitterOptions } from "@azure-tools/typespec-client-generator-core"

const UnbrandedPythonEmitterOptionsSchema: JSONSchemaType<UnbrandedPythonEmitterOptions> = {
  type: "object",
  additionalProperties: true,
  properties: {
    "python-specific-flag": {...},
    ...UnbrandedSdkEmitterOptions["api-version"],
    ...UnbrandedSdkEmitterOptions["license"],
  }
}

const BrandedPythonEmitterOptionsSchema: JSONSchemaType<BrandedPythonEmitterOptions> = {
  type: "object",
  additionalProperties: true,
  properties: {
    "branded-python-specific-flag": {...},
    ...BrandedSdkEmitterOptions["namespace"],
    ...UnbrandedPythonEmitterOptionsSchema.properties,
  }
}
```

### DIFF
--- a/packages/typespec-client-generator-core/src/context.ts
+++ b/packages/typespec-client-generator-core/src/context.ts
@@ -28,7 +28,7 @@ import {
   TCGCContext,
 } from "./interfaces.js";
 import {
-  BrandedSdkEmitterOptions,
+  BrandedSdkEmitterOptionsInterface,
   handleVersioningMutationForGlobalNamespace,
   parseEmitterName,
   TCGCEmitterOptions,
@@ -100,7 +100,7 @@ export interface CreateSdkContextOptions {
 }
 
 export async function createSdkContext<
-  TOptions extends Record<string, any> = BrandedSdkEmitterOptions,
+  TOptions extends Record<string, any> = BrandedSdkEmitterOptionsInterface,
   TServiceOperation extends SdkServiceOperation = SdkHttpOperation,
 >(
   context: EmitContext<TOptions>,
--- a/packages/typespec-client-generator-core/src/context.ts
+++ b/packages/typespec-client-generator-core/src/context.ts
@@ -38,12 +38,11 @@ export interface TCGCEmitterOptions extends SdkEmitterOptions {
   "emitter-name"?: string;
 }
 
-export interface SdkEmitterOptions {
+export interface UnbrandedSdkEmitterOptions {
   "generate-protocol-methods"?: boolean;
   "generate-convenience-methods"?: boolean;
   "api-version"?: string;
   "examples-dir"?: string;
-  namespace?: string;
   license?: {
     name: string;
     company?: string;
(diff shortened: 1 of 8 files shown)


packages/typespec-client-generator-core/src/context.ts
```diff
@@ -38,12 +38,11 @@ export interface TCGCEmitterOptions extends SdkEmitterOptions {
   "emitter-name"?: string;
 }
 
-export interface SdkEmitterOptions {
+export interface UnbrandedSdkEmitterOptions {
```
timotheeguerin, line 41: will all the emitters supprot those or it will still be pick an dchoose, might be beter to just export each one separately if thats the case

packages/typespec-client-generator-core/src/context.ts
```diff
@@ -38,12 +38,11 @@ export interface TCGCEmitterOptions extends SdkEmitterOptions {
...
   "generate-protocol-methods"?: boolean;
   "generate-convenience-methods"?: boolean;
   "api-version"?: string;
   "examples-dir"?: string;
```
tadelesh, line 45: this should be azure only.

packages/typespec-client-generator-core/src/context.ts
tadelesh: i agree. e.g., java does not want any flag for unbranded.

JoshLove-msft:
> We keep branded with the original name of SdkEmitterOptions to avoid any deprecations / breaking

Can't we still make breaking changes at this point?

packages/typespec-client-generator-core/src/context.ts
weidongxu-microsoft, line 45: For Java, currently we only expose `license` emitter option. What about .NET? Do they need to expose `generate-protocol-methods` `generate-convenience-methods`?

For input from Srikanta as well.

tadelesh: we are going to export each option separately which could give emitter enough flexibility.

iscai-msft:
> > We keep branded with the original name of SdkEmitterOptions to avoid any deprecations / breaking
> 
> Can't we still make breaking changes at this point?

We can, but I'm just trying to avoid it. If you feel strongly though, I'll go ahead and break the emitter options name. I'm going to actually work on exporting them each individually, which would also be a breaking change, so we might as well make everything how we want it to be

weidongxu-microsoft (approved):

tadelesh (approved):

Issue #2110, opened 2025-01-22
[tcgc]filter out empty namespaces in TCGC output


live1206:
### Clear and concise description of the problem

Namespaces with no operations and no sub-clients should be filtered out. 
Interfaces without operations should still be included because in that case the customer is explicitly defining the interface(garbage in, garbage out), 
whereas for namespaces they may be defining a models-only namespace in which case a client is unnecessary.
```
        @service
        namespace NotMyClient;
        model A {}
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

live1206:
shifted from https://github.com/microsoft/typespec/issues/5624#issuecomment-2605892497

JoshLove-msft:
> Namespaces with no operations and no sub-clients should be filtered out.

Even if it has a sub-client, if the top-level namespace has no operations AND it doesn't have the `@client` decorator, it should still be filtered out. The subclient should then be generated as a regular client with public ctor.

live1206:
> > Namespaces with no operations and no sub-clients should be filtered out.
> 
> Even if it has a sub-client, if the top-level namespace has no operations AND it doesn't have the `@client` decorator, it should still be filtered out. The subclient should then be generated as a regular client with public ctor.

Discussed with @tadelesh, this will cause breaking changes if user add operation to the top-level client later.
Then, the top-level client will be kept and the sub-client will be changed from public client to a sub-client of the top-level client.

live1206:
> > > Namespaces with no operations and no sub-clients should be filtered out.
> > 
> > 
> > Even if it has a sub-client, if the top-level namespace has no operations AND it doesn't have the `@client` decorator, it should still be filtered out. The subclient should then be generated as a regular client with public ctor.
> 
> Discussed with [@tadelesh](https://github.com/tadelesh), this will cause breaking changes if user add operation to the top-level client later. Then, the top-level client will be kept and the sub-client will be changed from public client to a sub-client of the top-level client.

[...]

PR #2479, opened 2025-04-01
Filter out empty client

live1206:
Resolves https://github.com/Azure/typespec-azure/issues/2110
- if user has defined `@client` or `@operationGroup`, skip the filtering as garbage-in-garbage-out
- filter out model-only clients
- if a top-level client contains no operation and only model-only sub-clients, it should be filtered out


packages/typespec-client-generator-core/test/clients/structure.test.ts
```diff
@@ -580,14 +580,48 @@ it("namespace", async () => {
...
+    autoUsings: ["Azure.Core"],
+    emitterName: "@azure-tools/typespec-java",
+  });
+  await runnerWithCore.compile(`
+    namespace Foo {
+      model B {}
+    }
+  `);
+  const sdkPackage = runnerWithCore.context.sdkPackage;
+  strictEqual(sdkPackage.clients.length, 0);
```
iscai-msft, line 604: can you also make sure if there's a model-only namespace with explicit usage marked that it's filtered out from the clients list, but not filtered out from `.namespaces` and that model is still exposed? Thanks!

live1206: test updated

packages/typespec-client-generator-core/src/methods.ts
```diff
@@ -727,6 +727,13 @@ export function createSdkMethods<TServiceOperation extends SdkServiceOperation>(
     const operationGroupClient = diagnostics.pipe(
       createSdkClientType<TServiceOperation>(context, operationGroup, sdkClientType),
     );
+    // skip if there is no operation or sub-client
+    if (
+      operationGroupClient.methods.length === 0 &&
+      (!operationGroupClient.children || operationGroupClient.children.length === 0)
+    ) {
+      continue;
+    }
```
tadelesh, line 736: you should add this logic to operation group related function. this place is just a consumer.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -27,8 +28,12 @@ export function createSdkPackage<TServiceOperation extends SdkServiceOperation>(
   diagnostics.pipe(handleAllTypes(context));
   const crossLanguagePackageId = diagnostics.pipe(getCrossLanguagePackageId(context));
   const allReferencedTypes = getAllReferencedTypes(context);
+  const allClients: SdkClientType<TServiceOperation>[] = listClients(context).map((c) =>
+    diagnostics.pipe(createSdkClientType(context, c)),
+  );
+
   const sdkPackage: SdkPackage<TServiceOperation> = {
-    clients: listClients(context).map((c) => diagnostics.pipe(createSdkClientType(context, c))),
+    clients: allClients.filter((c) => c.methods.length > 0 || c.children), // filter out clients without methods or children
```
tadelesh, line 36: same problem. you should change `listClients`.

packages/typespec-client-generator-core/src/methods.ts
live1206: This function is invoked recursively, this is the place the cut the unnecessary recursion properly.

live1206: And we don't have methods or children/sub-client from an operation group, we need to do this after L728 where client is constructed.

tadelesh (approved):

packages/typespec-client-generator-core/src/methods.ts
live1206: implemented in `listOperationGroups` instead

tadelesh (approved):

PR #2483, opened 2025-04-01
[tcgc] add linter rule against anonymous models

### LINKED ISSUE #1727: [linter]discourage to use anonymous model and always give a model name

Close [typespec one](https://github.com/microsoft/typespec/issues/4819) filed an issue in typespec-azure repo.

Offline discussed with @ArcturusZhang and we think anomymous model may not be a good practice and we should encourage to give a name for models especially for exposed public APIs.

Most of downstreaming language emitters may not have ability to handle inline expressions like typespec so they need to GUESS a name for anomymous models. This could introduce uncertainty if downstreaming name rule is different and for example the name rule introduced breakings between m4 and tcgc and one issue happened in common types is tracked [here](https://github.com/Azure/typespec-azure/issues/1717).

Even we could find a name rule and these names could be fragile and it would be changed with its context or code order if conflicting happened. 

For example in TCGC we try to guess this name according to context, including `parent model name + property name + model type`. So the name could be changed once any of these factors changed. Also we may name some models during conflicting and any code order changes would impact the generated code.

```
model Ab {
  c: {
    foo: string;
  };
}

// vs
model A {
  b: {
    c: {
      foo: string;
    };
  };
}
```

Can we add a linter to discourage the anonymous model in typespec? Especially in azure could we avoid using anonymous model then?

### Checklist

iscai-msft:
fixes #1727


packages/typespec-client-generator-core/src/rules/require-named-model.rule.ts
```diff
@@ -0,0 +1,34 @@
+import { createRule, Model, paramMessage } from "@typespec/compiler";
+import { createTCGCContext } from "../context.js";
+import { createSdkPackage } from "../package.js";
+
+export const requireNamedModelRule = createRule({
+  name: "require-named-model",
```
timotheeguerin, line 6:
```suggestion
  name: "no-unnamed-types",
```

I think this name would be better. There is also other cases than models like inline union and in the future we might allow any type to be created as expressions too

timotheeguerin (approved):

PR #2484, opened 2025-04-02
azure-http-specs, escape "scenario" as it is keyword

weidongxu-microsoft:
`scenario` now is a keyword.

With latest core, formatting would already report error, if this is not escaped https://github.com/Azure/typespec-azure/actions/runs/14210689268/job/39817212216?pr=2484

need to bump core first https://github.com/Azure/typespec-azure/pull/2485

### DIFF
--- a/packages/azure-http-specs/specs/client/structure/default/client.tsp
+++ b/packages/azure-http-specs/specs/client/structure/default/client.tsp
@@ -4,7 +4,7 @@ import "@typespec/spector";
 
 using Spector;
 
-@@scenario(Client.Structure.Service);
+@@`scenario`(Client.Structure.Service);
 @@scenarioDoc(Client.Structure.Service,
   """
     This is to show that if we don't do any customization. The client side should be able to call the api like
--- a/.chronus/changes/azure-http-specs_fix-keyword-2025-3-2-11-19-59.md
+++ b/.chronus/changes/azure-http-specs_fix-keyword-2025-3-2-11-19-59.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/azure-http-specs"
+---
+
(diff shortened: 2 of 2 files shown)


tadelesh (approved):

MaryGao (approved):

packages/azure-http-specs/specs/client/structure/default/client.tsp
```diff
@@ -4,7 +4,7 @@ import "@typespec/spector";
 
 using Spector;
 
-@@scenario(Client.Structure.Service);
+@@`scenario`(Client.Structure.Service);
```
MaryGao, line 7: @allenjzhang @timotheeguerin it seems the [recent pr](https://github.com/microsoft/typespec/pull/6794/files#diff-8685634079a4b1fdca6f014e6c1a079ab1e786c5689d26a56a16399c0684c793) introduced this changes.

And we have two questions here:
- why only `@@scenario` reported errors? we have a lot of `@scenario` with no errors?
- if we update this way we can't bypass the format, is the fix correct? how to bypass format check?

weidongxu-microsoft: I think we are not going to bypass format.

We will need to bump core first https://github.com/Azure/typespec-azure/pull/2485
(I guess I would actually merge the change in this PR into the above bump core PR)

weidongxu-microsoft:
would use https://github.com/Azure/typespec-azure/pull/2485 for this change

packages/azure-http-specs/specs/client/structure/default/client.tsp
allenjzhang: Yeah, do create a manual PR with this change and bump core to have green PR.

Issue #2388, opened 2025-03-14
[TCGC] Extend `@@alternateType` to support non-scalar types


live1206:
### Clear and concise description of the problem

Right now, `@@alternateType` only supports change type to scalar, we need to extend the usage for array of scalar and unknown.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

pshao25:
Customers need this to be enabled
```
model Test {
  @alternateType(unknown)
  thumbprint?: string;

  @alternateType(AzureLocation[])
  locations: string[];
}
```
So we could extend the type to "scalar, scalar[], unknown"

lirenhe:
Close it as resolved.

Issue #1757, opened 2024-10-28
[TCGC] Add @alternateType to change the type of a property (with scope)


lmazuel:
Example:
```typespec
model Bar {
  name: utcDateTime;
}

@@Tcgc.alternateType(Bar.name, string, "python");
```

Rationale is to be able to express in TypeScript autorest Swagger directive rules like this one:
```
- from: swagger-document
  where: $.definitions.AccessPolicy.properties
  transform: >
    $.Start.format = "str";
    $.Expiry.format = "str";
```

Side note, if I understood @timotheeguerin correctly, we should be able to make this work as well:
```typespec
scalar storageDateTime extends utcDataTime;
model Bar {
  name: storageDateTime;
}

@@Tcgc.alternateType(storageDateTime, string, "python");
```

which would avoid to replace all one by one.

I would expect this is done in a way that emitter do not need to change anything, since they will receive models with the new type auto-magically.

lmazuel:
I could imagine an operation type change indeed. 

It's not just format change in that scenario, it's literally that it can't be typed as a datetime in some language. So it's not encoding really, because it comes as a string on the wire, and we just want to keep it as a string unchanged.

As for other scenario, there are at least another one with an API in some AppInsight world were the type can be anything, and Python want to type it differently. I don't recall the exact case, it's somewhere on the repo, I can try some black magic Github query if you want to see it :)

tadelesh:
i got the idea. let me put it in nov sprint.

gracewilcox:
Adding to the convo that @richardpark-msft and I really need this for generating modules in Go with typespec.

tadelesh:
@gracewilcox could you elaborate more about your scenario

gracewilcox:
Hi @tadelesh! It'll be very useful for me to change the type of a field in a struct when generating Go SDK's. Right now, I'm working on converting a brown-field service to typespec, and I need to change the type of a field from `string` to my own custom defined type (`type ID string`). It would be nice to have a way to do this directly with typespec, at the moment I have to run my own post generation Go program to patch this.

PR #2486, opened 2025-04-02
[tcgc] extend decorator @alternateType supported types

haiyuazhang:
Extend `@alternateType`  to support `unknown` and `scalar array`: [issue2388](https://github.com/Azure/typespec-azure/issues/2388#event-16762415096)


website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/decorators.md
```diff
@@ -186,15 +188,25 @@ scalar storageDateTime extends utcDataTime;
...
 
+````typespec
+model Test {
+  @alternateType(unknown)
+  thumbprint?: string;
+
+  @alternateType(AzureLocation[], "csharp")
+  locations: string[];
+}
+} * ```
```
ArcturusZhang, line 199: here we have some format issues. Is this file generated? We should fix it.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,78 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
+      code: "invalid-alternate-type",
+      format: {},
+      target: alternate,
+    });
+    return;
+  }
+
+  let invalidSource = true;
+
+  if (alternate.kind === "Intrinsic" && alternate.name === "unknown") {
```
ArcturusZhang, line 991: similar here.

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/decorators.md
haiyuazhang: > here we have some format issues. Is this file generated? We should fix it.

Yes, it's generated.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,78 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
+      source.type.name === "Array" &&
+      source.type.indexer?.value.kind === "Scalar"
+    ) {
+      invalidSource = false;
+    }
+  }
+
+  if (invalidSource) {
     reportDiagnostic(context.program, {
       code: "invalid-alternate-source-type",
```
ArcturusZhang, line 1022: so do we care if the alternate type is not compatible with the type of the source? Such like I think in our current implementation, it would not fail if we do:
```
model Model {
	@alternateType(AzureLocation)
	prop: string[];
```
Is it a goal that we find such cases like this?

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/decorators.md
haiyuazhang: fixed the doc issue in decorators.tsp.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: we do have the check for it, if the alternate type is scalar then the source type must also be scalar.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,78 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
+  }
+
+  if (invalidAlternate) {
+    reportDiagnostic(context.program, {
+      code: "invalid-alternate-type",
+      format: {},
+      target: alternate,
+    });
+    return;
+  }
```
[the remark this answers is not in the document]
haiyuazhang: the "type" invalidation logic has been simplified.

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/decorators.md
```diff
@@ -186,15 +188,25 @@ scalar storageDateTime extends utcDataTime;
...
 ### `@apiVersion` {#@Azure.ClientGenerator.Core.apiVersion}
 
 Use to override default assumptions on whether a parameter is an api-version parameter or not.
 By default, we do matches with the `api-version` or `apiversion` string in the parameter name. Since api versions are
 a client parameter, we will also elevate this parameter up onto the client.
-
 ```typespec
 @Azure.ClientGenerator.Core.apiVersion(value?: valueof boolean, scope?: valueof string)
-```
+````
```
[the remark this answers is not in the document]
haiyuazhang: fixed.

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
```diff
@@ -561,11 +563,21 @@ export type ClientNamespaceDecorator = (
...
+ *   @alternateType(AzureLocation[], "csharp")
+ *   locations: string[];
+ * }
+ * ```
  */
 export type AlternateTypeDecorator = (
   context: DecoratorContext,
   source: ModelProperty | Scalar,
-  alternate: Scalar,
+  alternate: Type,
```
pshao25, line 580: Is this generated? If so, why a `Scalar | Model | unknown` will generate a `Type`?

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -572,8 +574,22 @@ extern dec clientNamespace(
...
+ *
+ *   @alternateType(AzureLocation[], "csharp")
+ *   locations: string[];
+ * }
+ * ```
  */
-extern dec alternateType(source: ModelProperty | Scalar, alternate: Scalar, scope?: valueof string);
+extern dec alternateType(
+  source: ModelProperty | Scalar,
+  alternate: Scalar | Model | unknown,
```
pshao25, line 590: Could you do a validation like 
```
@alternateType(int32)
someProperty: anyType;
```
will have compilation error?

I want to check if `int32` is compatible with `unknown`.

pshao25: @tadelesh Haiyuan tested if there is `unknown` here, this decorator could accept any type. Does that make sense?

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
haiyuazhang: It's generated part which we can't control, for the why part my bold guess is tcgc core think the latest base type for `Scalar | Model | unknown` is `Type`?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,71 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
+
+  if (invalidSource) {
     reportDiagnostic(context.program, {
       code: "invalid-alternate-source-type",
       format: {
-        typeName: source.type.kind,
+        typeName: source.kind,
```
pshao25, line 1017: Why we remove the type? If the source a model property, isn't the source.kind `property`?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,71 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
+      source.type.kind === "Model" &&
+      source.type.name === "Array" &&
+      source.type.indexer?.value.kind === "Scalar"
+    ) {
+      invalidSource = false;
+    }
+  }
+
+  // if alternate type is scalar, source type must be scalar
+  if (alternate.kind === "Scalar") {
```
pshao25, line 994: We can put this together with the above "if alternate type is unknown, source type must be scalar" ?

haiyuazhang: I think it's clearer and more readable to list three types of alternate types explicitly. 

> We can put this together with the above "if alternate type is unknown, source type must be scalar" ?

I think it's clearer and more readable to list three types of alternate types explicitly.

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
pshao25: I don't think we could put it aside. Could you try "if Scalar| Model what it procudes", "if unknown what it produces"?

packages/typespec-client-generator-core/lib/decorators.tsp
tadelesh: yes. `unknown` in ts means any types.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,71 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
   source: ModelProperty | Scalar,
-  alternate: Scalar,
+  alternate: Type,
   scope?: LanguageScopes,
 ) => {
-  if (source.kind === "ModelProperty" && source.type.kind !== "Scalar") {
+  let invalidAlternate = true;
+  let invalidSource = true;
+
+  // if alternate type is unknown, source type must be scalar
```
tadelesh, line 965: i'm wondering why this validation is needed? change to `unknown` could apply to any type in my opinion.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,19 +956,71 @@ const alternateTypeKey = createStateSymbol("alternateType");
...
+    invalidAlternate = false;
+    if (
+      (source.kind === "ModelProperty" && source.type.kind === "Scalar") ||
+      source.kind === "Scalar"
+    ) {
+      invalidSource = false;
+    }
+  }
+
+  // if alternate type is scalar array, source type must be scalar array
```
tadelesh, line 976: how about array of array, array of map? we do not support that for now? i don't like add too much validation for this decorator and let tsp author to take the responsibility.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: Per my understanding, the reason why we introduce the `unknown` type is to serialize some Scalar value into binary data. So I add the validation here to limit the usage of `unknown`. @pshao25 @ArcturusZhang @live1206  could you help to put your thoughts regarding whether we need this validation here?

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.ts
haiyuazhang: > I don't think we could put it aside. Could you try "if Scalar| Model what it procudes", "if unknown what it produces"?

`Scalar | Model` will produce `Scalar | Model` and single `unknown` will produce `Type`.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: > how about array of array, array of map? we do not support that for now? i don't like add too much validation for this decorator and let tsp author to take the responsibility.

array of array or array of map is not supported. @pshao25 , am I right here?

packages/typespec-client-generator-core/lib/decorators.tsp
haiyuazhang: @tadelesh @pshao25 ,  do we need to change the alternate type from `Scalar | Model | unknown` to `Type`?
My option is we can keep it as `Scalar | Model | unknown` although in generated type signature it will become `Type` any way. My point is that `Scalar | Model | unknown` has more type info which may prevent user from using the decorator incorrectly at the first place.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: `type` does not exist on `Scalar`.  If we want to check more "invalidate source type" cases, need to use `source.kind` instead of `source.type.kind` in the reportDiagnostic format.

packages/typespec-client-generator-core/lib/decorators.tsp
haiyuazhang: > @tadelesh @pshao25 , do we need to change the alternate type from `Scalar | Model | unknown` to `Type`? My option is we can keep it as `Scalar | Model | unknown` although in generated type signature it will become `Type` any way. My point is that `Scalar | Model | unknown` has more type info which may prevent user from using the decorator incorrectly at the first place.

offline synced with @tadelesh and @pshao25, it's ok to keep the alternat type as `Scalar | Model | unknown`.

packages/typespec-client-generator-core/src/decorators.ts
haiyuazhang: > Why we remove the type? If the source a model property, isn't the source.kind `property`?

make a improve here: if the source type is model property, using `source.type.kind` otherwise use `source.kind` as the source type name in the diagnostic format.

packages/typespec-client-generator-core/src/decorators.ts
live1206: The validation does look odd with all these limitations.
As offline discussed, we would like any/unknown for target of TypeProperty.
But for target of `Scalar`, we only allow `Scalar`.

haiyuazhang: @live1206 @tadelesh the unnecessary type validation logic has been removed, please help to review again.

.chronus/changes/tcgc-extend-supportedTypes-for-decorator-alternateType-2025-4-2-14-29-57.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+add support for `unknown` and `scalar array` in decorator `@alternateType`
```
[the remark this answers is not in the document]
haiyuazhang: updated.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -956,16 +956,14 @@ const alternateTypeKey = createStateSymbol("alternateType");
 export const $alternateType: AlternateTypeDecorator = (
   context: DecoratorContext,
   source: ModelProperty | Scalar,
-  alternate: Scalar,
+  alternate: Type,
   scope?: LanguageScopes,
 ) => {
-  if (source.kind === "ModelProperty" && source.type.kind !== "Scalar") {
+  if (source.kind === "Scalar" && alternate.kind !== "Scalar") {
```
tadelesh, line 962: so, after design change, we only check the scalar now?

haiyuazhang: > so, after design change, we only check the scalar now?

Yes.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -315,13 +315,7 @@ export const $lib = createTypeSpecLibrary({
     "invalid-alternate-source-type": {
```
tadelesh, line 315:
```suggestion
    "invalid-alternate-type": {
```

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: emmm, actually, i prefer to check if the alternate type kind is same with original type. then model to model, scalar to scalar, etc.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -315,13 +315,7 @@ export const $lib = createTypeSpecLibrary({
...
       severity: "error",
       messages: {
-        default: paramMessage`@alternateType only supports scalar types. The source type is '${"typeName"}'.`,
-      },
-    },
-    "invalid-alternate-type": {
-      severity: "error",
-      messages: {
-        default: paramMessage`@alternateType only supports scalar types, scalar array or unknown as the alternate type.`,
+        default: paramMessage`Can't apply none-Scalar type alternate to Scalar type source. The alternate type kind is '${"kindName"}'`,
```
tadelesh, line 318: can you help to rephase the word? it is not easy to understand.

packages/typespec-client-generator-core/src/decorators.ts
live1206: @tadelesh 
Based on a [user requirement](https://github.com/Azure/typespec-azure/issues/1757#issuecomment-2465623277) of the original issue, the user would like to alternate the type from string to a custom type.
For `TypeProperty`, I think we can allow the user to alternate the type as they wantas garbage-in-garbage-out. Would it cause any side-effect as you can see?

packages/typespec-client-generator-core/test/decorators/alternate-type.test.ts
```diff
@@ -160,6 +160,102 @@ describe.each([
   });
 });
 
+it.each([
+  ["string[]", "utcDateTime[]"],
+  ["int64[]", "utcDateTime[]"],
+  ["duration[]", "string[]"],
+])(
```
pshao25, line 167: Please make sure this work: original type is union and alternate type is string, like
```
model A {  
  property: SomeUnion;
}

union SomeUnion {
  string,
  SCSI: "SCSI",
  NVMe: "NVMe",
}
@@alternateType(A.property,
  string
);
```

tadelesh (approved):

packages/typespec-client-generator-core/test/decorators/alternate-type.test.ts
haiyuazhang: > union SomeUnion {
>   string,
>   SCSI: "SCSI",
>   NVMe: "NVMe",
> }

when the source is model property, alternate can be any type.

packages/typespec-client-generator-core/src/lib.ts
haiyuazhang: the diagnostic message has been rephased.

PR #2496, opened 2025-04-04
[spector] add tests for moving param from method to client

### LINKED ISSUE #2453: [spector] add test for `@clientInitialization`

iscai-msft:
fixes #2453 
fixes https://github.com/Azure/cadl-ranch/issues/766


lmazuel:
Also fix https://github.com/Azure/cadl-ranch/issues/766 :)

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
```diff
@@ -0,0 +1,185 @@
...
+  """)
+@scenario
+@global.Azure.ClientGenerator.Core.client({
+  name: "ParamAliasClient",
+  service: Service,
+})
+interface ParamAlias {
+  withAliasedName is Service.ParamAlias.withAliasedName;
+  withOriginalName is Service.ParamAlias.withOriginalName;
+}
```
weidongxu-microsoft, line 185: There is no `clientInitialization` on this API?

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/main.tsp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+model CommonHeaders {
+  @header("client-name")
+  clientName: string;
+
+  @header("correlation-id")
+  correlationId: string;
+}
+
```
tadelesh, line 36:
```suggestion
// Scenario 1: Header parameter moved to client level

```
nit

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
```diff
@@ -0,0 +1,185 @@
...
+  ```ts
+  const client = new HeaderParamClient({
+    name: "test-name-value"
+  });
+  
+  client.withQuery(id: "some-id");  // No need to pass name here
+  client.withBody({ name: "test-name" });  // No need to pass name here
+  ```
+  """)
+@scenario
```
tadelesh, line 55: can we add scenario doc for each operation? otherwise, it will only be one line in spector dashboard.

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
```diff
@@ -0,0 +1,185 @@
...
+  // Elevated to client level via original name
+  client.withOriginalName();
+  ```
+  """)
+@scenario
+@global.Azure.ClientGenerator.Core.client({
+  name: "ParamAliasClient",
+  service: Service,
+})
+interface ParamAlias {
```
tadelesh, line 182:
```suggestion
@global.Azure.ClientGenerator.Core.clientInitialization(ParamAliasClientOptions)
interface ParamAlias {
```

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/mockapi.ts
```diff
@@ -0,0 +1,202 @@
...
+    },
+    kind: "MockApiDefinition",
+  },
+  {
+    uri: "/azure/client-generator-core/client-initialization/param-alias/sample-blob/with-original-name",
+    method: "get",
+    request: {},
+    response: {
+      status: 200,
+      body: json("Successfully used aliased parameters"),
```
msyyc, line 198:
```suggestion
      body: json("Still used original parameters"),
```

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
iscai-msft: good catch, tahnks!

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
iscai-msft: That's what I was going for, I think it's fair to count being able to do a call with a query and with a body to mean "i can support moving a header parameter onto the client". If you feel strongly, I can change it.

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/main.tsp
```diff
@@ -0,0 +1,120 @@
...
+  name: string;
+}
+
+// Scenario 1: Header parameter moved to client level
+
+@route("/header-param")
+interface HeaderParam {
+  @route("/with-query")
+  @get
+  withQuery(...NameHeaderParam, @query id: string): string;
```
weidongxu-microsoft, line 35: For any of these `string` in response, if we intend it as JSON string (appears so in mockapi), we need to write a `@header contentType: "application/json"` now.
(found out when I try generate the code...)

Otherwise, they be interpretated as `application/text`.

see https://typespec.io/docs/release-notes/release-2025-03-18/

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
tadelesh: idk if spector framework will mark this scenario with green iff all the inner operation pass. if so, i'm ok with it.

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/main.tsp
```diff
@@ -0,0 +1,120 @@
...
+
+@doc("Test for client initialization decorator - moving parameters from method to client level")
+@scenarioService("/azure/client-generator-core/client-initialization")
+namespace Service;
+
+// Common parameter types and models
+
+model NameHeaderParam {
+  @header("name")
+  name: string;
```
msyyc, line 17: Curious why there is no change in spec-summary.md after new cases added.
https://github.com/Azure/typespec-azure/blob/c2d118aa8ac206a25537839fe8cf0fa9f422c40f/packages/azure-http-specs/spec-summary.md

tadelesh: i think doc is relay on `@scenario`. and izzy put it on interface level, not operation level.

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
iscai-msft: It will only be green iff all the inner operations pass

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/main.tsp
iscai-msft: Thanks for this catch @weidongxu-microsoft! Let me just switch it to `text/plain` responses

weidongxu-microsoft (approved):

PR #2510, opened 2025-04-09
[tcgc] fix clientNamespace bug when used earlier in namespace path

### LINKED ISSUE #2500: [Bug]: TCGC, clientNamespace on TypeSpec Namespace does not override "namespace" emitter option
### Describe the bug

In some cases, the `@clientNamespace` decorator on Namespace is not able to override `namespace` emitter options.

@iscai-msft 

### Reproduction

TCGC 0.54.0

Run Java emitter on
https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/client/namespace/client.tsp
with `namespace=client.noeffect` emitter option

(note that below would not show up in generated code, as Java emitter is WIP switching behavior, when both option and decorator provided)

The client.tsp have `@clientNamespace` override on all TypeSpec Namespaces.
Here is the override `@@clientNamespace(FirstModel, "client.clientnamespace.first", "java");`

However, from TCGC output, the client namespace of `FirstClientResult` model (it is under `FirstModel` Namespace) is `client.noeffect`
Image

---

If `namespace=client.noeffect` emitter option is not provided, `@clientNamespace` works fine. Here is a screenshot
Image

### Checklist

iscai-msft:
fixes #2500

### DIFF
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -1174,21 +1174,17 @@ export const $clientNamespace: ClientNamespaceDecorator = (
  * @param userDefinedNamespaces
  * @returns
  */
-function findShortestNamespaceOverlap(
+function findNamespaceOverlapClosestToRoot(
   override: string,
   userDefinedNamespaces: Namespace[],
 ): Namespace | undefined {
-  let shortestNamespace: Namespace | undefined = undefined;
-
   for (const namespace of userDefinedNamespaces) {
     if (override.includes(namespace.name)) {
-      if (!shortestNamespace || namespace.name.length < shortestNamespace.name.length) {
-        shortestNamespace = namespace;
-      }
+      return namespace;
     }
   }
 
-  return shortestNamespace;
+  return undefined;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1207,43 +1207,47 @@ export function getClientNamespace(
   entity: Namespace | Interface | Model | Enum | Union,
 ): string {
   const override = getScopedDecoratorData(context, clientNamespaceKey, entity);
-  if (override) {
+  const { namespace, isOverriddenInNamespace } = getNamespaceFullNameWithOverride(context, entity);
+  if (override || isOverriddenInNamespace) {
+    const overriddenNamespace = override || namespace;
```
weidongxu-microsoft, line 1212: I think the idea is correct, that it checks whether the model is contained in a Namespace that has `@clientNamespace`.

I am not sure whether the response of `namespace` var is correct. For
```ts
@clientNamespace("client.namespace")
namespace Client {
  namespace Sub {
    model Model {
       ...
    }
  }
}
```
It seems the func would return `client.namespace.Sub` as namespace for `Model`?

I am unsure whether `Sub` should be included in the result. (I guess this may be an ambiguity from the design)

Other than this ambiguity, PR looks good.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: we should generate it as `client.namesapce.Sub` I believe, i still think client namesapce should not flatten everything within its namespace

Issue #2495, opened 2025-04-04
[Bug]: optional path params aren't modeled as optional


jhendrixMSFT:
### Describe the bug

Now that https://github.com/microsoft/typespec/issues/4126 has landed, we can describe optional path params. For these cases `param.optional` should now be `true` (it's still `false` at present).

### Reproduction

tcgc version `0.54.0`
create tsp with an optional path parameter

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2520, opened 2025-04-10
[tcgc] fix optionality for path and body

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2495

for normal case, path and body parameter's optionality will follow typespec definition.
for spread body, only when all the properties are optional, the body is optional.


weidongxu-microsoft (approved):
code looks good.

PR #2523, opened 2025-04-10
[tcgc] add `@clientDoc` decorator

### LINKED ISSUE #2490: [TCGC] Add @clientDoc decorator
Quite simple, when used replace the doc of the element it's applied to.

Value is that this one has a scope, and therefore can be changed for a given language.

Note, this replaces https://github.com/Azure/typespec-azure/issues/2127

iscai-msft:
fixes #2490


iscai-msft:
I don't htink we need any specs tests bc as long as tcgc does the right thing here, all emitters should be fine. There's also not a great testserver way of testing this, since it is ultimately client-level doc generation

iscai-msft:
currently we're just adding for `doc`. I think if we get an ask for `@clientSummary` we can add it. @srnagar what are your thoughts?

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -663,3 +663,57 @@ extern dec deserializeEmptyStringAsNull(target: ModelProperty, scope?: valueof s
...
 extern dec responseAsBool(target: Operation, scope?: valueof string);
+
+/**
+ * Defines how client documentation should be applied
+ */
+enum DocumentationMode {
+  /**
+   * Append client documentation to the existing doc
+   */
+  append: "append",
```
srnagar, line 674: To address @tadelesh's comment, we can either introduce a new decorator `@clientSummary` or the other option is to use `appendToDoc`, `replaceDoc` values and then we can add `appendToSummary` and `replaceSummary` later if needed.

Also, `@summary` is used in 48 tsp files while `@doc` is used in 700+ tsp files in the rest-api-spec repo. `summary` is not used a lot so, we can add support for it when needed.

website/src/content/docs/docs/howtos/Generate client libraries/07types.mdx
```diff
@@ -166,6 +166,259 @@ Foo foo = new Foo("hello");
...
+</ClientTabs>
+
+The `@clientDoc` decorator can be used on:
+
+- Models
+- Model properties
+- Operations
+- Operation parameters
+- Enums
+- Enum members
```
srnagar, line 417: We should also make it available for anything that can have a `@doc` decorator. This includes `Namespace`, `Interface`, `Union` etc. I think the decorator's target is `unknown` which implies it works for all of the above. So, we should need to update this documentation.

srnagar (approved):

packages/typespec-client-generator-core/lib/decorators.tsp
iscai-msft: i think it would be better to add a `@clientSummary` decorator instead if we went down that route for greater symmetry, so I think we can keep the enum options the same for now

PR #2530, opened 2025-04-14
[azure-http-specs] fix client initialization spec problem

tadelesh:
1. wrong doc
2. wrong route

### DIFF
--- a/packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
@@ -1,9 +1,12 @@
 import "./main.tsp";
 import "@azure-tools/typespec-client-generator-core";
 import "@typespec/spector";
+import "@typespec/http";
 
 using Spector;
+using Http;
 
+@route("/azure/client-generator-core/client-initialization")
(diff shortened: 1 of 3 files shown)


packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
```diff
@@ -1,9 +1,12 @@
 import "./main.tsp";
 import "@azure-tools/typespec-client-generator-core";
 import "@typespec/spector";
+import "@typespec/http";
 
 using Spector;
+using Http;
 
+@route("/azure/client-generator-core/client-initialization")
```
iscai-msft, line 9: i'm curious why we need to update the route in the `client.tsp` as well? I'll go ahead and approve but am just curious

iscai-msft (approved):

iscai-msft:
i'm curious why we need to update the route in the client.tsp as well? I'll go ahead and approve but am just curious

tadelesh:
> i'm curious why we need to update the route in the client.tsp as well? I'll go ahead and approve but am just curious

autoroute combine the route for operation and its container (interface/namespace) till root. since we changed the container, the route changed.

PR #2535, opened 2025-04-14
[spector] add test for decorator deserializeEmptyStringAsNull.

haiyuazhang:
add spec test for decorator `@deserializeEmptyStringAsNull`

### DIFF
--- a/packages/azure-http-specs/specs/azure/client-generator-core/deserialize-empty-string-as-null/main.tsp
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/deserialize-empty-string-as-null/main.tsp
@@ -13,7 +13,7 @@ namespace _Specs_.Azure.ClientGenerator.Core.DeserializeEmptyStringAsNull;
 @doc("This is a Model contains a string-like property of type url.")
 model ResponseModel {
   @deserializeEmptyStringAsNull
-  serviceUrl: url;
+  sampleUrl: url;
 }
 
 @scenario
@@ -26,6 +26,6 @@ model ResponseModel {
   }
   ```
   """)
-@route("/emptyUrl")
+@route("/responseModel")
 @get
-op emptyUrl(): ResponseModel;
+op get(): ResponseModel;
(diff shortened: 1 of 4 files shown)


packages/azure-http-specs/specs/azure/client-generator-core/deserialize-empty-string-as-null/main.tsp
```diff
@@ -0,0 +1,31 @@
...
+  Expected response body:
+  ```json
+  {
+    "serviceUrl": ""
+  }
+  ```
+  """)
+@route("/emptyUrl")
+@get
+op emptyUrl(): ResponseModel;
```
iscai-msft, line 31: nit: I don't think `emptyUrl` is the best name, it implies that the url is empty. You can just call it `op get` or something like that

iscai-msft: I also think it's best not to name the property `serviceUrl`, I don't think we often see urls in returned bodies. If this is based off of a real service example, feel free to ignore my comment.

iscai-msft (approved):

packages/azure-http-specs/specs/azure/client-generator-core/deserialize-empty-string-as-null/main.tsp
```diff
@@ -0,0 +1,31 @@
...
+using global.Azure.ClientGenerator.Core;
+using Spector;
+
+@doc("Test decorator @deserializeEmptyStringAsNull.")
+@scenarioService("/azure/client-generator-core/deserialize-empty-string-as-null")
+namespace _Specs_.Azure.ClientGenerator.Core.DeserializeEmptyStringAsNull;
+
+@doc("This is a Model contains a string-like property of type url.")
+model ResponseModel {
+  @deserializeEmptyStringAsNull
```
MaryGao, line 15: What is the scope for this decorator? Do we expect all languages support this? If not, can we document this also?

Issue #2081, opened 2025-01-15
linter arm-resource-operation shouldn't be reported on non-resource operation


pshao25:
[...]

PR #2536, opened 2025-04-14
Fix arm-resource-operation rule to no get trigger for ArmProviderActionAsync and ArmProviderActionSync

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/2081


packages/typespec-azure-resource-manager/src/rules/core-operations.ts
```diff
@@ -31,7 +31,8 @@ export const coreOperationsRule = createRule({
       operation: (operation: Operation) => {
         if (
           !isInternalTypeSpec(context.program, operation) &&
-          !isSourceOperationResourceManagerInternal(operation)
+          !isSourceOperationResourceManagerInternal(operation) &&
+          !isArmProviderActionOperation(operation)
```
markcowl, line 35: I don't think we want to use this check here, for example, the `opOutsideInterface` check should still apply - this should narrowly prevent usage of the arm-resource-operation rule.

Also, I don't think we want to make exceptiosn to the rule this way - this should be about allowing provider actions,  meaning a static segment *after* the provider name in the operation route.  operations outside this, no matter what template they use should still be flagged.

packages/typespec-azure-resource-manager/test/rules/core-operations.test.ts
```diff
@@ -102,6 +102,55 @@ describe("typespec-azure-resource-manager: core operations rule", () => {
...
+      string,
+      @doc(".") Succeeded: "Succeeded",
+      @doc(".") Failed: "Failed",
+      @doc(".") Canceled: "Canceled",
+    }
+
+    op ComputeProviderActionAsync<
+      Request extends {} | void = void,
+      Response extends {} | void = void
+    > is ArmProviderActionAsync<Request, Response>;
```
markcowl, line 138: Also, I would include both valid and invalid provider actions in the tests.

markcowl (changes requested):

markcowl (approved):
This looks good - we should have a test PR against typespec-next (public and private) to verify

AlitzelMendez:
## Validation
- [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/actions/runs/15002420189/job/42152563643?pr=34644)
- [azure-rest-api-specs-pr](https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/15002888725/job/42154150408?pr=22656)

PR #2537, opened 2025-04-14
[specs] update python client namespace

iscai-msft:
cc @msyyc @tadelesh . We can piggyback onto what java has done, this way we don't have to set the `namespace` for all of our azure generations. This will have a big change in the autorest.python repo, since I think we can also get rid of the `specs.` prefix from the namespaces

### DIFF
--- a/packages/azure-http-specs/specs/azure/client-generator-core/access/main.tsp
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/access/main.tsp
@@ -8,7 +8,10 @@ using Spector;
 
 @doc("Test for internal decorator.")
 @scenarioService("/azure/client-generator-core/access")
-@global.Azure.ClientGenerator.Core.clientNamespace("azure.clientgenerator.core.access", "java")
+@global.Azure.ClientGenerator.Core.clientNamespace(
+  "azure.clientgenerator.core.access",
+  "java,python"
+)
 namespace _Specs_.Azure.ClientGenerator.Core.Access;
 
 @route("/publicOperation")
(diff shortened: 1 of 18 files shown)


packages/azure-http-specs/specs/azure/client-generator-core/access/main.tsp
```diff
@@ -8,7 +8,10 @@ using Spector;
 
 @doc("Test for internal decorator.")
 @scenarioService("/azure/client-generator-core/access")
-@global.Azure.ClientGenerator.Core.clientNamespace("azure.clientgenerator.core.access", "java")
+@global.Azure.ClientGenerator.Core.clientNamespace(
+  "azure.clientgenerator.core.access",
```
msyyc, line 12: NOTE: The namespace is different with we set in python emitter https://github.com/microsoft/typespec/blob/aac438be10c1a16eda3988a2a03b50a530da019d/packages/http-client-python/eng/scripts/ci/regenerate.ts#L41.

So after this PR merged, we need to update test cases in [microsoft/typespec](https://github.com/microsoft/typespec/blob/aac438be10c1a16eda3988a2a03b50a530da019d/packages/http-client-python/generator/test/azure/mock_api_tests/test_azure_client_generator_core_access.py#L7-L8) and [autorest.python](https://github.com/Azure/autorest.python/blob/312cf7867e4d9ae5820425b706cb4c283bb7b165/packages/typespec-python/test/azure/mock_api_tests/test_azure_client_generator_core_access.py#L7-L8).

msyyc: @iscai-msft Maybe we could define namespace same with current python emitter like:
@global.Azure.ClientGenerator.Core.clientNamespace(
  "specs.azure.clientgenerator.core.access",
  "python"
)

tadelesh (approved):

PR #2538, opened 2025-04-14
Fix `@identifiers` decorator, if decorator is passed it has a priority over what is resolved by keys

AlitzelMendez:
- Fixing gaps on `@identifiers` decorator


AlitzelMendez:
##Validation 
- azure-rest-api-specs: https://github.com/Azure/azure-rest-api-specs/pull/33973

packages/typespec-autorest/src/openapi.ts
```diff
@@ -2414,6 +2432,7 @@ export async function getOpenAPIForService(
     typespecType: Model,
     context: SchemaContext,
     namespace?: Namespace,
+    property?: ModelProperty,
```
timotheeguerin, line 2435: Feels like pasing the property around is a bit messy. Can't we just override whatever is set on `x-ms-identifiers` when resolving the property if it has `@identifiers` decorator set

packages/typespec-autorest/test/openapi-output.test.ts
```diff
@@ -1031,4 +1038,90 @@ describe("identifiers decorator", () => {
...
+        name: string;
+      }
+      model PetList {
+        @identifiers(#["name"])
+        value: Pet[];
+      }
+      `,
+    );
+
+    expectDiagnostics(ignoreDiagnostics(diagnostics, ["@typespec/http/no-service-found"]), [
```
markcowl, line 1121: Why is this important?  Why do we want to prevent usage of this decorator outside ARM?

markcowl (commented):
Not sure I understand why we limit the decorator usage to ARM namespaces.  Otherwise, looks ok

packages/typespec-autorest/test/openapi-output.test.ts
AlitzelMendez: we had a conversation some months ago where we decided this. I think it was because of something happening with the data-plane specs, but there was probably an underlying issue. Let me check that. I will be happy to remove this and make it not limited to ARM namespaces.

AlitzelMendez: ok confirmed that this is not needed, we can use it outside ARM, I did azure-rest-api-specs validation and everything looks good :)

AlitzelMendez: https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/14538571923/job/40791853430?pr=22342

markcowl (approved):

PR #2542, opened 2025-04-16
azure-http-specs, client initialization with child client

weidongxu-microsoft:
tested locally
```
var result = new ParentClientBuilder().buildClient()
  .getChildClient("sample-blob")
  .getStandalone();

new ChildClientBuilder().blobName("sample-blob").buildClient()
  .deleteStandalone();
```

### DIFF
--- a/packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
@@ -200,42 +200,44 @@ interface ParamAlias {
   withOriginalName is Service.ParamAlias.withOriginalName;
 }
 
-@scenarioDoc("""
-  Client for testing a path parameter (blobName) moved to client level, in child client.
-  
-  Parameters elevated to client level:
-  - blobName: "sample-blob" (path parameter)
-  
-  Expected client usage:
-  ```ts
-  // via ParentClient
-  const client = new ParentClient.getChildClient({
-    blobName: "sample-blob"
-  });
-  
-  // directly
-  const client = new ChildChild({
-    blobName: "sample-blob"
-  });
-  
-  // No need to pass blobName to any operations
-  client.withQuery(format: "text");  
-  client.getStandalone();
-  client.deleteStandalone();
-  ```
-  """)
-@scenario
 @global.Azure.ClientGenerator.Core.client({
   name: "ParentClient",
   service: Service,
 })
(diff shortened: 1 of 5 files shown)


iscai-msft:
@weidongxu-microsoft this looks good, but I also know there's controversy around hierarchy and client initialization, and it's not something we need to support for this quarter. Is there agreement on the side of the team in china in adding a test for this?

packages/azure-http-specs/specs/azure/client-generator-core/client-initialization/client.tsp
```diff
@@ -199,3 +199,48 @@ interface ParamAlias {
...
+    
+    Expected client usage:
+    ```ts
+    // via ParentClient
+    const client = new ParentClient.getChildClient({
+      blobName: "sample-blob"
+    });
+    
+    // directly
+    const client = new ChildChild({
```
anuchandy, line 224: wondering the reason for the name `ChildChild`, given the interface name used below is `ChildClient`

weidongxu-microsoft: Thanks, fixed the typo.

weidongxu-microsoft:
> @weidongxu-microsoft this looks good, but I also know there's controversy around hierarchy and client initialization, and it's not something we need to support for this quarter. Is there agreement on the side of the team in china in adding a test for this?

Face AI is using this exact pattern (just without the `intializedBy`) https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/Face/client.tsp#L121-L124

And I remember Laurent asked me to move [the old cadl-ranch test to specter](https://github.com/Azure/cadl-ranch/pull/727) last week, hence this PR :-)

I did discuss with Chenjie about whether we put this to a separate folder, so if emitter have problem of supporting this, they can skip this feature.
Let me check with languages.

ArcturusZhang (approved):
LGTM!

anuchandy (approved):

msyyc:
@weidongxu-microsoft Tried locally and find `ParentClient` lacks client level signature `blobName`. Is Java OK?
image

weidongxu-microsoft:
ParentClient does not have the client option, so it is without the `blob_name`.

It is the ChildChild has this in client param.

msyyc (approved):
Although Python does not work for hierarchy, we won't block this PR merge.

tadelesh (approved):

PR #2550, opened 2025-04-17
Use instantiated Typekits instead of default typekit

chrisradek:
Depends on https://github.com/microsoft/typespec/pull/7018/


timotheeguerin (approved):
quite simple this one

timotheeguerin (approved):

PR #2555, opened 2025-04-18
[specs] absolute url next link

iscai-msft:


packages/azure-http-specs/specs/azure/core/page/mockapi.ts
```diff
@@ -65,7 +65,8 @@ Scenarios.Azure_Core_Page_withParameterizedNextLink = passOnSuccess([
       body: json({
         value: [{ id: 1, name: "User1" }],
         // Include the original query parameters in the nextLink
-        nextLink: "/azure/core/page/with-parameterized-next-link/second-page?select=name",
+        nextLink:
+          "http://localhost:3000/azure/core/page/with-parameterized-next-link/second-page?select=name",
```
jhendrixMSFT, line 69: I see other tests use `${req.baseUrl}` for the host instead of a hard-coded `http://localhost:3000`

timotheeguerin: yeah I think this might break other uses which use custom ports

weidongxu-microsoft: Agree. .NET would use different port.

weidongxu-microsoft (approved):

tadelesh (approved):

packages/azure-http-specs/specs/azure/core/page/mockapi.ts
```diff
@@ -71,10 +71,6 @@ Scenarios.Azure_Core_Page_withParameterizedNextLink = passOnSuccess([
     handler: (req: MockRequest) => {
       return {
         status: 200,
-        body: json({
-          value: [{ id: 1, name: "User1" }],
-          nextLink: `${req.baseUrl}/azure/core/page/with-parameterized-next-link/second-page?select=${req.query.select}`,
-        }),
       };
     },
```
weidongxu-microsoft, line 75: @tadelesh Why remove the response? I assume you want to remove the nextLink in response, not handler?

PR #2560, opened 2025-04-21
Support pageable decorator in nested properties

allenjzhang:
Depends on core PR: https://github.com/microsoft/typespec/pull/7067
close: https://github.com/microsoft/typespec/issues/5208

### DIFF
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -40,6 +40,7 @@ import {
   NumericLiteral,
   Operation,
   PagingOperation,
+  PagingProperty,
   Program,
   Scalar,
   StringLiteral,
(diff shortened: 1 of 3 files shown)


packages/typespec-autorest/src/openapi.ts
```diff
@@ -531,11 +532,15 @@ export async function getOpenAPIForService(
     }
   }
 
+  function resolvePagableProperty(paging: PagingProperty): string {
+    return paging.path.length > 1 ? paging.path.map((p) => p.name).join(".") : paging.property.name;
```
timotheeguerin, line 536: shoudl we log a warning if the property already contains a `.`

packages/typespec-autorest/test/paging.test.ts
```diff
@@ -38,6 +38,28 @@ it("define a custom paged operation with custom next link", async () => {
...
+      };
+    }
+    @list op list(): CustomPageModel<{}>;
+    `,
+  );
+
+  const listThings = res.paths["/"].get;
+  ok(listThings);
+  deepStrictEqual(listThings["x-ms-pageable"], {
+    nextLinkName: "pagination.@odata.nextLink",
```
timotheeguerin, line 58: is that actually the correct path if the property contains a `.`? How does autorest differentiate

allenjzhang: @tadelesh Can you comment on the emitter '.' Handling behavior?

tadelesh: tcgc export the path with property array to prevent `.` ambiguous in string path.

allenjzhang: @tadelesh, it is unclear that `x-ms-pageable` supports nested property references AND TypeSpec autorest emitter had code explicitly take the last segment to output to the `x-ms-pageable` fields, I will close this PR and be done with the original issue. The full path is available on `PageProperty.path` if DPG/MPG decides to make use of it.

Closing this PR.

Issue #2559, opened 2025-04-21
[Bug][tcgc]: `null` value is not enriched in example


MaryGao:
With latest tcgc the `null` value is not enriched in [JS emitter case](https://github.com/v-jiaodi/autorest.typescript/blob/monitor-nightly43/packages/typespec-ts/test/modularUnit/scenarios/samples/parameters/parameterTypesCheck.md#should-generate-samples-for-different-types).

```
  nullValue: null;
```

The raw value has that value:

![Image](https://github.com/user-attachments/assets/3c4bab6b-773a-4bc5-b8c8-e64bc8fd1eb0)

But the example doesn't have that value:

![Image](https://github.com/user-attachments/assets/0d8e8f65-3d0f-4ec9-b97d-949c2e82c75e)

tadelesh:
could you provide the spec?

Issue #1758, opened 2024-10-28
[Client emitter] Generate a metadata file while emitting


lmazuel:
We want to save some metadata in a file along with the emitted code, that can be used for some automation purposes. For instance, @weshaggard would want to use it to build a mapping of package version with support api-version

# Proposal

Create a file `metadata.json` that contains:
```json
{
  "api-version": "2015-06-01"
}
```

This file is not generated by default, and is enabled if the flag `emit-metadata: true` is passed in the tspconfig.yaml.

Each emitter will need to do that code, as TCGC should not be responsible to write files on disk. We could have though TCGC building the dict for us, with a `getMetadataInfo()` that emitters just have to put on disk. This way, this is consistent cross-language.

msyyc:
I think the proposal makes sense. About the filename, since SDK users usually don't need care about it, we could mark it with `_` to declare it is inner file so maybe `_metadata.json` is better.

tadelesh:
though tcgc should be a lib depended by language's emitter, it has the ability to emit the metadata file if needed.

lmazuel:
@srnagar is suggesting we're merging that with the current `apiview_properties.json` file, and I feel it makes sense to avoid having too much files

tadelesh:
> [@srnagar](https://github.com/srnagar) is suggesting we're merging that with the current `apiview_properties.json` file, and I feel it makes sense to avoid having too much files

agree. emitter could use `client.apiVersions[-1]` to get the api version. @iscai-msft do you think we need to add more explicit property in `SdkClientType` to indicate the api version client generated from?

iscai-msft:
we have `getDefaultApiVersion` right now, I feel that's the best way to do it. We could add a `defaultApiVersion` property, but I think we're good for now

tadelesh:
`getDefaultApiVersion` need the raw typespec namespace. i don't think it is a good way to get default.

iscai-msft:
Hmm that's a very good point, yes we can go with `client.apiVersions[-1]` for now / add another property, but I think we can go without adding the property for now

tadelesh:
TCGC will export `apiVersion` in `SdkPacakge`.

MaryGao:
I reopened this issue to discuss some decisions cross languages and @iscai-msft  let me know if you have some concern!

MaryGao:
@weshaggard I need your input regarding this work. Two things we'd like to confirm with you:

[...]

PR #2562, opened 2025-04-22
[tcgc] add api version info to package and fix example issue

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/1758
fix: https://github.com/Azure/typespec-azure/issues/2559


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -65,10 +65,12 @@ export interface TCGCContext {
   __httpOperationExamples: Map<HttpOperation, SdkHttpOperationExample[]>;
   __pagedResultSet: Set<SdkType>;
   __mutatedGlobalNamespace?: Namespace; // the root of all tsp namespaces for this instance. Starting point for traversal, so we don't call mutation multiple times
+  __packageVersions?: string[]; // the package versions from the service versioning config and api version setting in tspconfig.
 
   getMutatedGlobalNamespace(): Namespace;
   getApiVersionsForType(type: Type): string[];
   setApiVersionsForType(type: Type, apiVersions: string[]): void;
+  getPackageVersions(): string[];
```
weidongxu-microsoft, line 73: Does it include those from `@clientApiVersions` and the filtering of `versioning -> previewStringRegex`?

PS: Should we have some docstring on the interface? This should be the most important interface of the TCGC, but there is zero doc.

tadelesh: currently, `@clientApiVersions` and `previewStringRegex` only impact api version enum. i'm not sure if we should take it into account for all type's api version list. @iscai-msft for idea. also, idk if we should export `getPackageVersions` helper to emitter since it is only for inner usage now.

weidongxu-microsoft: Got it.

I don't have strong opinion on it. Just want to make sure we know that the values in it be (hence the comment on doc).

So, generally we just pick the last of the items, as the expected api-version.

packages/typespec-client-generator-core/test/examples/types.test.ts
```diff
@@ -1000,6 +1000,43 @@ it("SdkModelExample with extra paramters", async () => {
...
+it("unknown type with null example value", async () => {
+  await runner.host.addRealTypeSpecFile(
+    "./examples/getUnknownNull.json",
+    `${__dirname}/example-types/getUnknownNull.json`,
+  );
+  await runner.compile(`
+    @service
+    namespace TestClient {
+      model Test {
+        prop: unknown;
```
MaryGao, line 1012: Could we add a property for null also?

```
foo: null;
```

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -869,6 +871,13 @@ export interface SdkPackage<TServiceOperation extends SdkServiceOperation> {
   crossLanguagePackageId: string;
   namespaces: SdkNamespace<TServiceOperation>[];
   licenseInfo?: LicenseInfo;
+  /**
+   * The version of the package.
+   * If undefined, the package is not versioned.
+   * If `all`, the package is versioned with all versions.
+   * If a string, the package is versioned with the specified version.
+   */
+  apiVersion?: string;
```
weidongxu-microsoft, line 880: I am not sure if `all` is useful here.

I'd assume even if we set `api-version=all` in config, the `api-version` we print out it may still be the `api-version` for the APIs. This depends on the intended use of the output.
- if the output is used for `regen` the SDK, e.g. if it integrated with `tsp-client update`, then `all` is what was expected
- if the output is used for indexing `api-version` -> `sdk version` map, the `all` would be useless

weidongxu-microsoft (approved):
Got one discussion on `all`, but it would depend on what's the expected usage of this `api-version` var.

Another thing of the handling in TCGC about `api-version` is that much of the logic assumes Azure and single api-version.
Unbranded can write [different api-version on different service/namespace](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=CmltcG9ydCAiQHR5cGVzcGVjL3Jlc3QiO8oZYXp1cmUtdG9vbHMvyCUtxhVjb3JlIjsKCnVzaW5nIFR5cGVTcGVjLkh0dHA70BVSZXN00RVWZXJzaW9uaW5nyBtBxFkuQ29yZdISLkZvdW5kYXRpb25z0x5UcmFpdHM7CgpAc2VydmljZSgjeyB0aXRsZTogIk11bHRpcGxlQXBpx3UiIH0pCkB2xg1lZChUc3BUZXN0LtIq6ACncykKbmFtZXNwYWNlINovIHsKICBAcmVzb3VyY2UoIsgKcyIpCiAgbW9kZWwgUscUxSwgIEB2aXNpYmlsaXR5KExpZmVjeWNsZS5SZWFkxDMgIGlkOiBzdHLlASjGMWtled86xjrkALHPPOQBxcoTICB9xBdpbnRlcmbkANnoAJtPcGVy5gFsxTZleHRlbmRz7AFx0io8Tm9SZXBlYXRhYmxlUmVxdWVzdHMgJsVBxAFOb0NvbmRpxCxhbNYgbGllbnTHG0lkPiB75QCfZW51bSDoAYfoATV1c2VEZXBlbmRlbmN5KGdsb2JhbC7rAKDIMC52MV8wX1ByZXZpZXdfMuYBGXYyMDIyXzA2XzAxX3DGGzogIsQVLTA2LTAxLccVIizmATbGMDEy0zAxMs4w5wFTQHJvdXRlKCIvY%2BUA1DHlAf%2FqAWjmAOrnAMxnZXQgaXPzAXzpAWXkAdM8yA0%2B5gG4ff8C7OkCkzL%2FAu3MK%2F8C7tEw%2FwLv7QLvxi3%2FArb%2FAvDeOv8C8P8C8P8C8P8C8P8C8NYg%2FwLw%2FwLw%2FwLw8QLwMTD0AsAw%2FwLw%2FwLw9gLwMv8C8P8C8O0C8DL%2FAvHxApYz%2FwLxzCv%2FAvHRMOUAw%2FQBof8BY%2B4BM%2F8BYv8BYuUBTDP8AWIvLyBubyBhcGkt5wEFIHF1ZXJ5IHBhcmFtZXRlciBpbiB0aGUgQVBJxTFAcG9zdMUKYWPkAYsoQGFkZGVkKOoCVvIAsCkgQMtWMT%2FoA2cpOiBPa1Jlc3BvbnNl6AG7&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D).
This is not something we need to solve in this PR, but just feel this `sdkContext.apiVersion` is really only the `sdkContext.azureFlavorApiVersion`.

packages/typespec-client-generator-core/test/examples/types.test.ts
tadelesh: it is tested in another test case.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -869,6 +871,15 @@ export interface SdkPackage<TServiceOperation extends SdkServiceOperation> {
   crossLanguagePackageId: string;
   namespaces: SdkNamespace<TServiceOperation>[];
   licenseInfo?: LicenseInfo;
+  metadata: {
+    /**
+     * The version of the package.
+     * If undefined, the package is not versioned.
+     * If `all`, the package is versioned with all versions.
```
MaryGao, line 878: With our current assumption for Azure, should the value only be single api-version? So `all` may not happen in reality?

Issue #1163, opened 2024-07-15
A model unioned with null introduces inheritance


pshao25:
This is what we found from the [unmerged pipeline](https://github.com/Azure/typespec-azure/pull/1099). We found this [change](https://github.com/Azure/azure-rest-api-specs/pull/29698/files#diff-81adef30b508d8dcb6b3878969b13a2cb45b5f790f7f55cb928b4ce68f1e4c7bL746-L751). The problem comes from this [ManagedServiceIdentity.userAssignedIdentities](https://github.com/Azure/typespec-azure/blob/f8ee0803166022ec66cc982222d1090aab10cacb/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp#L27) whose type is `Record<UserAssignedIdentity | null>`.
```
model ManagedServiceIdentity {
  /** The service principal ID of the system assigned identity. This property will only be provided for a system assigned identity. */
  @visibility("read")
  principalId?: uuid;

  /** The tenant ID of the system assigned identity. This property will only be provided for a system assigned identity. */
  @visibility("read")
  tenantId?: uuid;

  /** The type of managed identity assigned to this resource. */
  type: ManagedServiceIdentityType;

  /** The identities assigned to this resource by the user. */
  @typeChangedFrom(Versions.v5, Record<UserAssignedIdentity>)
  userAssignedIdentities?: Record<UserAssignedIdentity | null>;
}
```

See this [playground](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=bW9kZWwgQSB7CiAgbmFtZTogc3RyaW5nOwp9CgrGHULFHXByb3BlcnR5MTogQSB8IG51bGw7yxcyOiBSZWNvcmQ8yB4%2BzB8zyh8%2BOwp9&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-core%2Fall%22%5D%7D%7D), as long as a model is "unioned" with a `null`, its type is changed to an `allOf`. See the difference between `property2` and `property3` in `B`. This is not accurate,

MaryGao:
[...]

PR #2576, opened 2025-04-23
Modify how `x-nullable` is resolved next to `$ref` 

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1163#

### DIFF
--- a/packages/typespec-autorest/test/models.test.ts
+++ b/packages/typespec-autorest/test/models.test.ts
@@ -704,6 +704,27 @@ describe("typespec-autorest: model definitions", () => {
         type: "object",
       });
     });
+
+    it.only("defines nullable Array", async () => {
+      const res = await openApiFor(
+        `
+        model Pet {
+          name: string;
+        }
+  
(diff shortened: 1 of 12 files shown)


packages/typespec-autorest/test/models.test.ts
```diff
@@ -665,6 +665,45 @@ describe("typespec-autorest: model definitions", () => {
...
+        }
+        `,
+      );
+      deepStrictEqual(res.definitions.Dog.properties.type, {
+        $ref: "#/definitions/Pet",
+        "x-nullable": true,
+      });
+    });
+
+    it("defines nullable record", async () => {
```
markcowl, line 687: Should we have a test for array types as well (if we don't already)?

`arrayProp: Array<Pet | null>;`

markcowl (commented):
Looks good - another suggested test.

markcowl (approved):

timotheeguerin (approved):

PR #2591, opened 2025-04-25
[tcgc] fix source property comparison

iscai-msft:


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -634,6 +634,15 @@ export function findRootSourceProperty(property: ModelProperty): ModelProperty {
   return property;
 }
 
+export function compareRootSourceProperties(prop1: ModelProperty, prop2: ModelProperty): boolean {
+  const rootProp1 = findRootSourceProperty(prop1);
+  const rootProp2 = findRootSourceProperty(prop2);
+  if (rootProp1.node !== undefined && rootProp2.node !== undefined) {
+    return rootProp1.node !== undefined && rootProp1.node === rootProp2.node;
```
tadelesh, line 641: i have little concern about why we could not compare the root source property directly. do you have any idea about this?

tadelesh (commented):
i think we could just merge to main because we do not have any other ongoing work in main. we could just release hotfix from main directly.

packages/typespec-client-generator-core/src/internal-utils.ts
iscai-msft: I still don't fully understand the reasoning here over why this is happening, but this was the cause of `parameterizedNextLinkSegments` returning a list of a single value of undefined. The nodes of the root properties were the same, and so were the nodes of the root property's model type as well. But the two models weren't defined as being the same, which caused the original check to fail

iscai-msft: I'm just going to go ahead and merge this and release, and do a backmerge since I've already set up the release branch

Issue #2582, opened 2025-04-24
[tcgc] add initializedBy none flag in clientInitialization


qiaozha:
should we add InitializedBy.None to indicate an operation group which differentiate InitializedBy.Parent to indicate a sub client that could only initialized by parent client?

qiaozha:
Notes of dpg office hour meeting in 2025-04-24: this proposal is accepted.

PR #2592, opened 2025-04-27
[TCGC] Change default value of initializedBy for SdkOperationGroup from `InitializedByFlags.Parent` to `InitializedByFlags.Default`

msyyc:
fix https://github.com/Azure/typespec-azure/issues/2582

As discussed in dev office hours meeting, TCGC will add a new enum value `InitializedBy.none` to indicate an operation group which differentiate `InitializedBy.Parent` to indicate a sub client that could only initialized by parent client.

This change is breaking since default value of `initializedBy` for operationGroup will be changed from `InitializedBy.Parent` to `InitializedBy.None`.
Add language owners for awaress: @qiaozha / @iscai-msft / @JoshLove-msft / @tadelesh


packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -446,6 +446,11 @@ enum InitializedBy {
    * The client could be initialized by parent client.
    */
   parent: 2,
+
+  /**
+   * The client can't be initialized.
+   */
+  none: 4,
```
qiaozha, line 453: I just think of a question, can we mark the initializedBy as both none and (parent or individually)? I feel like none and other values are mutual exclusive?

msyyc: Yeah, none could only used independently and here is logic to check: https://github.com/Azure/typespec-azure/blob/90b771eef9f262150e111c80e9c1f5e1e900aaa0/packages/typespec-client-generator-core/src/clients.ts#L323-L333

tadelesh (approved):

tadelesh:
please do not merge this pr before the typespec ga

packages/typespec-client-generator-core/lib/decorators.tsp
qiaozha: I wonder if we can put None as 0, because if it's 0, then 
```
None | Parent = Parent
None | Individually = Individually
```
in this way we don't need to do extra checks maybe?

weidongxu-microsoft: Wondering whether `default` or `not_specified` be a better name.

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -489,7 +494,7 @@ model ClientInitializationOptions {
   parameters?: Model;
 
   /**
-   * Determines how the client could be initialized. Use `InitializedBy` enum to set the value. The value could be `InitializedBy.individually`, `InitializedBy.parent` or `InitializedBy.individually | InitializedBy.parent`.
+   * Determines how the client could be initialized. Use `InitializedBy` enum to set the value. The value could be `InitializedBy.individually`, `InitializedBy.parent`, `InitializedBy.none` or `InitializedBy.individually | InitializedBy.parent`.
```
weidongxu-microsoft, line 497: Do we expect user write `InitializedBy.none`?

Or it is just TCGC output that means user didn't config it?

tadelesh: user could not write `InitializedBy.none | InitializedBy.parent` which will cause diagnostic error.

packages/typespec-client-generator-core/test/clients/structure.test.ts
```diff
@@ -668,7 +686,7 @@ it("operationGroup", async () => {
   strictEqual(mainClient.crossLanguageDefinitionId, "TestService");
 
   strictEqual(operationGroup.clientInitialization.parameters.length, 1);
-  strictEqual(operationGroup.clientInitialization.initializedBy, InitializedByFlags.Parent);
+  strictEqual(operationGroup.clientInitialization.initializedBy, InitializedByFlags.None);
```
iscai-msft, line 689: to clarify, the default behavior is now a subclient by default will have initialized by none, and you have to specify `.parent` or `.individually` if you want to override? What will be the impact of this

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/data-types.md
```diff
@@ -57,6 +57,7 @@ enum Azure.ClientGenerator.Core.InitializedBy
 | ------------ | ----- | ------------------------------------------------- |
 | individually | `1`   | The client could be initialized individually.     |
 | parent       | `2`   | The client could be initialized by parent client. |
+| none         | `4`   | The client can't be initialized.                  |
```
JoshLove-msft, line 60: What does it mean that a client can't be initialized?

packages/typespec-client-generator-core/lib/decorators.tsp
msyyc: > I wonder if we can put None as 0, because if it's 0, then
> 
> ```
> None | Parent = Parent
> None | Individually = Individually
> ```
> 
> in this way we don't need to do extra checks maybe?

I am afraid the logic can't apply for all statements: https://github.com/Azure/typespec-azure/blob/90b771eef9f262150e111c80e9c1f5e1e900aaa0/packages/typespec-client-generator-core/src/clients.ts#L294-L337

packages/typespec-client-generator-core/test/clients/structure.test.ts
msyyc: Yeah. The impact depends on how language emitter use the value.
I take a rough look and Python emitter seems not use `InitializedByFlags.Parent`, so there shall be no impact for Python.

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/data-types.md
msyyc: There is concept "operation group" for Python/JS. Operation group can't be initialized independently and can only be accessed by its parent.

For example, class [Operations](https://github.com/Azure/azure-sdk-for-python/blob/21d8b165afb9497dafcce8702beb6bf5f4c1d881/sdk/trustedsigning/azure-mgmt-trustedsigning/azure/mgmt/trustedsigning/_client.py#L86) is operation group, you can use it by `client.operations.xxx` but it can't be initialized independently by user like `operations = Operations(...)`.

packages/typespec-client-generator-core/test/clients/structure.test.ts
msyyc: As claimed [here](https://github.com/Azure/typespec-azure/blob/1a164a48c6542788d9a9663438306e5703c129dd/packages/typespec-client-generator-core/src/decorators.ts#L472): `if there is no explicit client, we will treat non-client namespaces and all interfaces as operation group`. 

The default initializedBy for operation group is `none`. If users want to declare namespace or interface as sub client, they could override it in `client.tsp`.

As for impact, it depends on how language emitters use the value. I take a rough look, python emitter doesn't use `Parent` so there shall be no impact for python.

packages/typespec-client-generator-core/lib/decorators.tsp
msyyc: > Wondering whether `default` or `not_specified` be a better name that describes this config.

Yeah, `default` may be more general and not that misleading.

msyyc:
Thanks all your reviews @qiaozha / @iscai-msft / @JoshLove-msft / @tadelesh / @weidongxu-microsoft. According to your feedbacks, I update the PR with:
- Add enum value `Default` instead of `None` to avoid misleading. `Default` means `typespec doesn't set initialization way specifically`. The default value of initializeBy for non-client namespaces/interfaces change from `Parent` to `Default` so language emitters shall update if you consume `Parent` before. As for what `Default` means, it depends your language design and understanding.
- Not expose `Default` in decorator. This enum value will be internal between TCGC and langauge emitters to avoid confusion for customers.
- Since `Default` is only internal enum value, we don't need extra logic to check whether users' initialization configuration is right.

packages/typespec-client-generator-core/test/clients/structure.test.ts
msyyc: see https://github.com/Azure/typespec-azure/pull/2592#issuecomment-2837498699.
I take a rough look for python emitter and Python doesn't consume `Parent` before so there is no impact for Python. But in the future, python needs to distinguish the `Parent` and `Default` when design client hierarchy.

tadelesh (approved):

weidongxu-microsoft (approved):

qiaozha (approved):

PR #2600, opened 2025-04-29
eng, feeds, arm, update tspconfig template

weidongxu-microsoft:
The `service-name` in Java is recommended to be uppercase with whitespace.

related to https://github.com/Azure/azure-rest-api-specs/pull/34114

### DIFF
--- a/eng/feeds/arm/tspconfig.yaml
+++ b/eng/feeds/arm/tspconfig.yaml
@@ -19,7 +19,7 @@ options:
   "@azure-tools/typespec-java":
     package-dir: "azure-resourcemanager-{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
     namespace: "com.azure.resourcemanager.{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
-    service-name: "{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
+    service-name: "{{#toUpperCase}}{{parameters.ServiceNamespace}}{{/toUpperCase}}"
   "@azure-tools/typespec-ts":
     package-dir: "arm-{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
     is-modular-library: true


eng/feeds/arm/tspconfig.yaml
```diff
@@ -19,7 +19,7 @@ options:
   "@azure-tools/typespec-java":
     package-dir: "azure-resourcemanager-{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
     namespace: "com.azure.resourcemanager.{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
-    service-name: "{{#toLowerCase}}{{parameters.ServiceNamespace}}{{/toLowerCase}}"
+    service-name: "{{#toUpperCase}}{{parameters.ServiceNamespace}}{{/toUpperCase}}"
```
raych1, line 22: Is service-name supposed to be Pascal case or upper case?

weidongxu-microsoft: Pascal case.

Thanks. `#toUpperCase` would be wrong. How do I write "convert to Pascal case" here? I am still not sure what's the syntax in this file.

raych1: It's already with 'Pascal case'.  You could close this PR.
https://github.com/Azure/typespec-azure/blob/99af20a73bc3eff309ec54478d6da86f4e00ed6f/eng/feeds/azure-scaffolding.json#L240

weidongxu-microsoft: Unsure whether this is related.

ARM Resource Provider Name could be a same `Microsoft.Workloads` in all folders under https://github.com/Azure/azure-rest-api-specs/tree/main/specification/workloads, but each folder would generate different SDK and different service-name.

raych1: This scaffolding template only supports a RP with single service scenario.

markcowl:
Per the above conversation, I think this should be closed.  Please re-open or let me know if this is not the case.

PR #2607, opened 2025-04-30
Fix identifiers every array

AlitzelMendez:


AlitzelMendez:
## Candidate pull requests
- [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/pull/34659) / I had to do some changes on the couple of typespec files to avoid changing the swagger on non-data-plane specs

azure-test-api-specs-pr is in progress

packages/samples/common-types/openapi/v3/types.json
```diff
@@ -174,8 +174,7 @@
           "items": {
             "$ref": "#/definitions/ErrorAdditionalInfo"
           },
-          "readOnly": true,
-          "x-ms-identifiers": []
+          "readOnly": true
```
markcowl, line 177: How will this impact common-types generation?  Do we need to add the `@identifiers` decorator to this type?

packages/samples/common-types/openapi/v6/types.json
```diff
@@ -174,8 +174,7 @@
           "items": {
             "$ref": "#/definitions/ErrorAdditionalInfo"
           },
-          "readOnly": true,
-          "x-ms-identifiers": []
+          "readOnly": true
```
markcowl, line 177: same question for all of these common-types generated swaggers

markcowl (approved):

PR #2609, opened 2025-04-30
azure-http-specs, fix dyn in handler

weidongxu-microsoft:
@timotheeguerin

I've got 500 on this API, in nightly build (against unreleased azure-http-specs lib -- released one is good)
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4814852&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=1fb2445c-3c92-5f9b-bcad-710222d02bc3&l=274

So far, this appears to be the only place `dyn/dynItem` is used in handler. My guess is that this be a mistake.

Let me know if this is the not the correct fix.

### DIFF
--- a/.chronus/changes/azure-http-specs_fix-dyn-2025-3-30-18-28-19.md
+++ b/.chronus/changes/azure-http-specs_fix-dyn-2025-3-30-18-28-19.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/azure-http-specs"
+---
+
+Fix handler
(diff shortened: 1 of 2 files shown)


weidongxu-microsoft:
Something wrong. I cannot reproduce on local typespec-azure build.

Also, the 500 is on the POST, may not be related to the header in response.

timotheeguerin (approved):
Yeah was a mistake, not sure why the e2e call didn't catch it.

Th is just means those apis are not being called and verified correctly by the mock client

PR #2617, opened 2025-05-01
add copilot instructions

iscai-msft:


.github/copilot-instructions.md
timotheeguerin, line 1: Should this go in teh http-specs folder?

iscai-msft: nah i think all copilot instructions are supposed to go under `.github/` so the agents can find it. There can be different sections too

timotheeguerin (approved):

Issue #2595, opened 2025-04-28
[Bug]: tcgc generate different types for the same typespec dict type


qiaozha:
### Describe the bug

As title. a minimal repro tsp would be 
```
model A {
    propA: Record<A[]>;
}
model B {
   propB: Record<A[]>;
}
```
In this case, the raw typespec type is the same for Record<A[]>, the tcgc type name, kind is also the same, but tcgc return different types for model A's propA and model B's propB.
This is originally from Azure Search typespec https://github.com/Azure/azure-rest-api-specs/tree/search_tsp_2025_03_preview_multi_ns/specification/search/Azure.Search 

![Image](https://github.com/user-attachments/assets/fc14cb12-e466-4f24-8c90-99a8f4b8614d) 

This is problematic for us because in JS emitter, we have been using tcgc SDKType as reference management and we will only generate one serializer and deserializer for SDKType that will output the same serializer/deserializer name. Hence it causes us could not resolve the reference that we didn't generate because of duplication.

### Reproduction

N/A

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2626, opened 2025-05-06
[tcgc] convert to same type for array or dictionary of same type

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2595

### DIFF
--- a/.chronus/changes/tcgc-cache_array_map-2025-4-6-18-37-49.md
+++ b/.chronus/changes/tcgc-cache_array_map-2025-4-6-18-37-49.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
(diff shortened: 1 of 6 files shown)


qiaozha (approved):

weidongxu-microsoft (approved):
Should be fine for Java.

Does it also affect the generated name? E.g. previous the 2 `A[]` would be 2 different type and having different generated name (or tcgc didn't generate name for array?).

tadelesh:
> Should be fine for Java.
> 
> Does it also affect the generated name? E.g. previous the 2 `A[]` would be 2 different type and having different generated name (or tcgc didn't generate name for array?).

tcgc does not generate name for array or dictionary.

PR #2630, opened 2025-05-06
Adding billing-data common type.

wilcobmsft:
Adding a new billing-data common type. This provides standardization of billing properties in e.g. prepaid resource types. Examples of resource types expected to use this schema include:
- Microsoft.BillingBenefits/*
- Microsoft.SaaS/resources
- Microsoft.SaaSHub/cloudServices
- Microsoft.EnterpriseSupport/enterpriseSupports
- Oracle.Databases/accounts
- ...

This list will grow over time. RP Platform is enabling an integrated "(prepaid/commitment) billing" building block which utilizes this schema, meaning any RP which uses this building block is expected to use this schema.


markcowl (commented):
Can we provide some documentation on how to use this (either a sample using this data, or a section in the ARM getting started, or, preferrably, both).

Also, is this only related to the namespaces mentioned, or is it likely more RPs will want to use this feature for general billing needs?

wilcobmsft:
> Can we provide some documentation on how to use this (either a sample using this data, or a section in the ARM getting started, or, preferrably, both).
> 
> Also, is this only related to the namespaces mentioned, or is it likely more RPs will want to use this feature for general billing needs?

Updated how-to with an example. Once we land the RP Platform building block that utilizes this pattern also, we will include an example there as well.

packages/typespec-azure-resource-manager/lib/common-types/common-types.tsp
```diff
@@ -13,5 +13,5 @@ import "./versions.tsp";
 import "./mobo-ref.tsp";
 import "./network-security-perimeter-ref.tsp";
 import "./nsp-operations.tsp";
-
+import "./billing-data-ref.tsp";
 namespace Azure.ResourceManager.CommonTypes;
```
timotheeguerin, line 17:
```suggestion

namespace Azure.ResourceManager.CommonTypes;
```
nit

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -452,4 +452,22 @@ model AvailabilityZonesProperty {
...
+ *
+ * ```typespec
+ * model FooProperties {
+ *   ...BillingDataProperty,
+ * }
+ * ```
+ */
+@doc("Contains a billing data property to be spread into resource property types")
+model BillingDataProperty {
+  @doc("The billing data of the resource.")
```
timotheeguerin, line 469:
```suggestion
  /** The billing data of the resource. */
```

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -452,4 +452,22 @@ model AvailabilityZonesProperty {
...
+ *
+ * @example
+ *
+ * ```typespec
+ * model FooProperties {
+ *   ...BillingDataProperty,
+ * }
+ * ```
+ */
+@doc("Contains a billing data property to be spread into resource property types")
```
timotheeguerin, line 467:
```suggestion
```
don't need that one as its only meant to be used in spread

packages/typespec-azure-resource-manager/CHANGELOG.md
timotheeguerin, line 1: revert

packages/samples/specs/resource-manager/common-types/v6/privatelinks.json
timotheeguerin, line 1: I think those files are in the wrong folder 

Image

also we should include the exact copy in `existing` folder which should be synced by the common types script

timotheeguerin: I don't actually see billingdata.json in the spec repo common types yet, was the goal to add them in TypeSpec first and sync the swagger to spec repo after?

markcowl (commented):
The conversion of active specs to swagger means that we may not want to maintain swagger common-types going for new types going forward.  This means those types would be in the typespec repo, and the existing common-types swagger would remain frozen.

I don't expect we'll have a resolution on this until sometime in June.  What we *could* do is remove any connection to swagger common-types in these types.

markcowl:
This was taken up in another pr and merged.  closing

PR #2632, opened 2025-05-06
Common types - description feedback

AlitzelMendez:
I am getting more feedback on the common-types migration pr, specifically around descriptions, some were wrong, and others are unnecessarily impacting SDK team

### DIFF
--- a/.chronus/changes/common-types-latest-feedback-2025-4-6-12-54-34.md
+++ b/.chronus/changes/common-types-latest-feedback-2025-4-6-12-54-34.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Fixing common-types descriptions that are not accurate to the original Swagger specifications.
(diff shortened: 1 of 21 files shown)


markcowl (commented):
Do we have an accompanying PR for typespec-next?

This looks good, but comments are not breaking changes, nor are they reason to re-release SDKs.  IMO, we should err on the side of correctness of the language versus exact match

markcowl (commented):
Do we have an accompanying PR for typespec-next?

It looks like you need to regenerate samples, othherwise, this loo0ks fine, but comments are not breaking changes, nor are they reason to re-release SDKs.  IMO, we should err on the side of correctness of the language versus exact match

AlitzelMendez:
> Do we have an accompanying PR for typespec-next?
> 
> It looks like you need to regenerate samples, othherwise, this loo0ks fine, but comments are not breaking changes, nor are they reason to re-release SDKs. IMO, we should err on the side of correctness of the language versus exact match

I agree, it has been so push back from SDK team on not changing descriptions, but I am still in discussion about this, so I will keep this on draft until we can agree if this is really necessary

PR #2633, opened 2025-05-06
Bump versions for 0.56(Core 1.0.0) release

timotheeguerin:


chrisradek (approved):
CI failure (git diff) looks like it will be resolved when core is fully updated

Issue #2656, opened 2025-05-12
[Bug]: [tcgc] Missing base model mapping value for polymorphism in example generation


MaryGao:
### Describe the bug

![Image](https://github.com/user-attachments/assets/0b2bc46d-17c4-45d2-98d9-aa22a1279710)

missing https://github.com/Azure/azure-rest-api-specs/blob/main/specification/recoveryservicesdatareplication/DataReplication.Management/examples/2024-09-01/Fabric_Create.json#L13-L15 in tcgc example parameters
### Reproduction

NA

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2658, opened 2025-05-12
[tcgc] add fallback logic for example mapping of discriminated type

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2656


MaryGao (approved):
LGTM

Issue #2657, opened 2025-05-12
ErrorAdditionalInfo.info should be unknown


pshao25:
**Actual:**
https://github.com/Azure/typespec-azure/blob/0c7fe2a5caaf2958231459d7e7e41249d6f6d5f1/packages/typespec-azure-resource-manager/lib/common-types/types.tsp#L301

In original swagger, it is https://github.com/Azure/azure-rest-api-specs/blob/5802c95f18bfba1003be50e545d07f8bb679c857/specification/common-types/resource-management/v5/types.json#L76-L80

**Expected:**
As we agreed in https://github.com/Azure/typespec-azure/issues/1725#issuecomment-2448645928, it should be type of `unknown`.

PR #2669, opened 2025-05-13
ErrorAdditionalInfo.info should be unknown 

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/2657


markcowl:
This would seem to set us back on common types, based on this: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/types.json#L76-L80

Based on the common-types type, why did we decide this was `unknown` rather than `Record<unkinown>` ?  Since this is an 'object' type, it must be keyed and cannot have many types allowed by unknown (like string)

markcowl (commented):
This would seem to set us back on common types, based on this: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/types.json#L76-L80

Based on the common-types type, why did we decide this was unknown rather than Record<unkinown> ? Since this is an 'object' type, it must be keyed and cannot have many types allowed by unknown (like string)

AlitzelMendez:
> This would seem to set us back on common types, based on this: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/types.json#L76-L80
> 
> Based on the common-types type, why did we decide this was unknown rather than Record ? Since this is an 'object' type, it must be keyed and cannot have many types allowed by unknown (like string)


Having ` Record<unknown>` is going to add `additionalProperties`:
```diff
  "info": {
      "type": "object",
      "description": "The additional info.",
+     "additionalProperties": {},
      "readOnly": true
    }
```

I prefer not having any of these changes on the common-types swagger

AlitzelMendez:
I did a quick validation just in case: [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/actions/runs/15009012257/job/42174111037?pr=34663)

AlitzelMendez:
closing in favor of https://github.com/Azure/typespec-azure/pull/2686

markcowl:
Per discussion, we should:
- Verify if this type should be narrowed to `Record<unknown>`
- If not, use `unknown` and see if the ARM folks will agree that the swagger should be disambiguated
- In the last case, do post-processing to match the common types

packages/typespec-azure-resource-manager/lib/common-types/types.tsp
```diff
@@ -298,7 +298,7 @@ model ErrorAdditionalInfo {
 
   /** The additional info. */
   @visibility(Lifecycle.Read)
-  info?: {};
+  info?: Record<unknown>;
```
pshao25, line 301: It's breaking change to SDK. Why not just keep our previous decision?

timotheeguerin: I think its a breaking change in the SDK eitherway. JS generates that currently as `Record<string, unknown>` because they did respect the `type: object`

Regardless of that I do agree that I think this is supposed to be `unknown` anyway because there is a `type` right next to it that woudl seem to indicate the type of this property. However this is not super clear hence the comment from mark looking into more investigation

AlitzelMendez:
@markcowl / @timotheeguerin, I spoke with Johan, and his response was:

"It is not constrained to scalar or JSON object, so it should be considered unknown."

I've also been trying to get a response from Roopesh regarding whether this should be a simple type or a keyed type. I haven’t received an answer yet, but I’ll continue to follow up with him. I understand this is significantly impacting the migration effort, so I think Johan’s answer is quite valid.

Any thoughts on this?

AlitzelMendez:
I had a conversation with Wendy Chang (Roopesh redirected me to her) and she mentioned that:
`AdditionalInfo.Info is modeled as an object so I think that will cause the system to throw an error if the type doesn't match`
https://msazure.visualstudio.com/One/_git/AzureUX-AZSaaS?path=/src/common/errorresponses/data/typederrorinfo.cs&version=GBmaster&line=25&lineEnd=25&lineStartColumn=20&lineEndColumn=24&lineStyle=plain&_a=contents

markcowl (approved):

markcowl:
Per discussion, we have pretty good evidence that there are sometimes simple types like string in this collection, so this must be represented as 'unknown' in TypeSpec.  I think the common-types change is unfortunate, and we may be able to convince the ARM team to disambiguate in the future, but post-processing is a fine expedient for now.

AlitzelMendez:
closing in favor of: https://github.com/Azure/typespec-azure/pull/2724

Issue #1374, opened 2024-08-14
Singleton resource do not work with separate client namespace


timotheeguerin:
See the openapi3 emitter for client then route use `{employeeName}` instead of `default`

[Playground Link](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3yvIK3Jlc291cmNlLW1hbmFnZXIiOwoKdXNpbmcgVHlwZVNwZWMuSHR0cDvQFVJlc3TRFVbpAKnIG0HEZS5Db3JlzhJSx3dNxnY7CgovKiogQ29udG9zbyDIHiBQcm92aWRlciDmAJ5tZW50IEFQSS4gKi8KQGFybcggTmFtZXNwYWNlCkBzZXJ2acQJdXNlRGVwZW5kZW5jeSjVd%2BgAsXMudjFfMF9QcmV2aWV3XzEpxV1Db21tb27kAN1zxyjXRsspy1I1KQpu6ACXIFPmAJYgewogIOQA6kHoAOzoAMJIdWIg6AGM5ADeICBA5AEvbGV0b24KICBtb2RlbCBFbXBsb3llZSBpcyBUcmFja2Vk6ACMPHt9PsVeICAuLukAouQBGFBhcmFtZXRlcjzIPz47CiAgfQrEZGFybcgsT3BlcmF05AC8CiAgaW50ZXJm5AC6yDVzx15nZXTEf0HKOVJlYWTQXH0K6gGK6gEGQ2xp5AHCxEnfYXPoATIuyR8uZ2V0yF0%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%5D%7D%7D)

The problem is that we set this route resolution option on the namespace which is then not resolved in the client namespace.

weidongxu-microsoft:
Currently, this is the last remaining blocker for SDK team to generate SDK directly from TypeSpec, for a relatively large brownfield ([avs](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/vmware/Microsoft.AVS.Management)) that already finished converting to TypeSpec (and verified the result via Swagger and TypeSpec).

weidongxu-microsoft:
[...]

PR #2672, opened 2025-05-13
[arm lib] fix singleton resource autoroute issue

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/1374

### DIFF
--- a/packages/typespec-azure-resource-manager/src/namespace.ts
+++ b/packages/typespec-azure-resource-manager/src/namespace.ts
@@ -232,8 +232,12 @@ export const $armProviderNamespace: ArmProviderNamespaceDecorator = (
       });
     }
 
-    // Set route options for the whole namespace
-    setRouteOptionsForNamespace(program, program.getGlobalNamespaceType(), {
+    // Set route options for the top level namespace
+    let topLevelNamespace = entity;
+    while (topLevelNamespace.namespace) {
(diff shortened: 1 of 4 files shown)


packages/typespec-azure-resource-manager/src/namespace.ts
```diff
@@ -233,7 +233,7 @@ export const $armProviderNamespace: ArmProviderNamespaceDecorator = (
     }
 
     // Set route options for the whole namespace
-    setRouteOptionsForNamespace(program, entity, {
+    setRouteOptionsForNamespace(program, program.getGlobalNamespaceType(), {
```
timotheeguerin, line 236: that most likely will cause issue with versioning but if it works(intresting hack) then probably just traversing up the entity namespace till the top will work

tadelesh: it could work with setting to top level namespace.

weidongxu-microsoft (approved):

tadelesh:
@markcowl @timotheeguerin could you help to confirm if this fix is ok to you? it blocked one of our release.

packages/typespec-azure-resource-manager/src/namespace.ts
weidongxu-microsoft: @timotheeguerin Would this change look good for you?

timotheeguerin (approved):

PR #2677, opened 2025-05-14
[azure-http-specs] add test for mgmt-plane large LRO header

XiaofeiCao:
## Context
POST/DELETE LRO now requires that both `Location` and `AAO` headers are returned in response.

For each header value, [`Authorization`](https://eng.ms/docs/products/arm/api_contracts/asyncoperationsigningandvalidation) info will be appended at the end, making the URL much longer.

Thus, [doc](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations#url-to-monitor-status) requires that REST client should accept **minimum** header size of 4KB.

## This PR
Added a POST LRO which will return both `Location` and `AAO` headers with 6KB+ size each, 12KB+ size in total.

## To pass the test
Ensure your HttpClient accepts both
1. 7KB single header size
2. 13KB total header size

### DIFF
--- a/packages/azure-http-specs/specs/azure/resource-manager/large-header/mockapi.ts
+++ b/packages/azure-http-specs/specs/azure/resource-manager/large-header/mockapi.ts
@@ -12,7 +12,7 @@ export const Scenarios: Record<string, ScenarioMockApi> = {};
 
 const SUBSCRIPTION_ID_EXPECTED = "00000000-0000-0000-0000-000000000000";
 const RESOURCE_GROUP_EXPECTED = "test-rg";
-const SIX_KB_STRING = randomString(1024 * 6);
+const SIX_KB_STRING = "a".repeat(1024 * 6);
 let pollCount = 0;
 
 Scenarios.Azure_ResourceManager_LargeHeader_LargeHeaders_two6k = passOnSuccess([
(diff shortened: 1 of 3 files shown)


packages/azure-http-specs/specs/azure/resource-manager/large-header/mockapi.ts
```diff
@@ -0,0 +1,126 @@
...
+  passOnSuccess,
+  ScenarioMockApi,
+  ValidationError,
+} from "@typespec/spec-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+const SUBSCRIPTION_ID_EXPECTED = "00000000-0000-0000-0000-000000000000";
+const RESOURCE_GROUP_EXPECTED = "test-rg";
+const SIX_KB_STRING = randomString(1024 * 6);
```
weidongxu-microsoft, line 15: I am not sure whether randomness is required here. A char repeated 6k time should be good enough?

XiaofeiCao: Yeah, that should suffice. Let me update.

weidongxu-microsoft (approved):

PR #2680, opened 2025-05-14
Use new tester in autorest tests

timotheeguerin:
Speeds up the test significantly


packages/typespec-autorest/test/arm/arm.test.ts
```diff
@@ -1,9 +1,9 @@
 import { deepStrictEqual, ok, strictEqual } from "assert";
 import { it } from "vitest";
-import { openApiFor } from "../test-host.js";
+import { compileOpenAPI } from "../test-host.js";
 
 it("can share types with a library namespace", async () => {
-  const openapi = await openApiFor(
+  const openapi: any = await compileOpenAPI(
```
chrisradek, line 6: Nit: Is this cast to `any` needed? Looks like you use optional chaining in your tests anyway.

chrisradek (approved):

packages/typespec-autorest/test/arm/arm.test.ts
timotheeguerin: just went with the least modification possible here as it was getting tedious, when accessing params I would have had to add extra check or casting down the line anyway as it could also be a $ref

Issue #2655, opened 2025-05-09
[Bug]: @identifiers failing to resolve @key as expected when there is multi-level keys


AlitzelMendez:
### Describe the bug

The scenario only happens when there is a multi-level key usage

### Reproduction

[...]

PR #2691, opened 2025-05-16
Fix `@key` decorator resolution for `x-ms-identifiers

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/2655

## Validation
- https://github.com/Azure/azure-rest-api-specs-pr/pull/22761
- https://github.com/Azure/azure-rest-api-specs/pull/34789


packages/typespec-autorest/test/openapi-output.test.ts
```diff
@@ -823,6 +823,52 @@ describe("identifiers decorator", () => {
...
+        value: Bar[]
+      }
+      model Foo
+      {
+        @key
+        name:string;
+      }
+      `,
+    );
+    deepStrictEqual(oapi.definitions.BarList.properties.value["x-ms-identifiers"], []);
```
markcowl, line 846: This is not ignoring the values.  Ignoring the values would be not having an x-ms-identifiers extension.  If we want no identifiers in this case, we should use the decorstor override.

PR #2700, opened 2025-05-19
Add troubleshooting document

pshao25:
As titled.

### DIFF
--- a/website/src/content/docs/docs/migrate-swagger/01-get-started.md
+++ b/website/src/content/docs/docs/migrate-swagger/01-get-started.md
@@ -16,7 +16,7 @@ We have created a swagger to TypeSpec conversion tool to help take on the bulk o
 - Clone [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs).
 - Install dependencies:
   ```shell
-  npm install # Run at root
+  npm install # Run at root of the repository
   ```
 
 ### Generate TypeSpec with converter
@@ -84,14 +84,18 @@ You will need to compare the Swagger generated from TypeSpec with your original 
   1. Recompile your TypeSpec files with `tsp compile .` in your TypeSpec folder.
   2. Run the `npx tsmv` command again with the same parameters.
   3. Review the updated differences in VS Code.
-  4. Make further adjustments as needed. For more effective visualization, fix differences in this recommended order:
+  4. Make further adjustments as needed. Refer to [Resolving Swagger Breaking Change Violations](./faq/breakingchange.md) to understand expected breaking change and mitigation steps. For more effective visualization, fix differences in this recommended order:
      - Path (route) differences first
      - Definition (model) name differences next
      - Detail differences within paths and definitions last
 
 ### Create Spec PR with new TypeSpec project
 
-- Review CI checks such as breaking changes and other failures.
+- In your `readme.md` file, under your latest tag, change the `input-file` to the swagger generated from your TypeSpec.
+- Delete your previous swagger files if they are not referenced from the readme file.
+- Create PR with your TypeSpec files, changed swagger files (examples included) and readme file.
+- Check CI failures. Refer to 
+
 
 ## How to Get Help
 
--- a/website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
+++ b/website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
@@ -2,16 +2,44 @@
 title: Resolving Swagger Breaking Change Violations
 ---
 
-The Swagger Converter will not be able to accurately represent every part of every API in TypeSpec. This document
-outlines some common changes you might need to make to a converted TypeSpec to make it conform to your existing service API and  
-pass check-in validations.
+The Swagger Converter cannot perfectly represent every aspect of every API in TypeSpec. This document outlines common changes you might need to make to a converted TypeSpec to ensure compatibility with your existing service API and pass check-in validations.
 
-## Migrate ARM Specs
+## Using Resources from Common Types
 
-### Changing the Names of Request Payload Parameters
+If your resource definition already extends from a resource type in [common-types](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/common-types/resource-management), you can skip this section. Otherwise, you should understand how we identify resource models and select appropriate base models from the TypeSpec resource manager library.
 
-For operations with non-empty request bodies (PUT, POST, PATCH), the TypeSpec operation templates provide a default name for the
-request parameter corresponding to the request payload. You can use augment decorators to make changes to this parameter, and other parts of the operation signature.
+We identify a model as a resource when it meets all these criteria:
+
+1. There is a GET operation for it
+2. At least one operation returns a 200 response containing this model
+3. The model has properties named "id", "name", and "type"
+
+Once a model is identified as a resource, we represent it by extending an appropriate [resource model](../../howtos/ARM/resource-type.md#modeling-resources-in-typespec) from the TypeSpec library. This results in differences between your original Swagger and generated Swagger like:
+
+```diff
+"YourResource": {
+  "type": "object",
+  "properties": {
+    "properties": {
+      "$ref": "#/definitions/YourResourceProperties",
+    }
+  },
+  "allOf": [
+    {
+-      "$ref": "#/definitions/YourOwnProxyResourceDefinition"
++      "$ref": "../../../../../common-types/resource-management/v3/types.json#/definitions/ProxyResource"
+    }
+  ]
+}
+```
+
+We would suggest you to accept this expected change to align with ARM convention. However, if you have any strong business justification to keep original definition, you could use `@customAzureResource` to mark your model.
+
+## Migrating ARM Specifications
+
+### Customizing Request Payload Parameter Names
+
+For operations with request bodies (PUT, POST, PATCH), TypeSpec operation templates provide default names for request parameters. You can use augment decorators to customize these parameter names and other parts of the operation signature.
 
 The following sections show how to do this for each operation template.
 
--- a/website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
+++ b/website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
@@ -1,0 +1,38 @@
+---
+title: Resolving Pipeline failures
+---
+
+This document explains how to mitigate pipeline failures in TypeSpec migration PRs.
+
+## Swagger BreakingChange
+
+Skip this section if you only have one swagger file in your latest version.
+
+This pipeline will definately fail if you have more than one swagger files in your latest version. See detailed explanation in the discription of this [issue](https://github.com/Azure/typespec-azure/issues/2194#issue-2844564216). Therefore, we leverage "TypeSpec Migration Validation" pipeline to give us a signal if there is breaking change. Go to this pipeline and check the output from "Run TypeSpec Migration Validation". The output is expected to be exactly the same as this [step](../01-get-started.md#review-and-adjust-the-typespec) in your local machine. Review the changes to see if it's expected.
+
+## Swagger ModelValidation
+
+Some validation rules may fail because TypeSpec fixes certain legacy patterns. Review the known issues below and take appropriate action.
+
+### INVALID_FORMAT: Object didn't pass validation for format uri: {nextlink}
+#### Root cause
+The response model of a pageable operation uses the Azure.Core.Page template:
+```typespec
+model ResponseModel is Azure.Core.Page<ItemType>;
+```
+This standardized approach makes the nextLink type `uri` instead of plain `string`. If your previous example's `nextLink` value isn't a valid URI, it will cause a CI error.
+
+#### Mitigation
+Either:
+1. Update your example to include a valid URI value, or
+2. Modify your TypeSpec response model definition to:
+```typespec
+@pagedResult
+model ResponseModel {
+  @items
+  value: ItemType[];
+
+  @nextLink
+  nextLink?: string;
+}
+```


website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Resolving Pipeline failures
+---
+
+This document explains how to mitigate pipeline failures in TypeSpec migration PRs.
```
timotheeguerin, line 5: There is already a document in the spec repo for this, we should at the very least link it as it contains more info

timotheeguerin: we might also want to reorganize a bit here as this is at the same level and after the resolving "Breaking change failues" and resolving lint diff failures

pshao25: > There is already a document in the spec repo for this, we should at the very least link it as it contains more info

Which document you are saying? The content in this document only applies to migration cases.

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
```diff
@@ -2,16 +2,44 @@
...
+    {
+-      "$ref": "#/definitions/YourOwnProxyResourceDefinition"
++      "$ref": "../../../../../common-types/resource-management/v3/types.json#/definitions/ProxyResource"
+    }
+  ]
+}
+```
 
-## Migrate ARM Specs
+We would suggest you to accept this expected change to align with ARM convention. However, if you have any strong business justification to keep original definition, you could use `@customAzureResource` to mark your model.
```
live1206, line 36: link or example about the @customAzureResource usage?

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
```diff
@@ -2,16 +2,44 @@
...
 
-### Changing the Names of Request Payload Parameters
+## Migrating ARM Specifications
 
-For operations with non-empty request bodies (PUT, POST, PATCH), the TypeSpec operation templates provide a default name for the
-request parameter corresponding to the request payload. You can use augment decorators to make changes to this parameter, and other parts of the operation signature.
+### Customizing Request Payload Parameter Names
+
+For operations with request bodies (PUT, POST, PATCH), TypeSpec operation templates provide default names for request parameters. You can use augment decorators to customize these parameter names and other parts of the operation signature.
```
live1206, line 42: any example?

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
```diff
@@ -150,23 +180,32 @@ You can generally choose an asynchronous operation template that matches your op
...
 
-- `ArmResourceDeleteAsync`iis a DELETE operation that uses no request body, and return a `200` response in the case of a successful synchronous delete, a `202` response in the case of an Asynchronous delete operation, and a `204` response in case the resource does not exist.
+- `ArmResourceDeleteAsync` is a DELETE operation that uses no request body, and returns a `200` response in the case of a successful synchronous delete, a `202` response in the case of an asynchronous delete operation, and a `204` response in case the resource does not exist.
 
   ```tsp
-  op createOrUpdate is ArmResourceDeleteAsync<Resource>;
+  op delete is ArmResourceDeleteAsync<Resource>;
   ```
+
+### Visibility Changes for `nextLink` and `value` Properties
```
lmazuel, line 204: This has no SDK impact as far as I'm aware, since those properties are always internal. Should we accept the Swagger diff here, and remove the entire section?

I'm just worried we ask service teams to do useless work here, there is already enough of important work.

website/src/content/docs/docs/migrate-swagger/01-get-started.md
```diff
@@ -84,14 +84,17 @@ You will need to compare the Swagger generated from TypeSpec with your original
...
+  4. Make further adjustments as needed. Refer to [Understanding the Must-have Breaking Changes](./faq/mustread.md) to understand expected breaking change and mitigation steps. For more effective visualization, fix differences in this recommended order:
      - Path (route) differences first
      - Definition (model) name differences next
      - Detail differences within paths and definitions last
 
 ### Create Spec PR with new TypeSpec project
 
-- Review CI checks such as breaking changes and other failures.
+- In your `readme.md` file, under your latest tag, change the `input-file` to the swagger generated from your TypeSpec.
+- Delete your previous swagger files if they are not referenced from the readme file.
```
markcowl, line 95:
```suggestion
- If the generated swagger file(s) for the latest version changed name, delete the old swagger file(s) no longer referenced in README.md
```

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
```diff
@@ -2,16 +2,48 @@
...
+
+```tsp
+@action("customizedAction")
+move is ArmResourceActionSync<Employee, MoveRequest, MoveResponse>;
+```
+
+This produces the route:  
+`/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ContosoProviderHub/employees/{employeeName}/customizedAction`
+
+By default, the `@action` decorator converts the segment to camel case. If you want to preserve the exact casing, use `@Rest.Private.actionSegment`:
```
markcowl, line 30: We should not include this advice.  ARM APIs should be case insensitive.  Service teams should use the camelCase version and ensure it is supported by their service.

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
```diff
@@ -2,16 +2,48 @@
...
+
+```tsp
+@Rest.Private.actionSegment("UpperCase")
```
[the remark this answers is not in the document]
markcowl: We should try not to perpetuate this likely error.  Services should handle Uris case-insensitively.

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
```diff
@@ -2,16 +2,48 @@
...
+move is ArmResourceActionSync<Employee, MoveRequest, MoveResponse>;
+```
+
+This produces the route:  
+`/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ContosoProviderHub/employees/{employeeName}/UpperCase`
+
+### Customizing Scope Parameter Name for Extension Resources
```
markcowl, line 40: We should remove this section until there is advice that doesn't involve private decoratos  (not that LegacyOperations would allow using a Scope parameter without private decorators).

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
markcowl: Agreed that these changes make no difference to the SDK, or the API in general, because they are response-only types.  If possible, we should look at loosening this rule to exclude response-only types.

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
```diff
@@ -0,0 +1,88 @@
...
+title: Understanding the Must-have Breaking Changes
+---
+
+To fully leverage the benefits of TypeSpec and follow best practices, there are some unavoidable breaking changes you will encounter after migrating from Swagger to TypeSpec.
+
+## Using Resources from Common Types
+
+If your resource definition already extends from a resource type in [common-types](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/common-types/resource-management), you can skip this section. Otherwise, you should understand how we identify resource models and select appropriate base models from the TypeSpec resource manager library.
+
+A model is identified as a resource when it meets **all** of the following criteria:
```
markcowl, line 11: A model is a resource if it is registered with ARM as a resource.

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
```diff
@@ -0,0 +1,88 @@
...
+
+1. There is a GET operation for the model.
+2. At least one operation returns a 200 response containing this model.
+3. The model has properties named `id`, `name`, and `type`.
```
markcowl, line 15: 'id' , 'name',  and 'type'  have to do with when you can use common types, not whether a type is a resource.

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
```diff
@@ -0,0 +1,88 @@
...
++      "$ref": "../../../../../common-types/resource-management/v3/types.json#/definitions/ProxyResource"
+    }
+  ]
+}
+```
+
+We recommend accepting this change to align with ARM conventions. However, if you have a strong business justification to keep your original definition, you can use the `@customAzureResource` decorator to mark your model.
+
+```tsp
+model YourResource extends YourOwnProxyResourceDefinition {
```
markcowl, line 39: Note that the model doesn't necessarily need to extend anything.  Also, if they use this pattern, they will need to suppress the linting rule for legacy decorators.

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
```diff
@@ -0,0 +1,88 @@
...
+  name: string;
+}
+```
+
+## Using Page Model from Azure.Core Library
+
+For all pageable response models, we highly recommend using `Azure.Core.Page` to align with ARM conventions:
+
+```tsp
+model YourPageableModel is Azure.Core.Page<YourItemType>;
```
markcowl, line 57: When possible, Resource lists should use `ResourceListResult`, using the Azure.Core pageable should only be necessary for non-resource types

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+This makes the `value` property in `YourPageableModel` required, and the type of the `nextLink` property becomes `url`. If you need to keep the previous shape, define `YourPageableModel` as a regular model.
+
+## Misuse of "readOnly" on Model
```
markcowl, line 62:
```suggestion
## Handling "readOnly" in Model Schemas
```

website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Swagger BreakingChange
+
+This pipeline will definitely fail if you have more than one swagger files in your latest version. See detailed explanation in the description of this [issue](https://github.com/Azure/typespec-azure/issues/2194#issue-2844564216). Therefore, we leverage "TypeSpec Migration Validation" pipeline to give us a signal if there is breaking change. Go to this pipeline and check the output from "Run TypeSpec Migration Validation". The output is expected to be exactly the same as this [step](../01-get-started.md#review-and-adjust-the-typespec) in your local machine. Review the changes to see if it's expected.
```
markcowl, line 9:
```suggestion
This pipeline will fail if you have more than one swagger file in your latest version. See detailed explanation in the description of this [issue](https://github.com/Azure/typespec-azure/issues/2194#issue-2844564216). Therefore, we leverage the "TypeSpec Migration Validation" pipeline to properly identify real breaking changes. Go to this pipeline and check the output from "Run TypeSpec Migration Validation". The output is expected to be exactly the same as this [step](../01-get-started.md#review-and-adjust-the-typespec) in your local machine. Review the changes to see if they are expected.
```

website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
```diff
@@ -0,0 +1,55 @@
...
+
+However, if you have only one swagger file in your latest version, you should check out the results of this pipeline. Refer to [Resolving Swagger Breaking Change Violations](./faq/breakingchange.md) if it fails.
```
markcowl, line 11:
```suggestion
However, if you have only one swagger file in your latest version, you should use this pipeline to detect breaking changes. Refer to [Resolving Swagger Breaking Change Violations](./faq/breakingchange.md) if it fails.
```

website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
```diff
@@ -0,0 +1,55 @@
...
+  @nextLink
+  nextLink?: string;
+}
+```
+
+### INVALID_FORMAT: Object didn't pass validation for format arm-id
+
+#### Root cause
+
+It is possibly because your own defined resource is mapped to the resource defined in common-type. See [this](./breakingchange.md#using-resources-from-common-types) for details. The `id` type for `Resource` in common-types is of format `arm-id`.
```
markcowl, line 51: This is most likely because they didn't select the appropriate common-types version  (versions v5 and later use arm-id consistently).  We should suggest that they use the Azure.Core scalar `armResourceIdentifier` for these types (or the equivalent scalar aliased in Azure.ResourceManager.

markcowl (changes requested):

website/src/content/docs/docs/migrate-swagger/faq/breakingchange.md
lmazuel: +1 with Mark

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
pshao25: I'm expaining how we decide a model in a swagger as a resource model in TypeSpec.

markcowl, line 11:
```suggestion
An ARM Resource model should extend one of the typespec common-types Resource models when it meets **all** of the following criteria:
```

markcowl (changes requested):

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
pshao25: Yes, should we figure out a way since our goal is to remove the suppression?

website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+This standardized approach changes the `nextLink` type from plain `string` to `uri`. If your previous example's `nextLink` value isn't a valid URI, it will cause a validation error.
+
+#### Resolution
+
+Choose one of the following options:
+
+1. **Update your example** to include a valid URI value, or
+2. **Modify your TypeSpec response model** to use a custom definition:
```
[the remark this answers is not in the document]
pshao25: It makes more sense to change it to: 
2. **Modify your TypeSpec response model** to use a custom definition if the nextLink returned by the service is not a URI:

website/src/content/docs/docs/migrate-swagger/faq/conversionfix.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Fixing Converted TypeSpec
+---
+
+## Fixing Linter Rules
+
+Ideally, the original Swagger will be converted into equivalent TypeSpec following ARM conventions. However, many Swagger specifications predate current ARM conventions  and are 'grandfathered in' to prevent breaking changes. The  equivalent TypeSpec that represents the original Swagger will violate linter rules, and linter suppressions are added. Check out how to fix linter violations below.
```
markcowl, line 7:
```suggestion
Ideally, the original Swagger will be converted into equivalent TypeSpec following ARM conventions. However, many Swagger specifications predate current ARM conventions and are 'grandfathered in' to prevent breaking changes. The  equivalent TypeSpec that represents the original Swagger will violate linter rules, and linter suppressions must be added. Check out how to fix linter violations below.
```

website/src/content/docs/docs/migrate-swagger/faq/mustread.md
```diff
@@ -0,0 +1,92 @@
...
+
+```tsp
+@armResourceOperations
+interface YourResources {
+  @doc("List all resources")
+  list is ArmResourceListByParent<YourResource>;
+}
+```
+
+Both default `Azure.ResourceManager.ResourceListResult` make the `value` property in `{YourResource}ListResult` required, and the type of the `nextLink` property becomes `url`.
```
markcowl, line 64:
```suggestion
`Azure.ResourceManager.ResourceListResult` makes the `value` property in `{YourResource}ListResult` required, and the type of the `nextLink` property becomes `url`. 
```

markcowl (approved):

website/src/content/docs/docs/migrate-swagger/faq/pipeline.md
```diff
@@ -0,0 +1,158 @@
...
+  "x-ms-enum": {
+    "name": "EnumType",
+    "modelAsString": true
+  }
+}
+```
+
+#### 1023 - TypeFormatChanged
+
+The error message `The new version has a different format 'uri' than the previous one ''.` on the `nextLink` property is expected because the [page template defines the `nextLink` as `uri`](./mustread.md#using-page-model-from-azurecore-library).
```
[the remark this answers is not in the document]
pshao25: The error message does have `''`.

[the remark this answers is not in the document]
pshao25: meaning the previous format is empty.

PR #2701, opened 2025-05-19
Change the body name of legacy operations

pshao25:
I want to align the name to our previous design. Justifications:
1. Converter already aligns with these names. It costs to fix the logic in converter.
2. Current name is a breaking to all the ongoing migration PR. We need to rename a lot, including the operations and all the augmented decorators applied on the body name.
3. Current name is a breaking to haven't-merged Compute.

### DIFF
--- a/.chronus/changes/legacyOperationUpdate-2025-4-19-18-33-20.md
+++ b/.chronus/changes/legacyOperationUpdate-2025-4-19-18-33-20.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Change the body name of legacy operations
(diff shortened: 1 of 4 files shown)


timotheeguerin:
@markcowl is out until wednesday(21st) so I think we'll need to wait for him to weigh in on this. If we were to change that I assume this probably also would need ot be released as a hotfix on top of it so this should target the release branch.

markcowl (commented):
I have a mild opinion that it is better to have Delete and DeleteAsync in this interface than to have DeleteWithoutOkAsync, which matches the recommended operation name in Azure.ResourceManager.

I have a similarly mild opinion that it's better to use a consistent body property name across the operations here for consistency.  I would do the same across all operations except that would be a breaking change.

We should discuss.

Also, I *think* any impact of these name changes, if we go forward with them would be minimal (1 spec so far) so we wouldn't need another hotfix, but it wouldn't be difficult to release one if it were necessary.  There are a couple of other changes we could also consider, for a hotfix, if so.

pshao25:
@markcowl Yeah as you said "it's better to use a consistent body property name across the operations here for consistency", I just changed the name to keep consistence with its non-legacy version, like `CreateOrUpdateAsync`'s body name is `resource` in the `Azure.ResourceManager`. Other operations are same. 

As for `DeleteAsync`, I keep the same principle, in the `Azure.ResourceManager`, it should be `ArmResourceDeleteWithoutOkAsync` because `ArmResourceDeleteAsync`  is deprecated.

PR #2706, opened 2025-05-19
Remove swagger check from CI

timotheeguerin:
This is uncessary now, as none of the generated swagger should be used to generate SDK via autorest. We just need to make sure the swagger doesn't change unexpectedly which is caught in many other ways.


markcowl:
This is true for data plane, but not yet true for management plane, unless I am missing something

eng/pipelines/ci.yml
```diff
@@ -73,15 +73,4 @@ extends:
               - template: /eng/pipelines/jobs/website.yml
 
           - template: /eng/pipelines/jobs/e2e-job.yml@self
-          - job: AutorestChecks
```
markcowl, line 76: I believe that this is still used by management plane sdks, especially for updates to existing specs.

PR #2710, opened 2025-05-20
Add json value to usage decorator

JoshLove-msft:
This is needed for models-only libraries, like SystemEvents.


JoshLove-msft:
This decorator is owned by the compiler so we can't extend it in this way.

m-nash:
@JoshLove-msft  can we simply add an overload to the usage decorator that takes an int?  This way we can set any flags we want without adding them to the enum?

PR #2724, opened 2025-05-22
ErrorAdditionalInfo.info should be unknown 

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/2657


.chronus/changes/fix-addition-info-type-2025-4-22-13-45-19.md
```diff
@@ -0,0 +1,7 @@
+---
```
markcowl, line 1: This seems unneeded

PR #2772, opened 2025-06-03
Toggle auto

markcowl:


packages/typespec-azure-resource-manager/lib/Legacy/operations.tsp
```diff
@@ -15,11 +15,28 @@ using Rest;
  * @template ResourceTypeParameter The path parameter for the resource name
  * @template ErrorType Optional. The type of error models used in operations created form this template
  */
-@doc("")
+@autoRoute
```
timotheeguerin, line 18: is that not  a problem with client.tsp if we define that on the interfaces

AlitzelMendez (approved):

Issue #2683, opened 2025-05-15
[TCGC&ARM] introduce a new decorator to provide backward compatible to swagger `@operationId`


tadelesh:
Since we are going to remove the `@operationId` decorator of TypeSpec OpenAPI lib from all ARM specs, we need to introduce a new decorator to provide backward compatible.
Some background:
In swagger, we use `operationId` to indicate the client name and the operation name. For example, `Test_Operation` will generate a client named `TestClient` and an operation `operation` under that client.
When converting to TypeSpec, we use namespace, interface to represent a client and client name is inferred from that namespace and interface name. For example, an operation `op operation(): void` under `interface Test{}` will generate the same client and operation as below swagger case.
For ARM cases, we want all the operations for one resource is under the same interface to ensure a more accurate spec, which may cause discrepancy if the `operationId` in swagger is not the same as `InterfaceName_OperationName`. For such case, we will use `@operationId` to make them to be the same before and use `client.tsp` to redefine all the client and operation structure.
Then, two things we are facing:
1. We are going to remove the `@operationId` decorator of TypeSpec OpenAPI lib from all ARM specs.
2. `client.tsp` way to deal with restructuring is too complicated and is a great burden for future mantainence.

So, we are proposing to introduce a new decorator to solve it.

Solution 1:
Introduce `@operationId` under `Azure.ResourceManager.Legacy` namespace in `typespec-azure-resource-manager` library.
TCGC use it to reorganize the operations and also create new clients if not exist. Autorest emitter change to use it to generate `operationId` property.

Solution 2:
Introduce `@moveToClient(Interface/Namespace/string)` under TCGC library.
TCGC use it to reorganize the operations and also create new clients if not exist. Autorest emitter change to use it to generate `operationId` property.

markcowl:
I prefer the `@moveToClient` approach.  We might use the global namespace to designate operations without a group `@moveToClient(global)` or a sentinel value like the namespace marked as the arm provider namespace.

If we do use the operationId approach, we should choose another name to avoid confution with the OpenApi decorator, like `@operationLocation( clientName: string, operationName: string)`, or the very simple `@operationGroup(groupName: string)`

lmazuel:
I prefer option 2 moveToClient. More explicitly, I prefer non-option 1 :). My problem with option 1, is we perpetuating the `_` hack we did 10 years ago, and I'd rather things to be more explicit.

tadelesh:
I think we could continue with option 2.
[...]

PR #2785, opened 2025-06-04
[tcgc] `@clientLocation` decorator

tadelesh:
Resolve: https://github.com/Azure/typespec-azure/issues/2683

Things in this PR:
1. Introduce new `@moveTo` decorator.
2. Refactor all client and operation group logic in TCGC and consolidate all logic into several caches.
3. Operation ID calculation support new `@moveTo` logic.
4. Refine empty operation group and client judgement. cc: @live1206


packages/typespec-client-generator-core/test/decorators/move-to.test.ts
```diff
@@ -0,0 +1,360 @@
...
+
+  const sdkPackage = runner.context.sdkPackage;
+  const rootClient = sdkPackage.clients.find((c) => c.name === "TestServiceClient");
+  ok(rootClient);
+  strictEqual(rootClient.children?.length, 2);
+  const aClient = rootClient.children.find((c) => c.name === "A");
+  ok(aClient);
+  strictEqual(aClient.methods.length, 1);
+  strictEqual(aClient.methods[0].name, "a1");
+  const bClient = rootClient.children.find((c) => c.name === "B");
```
iscai-msft, line 146: in terms of the `@moveTo` design, would it be better to require that `B` exists as either an interface or namespace as well, instead of also allowing a string value for it? I think it would be more clear to see `B` described as an actual Typespec type instead of just a string value in a decorator, though it can be redundant, i.e.

```tsp
// main.tsp
namespace Test;
interface A {
  @route("a");
  op a(): void;
}
```

```tsp
// client.tsp
namespace Test;

interface B {};

@@moveTo(Service.A.a, B)
```

This might run into namespace issues though since client.tsp can be diff namespace than main.tsp

tadelesh: There are cases in ARM that the operation group has the same name as a model which makes it impossible to create an interface under the main namespace. So we need to add such support.

packages/typespec-client-generator-core/src/cache.ts
```diff
@@ -0,0 +1,339 @@
...
+  // if no explicit `@client` or `@operationGroup`
+  if (!hasExplicitClientOrOperationGroup(context)) {
+    const newOperationGroupNames = new Set<string>();
+    listScopedDecoratorData(context, moveToKey).map((target) => {
+      if (typeof target === "string") {
+        newOperationGroupNames.add(target);
+      }
+    });
+
+    for (const ogName of newOperationGroupNames) {
```
weidongxu-microsoft, line 84: What if `ogName` coincide with existing group?

PS: can `@clientName` rename a Namespace of certain language scope, which may also cause conflict in that scope?

packages/typespec-client-generator-core/src/cache.ts
```diff
@@ -0,0 +1,339 @@
...
+    const client = program.stateMap(clientKey).get(current);
+    if (client && (client[scope] || client[AllScopes])) {
+      return (client[scope] ?? client[AllScopes]).service;
+    }
+    current = current.namespace;
+  }
+  return undefined;
+}
+
+function buildHierarchyOfOperationGroups(
```
weidongxu-microsoft, line 310: Just checking if my understanding is correct.

We have only 1 level of client / clients. No hierarchy.
Each client can have operation groups, or hierarchy of operation groups (latter only happen on nested Namespace without explicit `@client`)

packages/typespec-client-generator-core/src/context.ts
```diff
@@ -135,6 +139,34 @@ export function createTCGCContext(
       }
       return getVersions(this.program, namespaces[0])[1]?.getVersions()?.[0].enumMember.enum;
     },
+    getClients(): SdkClient[] {
```
weidongxu-microsoft, line 142: Is it supposed to be called by emitter?

packages/typespec-client-generator-core/src/cache.ts
tadelesh: i'm adding such check logic. the calculation is based on the language scope. so i check with the final name for specfic language.

packages/typespec-client-generator-core/src/context.ts
tadelesh: emitter could call it. actually, it's kind of early version to tcgc typekit.

packages/typespec-client-generator-core/src/cache.ts
tadelesh: added new validation.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -389,6 +389,27 @@ export const $lib = createTypeSpecLibrary({
...
+          "`@moveTo` could only move operation to the interface or namespace belong to the root namespace with `@service`.",
+      },
+    },
+    "move-to-duplicate": {
+      severity: "warning",
+      messages: {
+        default:
+          "You could not `@moveTo` a new operation group which has duplicate name of namespace or interface under `@service` namespace.",
+      },
+    },
```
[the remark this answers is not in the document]
tadelesh: I have no strong opinion. Maybe we could report warning first to see if there will be too many misuses of this decorator.

packages/typespec-client-generator-core/test/decorators/move-to.test.ts
```diff
@@ -0,0 +1,426 @@
...
+
+    interface B {
+      @route("/b")
+      op b(): void;
+    }
+  `,
+  );
+
+  expectDiagnostics(diagnostics, {
+    code: "@azure-tools/typespec-client-generator-core/move-to-duplicate",
```
pshao25, line 111: Huh? This looks like myh expected usage. Why it is a duplicate?

.chronus/changes/move-to-decorator-2025-5-4-18-40-55.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Add `@moveTo` decorator to support move operation to another client or sub-client.
```
lirenhe, line 7: @clientLocation  ?

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -679,6 +679,19 @@ extern dec deserializeEmptyStringAsNull(target: ModelProperty, scope?: valueof s
...
+ * Change the operation location in client. If the target client is not defined, use `string` to indicate the client name.
+ *
+ * @param source The operation to change location for.
+ * @param target The target `Namespace`, `Interface` or a string which could indicate the client.
+ * @param scope The language scope for this decorator
+ */
+extern dec clientLocation(
+  source: Operation,
+  target: Interface | Namespace | (valueof string),
+  scope?: valueof string
```
joheredi, line 692: I know this is already an established pattern. However, if we add an options bag to hold all optional params we can get 2 main benefits:

1. Easy to add new params without breaking consumers.
2. Provide named params ({scope: "csharp"}) or ({isFoo: true}), which improves readability.

No need to change, but wanted to bring up this option for consideration at some point in the future

PR #2793, opened 2025-06-05
typespec-java, update client doc

weidongxu-microsoft:


website/src/content/docs/docs/emitters/clients/typespec-java/reference/emitter.md
```diff
@@ -126,6 +126,12 @@ Specify the package name for custom model classes.
 
 Specify the Java class that to be executed by emitter for [code customization](https://github.com/Azure/autorest.java/blob/main/customization-base/README.md), during post-process.
 
+### `rename-model`
+
+**Type:** `string,object`
```
weidongxu-microsoft, line 131: Not sure whether we want to document this.

BTW, string form is only for use of it on command line.

XiaofeiCao: We could leave out "string".

One concern from me, is this file generated by tspd? Will it be overriden in future tspd runs?

XiaofeiCao: I see. It should mostly be used by SDK devs, as one needs to know what generated model looks like. 
I feel like keeping it, as it's the last resort to rename the un-namable, though not sure either...

Issue #2731, opened 2025-05-23
Add overload to usage decorator that takes UsageFlag and Add/Remove


JoshLove-msft:
The usage decorator is owned by TCGC. The usage enum is owned by core. We can add a new overload for the usage decorator that takes [UsageFlags](https://github.com/Azure/typespec-azure/blob/a12937038d290356e19bb4e3a66daf52b2318f06/packages/typespec-client-generator-core/src/interfaces.ts#L109) and whether to add or remove the flag, e.g.

`
@usage(UsageFlags.Json, UsageAction.Add)
@usage(UsageFlags.Input, UsageAction.Remove)
model Foo {
}
`
It should also be possible to apply the decorator at the namespace level.

We should also consider deprecating the existing usage decorator overload as it would be superseded by this new overload.

tadelesh:
TCGC design:

1. Expand `Usage` enum defined for `@usage` decorator [here](https://github.com/Azure/typespec-azure/blob/a12937038d290356e19bb4e3a66daf52b2318f06/packages/typespec-client-generator-core/lib/decorators.tsp#L102) to be consistent with TCGC usage flag enum [here](https://github.com/Azure/typespec-azure/blob/a12937038d290356e19bb4e3a66daf52b2318f06/packages/typespec-client-generator-core/src/interfaces.ts#L109).
2. Change `@usage` to accept a new optional parameter action: 
```
extern dec usage(
  target: Model | Enum | Union | Namespace,
  value: EnumMember | Union,
  actionOrScope?: EnumMember | (valueof string),
  scope?: valueof string
);

enum ActionFlag {
  "add",
  "override",
}
```
In order to keep backward compatibility, the `action` parameter could also accept `scope` value. The decorator's code will take care of how to handle the `action` parameter.

`override` will be the default if no action provided. It is also the current `@usage` behavior.

tadelesh:
Three solutions after discussion:

1. Break the current decorator
```tsp
extern dec usage(
  target: Model | Enum | Union | Namespace,
  value: EnumMember | Union,
  action?: EnumMember,
  scope?: valueof string
);

enum ActionFlag {
  "add",
  "override",
}

enum Usage {
  input: 2,
  output: 4,
  jsonMergePatch: 16,
  multipartFormData: 32,
  json: 128,
  xml: 256
}
```

2. Put `action` after `scope` and make it to be an option bag
```tsp
extern dec usage(
  target: Model | Enum | Union | Namespace,
[...]

PR #2815, opened 2025-06-11
[TCGC] `@usage` change

tadelesh:
resolve: https://github.com/Azure/typespec-azure/issues/2731

### DIFF
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -663,36 +663,60 @@ export const $usage: UsageDecorator = (
   value: EnumMember | Union,
   scope?: LanguageScopes,
 ) => {
-  const isValidValue = (value: number): boolean => value === 2 || value === 4;
+  const isValidValue = (value: number): boolean => {
+    // Allow the new usage values: input(2), output(4), jsonMergePatch(16), multipartFormData(32), json(256), xml(512)
+    return (
+      value === UsageFlags.Input ||
+      value === UsageFlags.Output ||
+      value === UsageFlags.Json ||
+      value === UsageFlags.Xml
+    );
+  };
+
+  let newUsage = 0;
 
   if (value.kind === "EnumMember") {
     if (typeof value.value === "number" && isValidValue(value.value)) {
-      setScopedDecoratorData(context, $usage, usageKey, entity, value.value, scope);
+      newUsage = value.value;
+    } else {
+      reportDiagnostic(context.program, {
+        code: "invalid-usage",
+        format: {},
+        target: entity,
+      });
       return;
     }
   } else {
-    let usage = 0;
     for (const variant of value.variants.values()) {
       if (variant.type.kind === "EnumMember" && typeof variant.type.value === "number") {
         if (isValidValue(variant.type.value)) {
-          usage |= variant.type.value;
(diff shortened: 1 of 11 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1508,22 +1508,7 @@ export function updateUsageOrAccess(
...
-      // usage set
-      if (options.isOverride) {
-        // when a type has usage, it could not be override to narrow usage
-        if (
-          ((type.usage & UsageFlags.Input) > 0 && (value & UsageFlags.Input) === 0) ||
-          ((type.usage & UsageFlags.Output) > 0 && (value & UsageFlags.Output) === 0)
-        ) {
-          diagnostics.add(
-            createDiagnostic({
-              code: "conflict-usage-override",
```
weidongxu-microsoft, line 1520: do we delete it from options?

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -663,36 +663,60 @@ export const $usage: UsageDecorator = (
   value: EnumMember | Union,
   scope?: LanguageScopes,
 ) => {
-  const isValidValue = (value: number): boolean => value === 2 || value === 4;
+  const isValidValue = (value: number): boolean => {
+    // Allow the new usage values: input(2), output(4), jsonMergePatch(16), multipartFormData(32), json(256), xml(512)
```
JoshLove-msft, line 667: The comment doesn't line up with the code. Why not just use UsageFlags?

packages/typespec-client-generator-core/test/decorators/usage-extended.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    strictEqual(models.length, 1);
+    strictEqual(models[0].usage & UsageFlags.Json, UsageFlags.Json);
+    strictEqual(models[0].serializationOptions?.json?.name, "TestModel");
+    strictEqual(models[0].properties[0].kind, "property");
+    strictEqual(models[0].properties[0].serializationOptions.json?.name, "prop");
+  });
+
+  it("should support xml usage", async () => {
+    await runner.compileWithBuiltInService(`
+      @usage(Usage.xml)
```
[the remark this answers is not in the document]
iscai-msft: yep, we've been able to apply `@usage` to namespaces for a while, it originally came from event grid, since we wanted to default generate all of the models in the models-only typespec without having to write `@usage` on each of them. But yeah, `@usage` on a namespace sets the default for all models in a namespace, but it can be overwritten on individual models

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: I'm assuming this was more because the other options in usage flags are emitter-specific and we don't want to fully expose them yet. I think we can start smaller and then build, the other ones, like json merge patch and multipart form data are really tied to operations, so I think it's better to hide them for now

JoshLove-msft: It looks like the latest commit added JSON merge patch and multipart usages.

JoshLove-msft: Oh nevermind.

Issue #2478, opened 2025-04-01
[tcgc] refactor `SdkModelPropertyType`


tadelesh:
revisit all the typing related with method parameter, http parameter, http response header, model property and separate the typing. make sure no redundant typing in the usage.

PR #2817, opened 2025-06-11
[TCGC] Model property refactor and paging metadata migration

tadelesh:
Resolve: https://github.com/Azure/typespec-azure/issues/2478
Resolve: https://github.com/Azure/typespec-azure/issues/2876

In order to use the new typekit from http-client (see this [PR](https://github.com/microsoft/typespec/pull/7615)), we need to update some of the typing related with TypeSpec `ModelProperty` in TCGC.
The refactor removed `SdkModelPropertyType` and `SdkParameter` two type alias. For the place that used these two aliases, we changed to more explicit type, e.g., for the property of `SdkModelType`, the type now is `SdkBodyModelPropertyType` only.
The refactor contain many breaking changes, but only one thing should have impact for emitters:
Since current model property will never be `SdkHttpParameter`, emitter could not filter them out from a model.
So we introduced new `isHttpMetadata` helper to judge if it is a metadata for emitter migration convenience.

## Migration Guide
1. For the place that use `SdkParameter`, emitter needs to replace it with the concrete type: `SdkEndpointParameter`, `SdkCredentialParameter`, `SdkMethodParameter`. `SdkEndpointParameter` or `SdkCredentialParameter` could only be a client initialization parameter. `SdkMethodParameter` could be a client or method parameter.
2. For the place that use `SdkModelPropertyType`, emitter needs to ensure it only used in the place that represent a model's property. If emitter rely on `SdkHttpParameter` or `SdkServiceResponseHeader` to get the payload info of a model's property, then emitter needs to either switch to call `getHttpOperationParameter` to get the operation level parameter, or totally switch to use `SdkHttpOperation` to deal with payload info. If emitter only want to filter out metadata property in model, then emitter could call `isHttpMetadata` to judge it.
3. For the place that use `SdkBodyModelPropertyType`, emitter needs to change to `SdkModelPropertyType`.


weidongxu-microsoft (approved):
Java probably need to do some refactor upon this change, as we do quite some `property.type === "property"` or something like this.

I assume after this PR, we need to use `!isHttpMetadata(property)` for this?

heaths:
Why was a breaking change made in a semver patch release? This is causing major issues for `Azure/typespec-rust`. /cc @antkmsft @jhendrixMSFT

PR #2841, opened 2025-06-16
[TCGC] Fix operation missing issues after `@clientLocation`

tadelesh:
From nightly failure: https://dev.azure.com/azure-sdk/public/_build/results?buildId=4976875&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=1fb2445c-3c92-5f9b-bcad-710222d02bc3


packages/typespec-client-generator-core/src/cache.ts
```diff
@@ -107,8 +107,32 @@ export function prepareClientAndOperationCache(context: TCGCContext): void {
...
+      const operations = [...group.type.operations.values()];
+
+      // when there is explicitly `@operationGroup` or `@client`
+      // operations under namespace or interface that are not decoratored with `@operationGroup` or `@client`
+      // should be placed in the first accesster client or operation group
+      if (group.type.kind === "Namespace" && hasExplicitClientOrOperationGroup(context)) {
+        const innerQueue: Namespace[] = [group.type];
+        while (innerQueue.length > 0) {
+          const ns = innerQueue.shift()!;
+          for (const subNs of ns.namespaces.values()) {
```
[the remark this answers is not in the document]
tadelesh: Yes. So, I filtered them with `__rawClientsOperationGroupsCache` which has all namespaces with `@operationGroup`.

weidongxu-microsoft:
@microsoft-github-policy-service rerun

haolingdong-msft (approved):

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -165,18 +165,26 @@ export const $client: ClientDecorator = (
...
+  if (service?.kind !== "Namespace") {
+    service = findClientService(context.program, target);
+  }
+
+  if (
+    service === undefined ||
+    service.kind !== "Namespace" ||
+    !isService(context.program, service) ||
+    !service.decorators.some(
+      (d) => d.definition?.name === "@service" && d.definition?.namespace.name === "TypeSpec",
```
weidongxu-microsoft, line 179: Curious on what happened to `isService` from compiler lib.

nit, if we need write this test twice (also in `findClientService`), we may use a util function.

tadelesh: `isService` depend on the decorator apply sequence. it may return empty.

weidongxu-microsoft (approved):

PR #2859, opened 2025-06-17
[tcgc] add more lro metadata

### LINKED ISSUE #2848: [tcgc] Improve playground output for getLroMetadata
Recently spoke to @markcowl who mentioned that the finalStep field (and probably also pollingStep) returned by getLroMetadata is probably returning an underlying type with data that we're not surfacing in the playground through the tcgc output view. Can we check for the underlying types in getLroMetadata and surface it in the yaml we show in playground?

cc @iscai-msft

iscai-msft:
fixes #2848


packages/typespec-client-generator-core/src/methods.ts
```diff
@@ -517,7 +517,12 @@ function getServiceMethodLroMetadata(
...
     finalResponse: getFinalResponse(),
     finalStep:
-      rawMetadata.finalStep !== undefined ? { kind: rawMetadata.finalStep.kind } : undefined,
+      rawMetadata.finalStep !== undefined ? { 
+        responseModel: diagnostics.pipe(
+          getClientTypeWithDiagnostics(context, rawMetadata.finalStep.responseModel),
+        ) as SdkModelType,
+        kind: rawMetadata.finalStep.kind 
+      } : undefined,
     pollingStep: {
```
catalinaperalta, line 526: @iscai-msft just a heads up, I think polling step might have a lot of the same metadata returned on it

catalinaperalta (approved):

Issue #2862, opened 2025-06-18
[TCGC] Refine TCGC documentation to let AI understand better


tadelesh:
- [x] Decorator's documentation, especially `@scope`, etc.
- [x] Client related usage and concept consolidation.

lirenhe:
Some suggestion of this page:
1.  It is better to add more description of the sample code in this page to highlight the impact of the SDK of those decorators.
2. For EnumMember, can we list the possible values and the meaning of each enum member?
3. Describe the purpose of introducing those decorator, e.g.  what are the problem we could solve.

PR #2875, opened 2025-06-20
Refine tcgc doc

tadelesh:
Resolve: https://github.com/Azure/typespec-azure/issues/2862


packages/typespec-client-generator-core/README.md
```diff
@@ -359,37 +372,42 @@ op test(
 ): void;
 ```
 
+###### Mark a parameter as not an API version parameter
```
lirenhe, line 375:
```suggestion
###### Mark a parameter as not presenting an API version parameter
```

packages/typespec-client-generator-core/README.md
```diff
@@ -508,25 +535,32 @@ op myOperation(): void;
 
 #### `@clientInitialization`
 
-Customize the client initialization way.
+Customize the client initialization way in the generated client SDK.
```
lirenhe, line 538:
```suggestion
Customize the client initialization way in the generated client SDK.
```

Allows customization of how clients are initialized in the generated SDK.
By default, the root client is initialized independently, while sub-clients are initialized through their parent client.
Initialization parameters typically include endpoint, credential, and api-version.

lirenhe (approved):

Issue #2783, opened 2025-06-04
Improve error message of azureCutomResource without a name


pshao25:
[...]

PR #2882, opened 2025-06-20
Add arm-custom-resource-no-key linter rule

AlitzelMendez:
arm-custom-resource-no-key as not having a key can cause duplicate operations paths

issue: https://github.com/Azure/typespec-azure/issues/2783


packages/typespec-azure-resource-manager/src/rules/arm-custom-resource-no-key.ts
```diff
@@ -0,0 +1,30 @@
+import { createRule, isKey, Model } from "@typespec/compiler";
+import { isCustomAzureResource } from "../resource.js";
+import { getProperties } from "./utils.js";
+
+export const armCustomResourceNoKey = createRule({
+  name: "arm-custom-resource-no-key",
+  description: "Validate that custom resource contains a key property.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-core/rules/no-legacy-usage",
```
timotheeguerin, line 9: wrong url?

timotheeguerin (approved):

Issue #2885, opened 2025-06-23
[TCGC] duplicate-client-name logic should change after @clientLocation


pshao25:
A specific example is https://github.com/Azure/azure-rest-api-specs/pull/34150#discussion_r2157984311

Originally we have:
```
interface StorageTasks {
  @operationId("StorageTasksReport_List")
  list is ArmResourceActionSync<>;

  @operationId("StorageTaskAssignment_List")
  storageTaskAssignmentList is ArmResourceActionSync<>;
}
```

We are expected to change to 
```
@@clientLocation(StorageTasks.list, "StorageTasksReport");
@@clientLocation(StorageTasks.storageTaskAssignmentList, "StorageTaskAssignment");
@@clientName(StorageTasks.storageTaskAssignmentList, "list")
```

Actual: rule "duplicate-client-name" in TCGC fails.
Expected: no errors

PR #2889, opened 2025-06-23
[TCGC] Consider `@clientLocation` when validate client name collision

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/2885

### DIFF
--- a/packages/typespec-client-generator-core/test/validations/types.test.ts
+++ b/packages/typespec-client-generator-core/test/validations/types.test.ts
@@ -70,6 +70,27 @@ it("no duplicate operation with @clientLocation", async () => {
   expectDiagnosticEmpty(diagnostics);
 });
 
+it("no duplicate operation with @clientLocation another", async () => {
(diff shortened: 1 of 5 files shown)


packages/typespec-client-generator-core/test/validations/types.test.ts
```diff
@@ -47,3 +47,129 @@ it("no-discriminated-unions", async () => {
...
+      op list(): void;
+
+      @clientLocation("StorageTaskAssignment")
+      @clientName("list")
+      @route("/assignments")
+      op storageTaskAssignmentList(): void;
+    }
+    `,
+  );
+
```
pshao25, line 69: I think we need a test
```
 interface StorageTasks {
      @route("/list")
      op list(): void;
      @clientLocation("StorageTaskAssignment")
      @clientName("list")
      @route("/assignments")
      op storageTaskAssignmentList(): void;
    }
```

PR #2892, opened 2025-06-23
Add advanced extension resource support

markcowl:


.vscode/launch.json
```diff
@@ -144,6 +144,26 @@
...
+      "smartStep": true,
+      "sourceMaps": true,
+      "skipFiles": ["<node_internals>/**/*.js"],
+      "outFiles": [
+        "${workspaceFolder}/packages/*/dist/**/*.js",
+        "${workspaceFolder}/packages/*/dist-dev/**/*.js",
+        "${workspaceFolder}/core/packages/*/dist/**/*.js",
+        "${workspaceFolder}/core/packages/*/dist-dev/**/*.js"
+      ],
+      "cwd": "${workspaceRoot}/packages/samples/specs/resource-manager/resource-types/specific-extension",
```
timotheeguerin, line 162: wondering if we could set that up as a compile current opened file

packages/typespec-azure-resource-manager/lib/common-types/internal.tsp
```diff
@@ -18,6 +18,15 @@ enum ResourceHome {
 
   @doc("The resource is bound to an extension")
   Extension,
+
+  @doc("The resource is built in")
+  BuiltIn,
+
+  @doc("The resource is built in")
```
timotheeguerin, line 25: Are those the right comments?

packages/typespec-azure-resource-manager/lib/extension/parameters.tsp
```diff
@@ -0,0 +1,223 @@
...
+ * }
+ * ```
+ */
+@builtInResource
+model ScopeParameter<Type extends string = string> {
+  @path(#{ allowReserved: true })
+  @key
+  @doc("The fully qualified Azure Resource manager identifier of the resource.")
+  scope: Type;
+}
```
pshao25, line 26: Are we going to delete the existing `ExtensionResource`?

packages/typespec-azure-resource-manager/lib/extension/parameters.tsp
```diff
@@ -0,0 +1,223 @@
...
+ *  ...ResourceUriParameter;
+ * }
+ * ```
+ */
+@builtInResource
+model ScopeParameter<Type extends string = string> {
+  @path(#{ allowReserved: true })
+  @key
+  @doc("The fully qualified Azure Resource manager identifier of the resource.")
+  scope: Type;
```
pshao25, line 25: I'm curious why we changed the name to `scope`. How about the `resourceUri`?

packages/typespec-azure-resource-manager/lib/extension/parameters.tsp
markcowl: No

timotheeguerin (approved):

PR #2910, opened 2025-06-27
Add FAQ for changing `@pattern`

### LINKED ISSUE #2355: Document changing pattern
@johanste  input

>If bugfix, it goes into all. I suspect that is the most common.
>
> If restricting what was previously valid input, it is a breaking change. If expanding on what is valid output for a previously valid input, it is a breaking change. Breaking changes go through breaking change board. I hope that is the lest common.
>
> If it is a expanding outputs but only if the client sends something that was previously no possible to send (or, rather, the service would have rejected the call), it should be discussed in API review board. There are degrees of breakingness to this. I would make the case in review board to add to all, but I need a couple of concrete examples for the rest of the API review board to fully grok the implications I think. I have not a good picture of how common this is.'

Everytime this has come up it was a bug fix and so all specs should be fixed

an in more general document how to workaround the versioining of decorator

timotheeguerin:
fix #2355

### DIFF
--- a/website/src/content/docs/docs/howtos/ARM/versioning.md
+++ b/website/src/content/docs/docs/howtos/ARM/versioning.md
@@ -362,3 +362,60 @@ model EmployeeProperties {
   city?: string;
 }
 ```
+
+## FAQ
+
+### How do I version a pattern?
+
+I have this model and I would like to change the pattern in v2
+
+```tsp
+model EmployeeProperties {
+  @pattern("^[a-z]+$")
+  state: string;
+}
+```
+
+:::caution
+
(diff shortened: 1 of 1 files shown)


website/src/content/docs/docs/howtos/ARM/versioning.md
```diff
@@ -362,3 +362,60 @@ model EmployeeProperties {
   city?: string;
 }
 ```
+
+## FAQ
```
timotheeguerin, line 366: Its tagged at the end of the arm versioning doc but as we don't have a generic/data plane one this feels fine

Issue #2914, opened 2025-06-30
OperationId emitted from autorest emitter is not hornoring its original name


pshao25:
[...]

PR #2927, opened 2025-07-01
[typespec-autorest] Change operation ID normalization logic to only capitalize the first letter

tadelesh:
Fix: https://github.com/Azure/typespec-azure/issues/2914


timotheeguerin:
if there is no issues with the spec repos then this PR can be redone targetting the release branch

packages/typespec-autorest/src/utils.ts
```diff
@@ -16,8 +16,8 @@ import {
   Service,
   Type,
 } from "@typespec/compiler";
+import { capitalize } from "@typespec/compiler/casing";
 import { getOperationId } from "@typespec/openapi";
-import { pascalCase } from "change-case";
```
timotheeguerin, line 20: I think this also means this can be removed as a dependency(also in typespec-autorest-cannonical where it shouldn't even be there)

tadelesh:
> do you have a pr showing the effect of this in the spec repos, been slowly trying to get some builtin pipeline to validate that #2917 but for now need to do manually

Affected four services: https://github.com/Azure/azure-rest-api-specs/pull/35621.
- `awsconnector` does not release SDK yet
- `desktopvirtualization` preview version
- `migrate` preview version
- `workloads` SDK release from TypeSpec

The changes have few impacts for SDK side. I'm not sure if it has side effect for other sides such as doc site, etc.

packages/typespec-autorest/src/utils.ts
tadelesh: removed.

timotheeguerin:
that's quite a lot of changes, despite being from the same services mostly. Feel like that might need at least to be an option to keep the old behavior. I think this might need to wait on Mark to weight in

tadelesh:
I updated this PR: https://github.com/Azure/azure-rest-api-specs/pull/35621 and with the new `back-compatible.tsp`, the logic change in this PR should have no impact for swagger.

tadelesh:
@timotheeguerin I checked the change locally and it works well. But in the ci action, if failed. I'm wondering if the ci use the latest main?

markcowl (approved):
- I think we generally want to pascal case, with azure-specific exceptions (VM, IP, SAS, etc.)
- When trying to match existing SDKs, we are likely to need a mechanism to just state the casing of the operation,  IMO, this should be `@clientName`,  we should just substitute the clientName in these cases without applying any transform.

That said, this appears to be better than current behavior, so no objections.

PR #2954, opened 2025-07-08
Update doc for data plane

v-jiaodi:
related to https://github.com/Azure/autorest.typescript/issues/3329


qiaozha:
@v-jiaodi we don't need this, the is-modular-library option is still optional to the emitter.

Issue #1963, opened 2024-12-09
Migrate Azure Core library to use TypeSpec core paging


tadelesh:
Migrate the Azure core libraries to use typespec core paging
- Use `@pageItems` decorator
- Add `@list` for list operations
- Have the paging API sync both the Azure and core paging
- Deprecate the `@pagedResult` and `@items` decorators
- Migrate Spec repos

tadelesh:
also there are discrepancies for indicating the operation is pageable. in azure, we use `@pagedResult`, in unbranded, we use `@list`.

lmazuel:
@markcowl I think the `nextLink` simplification was a good step, but we need more consolidation :)

qiaozha:
I feel like we should also leverage the `@list` decorator in azure scope? using `@pagedResult` on a model to detect if it's a paging operation might be problematic if this model is being used in a non paging operation or being used as a property type of a non paging operation？

AlitzelMendez:
### Proposal

To consolidate decorators with shared functionality the first steps was to remove `@nextLink` from Azure.Core.nextLink in favor of TypeSpec.nextLink.

There are two other decorators in the same scenario:
- `@Azure.Core.items`/  `@TypeSpec.pageItems`
- `@Azure.Core.pagedResult` / `@TypeSpec.list`

These decorators share similar functionality, and the expectation is to consolidate them.

Since we are already using nextLink from TypeSpec, and the pageable functionalities of pageItems, list, and nextLink are closely related, we should support the entire workflow in TypeSpec and remove the duplicated functionality from Azure.Core.

Additionally, the Azure.Core decorators present some problems:
- As the `@pagedResult` is placed in the model, this can cause issues when the model is used by non-paging operations, as it is automatically identified as a pageable operation. This necessitates having a specific model for pageable operations and a similar one for non-pageable operations.
- When using `@pagedResult`, there is no validation to ensure that a @items model property exists.

The plan is to remove `@Azure.Core.pagedResult`/`@Azure.Core.items` decorators and keep `@TypeSpec.pageItems`/`@TypeSpec.list`, this will also involve updating azure-rest-api-specs that use them. This change will modify code that looks like this:
```tsp
@Azure.Core.pagedResult
model Pet {
  @Azure.Core.items
  names: string[];

  @TypeSpec.nextLink
  nextLink?: string;
}

@armResourceOperations
interface Pets {
  @autoRoute
  list is ArmProviderActionSync<
    Response = Pet
  >;
[...]

PR #2974, opened 2025-07-09
Deprecate paged result and items

AlitzelMendez:
Issue: https://github.com/Azure/typespec-azure/issues/1963

In this first pass of the problem we are deprecating the decorators `@pagedResult` and `@items`

- ❌ [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/pull/36153)
- ✅ [azure-rest-apis-specs-pr](https://github.com/Azure/azure-rest-api-specs-pr/pull/23668)


tadelesh (approved):

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -348,10 +348,9 @@ model ResourceListResult<Resource extends Foundations.SimpleResource> is Azure.C
  */
 @doc("The custom response of a list operation.")
 @friendlyName("{name}ListResult", Result)
-@pagedResult
```
weidongxu-microsoft, line 351: The removal of `@pagedResult` on the model would likely means break and require immediate action from TypeSpec source to adapt to `@list` -- some of the migration uses `ResourceActionSync` to specify a List API (i.e., existing code in specs repo, and all the migration PR that not switched to `@list` on API, would not work with the new version of typespec-azure-resource-manager lib)

Please let us know whether this is expected.
Or we prefer to keep this `@pagedResult` for a month or two, before a final switch.

weidongxu-microsoft: Yeah, I can get that breaks come sooner or later. I just thought there would be a few months buffer (that pending PR can still use the old pattern).

Unmerged PR to specs repo won't get fixed automatically.

SDK repo has tons of [commitID](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/storageactions/azure-resourcemanager-storageactions/tsp-location.yaml#L2) that need to be updated upon next release (and typically the TypeSpec source on HEAD has other difference from the pinned commitID).
And there are unmerged tsp-location files in pending PR.

All of them need to be handled upon the break. I just need to know when it happens... I don't have strong objection if "when" is "next release".

timotheeguerin: we brought this up a month and a half ago and talked about it, originally the plan was to have this last release and was postponed to next one(early august).

Now as said before this is not urgent and if adding `@pageItems` with `@azure.core.items` really helps out smooth the migration effort we can just do that now and do the breaking part in september

weidongxu-microsoft: Sure, let me circle it around in SDK devs. If no one have strong opinion, let's say we can just go break right now.

packages/typespec-azure-resource-manager/lib/operations.tsp
```diff
@@ -69,6 +71,7 @@ op ArmResourceListByParent<
  */
 @get
 @autoRoute
+@list
 @listsResource(Resource)
```
pshao25, line 75: I have a comment here: https://github.com/Azure/azure-rest-api-specs/pull/36014/files#r2221212583

timotheeguerin: yeah need to check if we can put `@list` on arm `ListBy` templates

packages/typespec-azure-resource-manager/lib/models.tsp
timotheeguerin: Its quite a simple change so i have a pr there too if we want to got that way first https://github.com/Azure/typespec-azure/pull/3028

weidongxu-microsoft: Thanks Timothee. So far, there is no outstanding feedback from dev asking for buffer time. I think we can go directly in.

packages/typespec-azure-resource-manager/lib/operations.tsp
timotheeguerin: checked and actually its because they are using a custom template with the same name, updating the pr to apply `@list` on the template instead

PR #2976, opened 2025-07-10
Introduce `@previewVersion` decorator

joheredi:
## Problem Description

Service teams often release API changes as **preview** features—unstable, subject to change—before they graduate to **GA** (General Availability). Currently, preview APIs can inadvertently appear in the GA build unless teams manually opt out. For example:

```ts
enum Versions {
  v1,
  v2_preview,
  v2
}

model Foo {
  id: string;
  @added(Versions.v2_preview)
  color: string;
}
```

By default, `color` leaks into `v2` unless you also add `@removed(Versions.v2)`, leading to an error-prone add/remove/add dance.

## Value of This Proposal

* **No silent leaks**: A preview API will not appear in GA unless explicitly promoted.
* **Simpler workflow**: Eliminate manual add/remove/add pattern when switching between preview and GA.
* **Intuitive decorators**: Preserve the existing `@added` and `@removed` semantics.
* **Consistent preview API authoring**: Use `@added(Versions.PreviewX)` in preview and in GA just `@added(Versions.vY)`.

### 1. Introduce `@previewVersion`

```diff
enum Versions {
  v1,
+ @previewVersion
+ v2_preview
}
```

* `@previewVersion` marks an enum value as preview-only.

### 2. Tag Preview-Only APIs

```diff
model Widget {
  id: string;
+ @added(Versions.v2_preview)
+ color: string
}
```

* `@added(Versions.v2_preview)` applies *only* to `v2_preview`.

### 3. Promote to GA and Roll Preview Forward

```diff
enum Versions {
- @previewVersion
- v2_preview,
+ v2,
+ @previewVersion
+ v3_preview
}
```

```diff
model Widget {
- @added(Versions.v2_preview)
+ @added(Versions.v3_preview)
  color: string
}
```

```diff
model Widget {
- @added(Versions.v2_preview)
+ @added(Versions.v2)
  analyze(...Widget): string;
}
```

* `color` exists only in `v3_preview`.
* `analyze` exists in `v2` and all later versions (including `v3_preview`).

## Guidance for Service Teams

- Always make the last enum the preview and apply @previewVersion to it.
- Apply @added(Versions.PreviewVersion) to all preview items - these items will not show up in any GA Version
- For a new GA, add a new version enum BEFORE the preview enum value. Manually change all preview items that are GA'ing so that the @added version value matches the new GA enum value)
- For any items remaining in preview, rename the old preview enum value to the new preview enum value.

*This rules are enforced by the decorator itself on the compiler*


packages/typespec-azure-core/test/decorators.test.ts
```diff
@@ -25,6 +27,82 @@ describe("typespec-azure-core: decorators", () => {
...
+        namespace DemoService;
+
+        enum Versions {
+          v1,
+          @Azure.Core.previewVersion
+          v2Preview: "2.0-preview",
+        }
+      `);
+
+      const [_, __, previewVersionsMap] = useStateMap(AzureCoreStateKeys.previewVersion);
```
timotheeguerin, line 99: Call the `isPreviewVersion` here instead

PR #3003, opened 2025-07-17
Experiement with fixes

timotheeguerin:


timotheeguerin:
Validated manually against `azure-rest-api-specs-pr` as well and this cause no changes

markcowl (approved):

PR #3012, opened 2025-07-18
[TCGC] Fix missing info for multipart cases after model property refactor

tadelesh:
This is a behavior breaking change. Currently, TCGC always replaces the property type of a multipart model with the real body type for the HTTP part. So, emitter could rely on the property type directly to serialize/deserialize an HTTP part. Also, the header parameters in one HTTP part are added to the `headers` property of `MultipartOptions` type.

For example:
```tsp
model MultiPartRequest {
  stringWithContentType: HttpPart<{
    @body body: string;
    @header contentType: "text/html";
  }>;
}
@post
op upload(
  @multipartBody body: MultiPartRequest,
): void;
```
The type of `stringWithContentType` is `string` instead of an anonymous model `{@body body: string; @header contentType: "text/html";}` as before.

## Migration Guide
For any property with multipart serialization option, use the property type directly instead of looking for body parameter as before.

### DIFF
--- a/packages/typespec-client-generator-core/src/types.ts
+++ b/packages/typespec-client-generator-core/src/types.ts
@@ -34,12 +34,9 @@ import {
 } from "@typespec/compiler";
 import {
   Authentication,
-  HttpOperationPart,
   Visibility,
   getAuthentication,
-  getHttpPart,
   getServers,
-  isBody,
   isHeader,
   isOrExtendsHttpFile,
   isStatusCode,
@@ -73,6 +70,7 @@ import {
   SdkDurationType,
   SdkEnumType,
   SdkEnumValueType,
+  SdkHeaderParameter,
   SdkModelPropertyType,
   SdkModelPropertyTypeBase,
   SdkModelType,
@@ -90,7 +88,7 @@ import {
   filterApiVersionsInEnum,
   getAvailableApiVersions,
   getClientDoc,
-  getHttpBodySpreadModel,
+  getHttpBodyType,
   getHttpOperationResponseHeaders,
   getNonNullOptions,
   getNullOption,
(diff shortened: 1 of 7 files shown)


packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1363,46 +1280,83 @@ export function getSdkModelPropertyType(
...
+      filename: part.filename
+        ? diagnostics.pipe(getSdkModelPropertyType(context, part.filename, operation))
+        : undefined,
+      contentType: part.body.contentTypeProperty
+        ? diagnostics.pipe(
+            getSdkModelPropertyType(context, part.body.contentTypeProperty, operation),
+          )
+        : undefined,
+      defaultContentTypes: part.body.contentTypes,
+      name: part.name!,
```
weidongxu-microsoft, line 1347: did we cover this case?
```
// Part name here is name2
op create(
  @multipartBody body: {
    name1: HttpPart<string, #{ name: "name2" }>;
  },
): void;
```

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1363,46 +1280,83 @@ export function getSdkModelPropertyType(
...
+      isMulti: part.multi,
+      filename: part.filename
+        ? diagnostics.pipe(getSdkModelPropertyType(context, part.filename, operation))
+        : undefined,
+      contentType: part.body.contentTypeProperty
+        ? diagnostics.pipe(
+            getSdkModelPropertyType(context, part.body.contentTypeProperty, operation),
+          )
+        : undefined,
+      defaultContentTypes: part.body.contentTypes,
```
weidongxu-microsoft, line 1346: Does a
```
  stringWithContentType: HttpPart<{
    @body body: string;
    @header contentType: "text/html";
  }>;
```

affect the `contentType` and `defaultContentTypes` here?

weidongxu-microsoft (commented):
Quite some changes seem not directly relate to multipart.

packages/typespec-client-generator-core/src/types.ts
tadelesh: Yes. The `contentType` is `contentType` property and `defaultContentTypes` is `["text/html"]`. But the headers will not contain the `contentType` header.

weidongxu-microsoft (approved):

PR #3026, opened 2025-07-22
Add ResolveArmResources method

markcowl:


packages/typespec-azure-resource-manager/src/operations.ts
```diff
@@ -158,6 +180,70 @@ function setResourceLifecycleOperation(
...
+    operation: target,
+    operationGroup: target.interface.name,
+  });
+}
+
+export function getArmResourceOperationList(
+  program: Program,
+  resourceType: Model,
+): Set<ArmOperationIdentifier> {
+  if (!program.stateMap(ArmStateKeys.resourceOperationList).has(resourceType)) {
```
timotheeguerin, line 195: You can use useStateMap and useStateSet to define setter and getter to the state and keep a typed version to make sure the setter and getter don't see the data differently. 

You can check http lib for a lot of examples or I migrated some of azure core Lro status

packages/typespec-azure-resource-manager/src/namespace.ts
```diff
@@ -299,3 +299,21 @@ function getArmProviderFromNamespace(
...
+
+export function resolveProviderNamespace(
+  program: Program,
+  ns?: Namespace | undefined,
+): Namespace | undefined {
+  ns = ns || program.getGlobalNamespaceType();
+  if (program.stateMap(ArmStateKeys.armProviderNamespaces).get(ns)) {
+    return ns;
+  }
+  for (const [_, child] of ns.namespaces) {
```
timotheeguerin, line 311:
```suggestion
  for (const child of ns.namespaces.values()) {
```

packages/typespec-azure-resource-manager/src/namespace.ts
```diff
@@ -299,3 +299,21 @@ function getArmProviderFromNamespace(
 
   return undefined;
 }
+
+export function resolveProviderNamespace(
+  program: Program,
+  ns?: Namespace | undefined,
+): Namespace | undefined {
+  ns = ns || program.getGlobalNamespaceType();
```
timotheeguerin, line 307:
```suggestion
  ns = ns ?? program.getGlobalNamespaceType();
```

packages/typespec-azure-resource-manager/src/operations.ts
markcowl: Yes, but in this case, these operations are only used by our internal decorators.  Happy to change if we want to keep an abstraction over the actual keys for future-proofing.

timotheeguerin: yeah just feels like its safer, so many decorators historically set different data that the getter would expect(specially around undefined). This is also IMO just cleaner and also shorter(if you do actually want an exposed getter/setter which is not your case)

PR #3031, opened 2025-07-22
[tcgc] add `@inheritsFrom` decorator

### LINKED ISSUE #3015: [TCGC] Support nested inheritance decorator
TypeSpec do not support several level of inheritance. I'm not going to go deep in the debate here, but I spent enough time with @johanste to feel it was fair.

Still, we may need at times to generates a hierarchy of class. TCGC should provide a decorator for that.

I reviewed this design with @lirenhe and @johanste already, and it seems a good starting point that solves our issues:

```tsp
@discriminator("kind")
model A {
  kind: string;
}

model BContent {
  foo: string;
}

model B extends A{
  kind: "B";
  ...BContent;
}

@tcgc.inheritfrom(B)
model C extends A{
  ...BContent;
  kind: "C";

}
```

iscai-msft:
fixes #3015

Some things to note:
1. This is the solution that it appears the TypeSpec team is more in favor of, bc it doesn't have bearing on the rest api definition
2. I didn't do any automatic property inheritance from parents, I don't think TCGC should do that. Instead, I do have warnings ensuring that the models are defined in the raw TypeSpec such that the properties are correctly inherited, otherwise the `@inheritsFrom` decorator gets ignored.
3. I do some post processing on `.discriminatedSubtypes` on the `SdkModelType`. I make sure that models from `@inheritsFrom` are added here. This is for ease of emitters implementing

### DIFF
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -1193,12 +1193,35 @@ export function getClientLocation(
 
 const inheritsFromKey = createStateSymbol("inheritsFrom");
 
+function isPropertySuperset(target: Model, value: Model): boolean {
+  // Check if all properties in value exist in target
+  for (const name of value.properties.keys()) {
+    if (!target.properties.has(name)) {
+      return false;
+    }
+  }
+  return true;
+}
+
 export const $inheritsFrom: InheritsFromDecorator = (
   context: DecoratorContext,
   target: Model,
   value: Model,
   scope?: LanguageScopes,
 ) => {
+  // Validate that target has all properties from value
+  if (!isPropertySuperset(target, value)) {
+    reportDiagnostic(context.program, {
+      code: "inherits-from-conflict",
+      format: {
+        childModel: target.name,
+        inheritsFromModelName: value.name,
+      },
+      target: context.decoratorTarget,
+    });
+    return;
+  }
+
   setScopedDecoratorData(context, $inheritsFrom, inheritsFromKey, target, value, scope);
 };
 
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -33,6 +33,7 @@ import {
   ConvenientAPIDecorator,
   DeserializeEmptyStringAsNullDecorator,
   FlattenPropertyDecorator,
+  InheritsFromDecorator,
   OperationGroupDecorator,
   ParamAliasDecorator,
   ProtocolAPIDecorator,
@@ -1188,6 +1189,21 @@ export function getClientLocation(
     | Interface
     | string
     | undefined;
+}
+
+const inheritsFromKey = createStateSymbol("inheritsFrom");
+
+export const $inheritsFrom: InheritsFromDecorator = (
+  context: DecoratorContext,
+  target: Model,
+  value: Model,
+  scope?: LanguageScopes,
+) => {
+  setScopedDecoratorData(context, $inheritsFrom, inheritsFromKey, target, value, scope);
+};
+
+export function getInheritsFrom(context: TCGCContext, target: Model): Model | undefined {
+  return getScopedDecoratorData(context, inheritsFromKey, target);
 }
 
 /**
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -1228,22 +1228,28 @@ export const $inheritsFrom: InheritsFromDecorator = (
 export function getInheritsFrom(context: TCGCContext, target: Model): Model | undefined {
   // have to check circular inheritance in getter because in the decorator stage, the circularity of the models isn't fully calculated yet
   const value = getScopedDecoratorData(context, inheritsFromKey, target);
-  let circular = false;
-  let current: Model | undefined = value;
-  while (current) {
-    if (current === target) {
-      circular = true;
-      break;
-    }
-    current = current.baseModel;
+  if (context.__typesCheckedForCircularInheritance === undefined) {
+    context.__typesCheckedForCircularInheritance = new Set();
   }
-  if (circular) {
-    reportDiagnostic(context.program, {
-      code: "inherits-from-circular",
-      format: { target: target.name, value: value.name },
-      target: value,
-    });
-    return undefined;
+  if (!context.__typesCheckedForCircularInheritance.has(value)) {
+    let circular = false;
+    let current: Model | undefined = value;
+    while (current) {
+      if (current === target) {
+        circular = true;
+        break;
+      }
+      current = current.baseModel;
+    }
+    context.__typesCheckedForCircularInheritance.add(value);
+    if (circular) {
+      reportDiagnostic(context.program, {
+        code: "inherits-from-circular",
+        format: { target: target.name, value: value.name },
+        target: value,
+      });
+      return undefined;
+    }
   }
   return value;
 }
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -1203,7 +1203,26 @@ export const $inheritsFrom: InheritsFromDecorator = (
 };
 
 export function getInheritsFrom(context: TCGCContext, target: Model): Model | undefined {
-  return getScopedDecoratorData(context, inheritsFromKey, target);
+  // have to check circular inheritance in getter because in the decorator stage, the circularity of the models isn't fully calculated yet
+  const value = getScopedDecoratorData(context, inheritsFromKey, target);
+  let circular = false;
+  let current: Model | undefined = value;
+  while (current) {
+    if (current === target) {
+      circular = true;
+      break;
+    }
+    current = current.baseModel;
+  }
+  if (circular) {
+    reportDiagnostic(context.program, {
+      code: "inherits-from-circular",
+      format: { target: target.name, value: value.name },
+      target: value,
+    });
+    return undefined;
+  }
+  return value;
 }
 
 /**
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -79,6 +79,7 @@ export interface TCGCContext {
   __mutatedGlobalNamespace?: Namespace; // the root of all tsp namespaces for this instance. Starting point for traversal, so we don't call mutation multiple times
   __packageVersions?: string[]; // the package versions from the service versioning config and api version setting in tspconfig.
   __packageVersionEnum?: Enum; // the enum type that contains all the package versions.
+  __typesCheckedForCircularInheritance?: Set<Type>; // used to track types that have been checked for circular inheritance to avoid redundant checks
 
   getMutatedGlobalNamespace(): Namespace;
   getApiVersionsForType(type: Type): string[];
--- a/packages/typespec-client-generator-core/src/public-utils.ts
+++ b/packages/typespec-client-generator-core/src/public-utils.ts
@@ -56,7 +56,6 @@ import {
 import {
   AllScopes,
   TspLiteralType,
-  getBaseModel,
   getHttpBodyType,
   getHttpOperationResponseHeaders,
   hasExplicitClientOrOperationGroup,
@@ -532,7 +531,7 @@ function getContextPath(
         if (result) return true;
       }
       // handle additional properties type: model MyModel extends Record<> {}
-      const baseModel = getBaseModel(context, currentType);
+      const baseModel = currentType.baseModel;
       if (baseModel) {
         if (baseModel.name === "Record") {
           const result = dfsModelProperties(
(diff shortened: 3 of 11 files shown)


packages/typespec-client-generator-core/test/decorators/inherits-from.test.ts
```diff
@@ -0,0 +1,375 @@
...
+  strictEqual(modelA.discriminatorProperty?.name, "kind");
+  strictEqual(modelB.discriminatorValue, "B");
+  strictEqual(modelC.discriminatorValue, "C");
+  strictEqual(modelA.discriminatorValue, undefined);
+
+  // Verify properties are correctly inherited
+  strictEqual(modelC.properties.length, 3);
+  strictEqual(modelC.properties[0].name, "kind");
+  strictEqual(modelC.properties[1].name, "foo");
+  strictEqual(modelC.properties[2].name, "bar");
```
tadelesh, line 62: I think there is a logic missing here. `C` should not have the property of `B`: `foo`. Also, as @ArcturusZhang mentioned, I wonder if we should take this [solution](https://github.com/Azure/typespec-azure/issues/3015#issuecomment-3099326745).

tadelesh:
Another concern: In m4, the middle layer will also have discriminated sub types start from the middle layer types. So, TCGC should apply the similar logic.

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -531,25 +531,23 @@ function getContextPath(
...
-        const result = dfsModelProperties(
-          expectedType,
-          currentType.baseModel,
-          currentType.baseModel.name,
-        );
+      if (baseModel) {
+        const result = dfsModelProperties(expectedType, baseModel, baseModel.name);
         if (result) return true;
       }
+      // TODO: come back and see if derived models are needed to change
```
tadelesh, line 550: The very corner case is the anonymous model happen in the spread alias `BBase` and the naming should follow the upper type `B`.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1737,6 +1739,37 @@ function updateSpreadModelUsageAndAccess(context: TCGCContext): void {
...
+    const legacyHierarchyBuilding = getLegacyHierarchyBuilding(context, sdkType.__raw as Model);
+    // must be done after discriminator is added
+    // Populate discriminated subtypes for legacy hierarchy building
+    if (legacyHierarchyBuilding && sdkType.discriminatorValue) {
+      let currBaseModel: SdkModelType | undefined = sdkType.baseModel;
+      while (currBaseModel) {
+        if (!currBaseModel.discriminatedSubtypes) {
+          currBaseModel.discriminatedSubtypes = {};
+        }
+        currBaseModel.discriminatedSubtypes[sdkType.discriminatorValue] = sdkType;
```
tadelesh, line 1756: In m4, all the layers have `discriminatedProperty`, I'm think of if TCGC needs to copy that?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1190,6 +1191,83 @@ export function getClientLocation(
     | undefined;
 }
 
+const legacyHierarchyBuildingKey = createStateSymbol("legacyHierarchyBuilding");
+
+function isPropertySuperset(target: Model, value: Model): boolean {
```
tadelesh, line 1196: If is possible to use `sourceProperty` along with `sourceModel` to handle comparison logic?

packages/typespec-client-generator-core/src/context.ts
```diff
@@ -217,6 +218,7 @@ export async function createSdkContext<
     previewStringRegex: options?.versioning?.previewStringRegex || tcgcContext.previewStringRegex,
     disableUsageAccessPropagationToBase: options?.disableUsageAccessPropagationToBase ?? false,
     flattenUnionAsEnum: options?.flattenUnionAsEnum ?? true,
+    respectLegacyHierarchyBuilding: options?.respectLegacyHierarchyBuilding ?? false,
```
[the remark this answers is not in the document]
iscai-msft: yeah, let me go ahead and make the default `true`

packages/typespec-client-generator-core/src/types.ts
iscai-msft: yeah we might as well, let me get that in

packages/typespec-client-generator-core/src/public-utils.ts
tadelesh:
```
@discriminator("kind")
model A {
  kind: string;
}

model BContent {
  foo: string;
  anonymous: {prop: string};
}

model B extends A {
  kind: "B";
  ...BContent;
}
@Legacy.legacyHierarchyBuilding(B)
model C extends A {
  kind: "C";
  ...BContent;
  bar: string;
}
```

The anonymous model should be named as `BAnonymous`, not `BContentAnonymous`. I doubt if we should handle such corner case. You could decide.

packages/typespec-client-generator-core/src/types.ts
tadelesh: I'm not sure about how each language deal with multiple layer discriminator. If they rely on discriminated property and treat each layer separately. Then we should add it.

tadelesh (commented):
Approved for the general impl. But please wait for language's review.

packages/typespec-client-generator-core/src/types.ts
weidongxu-microsoft: Impl in Java should be fine with current behavior (that only the most base class has `discriminatedProperty`) -- impl would find one from its parents, and combine it with `discriminatorValue`

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.Legacy.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This decorator is considered legacy functionality and may be deprecated in future releases.
+ *
+ * @param target The target model that will gain legacy inheritance behavior
+ * @param value The model whose properties should be inherited from
+ * @param scope Optional parameter to specify which language emitters this applies to
+ * @example Three-level inheritance
+ *
+ * ```typespec
+ * @discriminator("kind")
```
joheredi, line 16: Can we make this example a little bit more descriptive? For example, it would be super helpful if it illustrates clearly what is the user intent and how the new decorator enables that. From looking at the TypeSpec here I don't think that is clear.

packages/typespec-client-generator-core/generated-defs/Azure.ClientGenerator.Core.Legacy.ts
```diff
@@ -0,0 +1,46 @@
+import type { DecoratorContext, Model } from "@typespec/compiler";
+
+/**
+ * Adds support for client-level multiple levels of inheritance.
+ *
+ * This decorator will update the models returned from TCGC to include the multi-level inheritance information.
```
joheredi, line 6: Can we add more details om what scenario this is enabling? What does multi-level inheritance mean in this context. It might confuse users, for example multi-level inheritance is techincally possible in TypeSpec

```tsp
model A {
  id: string;
}

model B extends A{
  name: string;
}

model C extends B{
  birthday: utcDateTime;
}
```

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
```diff
@@ -0,0 +1,535 @@
...
+  ok(modelB);
+  ok(modelC);
+
+  // C should inherit from B instead of A due to @Legacy.legacyHierarchyBuilding
+  strictEqual(modelC.baseModel?.name, "B");
+  strictEqual(modelB.baseModel?.name, "A");
+
+  // Verify discriminator property is correctly identified
+  strictEqual(modelA.discriminatorProperty?.name, "kind");
+  strictEqual(modelB.discriminatorValue, "B");
```
MaryGao, line 55: do we guarantee that the discriminatorValue is a string?

```
@discriminator("kind")
model A {
  kind: string;
}
alias BContent = {
  foo: string;
};
model B extends A {
  kind: "B" | "C";
  ...BContent;
}
```

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
```diff
@@ -0,0 +1,535 @@
...
+  // Verify .discriminatedSubtypes is correctly populated
+  ok(salmonModel.discriminatedSubtypes);
+  strictEqual(salmonModel.discriminatedSubtypes["kingsalmon"], kingSalmonModel);
+  strictEqual(salmonModel.discriminatedSubtypes["smartkingsalmon"], smartKingSalmonModel);
+  ok(kingSalmonModel.discriminatedSubtypes);
+  strictEqual(kingSalmonModel.discriminatedSubtypes["smartkingsalmon"], smartKingSalmonModel);
+  ok(!smartKingSalmonModel.discriminatedSubtypes);
+});
+
+it("circular inheritance", async () => {
```
MaryGao, line 249: Should we really allow this case? First only B and A don't have the same root parent, how do we know the discrinminator property? A and B have circular inheritance, this is not allowed in typespec yet. Another inference is if A and B have circular relationship which means they should be the same, why we define this thing?

image

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
```diff
@@ -0,0 +1,535 @@
...
+      @route("/test")
+      op test(): C;
+    `);
+
+  // Should warn about conflicting inheritance
+  expectDiagnostics(diagnostics, {
+    code: "@azure-tools/typespec-client-generator-core/legacy-hierarchy-building-conflict",
+  });
+});
+
```
MaryGao, line 319: One case I'd like to confirm is if A extends B extends C, which means A could only override properties with narrow scopes e.g only allow to override one optional property to required. If with legacyHierarchyBuilding decorator do we gurantee that?

```
@discriminator("kind")
model C {
    kind: string;
}

model A extends B {
      propA: string;
}
@Legacy.legacyHierarchyBuilding(A)
model B extends C {
  propB: string;
  // override A's property
  propA?: string;
}
```

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -56,6 +56,7 @@ export interface TCGCContext {
   previewStringRegex: RegExp;
   disableUsageAccessPropagationToBase: boolean;
   flattenUnionAsEnum?: boolean;
```
MaryGao, line 58: JS would prefer to have discriminatorProperty for sub model also, could we add that support?  https://github.com/Azure/typespec-azure/blob/b02fc0b5a642c82018ae431da1ba9865a99e8c2e/packages/typespec-client-generator-core/src/interfaces.ts#L504

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
MaryGao: Okay, not sure we have any case which would requires more narrow scopes in swagger side? 

The `fail` here means a tcgc error?

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
```diff
@@ -0,0 +1,535 @@
...
+        foo: string;
+      };
+
+      model B extends A {
+        kind: "B";
+        ...BContent;
+      }
+
+      @Legacy.legacyHierarchyBuilding(B)
+      @usage(Usage.input)
```
MaryGao, line 366: why usage?

packages/typespec-client-generator-core/README.md
```diff
@@ -1197,3 +1197,57 @@ model MyModel {
...
 }
 ```
+
+### Azure.ClientGenerator.Core.Legacy
+
+- [`@legacyHierarchyBuilding`](#@legacyhierarchybuilding)
+
+#### `@legacyHierarchyBuilding`
+
+Adds support for client-level multiple levels of inheritance.
```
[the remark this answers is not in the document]
tadelesh: After remoing the stuttering `legacy`, current name is `@hierarchyBuilding`.

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
tadelesh: changed to add an op to prevent miss-leading.

packages/typespec-client-generator-core/test/decorators/legacy-hierarchy-building.test.ts
```diff
@@ -0,0 +1,535 @@
...
+      alias BContent = {
+        foo: string;
+      };
+
+      model B extends A {
+        kind: "B";
+        ...BContent;
+      }
+
+      @Legacy.legacyHierarchyBuilding(B)
```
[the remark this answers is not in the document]
tadelesh: Current impl does not limit it.

packages/typespec-client-generator-core/src/interfaces.ts
tadelesh: Added.

PR #3047, opened 2025-07-24
Same change as core to publish modified packages

timotheeguerin:


eng/scripts/check-docs.js
timotheeguerin, line 1: this isn't used anymore

packages/typespec-azure-core/CHANGELOG.json
timotheeguerin, line 1: for testing teh change but also don't need those old files so good to delete

markcowl (approved):

PR #3094, opened 2025-07-31
[Spector] Add test for overload method in C# client code.

tadelesh:
Spector test for this feature: https://github.com/Azure/typespec-azure/pull/3093

### DIFF
--- a/packages/azure-http-specs/specs/client/overload/client.tsp
+++ b/packages/azure-http-specs/specs/client/overload/client.tsp
@@ -1,0 +1,11 @@
+import "./main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+using Client.Overload;
+
+namespace Customization;
+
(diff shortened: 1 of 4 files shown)


packages/azure-http-specs/specs/client/overload/client.tsp
```diff
@@ -0,0 +1,11 @@
+import "./main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+using Client.Overload;
+
+namespace Customization;
```
weidongxu-microsoft, line 7: nit, Do we need this namespace?

weidongxu-microsoft (approved):
I guess we now missed a lot of cases in summary md :-)

packages/azure-http-specs/specs/client/overload/client.tsp
tadelesh: removed

PR #3098, opened 2025-08-01
Allow using scope with Provider Action

markcowl:


packages/samples/specs/resource-manager/legacy/legacy-operations/main.tsp
```diff
@@ -0,0 +1,63 @@
...
+  ...ResourceNameParameter<
+    Resource = BestPractice,
+    KeyName = "bestPracticeName",
+    SegmentName = "bestPractices",
+    NamePattern = ""
+  >;
+}
+
+@armResourceOperations
+interface BestPracticesVersionOps
```
timotheeguerin, line 39: Alias = I think is the correct way

markcowl: This shouldn't matter as there are no concrete operations, I think.

markcowl: Yeah, I was thinking that as I wrote this - we should probably make this the norm when using the legacy template (and when using the ResourceOperations template in azure.core)

packages/samples/specs/resource-manager/legacy/legacy-operations/main.tsp
```diff
@@ -0,0 +1,108 @@
...
+
+/** Resource-specific properties of a best practice resource */
+model BestPracticeProperties {
+  ...DefaultProvisioningStateProperty;
+
+  /** The description of the best practice */
+  description: string;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Resource template"
```
timotheeguerin, line 48: does this actually emit a warning here?
Seems like we should file an issue if it does

Issue #3091, opened 2025-07-31
[TCGC] Support download files/contents in multipart/form-data


MaryGao:
### Clear and concise description of the problem

Not only uploading we could also download by multipart form data and below is the relevant typespec. TCGC also need to handle the response models correctly.

```tsp
@route("/downloadFile")
@post
op downloadFile(): {
  @header contentType: "multipart/form-data";
  @multipartBody body: {
    name: HttpPart<string>;
    file: HttpPart<bytes>[];
  };
};
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #3088, opened 2025-07-31
[Bug]: [TCGC] the model name is empty in multipart operation


MaryGao:
### Describe the bug

Once we adopted to the latest dev version, we can't get a name for TodoItem. In our previous version we could have a name for that model.

<img width="2244" height="1910" alt="Image" src="https://github.com/user-attachments/assets/3af08825-599f-46f3-98dd-0f48ed9f9be1" />

```tsp
model TodoItem {
  /** The item's unique id */
  @visibility(Lifecycle.Read) @key id: safeint;

  /** The item's title */
  @maxLength(255)
  title: string;

  /** User that created the todo */
  @visibility(Lifecycle.Read) createdBy: string;

  /** User that the todo is assigned to */
  assignedTo?: string;

  /** A longer description of the todo item in markdown format */
  description?: string;

  /** The status of the todo item */
  status: "NotStarted" | "InProgress" | "Completed";

  /** When the todo item was created. */
  @visibility(Lifecycle.Read) createdAt: utcDateTime;

  /** When the todo item was last updated */
  @visibility(Lifecycle.Read) updatedAt: utcDateTime;

  /** When the todo item was makred as completed */
  @visibility(Lifecycle.Read) completedAt?: utcDateTime;

  // Want the read form to be normalized to TodoLabelRecord[], but can't
  // https://github.com/microsoft/typespec/issues/2926
  labels?: string[];

  // hack to get a different schema for create
  // (fastify glue doesn't support readonly)
  @visibility(Lifecycle.Create) dummy?: string;
}

op try(@header contentType: "multipart/form-data",
    @multipartBody body: {
  item: HttpPart<TodoItem>;
}): void;
```

### Reproduction

NA

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For bug in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #3107, opened 2025-08-01
[TCGC] Support multipart response and fix multipart anonymous model naming empty.

tadelesh:
fix: https://github.com/Azure/typespec-azure/issues/3091
fix: https://github.com/Azure/typespec-azure/issues/3088

### DIFF
--- a/.chronus/changes/fix_multipart-2025-7-1-16-32-10.md
+++ b/.chronus/changes/fix_multipart-2025-7-1-16-32-10.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Support multipart response and fix multipart anonymous model naming empty.


packages/typespec-client-generator-core/test/types/multipart.test.ts
```diff
@@ -785,3 +785,83 @@ it("check header in multipart with @multipartBody for model", async function ()
...
+      item: HttpPart<TodoItem>;
+    }): void;
+  `);
+
+  const models = runner.context.sdkPackage.models;
+  strictEqual(models.length, 2);
+  const model = models[0];
+  strictEqual(model.kind, "model");
+  ok((model.usage & UsageFlags.MultipartFormData) > 0);
+  strictEqual(model.name, "TryRequest");
```
weidongxu-microsoft, line 851: Emm, we now had a generated model named `TryRequest`.

Emitter may see if they want to map it back to `TodoItem`. or not. (a bit similar to round-trip resource on PUT).

Though I'd assume we are less likely to encounter this case in real life, as usually the model for multipart won't be shared, and it not very meaningful to have visibility over a non-shared model.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -837,11 +837,30 @@ export function getSdkModelWithDiagnostics(
...
+        requestBody &&
+        requestBody.bodyKind === "multipart" &&
+        getHttpBodyType(requestBody) === type
+      ) {
+        // handle multipart request body model properties
+        diagnostics.pipe(
+          addMultipartPropertiesToModelType(context, sdkType, requestBody, operation),
+        );
+      } else if (multipartResponseBodies.length > 0) {
+        // handle multipart response body model properties
```
weidongxu-microsoft, line 856: I am not sure why this 2 block is `if .. else ..`, as one handle request, one handle response?

packages/typespec-client-generator-core/test/types/multipart.test.ts
tadelesh: Current fix is actually accord to the previous behavior before I did the refactor. For visibility, the model is changed from the Http lib, TCGC have no idea about the source.

packages/typespec-client-generator-core/src/types.ts
tadelesh: If a model is used both in request and response, we could do the properties handling only once.

weidongxu-microsoft (approved):

MaryGao (approved):

PR #3111, opened 2025-08-02
Move external integration to a dedicated pacakage

timotheeguerin:
This allows the test to be run very easily locally and configure private repos that wouldn't be safe to do in the public CI.

Features: 
- Clone repo at current branch, by default reuse previously cloned ones (quite time saving locally for large repos like azure-rest-api-specs)
- Link local packages to the package.json
- Install dependencies
- Scan for all tsp project with the given pattern
- Run `tsp compile .` for all specs in there
- Check no git changes
 
Config example:
```yaml
suites:
  azure-specs:
    repo: https://github.com/Azure/azure-rest-api-specs
    branch: typespec-next
    pattern: "specification/**/tspconfig.yaml"
    entrypoints: ["client.tsp", "main.tsp"]
  azure-specs-pr:
    repo: https://github.com/Azure/azure-rest-api-specs-pr
    branch: typespec-next
    pattern: "specification/**/tspconfig.yaml"
    entrypoints: ["client.tsp", "main.tsp"]
```


packages/integration-tester/src/git.ts
mikeharder, line 1: I recommend using `simple-git` rather than exec git.  Examples:

https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20simple-git&type=code

packages/integration-tester/src/validate.ts
```diff
@@ -0,0 +1,148 @@
...
+import { readdir } from "fs/promises";
+import { globby } from "globby";
+import { dirname, relative } from "pathe";
+import pc from "picocolors";
+import type { IntegrationTestSuite } from "./config/types.js";
+import type { TaskRunner } from "./runner.js";
+import { log } from "./utils.js";
+
+// Number of parallel TypeSpec compilations to run
+const COMPILATION_CONCURRENCY = 4;
```
mikeharder, line 11:
```suggestion
const COMPILATION_CONCURRENCY = os.cpus().length;
```

Why not default to number of cores?

packages/integration-tester/src/validate.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  return execTspCompile(dir, entrypoint, entrypoint === "client.tsp" ? ["--dry-run"] : []);
+}
+async function execTspCompile(
+  directory: string,
+  file: string,
+  args: string[] = [],
+): Promise<{ success: boolean; output: string }> {
+  const { failed, all } = await execa("npx", ["tsp", "compile", file, "--warn-as-error", ...args], {
```
mikeharder, line 120:
```suggestion
  const { failed, all } = await execa("npm", ["exec", "--no", "--", "tsp", "compile", file, "--warn-as-error", ...args], {
```

This is the safest and most reliable way to exec a package that should already be installed, without any chance of invalid arg parsing.  We have a helper for this:

https://github.com/Azure/azure-rest-api-specs/blob/48cba8646b6cb51ce77bd95a2a8c3dddca068459/.github/shared/src/exec.js#L121-L123

mikeharder (approved):
This has similiarities to TSV in specs repo, at some point in the future we might want to refactor / share code, but this looks good for now.

packages/integration-tester/src/validate.ts
```diff
@@ -0,0 +1,148 @@
...
+
+// Function to run tasks with limited concurrency
+async function runWithConcurrency<T, R>(
+  items: T[],
+  concurrency: number,
+  processor: (item: T) => Promise<R>,
+): Promise<R[]> {
+  const results: R[] = [];
+
+  for (let i = 0; i < items.length; i += concurrency) {
```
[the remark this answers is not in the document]
timotheeguerin: Yeah this was just done as a basic parallelisation in the original script. I'll update with using number of cores as the default number

PR #3118, opened 2025-08-04
[azure-core] Change type of etagHeader to eTag

### LINKED ISSUE #3101: [azure-core] Type inconsistent of `eTag`
https://github.com/Azure/typespec-azure/blob/6d8cf581f650eaac7297296fcf684ff4252ffb2f/packages/typespec-azure-core/lib/models.tsp#L164
The `eTag` header does not use `eTag` scalar, but use `string` directly which makes emitter could not get the correct type for eTag. Is it by design?

witemple-msft:
Let's just see if this affects anything. This probably should have just been using the eTag type.

Closes #3101

### DIFF
--- a/packages/typespec-azure-core/lib/models.tsp
+++ b/packages/typespec-azure-core/lib/models.tsp
@@ -161,7 +161,7 @@ model EtagResponseEnvelope {
   @header("ETag")
   @visibility(Lifecycle.Read)
   @doc("The entity tag for the response.")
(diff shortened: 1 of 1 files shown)


markcowl (commented):
Need to add a changelog, and merge a PR into typespec-next that fixes any failures.  

This should prsent no API breaking changes, but might present SDK breaking changes if any SDK has an `eTag` type

Issue #1138, opened 2024-07-09
[Bug][ARM] LocationResourceParameter should be azureLocation instead of string


live1206:
### Clear and concise description of the problem

For .NET SDK generation, we would like to have the correct type in spec.
location should be `azureLocation` instead of `string`  in [common-types](https://github.com/Azure/typespec-azure/blob/edf8c73eb61e261cf66adb1a5959a43e730ce12f/packages/typespec-azure-resource-manager/lib/common-types/types.tsp#L515)

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3136, opened 2025-08-08
update location type to azureLocation

live1206:
`location` in `TrackedResource` should be `azureLocation`
This change is follow-up for https://github.com/Azure/typespec-azure/issues/1138, we just found more cases.


markcowl (changes requested):
I think there are too many changes in this - can we scope the changes to just this change?

PR #3138, opened 2025-08-08
update location type to azureLocation

live1206:
`location` in `TrackedResource` should be `azureLocation`
This change is follow-up for https://github.com/Azure/typespec-azure/issues/1138, we just found more cases.


markcowl:
closing as this was already released.

PR #3143, opened 2025-08-10
Add legacy templates for optional location and etags

markcowl:


packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -302,6 +302,25 @@ model EntityTagProperty {
...
+ * ```typespec
+ * model Foo is TrackedResource<FooProperties> {
+ *   // Only have standard Succeeded, Failed, Cancelled states
+ *   ...ETagProperty;
+ * }
+ * ```
+ */
+@doc("The etag property envelope.")
+model ETagProperty {
+  @doc("If etag is provided in the response body, it may also be provided as a header per the normal etag convention.  Entity tags are used for comparing two or more entities from the same requested resource. HTTP/1.1 uses entity tags in the etag (section 14.19), If-Match (section 14.24), If-None-Match (section 14.26), and If-Range (section 14.27) header fields.")
```
timotheeguerin, line 319: Doc comment instead?

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -302,6 +302,25 @@ model EntityTagProperty {
...
+ *
+ * ```typespec
+ * model Foo is TrackedResource<FooProperties> {
+ *   // Only have standard Succeeded, Failed, Cancelled states
+ *   ...ETagProperty;
+ * }
+ * ```
+ */
+@doc("The etag property envelope.")
+model ETagProperty {
```
timotheeguerin, line 318: Worried here this will be easily used instead of the other one, this feels like a more intuitive name.

Is that for legacy scenario only? Do we really need a template just that that?

packages/typespec-azure-resource-manager/lib/models.tsp
```diff
@@ -302,6 +302,25 @@ model EntityTagProperty {
...
+ * @example
+ *
+ * ```typespec
+ * model Foo is TrackedResource<FooProperties> {
+ *   // Only have standard Succeeded, Failed, Cancelled states
+ *   ...ETagProperty;
+ * }
+ * ```
+ */
+@doc("The etag property envelope.")
```
timotheeguerin, line 317: I think this doc is unnecessary as it should always be spread which make me think maybe we should make those aliases of model expr instead

Issue #3162, opened 2025-08-14
Need linter warning for namespace and models with the same name


JoshLove-msft:
In C# (and possibly other languages) having a namespace named the same as a model (whether via clientName or directly in the spec) will result in build errors. We should warn about this up front to save time and rework.

markcowl:
@markcowl see if any duplicate issues

PR #3163, opened 2025-08-14
Rename namespace to avoid conflict

JoshLove-msft:
Renaming the namespace to "ClientModel" causes a conflict with a model in the spec. This causes build errors in C# and it is not necessary for what this scenario is trying to exercise.

Filed https://github.com/Azure/typespec-azure/issues/3162

### DIFF
--- a/packages/azure-http-specs/specs/client/naming/main.tsp
+++ b/packages/azure-http-specs/specs/client/naming/main.tsp
@@ -144,7 +144,7 @@ namespace Header {
 
 @route("/model")
 @operationGroup
-@clientName("ClientModel")
+@clientName("ModelClient")
 namespace Model {
   @clientName("CSModel", "csharp")
   @clientName("GoModel", "go")


weidongxu-microsoft:
Can we trigger a release of http-specs lib?

The change breaks Java nightly.
Or, can I revert Java scope to original name?

weidongxu-microsoft:
https://github.com/Azure/typespec-azure/pull/3164

JoshLove-msft:
> Can we trigger a release of http-specs lib?

Yes, I can queue a new release with these changes. IMO it is not worth special-casing specific languages to avoid updating a couple tests. Since the scenario is not explicitly exercising the fact that the name of the namespace and the model can match, we can remove this from the test. That said, I don't think it is a big deal one way or the other - but my preference is to take the one time hit rather than adding drift that will exist indefinitely.

JoshLove-msft:
> > Yes, I can queue a new release with these changes. IMO it is not worth special-casing specific languages to avoid updating a couple tests. Since the scenario is not explicitly exercising the fact that the name of the namespace and the model can match, we can remove this from the test. That said, I don't think it is a big deal one way or the other - but my preference is to take the one time hit rather than adding drift that will exist indefinitely.
> 
> Yeah, please trigger a release. Thanks.
> 
> PS: The reason we'd prefer not to break in http-specs / azure-http-specs is that
> 
> 1. If we do not release azure-http-specs after break, the nightly build of many emitters would fail (because test code still uses the pre-break code). Then we can no longer catch other breaks (e.g. from compiler or from TCGC).
> 2. If we do a release, we risk shipping changes from other source (which you already saw in last month, that a release of azure-http-specs causes ` ../node_modules/@azure-tools/azure-http-specs/specs/azure/core/page/main.tsp:89:4 - error missing-paging-items: Paged operation 'listWithPage' return type must have a property annotated with @pageItems.` to emitters

https://www.npmjs.com/package/@azure-tools/azure-http-specs/v/0.1.0-alpha.27 is available now.

PR #3164, opened 2025-08-15
Fix spector

msyyc:
fix nightly build failure caused by https://github.com/Azure/typespec-azure/pull/3163


weidongxu-microsoft (approved):

JoshLove-msft:
IMO it is not worth special-casing specific languages to avoid updating a couple tests. Since the scenario is not explicitly exercising the fact that the name of the namespace and the model can match, we can remove this from the test. That said, I don't think it is a big deal one way or the other - but my preference is to take the one time hit rather than adding drift that will exist indefinitely.

PR #3173, opened 2025-08-19
Allow autorest to inline Azure.Core.azureLocation

markcowl:


packages/typespec-autorest/src/openapi.ts
```diff
@@ -965,6 +967,16 @@ export async function getOpenAPIForService(
...
   }
+  function shouldInlineCoreScalarProperty(type: Type): boolean {
+    if (
+      type.kind !== "ModelProperty" ||
+      type.type.kind !== "Scalar" ||
+      type.type.namespace === undefined
+    )
+      return false;
+    const nsName = getNamespaceFullName(type.type.namespace);
+    return nsName === "Azure.Core" && getInlineAzureType(program, type) === true;
```
timotheeguerin, line 978: Wouldn't just checking the decorator be enough?

timotheeguerin (approved):

packages/typespec-autorest/src/openapi.ts
markcowl: I don't want to change the way it processes other kinds of scalars.  This is kind of a compromise to allow updating some common-types to using azure.Core types without impacting any existing specs (whichh may also, separately use azure.Core types)

PR #3178, opened 2025-08-20
Allow renaming path parameters in an ARM operation

markcowl:


packages/typespec-azure-resource-manager/src/operations.ts
```diff
@@ -513,3 +514,28 @@ export function getRouteOptions(program: Program, target: Operation): ArmOperati
...
+ * @returns
+ */
+export const $renamePathParameter: RenamePathParameterDecorator = (
+  context: DecoratorContext,
+  target: Operation,
+  sourceParameterName: string,
+  targetParameterName: string,
+) => {
+  if (target.parameters.properties.has(targetParameterName)) {
+    return;
```
AlitzelMendez, line 533: should we have a linter rule that says that the name is already in use? because just returning is not given any information on why the name wasn't change

AlitzelMendez (approved):

Issue #3071, opened 2025-07-29
[TCGC] Wrong `correspondingMethodParams` for `@override`


msyyc:
For [playground](https://azure.github.io/typespec-azure/playground/?e=%40azure-tools%2Ftypespec-client-generator-core&c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsNCsk3yCsvaHR0cMQaDQp1c2luZyBBxFAuQ8VBR8hALkNvcmU7yCNIdHRwxTJAc2VydmljZQ0KbmFtZXNwYWNlIFRlc3TFHW1vZGVsIE3FBnsNCiAgcHJvcDogc3RyaW5nOw0KfcZCaGFyZWRSb3V0ZQ0KQGdldCBvcCBvcmlnaW5hbChAYm9keSDEBTrGSik6IHZvaWTHftQ8bmV3xzd0ZXN01DdAb3ZlcnJpZGUoyGgsxDYpOw%3D%3D&options=%7B%7D&vs=%7B%7D), TCGC outputs wrong `correspondingMethodParams`:

<img width="817" height="586" alt="Image" src="https://github.com/user-attachments/assets/5a4f7dcb-996c-4596-9664-6fb3ceaf705a" />

<img width="1048" height="695" alt="Image" src="https://github.com/user-attachments/assets/5c1d9ab8-059c-42a5-9b92-ce26c60bdb22" />

tadelesh:
[...]

PR #3189, opened 2025-08-21
[TCGC] Compare type when map body parameter to method parameter.

tadelesh:
This is a behavior breaking change. The Http parameter's corresponding parameter may change. It mostly happens in `@override` scenario. See the test change in the PR for the breaking change impact.

fix: https://github.com/Azure/typespec-azure/issues/3071

### DIFF
--- a/packages/typespec-client-generator-core/src/http.ts
+++ b/packages/typespec-client-generator-core/src/http.ts
@@ -716,6 +716,10 @@ function findMapping(
     ) {
       return methodParam;
     }
+    // If the service parameter is a body parameter, try to see if we could find a method parameter with same type of the body parameter.
+    if (serviceParam.kind === "body" && serviceParam.type === methodParam.type) {
+      return methodParam;
+    }
     // BFS to find the mapping.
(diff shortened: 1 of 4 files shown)


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -716,6 +716,10 @@ function findMapping(
     ) {
       return methodParam;
     }
+    // If the service parameter is a body parameter, try to see if we could find a method parameter with same type of the body parameter.
+    if (serviceParam.kind === "body" && serviceParam.type === methodParam.type) {
+      return methodParam;
+    }
```
weidongxu-microsoft, line 722: I am not sure why the BFS approach produce different result than this "shortcut" approach.

weidongxu-microsoft (approved):

msyyc (approved):
Test this PR locally and it works as expected for python emitter.

packages/typespec-client-generator-core/src/http.ts
tadelesh: The BFS only compare the parameter/property, but for this special case, we use the parameter's type directly.

PR #3199, opened 2025-08-26
[tcgc] add `.pageSizeParameter` to paging metadata

### LINKED ISSUE #2778: [tcgc] label page size parameter
Some languages need to know if a parameter is the page size parameter for correct generated code mapping. Consider adding a `.pageSizeParameter` property either on parameters, or on `.pagingMetadata` (I think this latter is preferred so paging information is coalesced in one place).

Use the decorator [`@pageSize`](https://typespec.io/docs/standard-library/built-in-decorators/#@pageSize) to determine whether the parameter is the page size param or not

iscai-msft:
fixes #2778

### DIFF
--- a/packages/typespec-client-generator-core/src/interfaces.ts
+++ b/packages/typespec-client-generator-core/src/interfaces.ts
@@ -893,6 +893,8 @@ export interface SdkPagingServiceMetadata<TServiceOperation extends SdkServiceOp
   continuationTokenResponseSegments?: (SdkServiceResponseHeader | SdkModelPropertyType)[];
   /** Segments to indicate how to get page items from response. */
   pageItemsSegments?: SdkModelPropertyType[];
+  /** Denotes which parameter is the page size parameter */
+  pageSizeParameter?: SdkModelPropertyType;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -893,6 +893,8 @@ export interface SdkPagingServiceMetadata<TServiceOperation extends SdkServiceOp
   continuationTokenResponseSegments?: (SdkServiceResponseHeader | SdkModelPropertyType)[];
   /** Segments to indicate how to get page items from response. */
   pageItemsSegments?: SdkModelPropertyType[];
+  /** Denotes which parameter is the page size parameter */
+  pageSizeParameter?: SdkMethodParameter;
```
tadelesh, line 897: Since `@pageSize` could be defined nested, can we keep the same patten `pageSizeSegments` as other paging metadata?

iscai-msft: ooh that's a good point, thanks!

JoshLove-msft (approved):

PR #3202, opened 2025-08-27
[tcgc] move `@flattenProperty` to legacy namespace

### LINKED ISSUE #2791: [tcgc] transfer `@flattenProperty` to `.Legacy` namespace
Right now, we have the `@flattenProperty` dec in tcgc marked as deprecated to discourage usage for greenfield services. However, this is confusing since it seems the documentation makes it seem like the dec will be removed at a later point, when that won't be the case since so many brownfield services depend on it. To suppress usages right now, spec authors have to write `#suppress deprecated` instead of a more specific linter warning as well.

Instead, what we should do is move the property to `Azure.ClientGenerator.Core.Legacy`, (ex for `Azure.Core.Legacy` [here](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/legacy.tsp)). We are already throwing linter errors if anyone uses something from the `.legacy` namesapce ([here](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/src/rules/no-legacy-usage.ts)), so they will suppress it with `#suppress @azure-tools/typespec-azure-core/no-legacy-usage`, which seems good.

TODO:
1. Create `@Azure.ClientGenerator.Core.Legacy.flattenProperty`
2. Deprecate `@Azure.ClientGenerator.Core.flattenProperty` and move people to the above

iscai-msft:
fixes #2791

### DIFF
--- a/packages/typespec-client-generator-core/test/decorators/flatten-property.test.ts
+++ b/packages/typespec-client-generator-core/test/decorators/flatten-property.test.ts
@@ -3,6 +3,7 @@ import { strictEqual } from "assert";
 import { beforeEach, it } from "vitest";
 import { getAllModels } from "../../src/types.js";
 import { createSdkTestRunner, SdkTestRunner } from "../test-host.js";
+import { AzureCoreTestLibrary } from "@azure-tools/typespec-azure-core/testing";
 
 let runner: SdkTestRunner;
 
@@ -13,8 +14,7 @@ beforeEach(async () => {
 it("marks a model property to be flattened with suppression of deprecation warning", async () => {
   await runner.compileWithBuiltInService(`
     model Model1{
-      #suppress "deprecated" "@flattenProperty decorator is not recommended to use."
-      @flattenProperty
+      @Azure.ClientGenerator.Core.Legacy.flattenProperty
       child: Model2;
     }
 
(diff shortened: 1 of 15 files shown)


live1206:
I don't think this is a blocker for .NET since we handle breaking changes from time to time with TCGC.
But it does take much extra effort than Java, because we can't simply bump the commit id to latest main in spec for all RPs like Java. Because the latest commit might have breaking changes for that RP, .NET need to mitigate the breaking changes after bumping to the latest commit.
The process we used to do is to fork from the previous commit for that RP, and apply the changes to the spec and use that commit id instead.
I wish TCGC could be stable at some point, so .NET wouldn't suffer from this anymore.

packages/typespec-client-generator-core/test/decorators/flatten-property.test.ts
```diff
@@ -133,3 +108,44 @@ it("throws error when used on a polymorphism type", async () => {
     code: "@azure-tools/typespec-client-generator-core/flatten-polymorphism",
   });
 });
+
+it("verify diagnostic gets raised for usage", async () => {
```
tadelesh, line 112:
```suggestion
it("verify diagnostic gets raised for flatten", async () => {
```

tadelesh (approved):

tadelesh:
Before merge, do we need to have a PR to migrate all cases in current spec repo?

iscai-msft:
will do those now, wanted to hold off on discussion on whether we would make this change. Seems like we're good to go ahead with the change, so let me make the PRs now

PR #3232, opened 2025-09-04
[tcgc] raise validation warning for `@clientName` applied to override methods

### LINKED ISSUE #3231: [Bug]: clientName can't be applied to the override method
### Describe the bug

When specifying `clientName` on an operation that's also `override`, you must specify the client name on the original operation, not the override. This is unintuitive. Ideally the `clientName` should apply to the override and if not, then an error diagnostic should be produced.

### Reproduction

https://azure.github.io/typespec-azure/playground/?e=%40azure-tools%2Ftypespec-client-generator-core&c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsNCsk3yCsvaHR0cMQaDQp1c2luZyBBxFAuQ8VBR8hALkNvcmU7yCNUeXBlU3BlYy5IdHRwxTtAc2VydmljZQ0KbmFtZXNwYWNlIEtleVZhdWx0IHsNCiAgQHJvdXRlKCIvc2VjcmV0cy97xgktxDF9yQ52ZXJzaW9ufSIpxTVwYXRjaCgjeyBpbXBsaWNpdE9wdGlvbmFsaXR5OiB0cnVlIH3FKm9wIHVwZGF0ZVPFSCjEFMZAaCgiy2oiKSDGDk5hbWU6IHN0cmluZyzULucAiskxVsYQP801KTogdm9pZDsNCn3kAQfqAP1DdXN0b21pemHkAK9z%2FwED%2FwED%2FwED%2FwED%2FwED%2FwED9AED9wECxA5AQG92ZXJyaWRlKOgB%2Fy7sAJUs7QCjKTvGNOYCruQAjyjOLSLMD1Byb3BlcnRpZXMixDl9DQo%3D&options=%7B%7D&vs=%7B%7D

Note that the method name is still `updateSecret`, it should be `updateSecretProperties`.

### Checklist

iscai-msft:
fixes #3231

### DIFF
--- a/.chronus/changes/tcgc-clientNameOverride-2025-8-4-13-24-44.md
+++ b/.chronus/changes/tcgc-clientNameOverride-2025-8-4-13-24-44.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Add warning for `@clientName` that is applied to override methods bc they're ineffectual


jhendrixMSFT:
Do we lint for such warnings in the specs repo?  Maybe a better way to say it is, how can we catch such authoring errors before merging a tsp change where `@@clientName` will have no effect?

PR #3233, opened 2025-09-04
[tcgc] add retype capability to `@alternateType`

### LINKED ISSUE #3146: Add refType decorator
Add a refType decorator that can be applied to properties/models to indicate to emitters that an existing reference type should be used rather than the type specified in the spec.

e.g.

```
@@refType(brokerProperties.event, "azure.core.messaging.CloudEvent", "python")
```

or 
```
@@refType(dfeModel.intProperty, "Azure.Core.Expressions.DataFactoryExpression<int>", "csharp")
```

We may want to allow it to apply to entire models as well.

iscai-msft:
fixes #3146

### DIFF
--- a/packages/typespec-client-generator-core/README.md
+++ b/packages/typespec-client-generator-core/README.md
@@ -280,24 +280,25 @@ op func8(@body body: Test5): void;
 
 #### `@alternateType`
 
-Set an alternate type for a model property, Scalar, or function parameter. Note that `@encode` will be overridden by the one defined in the alternate type.
+Set an alternate type for a model property, Scalar, Model, Enum, Union, or function parameter. Note that `@encode` will be overridden by the one defined in the alternate type.
 When the source type is `Scalar`, the alternate type must be `Scalar`.
+The replaced type could be a type defined in the TypeSpec or an external type declared by type identity, package that export the type and package version.
 
 ```typespec
-@Azure.ClientGenerator.Core.alternateType(alternate: unknown, scope?: valueof string)
+@Azure.ClientGenerator.Core.alternateType(alternate: unknown | Azure.ClientGenerator.Core.ExternalType, scope?: valueof string)
 ```
 
 ##### Target
 
 The source type to which the alternate type will be applied.
-`ModelProperty | Scalar`
+`ModelProperty | Scalar | Model | Enum | Union`
 
 ##### Parameters
 
-| Name      | Type             | Description                                                                                                                                                                                                                                                |
-| --------- | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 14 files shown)


packages/typespec-client-generator-core/README.md
```diff
@@ -280,24 +280,25 @@ op func8(@body body: Test5): void;
...
+| Name      | Type                                           | Description                                                                                                                                                                                                                                                |
+| --------- | ---------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| alternate | `unknown \| ClientGenerator.Core.ExternalType` | The alternate type to apply to the target. Can be a TypeSpec type or an ExternalType.                                                                                                                                                                      |
```
johanste, line 300: Why is this typed as `unknown`?

iscai-msft: it should be `Type` but I'm not able to do `TypeSpec.Reflection.Type`, and the guidance at one point was `unknown`. Is there updated guidance in regards to this?

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/data-types.md
```diff
@@ -34,6 +34,20 @@ model Azure.ClientGenerator.Core.ClientOptions
...
+
+```typespec
+model Azure.ClientGenerator.Core.ExternalType
+```
+
+#### Properties
+
+| Name               | Type     | Description |
+| ------------------ | -------- | ----------- |
+| fullyQualifiedName | `string` |             |
```
JoshLove-msft, line 47: Do we need to fill out the descriptions? In particular, I think that for package/version these may or may not be respected by individual languages so it would be good to use language like "If supported" for these two properties.

iscai-msft:
> I would consider supporting a type moniker for well-known types and not require a direct reference to the type in question. Consider services that expose [fhir](https://www.hl7.org/fhir/) or GeoJSON models - being able to annotate that this is a model in said space without tying it to a specific library type allows the spec author to indicate what well-known schema this represents without having to map it to a concrete library type. An emitter can then use a mapping from well-known schema types to the chosen library to map it into independently.
> 
> This doesn't have to be supported from day 1, but we should design in a way that the support can be added in the future.

That makes sense. If I switch the name to `identity` instead of `fullyQualifiedName`, this scenario will make more sense during authoring, because what we require would then be some form of identification of a known object, and emitters can do with that object  as they please

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -612,11 +612,18 @@ extern dec clientNamespace(
   scope?: valueof string
 );
 
+model ExternalType {
+  fullyQualifiedName: string;
+  package?: string;
+  version?: string;
```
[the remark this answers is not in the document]
iscai-msft: that's a great point, updating

PR #3300, opened 2025-09-16
[tcgc] add `@markAsLro` decorator

### LINKED ISSUE #3256: [TCGC] Add new decorator @legacy.lro to force generation of LRO SDK pattern while the spec doesn't specify it
There are situations where we can't declare a LRO in TypeSpec. The point of this issue is not to debate that point, so I won't go there here. Please don't comment on this issue about why TypeSpec can't declare some operation LRO.

In those situations, sometimes we want the SDK to generate a LRO pattern (using a poller) anyway, for backward compatibility of SDK.

This issue is to add a new decorator, that will indicate to TCGC to expose an operation a as LRO one. 

For implementation detail, my hope is that TCGC will be able to return a metadata structure similar to getLROMetadata, so emitters don't have to do the work to revamp anything, and it just works. If that's not possible for some reason, we should still aim to have TCGC mock as best as it could the getLROMetadata, maybe adding a wrapper like:

```ts
def getTCGCLROMetadata(operation):
    if hasTCGCLRODecorator(operation):
        return mockLROMetadata(operation)
    return getLroMetadata(operation)
```

And all emitters would call that.

iscai-msft:
fixes #3256


packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1419,6 +1421,40 @@ export function getLegacyHierarchyBuilding(context: TCGCContext, target: Model):
...
+    reportDiagnostic(context.program, {
+      code: "invalid-mark-as-lro-target",
+      format: {
+        operation: target.name,
+      },
+      target: context.decoratorTarget,
+    });
+    return;
+  }
+  if (getLroMetadata(context.program, target)) {
```
tadelesh, line 1441: If an operation is not defined according to the LRO pattern, I doubt if `getLroMetadata` could get the correct result. It is not safe to use the result directly, especially for the final result.

packages/typespec-client-generator-core/test/methods/lro.test.ts
```diff
@@ -1162,3 +1162,155 @@ describe("getLroMetadata", () => {
...
+
+    const methods = basicRunner.context.sdkPackage.clients[0].methods;
+    strictEqual(methods.length, 1);
+
+    const method = methods[0];
+    strictEqual(method.kind, "lro");
+    strictEqual(method.name, "startDeployment");
+
+    // Check LRO metadata exists
+    const metadata = method.lroMetadata;
```
tadelesh, line 1208: I doubt if the final result is correct.

iscai-msft: it has to be, since that's what we return for the final result. I'll add another check to the test though

website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/decorators.md
```diff
@@ -1177,3 +1177,43 @@ model SportsCar extends Vehicle {
...
+even when the operation is not long-running on the service side.
+
+When applied, TCGC will treat the operation as an LRO and SDK generators should:
+
+- Generate polling mechanisms (pollers)
+- Return appropriate LRO-specific return types
+- Handle the operation as an asynchronous long-running process
+
+This decorator is considered legacy functionality and should only be used when
+standard TypeSpec LRO patterns are not feasible.
```
johanste, line 1193: I would add a bigger warning blurb about the requirement to verify the operation/add tests for the generated code. Else, there is a definite risk that it gets added to operations that only worked by accident in autorest classic.

PR #3306, opened 2025-09-17
[specs] add reftype tests

### LINKED ISSUE #3259: Doc and spector test for `refType`
Fork from https://github.com/Azure/typespec-azure/issues/3146.

iscai-msft:
fixes #3259


packages/azure-http-specs/specs/azure/client-generator-core/alternate-type/client.tsp
```diff
@@ -0,0 +1,40 @@
...
+  "python"
+);
+
+@@alternateType(_Specs_.Azure.ClientGenerator.Core.AlternateType.ExternalType.Feature,
+  {
+    identity: "com.bedatadriven.jackson.datatype.geo.Feature",
+    package: "com.bedatadriven.jackson.datatype.geo",
+    minVersion: "1.0.0",
+  },
+  "java"
```
weidongxu-microsoft, line 21: Unsure where we get these. The `package` appears not a valid one. And mvn artifact needs a groupId too.

The nearest one I can find according to the namespace is this
https://mvnrepository.com/artifact/com.bedatadriven/jackson-datatype-jts/2.4
```
<dependency>
    <groupId>com.bedatadriven</groupId>
    <artifactId>jackson-datatype-jts</artifactId>
    <version>2.4</version>
</dependency>
```
which is years old and got lots of CVE.

@srnagar thoughts?
PS: Type "Feature" seems not in our core https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/models/GeoObject.java
But it does exist in other libs https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/maps/azure-maps-search/src/main/java/com/azure/maps/search/models/GeoJsonFeature.java

iscai-msft: @srnagar how does this look to you? We don't have to go 3p, it's more about referencing a specific type in your language that already exists. Thanks so much for your help!

weidongxu-microsoft: I am kind of fine with this. The `GeoObject` here cannot be serialized or de-serialized to `Feature` type, so Java will not pass the test. But I guess the codegen should work. If it still does not work (as `GeoObject` is an abstract class), I can change it to `GeoPoint` later.

weidongxu-microsoft (approved):

PR #3315, opened 2025-09-19
refactor the output models of resolveArmResources

ArcturusZhang:


packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -115,37 +115,33 @@ export interface ArmVirtualResourceDetails {
...
   /** The model type for the resource */
   type: Model;
   /** The kind of resource (extension | tracked | proxy | custom | virtual | built-in) */
   kind: ArmResourceKind;
   /** The provider namespace */
   providerNamespace: string;
-}
-
-/** New details for a resolved resource */
-export interface ResolvedResource extends ResourceMetadata {
```
markcowl, line 129: This won't work.  `ResolvedResource` is a set of operations that applies over a single resource/type or resource instance path.

Here there is one ResourceModel  that corresponds to multiple resource types, and each of those resource types has a set of operations.  The interfrace containg the actual ResourceModel needs to be at the top level.

We can reorganize where some of the properties are, but the basic one-> many relationship betweem ResourceModels -> ResourceTypes -> ResourceOperations  won't change.

markcowl: Renaming to make this more intuitive is a good thing to do,  and moving some of the metadata could make sense as long as it doesn't break the one-to-many relationships.

ArcturusZhang:
Close this since Mark has done some renames.

PR #3330, opened 2025-09-24
Fix #3294 Add support for private links

### LINKED ISSUE #3294: Support privateLinkResource as resource
Similar to PrivateEndpintConnection as required by containerregistry https://github.com/Azure/azure-rest-api-specs/pull/37169#discussion_r2350222719

markcowl:


kazrael2119:
I try to use this pr to update my typespec code like:
image

then I get this diff between new generated swagger and origin generated swagger:
image

could I customize the name of the resource model for when using arm privateLinkResource template？

PR #3339, opened 2025-09-25
Migrate arm testerv2 

timotheeguerin:


packages/typespec-azure-resource-manager/test/resource.test.ts
```diff
@@ -823,68 +672,41 @@ interface RestorePointOperations {
 
     `);
 
-    const resources = getArmResources(program);
-    expectDiagnosticEmpty(diagnostics);
```
markcowl, line 827: I notice the expectDiagnosticEMpty being removed in some of these tests, is this because of the removal of docs, or does the new tester automatically check for this so it's no longer necessary?

packages/typespec-azure-resource-manager/test/resource.test.ts
```diff
@@ -1533,127 +1105,11 @@ it("recognizes resource with customResource identifier", async () => {
...
       model Person {
-        /** The parent name */
         name: string;
       }
     }
 `);
   expectDiagnosticEmpty(diagnostics);
 });
-
-describe("typespec-azure-resource-manager: identifiers decorator", () => {
```
markcowl, line 1547: where did these tests end up?

packages/typespec-azure-resource-manager/test/resource.test.ts
timotheeguerin: Nothing actually changed here with the new tester but if you use .compile it automatically validate no failures which I changed to use that instead of compileAndDiagnose

markcowl (approved):

PR #3350, opened 2025-09-29
Remove openapi from azure core and arm libraries

timotheeguerin:


weidongxu-microsoft:
I've took a quick search, Java emitter does not use openapi.

PS: we had it in emitter dependency but mostly just to pass the dependency to emitter-package in SDK repo and make compiler pass when source uses this lib (e.g. `@OpenAPI.extension("x-ms-identifiers", #[])`).

markcowl (approved):

PR #3358, opened 2025-10-01
Brownfield additions for typespec-autorest

### LINKED ISSUE #3313: add support for object secret

In Container Registry service, there are a property customHeaders that is marked as x-ms-secret over an object with additional properties [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/preview/2025-05-01-preview/containerregistry.json#L7529-L7536)

```
        "serviceUri": {
          "description": "The service URI for the webhook to post notifications.",
          "type": "string",
          "x-ms-secret": true
        },
        "customHeaders": {
          "description": "Custom headers that will be added to the webhook notifications.",
          "type": "object",
          "additionalProperties": {
            "type": "string"
          },
          "x-ms-secret": true
        }
```

But if we convert this customHeaders to typespec, the type will be a `Record<string>`, while will report errors 

Image

I have confirmed with service team that this is intentional, because the customer specifies the endpoint and several properties used in the request, including request headers, secrets may be present in either the ServiceUri or any of the custom header. 
- It could be any of the properties provided, including the ServiceURI or any of the header keys or values. In TypeSpec what is the recommendation/guideline on how we should indicate that all properties of an object (or all keys and values in a map) may be secrets?

This x-ms-secret is required from [ARM guideline](https://eng.ms/docs/products/arm/api_contracts/guidelines/openapi#oapi013-secrets-must-be-annotated-with-x-ms-secret-and-x-ms-mutability ) for secrets related types. 

Therefore, it's a valid feature ask and currently it's blocking the container registry service typespec conversion. 

please proceed it with high priority. Thank you!

### Checklist

### LINKED ISSUE #3322: [autorest] integrate `@Azure.ClientGenerator.Core.Legacy.markAsLro` in emitter
The [`@markAsLro`](https://github.com/Azure/typespec-azure/blob/release/september-2025/packages/typespec-client-generator-core/lib/legacy.tsp#L99) decorator in hot fix version `0.60.2` should be integrated into the auto rest emitter, with operations marked with `@markAsLro` adding `x-ms-long-running-operation: true` into the emitted swaggers

markcowl:
Fix #3322 Support legacy get lro
Fix #3313 Support for object secret


packages/typespec-autorest/test/models.test.ts
```diff
@@ -22,6 +22,26 @@ describe("typespec-autorest: model definitions", () => {
...
+  it("allows secrets on models", async () => {
+    const res = await oapiForModel(
+      "Foo",
+      `@secret
+       model Foo {
+        x: int32;
+      };`,
+    );
+
+    ok(res.isRef);
```
timotheeguerin, line 34:
```suggestion
```
don't feel like those are really necessary checks

timotheeguerin (approved):

PR #3410, opened 2025-10-17
Uptake fix from compiler

timotheeguerin:
Arm library was using a broken pattern with interface and operations that was fixed to error out in the compiler, this migrate to an alternative until we have a better solution to do that inline

### DIFF
--- a/.chronus/changes/copilot-fix-inconsistency-main-tsp-mockapi-2025-9-21-6-59-44.md
+++ b/.chronus/changes/copilot-fix-inconsistency-main-tsp-mockapi-2025-9-21-6-59-44.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/azure-http-specs"
+---
+
+Fix URI inconsistency in alternate-type mockapi.ts to match main.tsp definition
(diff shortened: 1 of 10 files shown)


packages/typespec-client-generator-core/test/package/azure-widget-service.test.ts
```diff
@@ -474,19 +474,19 @@ it("lro delete", async () => {
...
-      op delete is ResourceOperations.LongRunningResourceDelete<Widget>;
+      op delete is Operations.LongRunningResourceDelete<Widget>;
       `,
   );
   const method = getServiceMethodOfClient(runnerWithCore.context.sdkPackage);
   strictEqual(method.name, "delete");
   strictEqual(method.kind, "lro");
   strictEqual(method.response.type, undefined);
-  strictEqual(runnerWithCore.context.sdkPackage.models.length, 3);
+  strictEqual(runnerWithCore.context.sdkPackage.models.length, 4);
```
timotheeguerin: this now adds the ErrorResponseBase which does get instantiated when we correctly instantiate the interface.

I don't really get what those test are really testing with checking the number of models

PR #3411, opened 2025-10-17
Add new secret prop rule

### LINKED ISSUE #3403: Add linting tule for new RPC rule RPC-v1-13
This requires any property that occurs in a response that looks like a credential to be marked as `x-ms-secret`  (in TypeSpec, this means using the password type or`@secret` decorator).

In this case 'looking like' a credential means that the property name contains one of the following values, either as a substring, or as the entire name:

- `password`
- `key`
- `access`
- `credential`
- `token`
- `secret`
- `auth`

The corresponding ci check is here: https://github.com/Azure/azure-openapi-validator/pull/800

Also add appropriate documentation page and codefix

timotheeguerin:
fix #3403


packages/typespec-azure-resource-manager/src/rules/secret-prop.ts
```diff
@@ -0,0 +1,76 @@
...
+  "password",
+  "key",
+  "token",
+  "auth",
+  "connection",
+].map((keyword) => keyword.toUpperCase());
+
+function isPotentialSensitiveProperty(propertyName: string): boolean {
+  const upperName = propertyName.toUpperCase();
+  return sensitiveKeywords.some((keyword) => upperName.includes(keyword));
```
markcowl: I think we decided on endsWith for this

timotheeguerin: yeah forgot to push

packages/typespec-azure-resource-manager/README.md
```diff
@@ -60,6 +60,7 @@ Available ruleSets:
...
 | [`@azure-tools/typespec-azure-resource-manager/unsupported-type`](https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/unsupported-type)                                   | Check for unsupported ARM types.                                                                                                                                                                                                                      |
+| [`@azure-tools/typespec-azure-resource-manager/secret-prop`](https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/secret-prop)                                             | Check that property with names indicating sensitive information(e.g. contains auth, password, token, secret, etc.) are marked with @secret decorator.                                                                                                 |
```
markcowl: Add RPC-v1-13  into the text of the message, so we can track that this is an RPC violation

packages/typespec-azure-resource-manager/test/rules/secret-prop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    )
+    .toBeValid();
+});
+
+it("doesn't flag if the property is not used in output", async () => {
+  await tester
+    .expect(
+      `
+    model Test { password: string; }
+    op test(test: Test, key: string): void;
```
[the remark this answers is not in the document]
timotheeguerin: hhm I think this rule doesn't really matter how things are built or related too much to arm (it even feels like we could make this an azure core rule)

timotheeguerin:
@microsoft-github-policy-service run

PR #3448, opened 2025-10-27
[tcgc] add files as singular input / output

### LINKED ISSUE #3446: [tcgc] add full support for files as input or output
Right now, tcgc supports files as part of a multipart request. However, we'd also like to add support for this as just a file in the request or response, i.e. `op foo(request: Http.File): void`. Right now, if someone defines this, we throw [`unsupported-http-file-body`](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/lib.ts#L368).

We want to add support for this, we should do what we do for files in multipart bodies without the multipart wrapping (see [here](https://azure.github.io/typespec-azure/playground/?e=%40azure-tools%2Ftypespec-client-generator-core&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CnVzaW5nIEh0dHA7CgpAZG9jKCJUZXN0IGZvciBtdWx0aXBhcnQiKQpAc2VydmljZQpuYW1lc3BhY2UgUGF5bG9hZC5NxCdQYXJ0OwrNQ2NvbnRlbnQtxHM6yk0vZm9ybS1kYXRhxFdwb3N0Cm9wIHRlc3QoCiAgQGhlYWRlcsg9VMU8ItQ9LMQuyRlCb2R5IGJvZHk65QDN5ACPPEZpbGU%2BLAopOiBOb0PGUFJlc3BvbnNlOw%3D%3D&options=%7B%7D&vs=%7B%7D))

iscai-msft:
fixes #3446

### DIFF
--- a/website/src/content/docs/docs/howtos/Generate client libraries/10versioning.mdx
+++ b/website/src/content/docs/docs/howtos/Generate client libraries/10versioning.mdx
@@ -464,6 +464,286 @@ func (client *ServiceClient) StableFunctionality(ctx context.Context, stableMode
 
 </ClientTabs>
 
+### New model/operation added in new api version
+
+You can use `@added` decorator to indicate when a specific Api feature was first introduced.
+When applied to models, operations, or parameters, `@added` marks the Api Version at which these elements became available. This prevents confusing error messages for customers using older version clients.
+
+#### New operation added in new api version
+
+In this example, `createUserProfile` is marked with `@added(Versions.v2024_02_01)`, meaning this operation only becomes available starting from the `2024-02-01` version. Elements without the `@added` decorator, like `getUserInfo`, are available across all defined versions.
+
+<ClientTabs>
+
+<ClientTabItem lang="typespec">
+
+```typespec
+import "@typespec/versioning";
+import "@typespec/http";
+
+using Versioning;
+using Http;
+
+@versioned(Versions)
+@service
+namespace My.Service;
+
+enum Versions {
+  v2024_01_01: "2024-01-01",
+  v2024_02_01: "2024-02-01",
+}
+
+model UserInfo {
+  username: string;
+}
+
+model UserProfile {
+  profileName: string;
+}
+
+@route("/users/info")
+op getUserInfo(@body userInfo: UserInfo): void;
+
+@added(Versions.v2024_02_01)
+@route("/users/profile")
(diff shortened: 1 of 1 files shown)


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -180,20 +183,16 @@ function getSdkHttpParameters(
...
-      return diagnostics.wrap(retval);
-    }
     if (tspBody.property && !isNeverOrVoidType(tspBody.property.type)) {
       const bodyParam = diagnostics.pipe(
         getSdkHttpParameter(context, tspBody.property, httpOperation.operation, undefined, "body"),
       );
+      if (tspBody.bodyKind === "file" && bodyParam.kind === "body") {
+        if (!bodyParam.serializationOptions) {
+          bodyParam.serializationOptions = {};
+        }
```
joheredi: nit: I think this could also be expressed as
```suggestion
        bodyParam.serializationOptions = bodyParam.serializationOptions ??  {};
```

packages/typespec-client-generator-core/test/methods/file.test.ts
```diff
@@ -8,16 +8,83 @@ beforeEach(async () => {
...
+  ok(fileModel.properties.find((p) => p.name === "contents"));
+  ok(fileModel.properties.find((p) => p.name === "filename"));
+});
+
+it("file input with content type", async () => {
+  await runner.compile(
+    `
+      @service
+      namespace TestService {
+        op uploadFile(@body file: File<"application/yaml">): void;
```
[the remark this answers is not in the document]
tadelesh: From the [definition](https://github.com/microsoft/typespec/blob/5d549c414f9a359fd768f5130de4f029c8cb67f9/packages/http/lib/main.tsp#L221), it is not the same. It looks like it is not a header, just the metadata.

packages/typespec-client-generator-core/src/interfaces.ts
```diff
@@ -631,6 +632,11 @@ export interface XmlSerializationOptions {
   };
 }
 
+export interface BinarySerializationOptions {
+  /** Whether this is a file/stream input */
+  isFile: boolean;
```
tadelesh: Do we need to add file info such as content type, etc. here?

PR #3455, opened 2025-10-29
Spec repo structure v2

### LINKED ISSUE #3425: tsp init still uses folder structure v1
When running `npx tsp init https://aka.ms/typespec/azure-init`, it will try to initialize the tspconfig.yaml with the following content
```
options:
  "@azure-tools/typespec-autorest":
    use-read-only-status-schema: true
    emitter-output-dir: "{project-root}/.."
    azure-resource-provider-folder: "resource-manager"
    output-file: "{azure-resource-provider-folder}/{service-name}/{version-status}/{version}/openapi.json"
```
which should be like this https://github.com/Azure/azure-rest-api-specs/blob/main/specification/widget/resource-manager/Microsoft.Widget/Widget/tspconfig.yaml#L7-L13 for folder structure v2. 
```
options:
  "@azure-tools/typespec-autorest":
    use-read-only-status-schema: true
    emitter-output-dir: "{project-root}"
    azure-resource-provider-folder: "resource-manager"
    output-file: "{version-status}/{version}/widget.json"
    arm-types-dir: "{project-root}/../../../../common-types/resource-management"
```

Another question is it's also trying to ask information like project name and ARM RP Namespace 

Image

But that information doesn't seem to be used in the folder structure v2?

timotheeguerin:
fix #3425

### DIFF
--- a/eng/feeds/arm/main.tsp
+++ b/eng/feeds/arm/main.tsp
@@ -16,7 +16,7 @@ using Azure.ResourceManager;
 @service(#{
   title: "Microsoft.{{parameters.ServiceNamespace}} management service",
 })
-@versioned(Microsoft.Contoso.Versions)
+@versioned(Microsoft.{{parameters.ServiceNamespace}}.Versions)
 namespace Microsoft.{{parameters.ServiceNamespace}};
 
 /** The available API versions. */


eng/feeds/__snapshots__/azure-core_stand_alone/tspconfig.yaml
```diff
@@ -11,7 +11,7 @@ options:
     output-file: "{azure-resource-provider-folder}/{service-name}/{version-status}/{version}/openapi.json"
   "@azure-tools/typespec-python":    
     emitter-output-dir: "{output-dir}/{service-dir}/azure-contoso"
-    namespace: "azure.contoso"
+    namespace: azure.azure.contoso
```
markcowl: intentional?

eng/feeds/azure-scaffolding.json
```diff
@@ -47,7 +47,7 @@
     ],
     "inputs": {
       "ServiceNamespace": {
-        "description": "Please provide service namespace in Pascal case:",
+        "description": "Please provide service namespace in Pascal case(e.g. Azure.Contoso):",
```
qiaozha: I wonder why we still need the service namespace.

timotheeguerin: This one is for data plane where the full namespace is expected for arm template just the rp name is expected without Microsoft.

qiaozha: I hope we are not using this as the service name? `specification/contoso/data-plane/Azure.Contoso` or `specification/Azure.Contoso` are not recommended in folder structure v2.

timotheeguerin: tsp init has to be run in the folder already created so this is not used

eng/feeds/__snapshots__/azure-arm/tspconfig.yaml
```diff
@@ -3,9 +3,39 @@ emit:
...
+  "@azure-tools/typespec-python":
+    emitter-output-dir: "{output-dir}/{service-dir}/azure-mgmt-contoso"
+    namespace: "azure.mgmt.contoso"
+    generate-test: true
+    generate-sample: true
+  "@azure-tools/typespec-java":
+    emitter-output-dir: "{output-dir}/{service-dir}/azure-resourcemanager-contoso"
+    namespace: "com.azure.resourcemanager.contoso"
+    service-name: "contoso"
+  "@azure-tools/typespec-ts":
```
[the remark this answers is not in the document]
v-jiaodi: It is correct.

PR #3460, opened 2025-10-30
[tcgc] one client multiple services

### LINKED ISSUE #3536: Multiples services for one client TCGC Author doc

### LINKED ISSUE #3537: Multiples services for one client TCGC Implementation

iscai-msft:
fixes #3536, fixes #3537

### DIFF
--- a/website/src/content/docs/docs/howtos/Generate client libraries/03client.mdx
+++ b/website/src/content/docs/docs/howtos/Generate client libraries/03client.mdx
@@ -706,17 +706,17 @@ namespace MultiService
             Widget = new WidgetClient(endpoint, "2.0", options);
             Gadget = new GadgetClient(endpoint, "1.1", options);
         }
-        
+
         public virtual WidgetClient Widget { get; }
         public virtual GadgetClient Gadget { get; }
     }
-    
+
     public partial class WidgetClient
     {
         public virtual async Task<Response<IReadOnlyList<Widget>>> ListWidgetsAsync(RequestContext context = null) {}
         public virtual Response<IReadOnlyList<Widget>> ListWidgets(RequestContext context = null) {}
     }
-    
+
     public partial class GadgetClient
     {
         public virtual async Task<Response<IReadOnlyList<Gadget>>> ListGadgetsAsync(RequestContext context = null) {}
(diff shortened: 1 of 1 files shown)


packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -89,6 +89,39 @@ export function getDefaultApiVersion(
...
+  // Fallback: try to get from any service namespace
+  const services = listAllServiceNamespaces(context);
+  if (services.length > 0) {
+    const versions = getVersions(context.program, services[0])[1]?.getVersions();
+    if (versions?.length) {
+      const versionEnum = versions[0].enumMember.enum;
+      context.__packageVersionEnum = versionEnum;
+      return versionEnum;
+    }
+  }
```
tadelesh: We do not need this fallback. This helper is only used for the case:
```
op test(apiVersion: VersionEnum): void;
```

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -780,7 +780,7 @@ function getVersioningMutator(
 
 export function handleVersioningMutationForGlobalNamespace(context: TCGCContext): Namespace {
   const globalNamespace = context.program.getGlobalNamespaceType();
-  const allApiVersions = context.getPackageVersions();
+  const allApiVersions = context.getApiVersions();
```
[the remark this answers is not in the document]
iscai-msft: let me work on the second one, that's `@useDependency` right?

lmazuel (approved):

packages/typespec-client-generator-core/src/context.ts
```diff
@@ -113,50 +204,116 @@ export function createTCGCContext(
...
+      if (cachedVersions?.length) {
+        return cachedVersions;
+      }
+
+      const versions = getVersions(program, service)[1]?.getVersions();
       if (!versions) {
-        this.__packageVersions = [];
-        return this.__packageVersions;
+        return [];
       }
```
tadelesh: For services that without versioning, we should also save the cache.

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -46,6 +52,47 @@ export function createSdkPackage<TServiceOperation extends SdkServiceOperation>(
   return diagnostics.wrap(sdkPackage);
 }
 
+function createClients<TServiceOperation extends SdkServiceOperation>(
```
tadelesh: Can we move this client hierarchy build process into cache builder section. I prefer after we get the client hierarchy info, all the following calculation does not have special logic for multi-service scenario. The only exception could be API version cache builder.

PR #3465, opened 2025-11-03
Deprecate arm-resource-flattening

timotheeguerin:
close [#3485](https://github.com/Azure/typespec-azure/issues/3485)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -486,21 +486,21 @@ importers:
   core/packages/emitter-framework:
     dependencies:
       '@alloy-js/csharp':
-        specifier: ^0.20.0
-        version: 0.20.0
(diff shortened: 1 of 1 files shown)


packages/typespec-azure-resource-manager/lib/foundations/arm.foundations.tsp
```diff
@@ -121,6 +121,7 @@ model ResourceUpdateModel<
     "Name" | "name" | "properties"
   >>> {
   /** The resource-specific properties for this resource. */
+  #suppress "deprecated" "https://github.com/Azure/typespec-azure/issues/2284"
```
markcowl: We should remove these, I think, but this might require a new issue next sprint

PR #3506, opened 2025-11-11
[specs] update arm LRO test to correspond to mockapi

iscai-msft:
This test is causing issues for the Rust team because the spec specifies that the response will be a `CancelResult`, while the initial response is `void`. We think this test is trying to go for specific ARM behavior, as such, we are not changing the mockapi response, instead updating the spec to reflect what the mockapi returns

### DIFF
--- a/.chronus/changes/specs-updateLroInitialResponse-2025-10-11-13-31-48.md
+++ b/.chronus/changes/specs-updateLroInitialResponse-2025-10-11-13-31-48.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/azure-http-specs"
+---
+
+Update mgmt spec LRO to correspond to initial response being void


markcowl (approved):

markcowl (commented):
Not sure about this scenario.  Generally speaking, the POST response will never contain the final result, this is only returned after polling.  We include the final result essentially just for swagger processors that don't understand x-ms-long-running-operation-options.

Having the poller return a 202 is not standard, but might be part of the required handling.  In any case, void is not a final result, so it seems wrong to include it in the 'fake' 200 responses of the POST endpoint

weidongxu-microsoft (commented):
The problem may be on the mockapi.

@XiaofeiCao Can you check about live service response of action LRO with `Location` header? Do them typically have a body or not?

On the other hand, I don't think the initial response would be the form of `CancelResult` model (if it has a body).

XiaofeiCao:
> Hmm that seems to contradict the [vNext design guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-existing-resource) (I'm familiar with the one you cited). Which one is correct?

The Spector test cases under azure/resource-manager are for management-plane, hence the ARM guideline should be followed here.

The vNext design guideline you mentioned is for _data-plane_. You can tell from its introduction:
image

> IMHO, one should not rely on the response body even if it's returned, since I don't see any schema enforcement on the response body schema.

By this I mean under ARM guideline, `202 + Location header` should have no response body, and response body should be ignored by SDK even if service returns it.

jhendrixMSFT:
Thanks this helps.  I think this can be closed and we'll follow up on our end.

PR #3543, opened 2025-11-19
Tsp integration interactive mode and tweaks

timotheeguerin:
Added new interactive mode that allow reruning only the failed specs. This makes it much easier to fix things in the spec repo where you can now run the original list and keep iterating witout having to rerun all
Other minor changes
- Actually build all configured entrypoints
- Run compile from the workspace wd for error reporting

https://github.com/user-attachments/assets/70cdd7a7-120e-4b4b-9ec5-072c429e6b9a

### DIFF
--- a/packages/integration-tester/src/keyboard-api.test.ts
+++ b/packages/integration-tester/src/keyboard-api.test.ts
@@ -60,7 +60,7 @@ describe("when tests are running", () => {
     });
   });
 
-  it("does NOT calls cancelCurrentRun when other keys are pressed", () => {
+  it("does NOT call cancelCurrentRun when other keys are pressed", () => {
     stdin.emit("data", "b");
     stdin.emit("data", "d");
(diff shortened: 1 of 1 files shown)


packages/integration-tester/src/keyboard-api.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    runner.runningPromise = Promise.resolve() as any;
+  });
+  describe("calls cancelCurrentRun when cancel keys are pressed", () => {
+    it.each(["q", "c", "a", "f", "space", "\x03"])(`%s`, (key) => {
+      stdin.emit("data", key);
+      expect(runner.cancelCurrentRun).toHaveBeenCalled();
+    });
+  });
+
+  it("does NOT calls cancelCurrentRun when other keys are pressed", () => {
```
markcowl:
```suggestion
  it("does NOT call cancelCurrentRun when other keys are pressed", () => {
```

markcowl (approved):

PR #3550, opened 2025-11-20
[specs] add multi-service test

iscai-msft:
Fix https://github.com/microsoft/typespec/issues/9033

### DIFF
--- a/packages/azure-http-specs/specs/service/multi-service/main.tsp
+++ b/packages/azure-http-specs/specs/service/multi-service/main.tsp
@@ -16,11 +16,11 @@ namespace Service.MultiService.FirstService {
   interface Operations {
     @doc("Operation from first service")
     @route("/one")
-    op firstOne(@query("api-version") apiVersion: FirstServiceVersions): void;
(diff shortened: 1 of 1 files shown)


packages/azure-http-specs/specs/service/multi-service/client.tsp
```diff
@@ -0,0 +1,55 @@
...
+  // Call operations from second service - uses api-version 2.1.0
+  client.secondOne();
+  client.secondTwo();
+  ```
+  """)
+@scenario
+@route("/multi-service")
+namespace Service.MultiService.Scenario {
+  @get
+  @route("/first/one")
```
weidongxu-microsoft: I assume we should not write these REST decorators in client.tsp

They belong to the definition of the REST in main.tsp
If we are duplicating all of them here, something likely not right.

---

Also, for the op in client.tsp, e.g. this `firstOne`, it also not referring to the op in main.tsp (not using `firstOne is Service.MultiService.FirstService.Operations.firstOne`

---

Overall, this test kind of just define new op in client.tsp

PR #3582, opened 2025-11-28
update tests for client-location

kazrael2119:
split client-location test cases to individual clients

### DIFF
--- a/packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-method-parameter-to-client/mockapi.ts
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-method-parameter-to-client/mockapi.ts
@@ -6,7 +6,7 @@ export const Scenarios: Record<string, ScenarioMockApi> = {};
 Scenarios.Azure_ClientGenerator_Core_ClientLocation_MoveMethodParameterToClient_BlobOperations =
   passOnSuccess([
     {
-      uri: "/azure/client-generator-core/client-location/blob",
+      uri: "/azure/client-generator-core/client-location/move-method-parameter-to-client/blob",
       method: "get",
       request: {
         query: {
--- a/packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-to-existing-sub-client/mockapi.ts
+++ b/packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-to-existing-sub-client/mockapi.ts
@@ -6,7 +6,7 @@ export const Scenarios: Record<string, ScenarioMockApi> = {};
 Scenarios.Azure_ClientGenerator_Core_ClientLocation_MoveToExistingSubClient_UserOperations =
   passOnSuccess([
     {
-      uri: "/azure/client-generator-core/client-location/user",
+      uri: "/azure/client-generator-core/client-location/move-to-existing-sub-client/user",
       method: "get",
       request: {},
       response: {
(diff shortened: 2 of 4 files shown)


weidongxu-microsoft:
This is a breaking change to spec test and those uses this spec test.

qiaozha:
@weidongxu-microsoft  the current test is problematic because without specifying initializedBy, the namespace should be treated as operation group instead of client. the clientLocation decorator can be only applied to clients. However, if we mixed the test with clientInitialization and clientLocation, it would be too heavy for a specific scenario test. Therefore, it's necessary to break them into each client per scenario.

How should we introduce this breaking change?

packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-method-parameter-to-client/main.tsp
```diff
@@ -0,0 +1,50 @@
+import "@typespec/http";
+import "@typespec/spector";
+import "@azure-tools/typespec-client-generator-core";
+
+using Http;
+using Spector;
+using Azure.ClientGenerator.Core;
+
+@doc("Test for @clientLocation decorator - moving operations between clients")
+@scenarioService("/azure/client-generator-core/client-location")
```
weidongxu-microsoft: unsure if we can use same scenarioService value in different main.tsp

packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-to-existing-sub-client/main.tsp
```diff
@@ -0,0 +1,43 @@
...
+import "@azure-tools/typespec-client-generator-core";
+
+using Http;
+using Spector;
+using Azure.ClientGenerator.Core;
+
+@doc("Test for @clientLocation decorator - moving operations between clients")
+@scenarioService("/azure/client-generator-core/client-location/move-to-existing-sub-client")
+@global.Azure.ClientGenerator.Core.clientNamespace(
+  "azure.clientgenerator.core.clientlocation.MoveToExistingSubClient",
```
weidongxu-microsoft:
```suggestion
  "azure.clientgenerator.core.clientlocation.subclient",
```

packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-to-new-sub-client/main.tsp
```diff
@@ -0,0 +1,37 @@
...
+import "@azure-tools/typespec-client-generator-core";
+
+using Http;
+using Spector;
+using Azure.ClientGenerator.Core;
+
+@doc("Test for @clientLocation decorator - moving operations between clients")
+@scenarioService("/azure/client-generator-core/client-location/move-to-new-sub-client")
+@global.Azure.ClientGenerator.Core.clientNamespace(
+  "azure.clientgenerator.core.clientlocation.MoveToNewSubClient",
```
weidongxu-microsoft:
```suggestion
  "azure.clientgenerator.core.clientlocation.newsubclient",
```

packages/azure-http-specs/specs/azure/client-generator-core/client-location/move-to-root-client/main.tsp
```diff
@@ -0,0 +1,39 @@
...
+import "@azure-tools/typespec-client-generator-core";
+
+using Http;
+using Spector;
+using Azure.ClientGenerator.Core;
+
+@doc("Test for @clientLocation decorator - moving operations between clients")
+@scenarioService("/azure/client-generator-core/client-location/move-to-root-client")
+@global.Azure.ClientGenerator.Core.clientNamespace(
+  "azure.clientgenerator.core.clientlocation.MoveToRootClient",
```
weidongxu-microsoft:
```suggestion
  "azure.clientgenerator.core.clientlocation.rootclient",
```

weidongxu-microsoft (commented):
@haolingdong-msft for review too.

One concern is name like `movemethodparametertoclient` would trigger spelling check failure.

PR #3583, opened 2025-11-28
[TCGC] Support array encode on model property

tadelesh:
Resolve: https://github.com/Azure/typespec-azure/issues/3534

### DIFF
--- a/packages/typespec-client-generator-core/test/types/array.test.ts
+++ b/packages/typespec-client-generator-core/test/types/array.test.ts
@@ -181,5 +181,6 @@ it("array with encode", async () => {
   strictEqual(model.name, "Foo");
   strictEqual(model.properties.length, 1);
   const modelProp = model.properties[0];
+  strictEqual(modelProp.type.kind, "array");
   strictEqual(modelProp.encode, "commaDelimited");
 });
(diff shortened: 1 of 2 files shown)


packages/typespec-client-generator-core/test/types/array.test.ts
```diff
@@ -164,3 +166,20 @@ it("recursive array type", async () => {
   const prop = testArrayModel.properties[0];
   strictEqual(prop.type, modelProp.type);
 });
+
+it("array with encode", async () => {
```
lirenhe: Can we add more test on this?  I think we only support array of string to have this encoding.   Shall we report error/warning if this is used on other types?

tadelesh: Link [here](https://github.com/microsoft/typespec/pull/9002#discussion_r2570297016).

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -1245,6 +1246,20 @@ export function getSdkModelPropertyTypeBase(
   diagnostics.pipe(addEncodeInfo(context, alternateType ?? type, propertyType));
   const name = getPropertyNames(context, type)[0];
   const onClient = isOnClient(context, type, operation, apiVersions.length > 0);
+  let encode: ArrayKnownEncoding | undefined = undefined;
+  // We only support array encoding at property level for now
+  if ($(context.program).array.is(type.type)) {
```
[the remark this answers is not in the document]
tadelesh: I'm asking TypeSpec core to add lint for this. See comment [here](https://github.com/microsoft/typespec/pull/9002#discussion_r2570297016).

msyyc (approved):

PR #3594, opened 2025-12-01
Fix issues when we have singleton resource with a customized name

ArcturusZhang:
Fixes https://github.com/Azure/typespec-azure/issues/3595
Fixes https://github.com/Azure/typespec-azure/issues/3623

Previously the segment default was hard-coded in a function checking if a segment in path is a variable.
This PR modifies it to accept an optional extra singletonResourceName parameter - if match, we will return true as variable segment.

This PR also updates the `ResolvedResource` API to add a new property `singletonResourceNames`

### DIFF
--- a/packages/typespec-azure-resource-manager/src/resource.ts
+++ b/packages/typespec-azure-resource-manager/src/resource.ts
@@ -632,9 +632,13 @@ function isVariableSegment(segment: string, singletonResourceName?: string): boo
 function getResourceInfo(
   program: Program,
   operation: ArmResourceOperation,
-  singletonResourceName: string | undefined
+  singletonResourceName: string | undefined,
 ): ResolvedResourceInfo | undefined {
-  const pathInfo = getResourcePathElements(operation.httpOperation.path, operation.kind, singletonResourceName);
+  const pathInfo = getResourcePathElements(
+    operation.httpOperation.path,
+    operation.kind,
+    singletonResourceName,
+  );
   if (pathInfo === undefined) return undefined;
   return {
     ...pathInfo,
@@ -659,7 +663,10 @@ export function getResourcePathElements(
     }
 
     // if the next segment is the last segment
-    if (i + 1 === segments.length - 1 && isVariableSegment(segments[i + 1], singletonResourceName)) {
+    if (
+      i + 1 === segments.length - 1 &&
+      isVariableSegment(segments[i + 1], singletonResourceName)
+    ) {
       typeSegments.push(segments[i]);
       instanceSegments.push(segments[i]);
       instanceSegments.push(segments[i + 1]);
@@ -873,7 +880,7 @@ export function resolveArmResourceOperations(
 
     if (armOperation === undefined) continue;
     armOperation.kind = operation.kind;
-    const singletonResourceName = getSingletonResourceKey(program, resourceType)
+    const singletonResourceName = getSingletonResourceKey(program, resourceType);
 
     armOperation.resourceModelName = operation.resource?.name ?? resourceType.name;
     const resourceInfo = getResourceInfo(program, armOperation, singletonResourceName);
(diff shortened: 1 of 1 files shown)


packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -652,7 +658,12 @@ export function getResourcePathElements(
       break;
     }
 
-    if (i + 1 < segments.length && isVariableSegment(segments[i + 1])) {
+    // if the next segment is the last segment
+    if (i + 1 === segments.length - 1 && isVariableSegment(segments[i + 1], singletonResourceName)) {
+      typeSegments.push(segments[i]);
+      instanceSegments.push(segments[i]);
+      instanceSegments.push(segments[i + 1]);
```
ArcturusZhang: so we only consider the singletonResourceName when the current segment is the last segement.

packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
```diff
@@ -257,39 +258,42 @@ describe("unit tests for resource manager helpers", () => {
...
+        singletonResourceName: "default",
         expected: {
           resourceType: {
             provider: "Microsoft.Bar",
-            types: ["bars", "basses"],
+            types: ["bars"],
           },
           resourceInstancePath:
-            "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/{providerName}/{resourceType}/{resourceName}/{childResourceType}/{childResourceName}/providers/Microsoft.Bar/bars/{barName}/basses/default",
+            "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/{providerName}/{resourceType}/{resourceName}/{childResourceType}/{childResourceName}/providers/Microsoft.Bar/bars/{barName}",
```
ArcturusZhang: the result changed because of [the change above](https://github.com/Azure/typespec-azure/pull/3594/files#r2591600872).
In this path, we have a `default` but it is not at the end.

Also I have concerns: why would we try to rebuild the resourceInstancePath from the path any way? Should we just use the READ request path as the resource instance path? I think ARM did this in their logic, and resource type is then built out of this path.

packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -652,7 +662,15 @@ export function getResourcePathElements(
       break;
     }
 
-    if (i + 1 < segments.length && isVariableSegment(segments[i + 1])) {
+    // if the next segment is the last segment
+    if (
```
markcowl: I think if we change this logic, we will fix the issues below - it makes sense that the singleton resource name should just *always* be checked when looking for a variable segment.  

I think there is still a case to think about, when multiple resources in the resource type are singletons

packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -621,15 +625,20 @@ function getResourceScope(
   return undefined;
 }
 
-function isVariableSegment(segment: string): boolean {
-  return (segment.startsWith("{") && segment.endsWith("}")) || segment === "default";
+function isVariableSegment(segment: string, singletonResourceName?: string): boolean {
```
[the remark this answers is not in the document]
markcowl: I think we should also continue to use 'default'  as a singleton resource name, *in addition to* the singleton resource name from getSingletonKey

markcowl (commented):
First, we need to target this PR at release/december-2025,

scond I have a couple of minor comments, and a slight algorithmic change:

- We should check for 'default' as well as the resourceInstance Name from the resource type
- We should check against the resourceInstanceName (and default) *whwnever* we are looking to see that an enven-numbered segment is a variable segment

I think this will give us better results in legacy scenarios

PR #3609, opened 2025-12-02
[Spector] Test case for data plane multiple services

iscai-msft:
Resolve: https://github.com/microsoft/typespec/issues/9033

### DIFF
--- a/.chronus/changes/fix-array-encoding-wrong-type-2025-11-16-8-34-34-2.md
+++ b/.chronus/changes/fix-array-encoding-wrong-type-2025-11-16-8-34-34-2.md
@@ -1,7 +1,0 @@
----
-changeKind: fix
-packages:
-  - "@azure-tools/typespec-autorest"
----
-
-Allow explicit `ArrayEncoding.commaDelimited` on parameters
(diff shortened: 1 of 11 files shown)


packages/azure-http-specs/specs/service/multi-service/main.tsp
```diff
@@ -0,0 +1,51 @@
...
+  ```ts
+  const client = new MultiServiceClient();
+  
+  // Call operation from first service - uses api-version av2
+  client.aTest();
+  
+  // Call operation from second service - uses api-version bv2
+  client.bTest();
+  ```
+  """)
```
weidongxu-microsoft: We probably do not need scenarioDoc here on ServiceA (there is none on ServiceB)?

packages/azure-http-specs/specs/service/multi-service/main.tsp
```diff
@@ -0,0 +1,59 @@
...
+ */
+@scenarioService("/service/multi-service/service-a")
+@versioned(VersionsA)
+namespace ServiceA {
+  enum VersionsA {
+    av1,
+    av2,
+  }
+
+  interface AI {
```
weidongxu-microsoft: nit, not sure what `AI` and `BI` means. but this is just a test so OK for me.

msyyc (approved):

packages/azure-http-specs/specs/service/multi-service/main.tsp
tadelesh: A interface. Let me refine the naming.

weidongxu-microsoft (approved):

PR #3613, opened 2025-12-03
Use auth id resolved by http library to dedup the auth with same model name

### LINKED ISSUE #3606: [Bug]: typespec-autorest generates duplicate "ApiKeyAuth" entry under "security"
### Describe the bug

Given the following auth definition, the typespec-autorest emitter outputs duplicate ApiKeyAuth entries under security which causes swagger validations to fail in the rest api specs repo: [Playground example](https://azure.github.io/typespec-azure/playground/?e=%40azure-tools%2Ftypespec-autorest&c=aW1wb3J0ICJAdHlwZXNwZWMvcmVzdCI7DQrSGmh0dHDNGmF6dXJlLXRvb2xzL8gmLcYVY29yZcQsDQp1c2luZyBUeXBlU3BlYy5SZXN0O9EWSHR0cNIWVmVyc2lvbmluZ8kcQcReLkNvcmXFXS8qKg0KICrGFyBNYXBzIEdlb2xvY2F0aW9uIFJFU1QgQVBJc8QlLw0KQHVzZUF1dGgoDQogIEFwaUtlecQPPMQPyBFMxz0ucXVlcnksxhsic3Vic2NyaXDEGi1rZXkixBg%2BIHzMRs9AaGVhZGVyLCAiU0FTIFRva2VuIj4NCikNCkBzZXJ2aWNlKCN7IHRpdGxlOiAi8AC7IiB9xCp25gECZWQo5wEMc8hAZXLlAMUie2VuZHBvaW50fSLlAKwi%2FwELc8Yne8QFICDIPTogIuQB3nM6Ly9hdGxhcy5taWNyb3NvZnQuY29txjN95QC9bmFtZXNwYWPGY8ti7QGQVGhlIGF2YWlsYWJsZcR3IOcAynMu5wGLZW51bSDoANXmAI%2FGPSDHPzEuMMw5xDggxTogIHYxXzA6ICIxLjDkAJp9DQo%3D&options=%7B%7D&vs=%7B%7D)

### Reproduction

See playground example above

### Checklist

timotheeguerin:
fix  #3606

The dedup id just have `_` added to them which is not great but this is what openapi3 does now and if we want to change probably cna improve in http library instead after. 

That code change should have happened before anyway and set ground work for adding the work for per operation auth(which we might never need for autorest)

### DIFF
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -108,6 +108,7 @@ import {
   AuthenticationOptionReference,
   AuthenticationReference,
   HttpAuth,
+  HttpAuthRef,
   HttpOperation,
   HttpOperationBody,
   HttpOperationMultipartBody,
@@ -2714,7 +2715,7 @@ export async function getOpenAPIForService(
       }
     }
 
-    const security = getOpenAPISecurity(authentication.defaultAuth);
+    const security = getOpenAPISecurity(oaiSchemes, authentication.defaultAuth);
 
     return { securitySchemes: oaiSchemes, security };
   }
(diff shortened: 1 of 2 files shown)


markcowl (approved):
Approved once the azure=specs check passes

Issue #3587, opened 2025-11-28
[TCGC] Refining multiple services support


tadelesh:
1. Example loading for different services should use different versioning.
2. Do we need to support pure multi-service scenario?
3. Could mutator support `useDependency` logic?
4. I don't think we should define `@service` for the wrapper client since it is not a service, just a client. We need to refine related logic.
5. Change rule of `invalid-initialized-by`.
6. Consolidate client hierarchy building logic all into cache builder to decouple logic.

Issue #1039, opened 2024-06-19
[Linter Rule] Add warning if service spec has naming conflict within one client or with azure core models


MaryGao:
### Clear and concise description of the problem

Any rule to resolve the naming conflict will be wired and just workaround no matter we use `AutoGenerated` in m4 or any prefix we may use in language emitter.

I would suggest we add a warning for any naming conflicts if happening within one client or within referred azure core models see an [example](https://github.com/Azure/azure-rest-api-specs/blob/0031475e8ae9bc16936bd8d88a2cf1981bf4c510/specification/scvmm/ScVmm.Management/models.tsp#L35-L37).

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [X] Read the [docs](https://azure.github.io/typespec-azure/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

live1206:
@MaryGao 
We have implemented the naming conflicts check in TCGC, within the same namespace union & model & enum should not share the same name.
For your [example](https://github.com/Azure/azure-rest-api-specs/blob/0031475e8ae9bc16936bd8d88a2cf1981bf4c510/specification/scvmm/ScVmm.Management/models.tsp#L35-L37), the union and model are in different namespaces. 
From TypeSpec respective, models in different namespaces can share the same name.

tadelesh:
i doubt if it is a valid lint for azure?

MaryGao:
I was thinking about the scope for azure with name conflict handling if happened cross namespaces. I know from typespec side there is no validation for models but I was wondering if we could have one in azure scope.

For JS we would not respect models namespace in azure the name collisions would be an issue for us to generate model with correct name. 

@weidongxu-microsoft @tadelesh @msyyc Not sure this would be an issue for other languages also?

msyyc:
For python, since all models will be generated in same namespace so I also hope we could avoid name conflict even across namespaces.

live1206:
[...]

PR #3617, opened 2025-12-04
[TCGC] Support multiple service

iscai-msft:
Resolve: https://github.com/Azure/typespec-azure/issues/3587

### DIFF
--- a/packages/typespec-client-generator-core/src/cache.ts
+++ b/packages/typespec-client-generator-core/src/cache.ts
@@ -8,6 +8,7 @@ import {
   Namespace,
   Operation,
 } from "@typespec/compiler";
+import { unsafe_Realm } from "@typespec/compiler/experimental";
 import { getVersionDependencies, getVersions } from "@typespec/versioning";
 import { getClientLocation, getClientNameOverride, isInScope } from "./decorators.js";
 import { SdkClient, SdkOperationGroup, TCGCContext } from "./interfaces.js";
@@ -159,7 +160,10 @@ export function prepareClientAndOperationCache(context: TCGCContext): void {
     const newOperationGroupWithService = new Map<string, Namespace>();
     listScopedDecoratorData(context, clientLocationKey).forEach((v, k) => {
       // only deal with mutated types or without mutation
-      if (!context.__mutatedRealm || context.__mutatedRealm.hasType(k)) {
+      if (
+        (!context.__mutatedRealm && !unsafe_Realm.realmForType.has(k)) ||
+        (context.__mutatedRealm && context.__mutatedRealm.hasType(k))
+      ) {
         if (typeof v === "string") {
           if (
             clients[0].subOperationGroups.some(


packages/typespec-client-generator-core/design-docs/mutliple-services.md
```diff
@@ -0,0 +1,174 @@
...
+  - If any service is versioned and `@useDependency` is not set, TCGC will use the latest version for each service by default.
+- Only one `@client` with multiple services can be defined per package. Defining multiple such clients or mixing with `@operationGroup` is not supported.
+
+### TCGC Behavior
+
+When TCGC detects multiple services in one client, it will:
+
+1. Create the root client for the combined client. If any service is versioned, the root client's initialization method will have an `apiVersion` parameter with no default value. The `apiVersions` property and the `apiVersion` parameter for the root client will be empty (since multiple services' API versions cannot be combined).
+2. Create sub-clients for each service's nested namespaces or interfaces. Each sub-client will have its own `apiVersion` property and initialization method if the service is versioned.
+3. Operations directly under each service's namespace are placed under the root client. Operations under nested namespaces or interfaces are placed under the corresponding sub-clients.
```
joheredi: First time I read this I understood that operations directly under the service namespace are placed in the root/combined client. But I think they will be within the service subclient right?

```ts
namespace ServiceA {
  op createA(): void;
}

namespace ServiceB {
   op createB(): void;
}
```

We get 

```ts
client.serviceA.createA()
client.serviceB.createB()
```

Not 
```ts
client.createA()
client.createB()
```

packages/typespec-client-generator-core/src/clients.ts
```diff
@@ -133,7 +133,9 @@ function getSdkEndpointParameter<TServiceOperation extends SdkServiceOperation =
 ): [SdkEndpointParameter, readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
   const rawClient = client.__raw;
-  const servers = getServers(context.program, client.__raw.service);
+  // For multiple services, just take the first one to get servers
+  const service = Array.isArray(rawClient.service) ? rawClient.service[0] : rawClient.service;
+  const servers = getServers(context.program, service);
```
ArcturusZhang: I think we should report something if we are taking the first ignoring others.
I am not sure I know what the "servers" here represent, but if we are ignoring others, we should report something when we ignore them and our user could rethink.

packages/typespec-client-generator-core/design-docs/mutliple-services.md
```diff
@@ -0,0 +1,174 @@
...
+
+### TCGC Behavior
+
+When TCGC detects multiple services in one client, it will:
+
+1. Create the root client for the combined client. If any service is versioned, the root client's initialization method will have an `apiVersion` parameter with no default value. The `apiVersions` property and the `apiVersion` parameter for the root client will be empty (since multiple services' API versions cannot be combined). The root client's endpoint and credential parameters will be created based on the first sub-service, which means all sub-services must share the same endpoint and credential.
+2. Create sub-clients for each service's nested namespaces or interfaces. Each sub-client will have its own `apiVersion` property and initialization method if the service is versioned.
+3. Operations directly under each service's namespace are placed under the root client. Operations under nested namespaces or interfaces are placed under the corresponding sub-clients.
+4. Decorators such as `@clientLocation`, `@convenientAPI`, `@protocolAPI`, `@moveTo`, and `@scope` work as usual.
+5. All other TCGC logic remains unchanged.
```
lirenhe: For, you highlighted  some decorators work as usual.  Is there any decorator that could not work under multiple service?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -179,17 +182,85 @@ export const $client: ClientDecorator = (
...
+      } else {
+        isSame = isSameAuth(auth, currentAuth);
+        if (!isSame) {
+          break;
+        }
+      }
+    }
+    if (!isSame) {
+      reportDiagnostic(context.program, {
+        code: "inconsistent-multiple-service",
```
lirenhe: As there are different cases of 'isSame',  could you provide more detailed information in the diagnostics?

packages/typespec-client-generator-core/lib/decorators.tsp
```diff
@@ -163,10 +163,10 @@ extern dec client(target: Namespace | Interface, options?: ClientOptions, scope?
  */
 model ClientOptions {
   /**
-   * The service that this client is generated for. If not specified, TCGC will look up the first parent namespace decorated with `@service` for the target.
+   * The services that this client is generated for. If not specified, TCGC will look up the first parent namespace decorated with `@service` for the target.
    * The namespace should be decorated with `@service`.
    */
-  service?: Namespace;
+  service?: Namespace | Namespace[];
```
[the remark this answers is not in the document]
tadelesh: Keep this naming is just prevent breaking changes.

packages/typespec-client-generator-core/design-docs/mutliple-services.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Background
+
+Previously, TCGC [client](./client.md) only supported generating a client from a single service. However, in real-world scenarios, a single package often contains multiple services. TCGC must support multiple services within one package to address these needs.
+
+## User Scenario
+
+1. Merging multiple services' namespaces into one client
+
+This scenario is common in Azure management services. For example, the compute team maintains several services: `Compute`, `Disk`, `Gallery`, and `Sku`. These services share the same endpoint and credential but have different versions. Users expect a single SDK package to manage all these services, not multiple packages. According to Azure SDK guidelines, the package should expose a single client. Therefore, TCGC must support auto-merging operations and nested namespaces/interfaces from multiple services into one client.
```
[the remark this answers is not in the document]
tadelesh: Done.

packages/typespec-client-generator-core/test/examples/load.test.ts
```diff
@@ -403,3 +403,69 @@ it("teamplate case", async () => {
   strictEqual(operation.examples?.length, 1);
   strictEqual(operation.examples![0].filePath, "template.json");
 });
+
+it("multiple services without versioning", async () => {
```
lirenhe: It is better to add more cases, including positive and negative cases to verify the behavior.

packages/typespec-client-generator-core/design-docs/multiple-services.md
```diff
@@ -0,0 +1,176 @@
+# Multiple Service Support in TypeSpec Client Generator Core
+
+## Background
+
+Previously, TCGC [client](./client.md) only supported generating a client from a single service. However, in real-world scenarios, a single package often contains multiple services. TCGC must support multiple services within one package to address these needs.
+
+## User Scenario
+
+1. Merging multiple services' namespaces into one client
```
MaryGao: Do we have conflicts? For example do we have duplicated operations_list among sub services?

packages/typespec-client-generator-core/design-docs/multiple-services.md
```diff
@@ -0,0 +1,176 @@
...
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+@client({
+  name: "CombineClient",
+  service: [ServiceA, ServiceB],
+})
+@useDependency(ServiceA.VersionsA.av2, ServiceB.VersionsB.bv2) // optional
+namespace CombineClient;
```
MaryGao: Trying to map this case to compute one, do we have followings?
- CombineClient => Compute;
- ServiceA => Compute.Disk;
- ServiceB => Compute.Gallary;

So we need to update the existing typespec?

packages/typespec-client-generator-core/design-docs/multiple-services.md
```diff
@@ -0,0 +1,176 @@
...
+2. Create sub-clients for each service's nested namespaces or interfaces. Each sub-client will have its own `apiVersion` property and initialization method if the service is versioned.
+3. Operations directly under each service's namespace are placed under the root client. Operations under nested namespaces or interfaces are placed under the corresponding sub-clients.
+4. Decorators such as `@clientLocation`, `@convenientAPI`, `@protocolAPI`, `@moveTo`, and `@scope` work as usual.
+5. All other TCGC logic remains unchanged.
+6. Since TCGC does not check if merging multiple services will cause sub-clients, models, operations, or other name conflicts, emitters must handle these conflicts appropriately.
+
+For the example above, TCGC will generate types like:
+
+```yaml
+clients:
```
MaryGao: From emitter side, can we distingush the below cases?
- General mgmt SDK with ClientA which has two operation groups defined;
- A combined ClientA which includes two sub services merged which has one operation group per service.

packages/typespec-client-generator-core/src/example.ts
```diff
@@ -60,79 +60,111 @@ async function checkExamplesDirExists(host: CompilerHost, dir: string) {
  */
 async function loadExamples(
```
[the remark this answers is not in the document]
tadelesh: This function is used to load all examples, not for single service.

packages/typespec-client-generator-core/test/examples/load.test.ts
tadelesh: Added.

packages/typespec-client-generator-core/design-docs/multiple-services.md
```diff
@@ -0,0 +1,176 @@
...
+import "./ServiceB/main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+@client({
+  name: "CombineClient",
+  service: [ServiceA, ServiceB],
+})
+@useDependency(ServiceA.VersionsA.av2, ServiceB.VersionsB.bv2) // optional
```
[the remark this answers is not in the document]
tadelesh: For the first question, please check the rest doc.
For the second question, TCGC rely on TypeSpec versioning to handle different versioning. Then it will come out two model with same name in the exported code model for this case.

packages/typespec-client-generator-core/design-docs/multiple-services.md
tadelesh: Yes. Original one.

packages/typespec-client-generator-core/design-docs/multiple-services.md
tadelesh: No explicit property will show that. You could only infer that from root client without api versions.

packages/typespec-client-generator-core/design-docs/multiple-services.md
tadelesh: For duplication, there is an issue to track: https://github.com/Azure/typespec-azure/issues/1039.

packages/typespec-client-generator-core/lib/decorators.tsp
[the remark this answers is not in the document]
lirenhe: I asked copilot for this case, here is some suggestion:
-  If you control the next API version: migrate to services: Namespace[]
- If you must keep the existing version wire‑compatible but need to handle multiple items: add serviceList: Namespace[] and keep service for legacy
- Use the union form only when there is a hard requirement to preserve mixed wire shapes

I think if we want to ensure the naming consistent in the long term, shall we do the change now?

packages/typespec-client-generator-core/src/cache.ts
```diff
@@ -382,7 +493,10 @@ function buildHierarchyOfOperationGroups(
   return undefined;
 }
 
-function isArm(service: Namespace): boolean {
+function isArm(service: Namespace[] | Namespace): boolean {
+  if (Array.isArray(service)) {
```
lirenhe: I go through the code,  the function parameter of service would `Namespace`  or  `Namespace[] | Namespace`.
And we use different logic to handle 'service' throughout the code.  I suggest to make it consistent at the early stage.

weidongxu-microsoft (approved):

packages/typespec-client-generator-core/design-docs/multiple-services.md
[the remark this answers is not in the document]
MaryGao: @tadelesh Could we create an issue to track this topic? Maybe a linter first to avoid this case?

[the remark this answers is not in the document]
tadelesh: [Here](https://github.com/Azure/typespec-azure/issues/3680).

PR #3644, opened 2025-12-12
Fix autorest omit unknown format

timotheeguerin:


markcowl (commented):
It looks like one instance of this in main, and we should have a check on RPSaaSMaster before we merge this

PR #3657, opened 2025-12-15
[tcgc] set array encoding for properties that are array types through `@alternateType`

iscai-msft:


timotheeguerin (changes requested):
Per the discussion I don't think this is how it should be done.

If having a different type is needed for some reason then the encode should be on a model is that is the array type. The sample code below will generate a warning/error when issue is fixed

PR #3734, opened 2026-01-08
Add typespec-metadata emitter

tjprescott:
Closes https://github.com/Azure/azure-sdk-tools/issues/13302

Here's an example output:
```yaml
emitterVersion: 0.1.0
generatedAt: 2026-01-16T20:02:03.538Z
typespec:
  namespace: Storage.Blob
  type: data
languages:
  python:
    emitterName: "@azure-tools/typespec-python"
    packageName: azure-storage-blob
    namespace: azure.storage.blob
    outputDir: "{output-dir}/sdk/storage/azure-storage-blob"
    flavor: azure
    serviceDir: sdk/storage
  csharp:
    emitterName: "@azure-tools/typespec-csharp"
    packageName: Azure.Storage.Blob
    namespace: Azure.Storage.Blob
    outputDir: "{output-dir}/sdk/storage/Azure.Storage.Blob"
    flavor: azure
    serviceDir: sdk/storage
  typescript:
    emitterName: "@azure-tools/typespec-ts"
    packageName: "@azure-rest/azure-storage-blob"
    namespace: "@azure-rest/azure-storage-blob"
    outputDir: "{output-dir}/sdk/storage/azure-storage-blob-rest"
    flavor: azure
    serviceDir: sdk/storage
  java:
    emitterName: "@azure-tools/typespec-java"
    packageName: azure-storage-blob
    namespace: com.azure.storage.blob
    outputDir: "{output-dir}/sdk/storage/azure-storage-blob"
    flavor: azure
    serviceDir: sdk/storage
  rust:
    emitterName: "@azure-tools/typespec-rust"
    packageName: azure_storage_blob
    namespace: azure_storage_blob
    outputDir: "{output-dir}/sdk/storage/azure_storage_blob"
    flavor: azure
    serviceDir: sdk/storage
sourceConfigPath: C:/repos/azure-rest-api-specs/specification/storage/Microsoft.BlobStorage/tspconfig.yaml
```

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,28 +27,28 @@ importers:
         version: 9.39.2
       '@pnpm/workspace.find-packages':
         specifier: ^1000.0.24
-        version: 1000.0.49(@pnpm/logger@1001.0.1)(@pnpm/worker@1000.4.1(@pnpm/logger@1001.0.1)(@types/node@25.0.2))(typanion@3.14.0)
+        version: 1000.0.55(@pnpm/logger@1001.0.1)(@pnpm/worker@1000.6.2(@pnpm/logger@1001.0.1)(@types/node@25.0.9))(typanion@3.14.0)
       '@types/node':
         specifier: ~25.0.2
-        version: 25.0.2
+        version: 25.0.9
       '@vitest/coverage-v8':
         specifier: ^4.0.15
-        version: 4.0.15(vitest@4.0.15)
+        version: 4.0.17(vitest@4.0.17)
       '@vitest/eslint-plugin':
         specifier: ^1.5.2
-        version: 1.5.2(eslint@9.39.2)(typescript@5.9.3)(vitest@4.0.15)
+        version: 1.6.6(eslint@9.39.2)(typescript@5.9.3)(vitest@4.0.17)
       c8:
         specifier: ^10.1.3
         version: 10.1.3
       cspell:
         specifier: ^9.4.0
-        version: 9.4.0
+        version: 9.6.0
       eslint:
         specifier: ^9.39.2
         version: 9.39.2
       eslint-plugin-import:
         specifier: ^2.31.0
-        version: 2.32.0(@typescript-eslint/parser@8.49.0(eslint@9.39.2)(typescript@5.9.3))(eslint@9.39.2)
+        version: 2.32.0(@typescript-eslint/parser@8.53.1(eslint@9.39.2)(typescript@5.9.3))(eslint@9.39.2)
       eslint-plugin-unicorn:
         specifier: ^62.0.0
         version: 62.0.0(eslint@9.39.2)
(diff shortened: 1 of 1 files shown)


packages/typespec-metadata/src/collector.ts
```diff
@@ -0,0 +1,509 @@
...
+  optionMap: Record<string, Record<string, unknown>>,
+  params: Record<string, unknown>,
+  baseOutputDir: string,
+  defaultServiceDir?: string,
+): Record<string, LanguagePackageMetadata> {
+  // Only include SDK language emitters
+  const includedEmitters = [
+    "@azure-tools/typespec-python",
+    "@azure-tools/typespec-java",
+    "@azure-tools/typespec-csharp",
```
weshaggard: Should we use a naming pattern instead of a hard-coded list? This already doesn't look complete as I know we have newer emitters for csharp.

tjprescott: I figured an include list made more sense than an exclude list. Otherwise it will collect information for any emitters in tsp-config including ones we don't care (or possibly know) about.

packages/typespec-metadata/README.md
```diff
@@ -0,0 +1,54 @@
...
+  - "@azure-tools/typespec-metadata"
+options:
+  "@azure-tools/typespec-metadata":
+    output-dir: "./metadata"
+    output-file: "api-metadata.yaml"
+```
+
+## Output
+
+The emitter generates structured metadata in YAML or JSON format that includes:
```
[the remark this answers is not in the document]
tjprescott: Yes, I'll add an example of the output.

Would one format or another be preferable for pipeline consumption (i.e. as the default)?

packages/typespec-metadata/test/collector.test.ts
```diff
@@ -0,0 +1,232 @@
...
+
+describe("outputDir path handling", () => {
+  it("should replace absolute base path with {output-dir} placeholder", () => {
+    const baseDir = "c:/repos/project/tsp-output";
+    const absolutePath = "c:/repos/project/tsp-output/sdk/keyvault/secrets";
+    const expected = "{output-dir}/sdk/keyvault/secrets";
+
+    // Test the logic that converts absolute to placeholder
+    const normalizedBase = baseDir.replace(/\\/g, "/").replace(/\/$/, "");
+    const normalizedPath = absolutePath.replace(/\\/g, "/");
```
timotheeguerin: seems like all the test are doing this, can the test path not be a unix path so start with and use path utils from the compiler or `pathe` package to normalize

packages/typespec-metadata/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "main": "dist/src/index.js",
+  "types": "dist/src/index.d.ts",
+  "files": [
+    "dist/**",
+    "!dist/test/**"
+  ],
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
```
timotheeguerin:
```suggestion
      "import": "./dist/src/index.js"
```

packages/typespec-metadata/README.md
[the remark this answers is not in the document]
tjprescott: Added an example. You can configure YAML or JSON as an emitter option, but it defaults to YAML.

packages/typespec-metadata/src/collector.ts
tjprescott: Arguably no, but if it's filled with weird random stuff, I figured it would be confusing for consumers.

tjprescott: I'll remove it and see if it causes any problems.

PR #3736, opened 2026-01-09
Add Action templates and allow name parameter changes in NSP operations

markcowl:


packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
```diff
@@ -3504,6 +3504,206 @@ model DependentProperties {
...
+model EmployeeProperties {
+  age?: int32;
+
+
+  @visibility(Lifecycle.Read)
+  provisioningState?: ProvisioningState;
+}
+
+@lroStatus
+union ProvisioningState {
```
timotheeguerin: for the sake of keeping those test a little simpler I think we can just use `ResourceProvisioningState` directly

PR #3747, opened 2026-01-10
Allow final-state-via: original-uri for PATCH operations

markcowl:


qiaozha (approved):
wonder if we should add a test somewhere?

PR #3768, opened 2026-01-14
[TCGC] Refine diagnostic for LRO and paging metadata generation.

tadelesh:
Fix regression for latest release.

### DIFF
--- a/packages/typespec-client-generator-core/src/methods.ts
+++ b/packages/typespec-client-generator-core/src/methods.ts
@@ -515,7 +515,7 @@ function getServiceMethodLroMetadata<TServiceOperation extends SdkServiceOperati
     }
     return {
       kind: "reference",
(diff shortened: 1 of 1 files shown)


packages/typespec-client-generator-core/src/methods.ts
```diff
@@ -515,7 +515,8 @@ function getServiceMethodLroMetadata<TServiceOperation extends SdkServiceOperati
     }
     return {
       kind: "reference",
-      operation: diagnostics.pipe(getSdkBasicServiceMethod(context, reference.operation, client))
+      // since these operations may not be included in the client customization, we ignore diagnostics here
+      operation: ignoreDiagnostics(getSdkBasicServiceMethod(context, reference.operation, client))
```
weidongxu-microsoft: But I am still not sure why diagnostics would be reported in the first place.

The `ResourceRead` template should contain the api-version?

PR #3775, opened 2026-01-14
[tcgc] crash if duplicate models / enums across namespaces when flattened with namespace flag or multi service

iscai-msft:
Resolve: https://github.com/Azure/typespec-azure/issues/1039

### DIFF
--- a/packages/typespec-client-generator-core/test/clients/structure.test.ts
+++ b/packages/typespec-client-generator-core/test/clients/structure.test.ts
@@ -1560,9 +1560,6 @@ it("one client from multiple services with models shared across services", async
     )
     @useDependency(ServiceA.VersionsA.av2, ServiceB.VersionsB.bv2)
     namespace CombineClient;
-
-    // Rename ServiceB's SharedModel to avoid naming conflict
-    @@clientName(ServiceB.SharedModel, "SharedModelB");
   `,
     ),
   );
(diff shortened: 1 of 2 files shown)


packages/typespec-client-generator-core/src/validations/types.ts
tadelesh: The following `@clientLocation` logic for operation has one legacy issue: we now allow to move to existed ns with `string` (`@@clientLocation(op, "Test"); namespace Test {}`), which is not handled in the code.

packages/typespec-client-generator-core/src/validations/types.ts
```diff
@@ -80,14 +112,27 @@ function validateClientNames(tcgcContext: TCGCContext) {
...
 
-    // Validate client names for the current scope
-    validateClientNamesPerNamespace(
-      tcgcContext,
-      scope,
-      moved,
-      movedTo,
-      tcgcContext.program.getGlobalNamespaceType(),
-    );
+    if (isMultiService) {
```
tadelesh: I don't think multiple service is a criteria for this check. Shall we use the criteria: whether honor TypeSpec namespace?

packages/typespec-client-generator-core/src/package.ts
```diff
@@ -27,6 +28,11 @@ export function createSdkPackage<TServiceOperation extends SdkServiceOperation>(
   context: TCGCContext,
 ): [SdkPackage<TServiceOperation>, readonly Diagnostic[]] {
   const diagnostics = createDiagnosticCollector();
+
+  // Validate cross-namespace names if namespace flag is set (flattens namespaces)
+  // Can't validate in $onValidate bc we don't have access to the namespace flag
+  validateCrossNamespaceNamesWithFlag(context, diagnostics);
```
tadelesh: We always have duplicate logic because of the config for emitter could not be get when `compile` or `validate`. Could you help to confirm with TypeSpec team to see if there is any solution that we could have such info.

packages/typespec-client-generator-core/src/validations/types.ts
```diff
@@ -191,6 +236,116 @@ function validateClientNamesPerNamespace(
...
+  movedTo: Map<Namespace | Interface, Operation[]>,
+  serviceNamespaces: Namespace[],
+) {
+  // Collect all types from all service namespaces
+  const allModels: Model[] = [];
+  const allEnums: Enum[] = [];
+  const allUnions: Union[] = [];
+  const allScalars: Scalar[] = [];
+
+  for (const serviceNs of serviceNamespaces) {
```
tadelesh: This missed the check for imported lib: `Azure.Core`, `Azure.ResourceManager`, etc.

packages/typespec-client-generator-core/src/validations/types.ts
tadelesh: I'm designing extending multiple service to all kinds of scenarios. In that case, it depends on the spec author if the client should follow "namespace is namespace".

tadelesh (commented):
If we decide to check the cross namespace collision after `createSdkContext`, I believe the algorithm could be much easier, just check all types in `SdkPackage`.

tadelesh (approved):

Issue #3276, opened 2025-09-12
Rollout of breaking changes(or not) and integration with client emitters


timotheeguerin:
This is an issue to establish an official plan for how we should update the spec repo and the sdk repo to reduce the likelihood of the SDK CI failing due to a breaking change (or a new type/decorator that a spec is trying to use).

Right now we release TypeSpec at a fixed schedule
1. First Tuesday at the end of the sprint
2. The spec repo gets updated either that Tuesday or the next(Wednesday)
3. SDK repo have been inconsistent on the speed at which they can update

Spec repo ci will also improve to give better message when those error happen https://github.com/Azure/azure-rest-api-specs/issues/37391

## Proposal
- Update spec repo day after the release giving time for SDK emitter to be released and SDK repos to be updated
- Breaking changes should as much possible be doing a smooth rollout(introduce new capability in release 1, migrate everything(Later not during spec repo upgrade), make breaking change in release 2, by that point no spec should have been using the breaking change)
- Breaking change should be done early in a sprint, to give time to nightly and SDK emitters to have a PR ready
- For hard breaking change before merging to typespec(-azure) a pr handling the change (if needed) should be ready for all emitters

PR #3779, opened 2026-01-15
Revert @identifiers decorator and deprecate x-ms-identifiers generation

qiaozha:
## Summary
This PR reverts the changes introduced in PR #1983 which added the @identifiers decorator and automatic x-ms-identifiers generation from @key decorator.

## Changes
- ✅ Removed @identifiers decorator from typespec-azure-resource-manager
- ✅ Removed automatic x-ms-identifiers generation from @key decorator  
- ✅ Removed x-ms-identifiers generation logic from typespec-autorest emitter
- ✅ Reverted to using @OpenAPI.extension("x-ms-identifiers", ...) for explicit specification
- ✅ Removed related tests for @identifiers decorator
- ✅ Updated documentation to reflect the changes
- ✅ Added deprecation changelog entry

## Related Issues
Reverts: #1983

## Breaking Changes
- The @identifiers decorator is no longer available
- x-ms-identifiers is no longer automatically generated from @key decorators
- Users must use @OpenAPI.extension("x-ms-identifiers", ...) to specify identifiers explicitly when needed

### DIFF
--- a/packages/typespec-autorest/CHANGELOG.md
+++ b/packages/typespec-autorest/CHANGELOG.md
@@ -7,6 +7,10 @@
 - [#3622](https://github.com/Azure/typespec-azure/pull/3622) Add support for multiple output files in typespec-autorest
 - [#3746](https://github.com/Azure/typespec-azure/pull/3746) #3693 Allow array and unknown result for async operations
 
+### Deprecations
+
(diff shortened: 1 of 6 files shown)


timotheeguerin (changes requested):
@qiaozha before this happens we need full confirmation from ARM team that this is uncesssary and this change cannot be done in such a breaking manner. It needs to be rolled out following the [process we've established](https://github.com/Azure/typespec-azure/issues/3276).
1. deprecate only
2. migrate everything in spec repo
3. remove 1-2 sprint later

PR #3813, opened 2026-01-21
Add detailed versioning docs for single preview

markcowl:


website/src/content/docs/docs/howtos/ARM/versioning/02-preview-after-preview.md
```diff
@@ -0,0 +1,61 @@
...
+  - In any editor, search and replace the latest preview version in the spec with the new preview version
+- Update the version value of this version to match the new api-version string, for example:
+
+  ~~v2025_12_01_preview: "2025-12-01-preview",~~
+  v2026_01_01_preview: "2026-01-01-preview",
+
+- Update any version documentation to use the new version
+- Change the name of the `examples` version folder for the latest preview to match the new preview version
+- Make changes to the API description based on how the API has changed
+  - If any type that was introduced in the latest preview is _not_ in the new preview, simply remove the type
```
haolingdong-msft: >  simply remove the type

just would like to clarify my understanding, does 'remove' here mean we just delete the type, rather than adding a the `@removed` decorator?

website/src/content/docs/docs/howtos/ARM/versioning/02-preview-after-preview.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Rename the latest preview version to match the new preview version, in all instances in the spec
+  - In vscode, highlight the version name and select `rename symbol` from the context menu to rename the version throughout your spec
+  - In any editor, search and replace the latest preview version in the spec with the new preview version
+- Update the version value of this version to match the new api-version string, for example:
+
+  ~~v2025_12_01_preview: "2025-12-01-preview",~~
+  v2026_01_01_preview: "2026-01-01-preview",
+
+- Update any version documentation to use the new version
```
haolingdong-msft: > version documentation
Could I understand more on what does version documentation mean here? Does it mean all the `@doc` ?

website/src/content/docs/docs/howtos/ARM/versioning/07-perpetual-preview.md
```diff
@@ -0,0 +1,47 @@
...
+  - Remove its example folder
+  - Remove all references to version A in `README.md`
+
+## Create A Copy of the Spec for the Stable Version only
+
+- Create two copies of your spec
+  - One which will contain just the new stable version (use this to create the first PR into the specs repo). Call this Copy 1.
+  - The other will contain both the new stable _and_ preview versions (use this to create the final PR after the first PR is merged). Call this Copy 2.
+- Do the following with Copy 1
+  - Remove version `A + 2` and all decorators that reference version `A + 2` from the spec.
```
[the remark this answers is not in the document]
markcowl: Correct.  I think this is a little oversimplified, so will update with instructions on what to do with specific decorators, this should essentially be removing any decorators in the new preview (A+2) that are *duploicated* in the new stable (A + 1).  Any decorators in (A + 2) and *not* A + 1 should be reversed and removed.

website/src/content/docs/docs/howtos/ARM/versioning/07-perpetual-preview.md
```diff
@@ -0,0 +1,47 @@
...
+  - Compile the spec to produce artifacts (especially the new stable version (`A + 1`) openapi )
+  - Add the new stable version (`A + 1`) to the README.md file.
+  - Create and merge the PR
+
+## Create a PR with the Combined Spec
+
+- Do the following with Copy 2
+  - Follow the instructions for normalizing decoration in the [converting specifications](./06-converting-specs.md#normalizing-version-decoration-optional) document. This will remove any redundant decoration between the new stable and preview versions (`A + 1` and `A + 2`).
+  - Compile the spec to produce artifacts (especially the new stable and preview version (`A + 1` and `A + 2` ) apis).
```
markcowl: Need to add any type changes in the new preview here.  Also, this is a better place to add examples for the new preview

website/src/content/docs/docs/howtos/ARM/versioning/07-perpetual-preview.md
```diff
@@ -0,0 +1,59 @@
...
+  - if a type was made optional in `A` and this change is now stable (`@madeOptional(T, A)`), add the new decorator `@madeOptional(T, A + 1)`
+  - if a type was renamed in `A` and this change is now stable (`@renamedFrom(T, A, Name)`), add the new decorator `@typeChangedFrom(T, A + 1, Name)`
+  - if a property or parameter had its type changed in `A` and is now stable (`@typeChangedFrom(T, A, U)`), add the new decorator `@typeChangedFrom(T, A + 1, U)`
+  - if an operation returnType was changed in `A` and this change is now stable (`@returnTypeChangedFrom(T, A, U)`), add the new decorator `@returnTypeChangedFrom(T, A + 1, U)`
+  - If a type was removed in `A` and this change is now stable (`@removed(T, A)`), add the new decorator `@removed(T, A + 1)`
+- Change all versioning decorators `dec(T, A, args)` to `dec(T, A + 2, args)` where `T` is a type, `A` is the latest preview version, `A + 2` is the new preview version you added in the first step and `args` are any additional arguments to the decorator. Note that, after this change, some decorators will be duplicated in version `A + 1` and version `A + 2`. This is expected.
```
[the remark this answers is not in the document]
markcowl: Unfortunately, the breaking change tools will not operate on PRs that introduce 2 api-versions, so we *have to* create 2 PRs.

I don't see any way to make this uncomplicated (although, if there is one, I'd prefer that, too),  which is why I think this is a good candidate for an AI agent or the like.

website/src/content/docs/docs/howtos/ARM/versioning/01-about-versioning.md
```diff
@@ -0,0 +1,30 @@
...
+TypeSpec uses a versioning library that models the changes in each new version of the API, rather than having a separate api description for each api-version.  
+This works well when APIs evolve according to versioning guidelines, without breaking changes. For the most part, this means that this system is very good at
+modeling differences between stable api-versions for Azure APIs, but can be cumbersome when describing differences between preview APIs.
+
+Additionally, in Azure. preview APIs have a limited lifespan and limited support in SDKs and other tooling. For this reason and others, specs should only have a _single active preview_ at any point during the spec development process.
+
+At the same time, Azure ResourceManager teams may need to maintain OpenAPI files for preview versions until they are retired, some reasons for this include:
+
+- RPaaS live validation support
+- ARM registration support
```
qiaozha: I think Docs team is also leveraging OpenAPI files.

website/src/content/docs/docs/howtos/ARM/versioning/02-preview-after-preview.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Making Changes to your TypeSpec spec
+
+- Rename the latest preview version to match the new preview version, in all instances in the spec
+  - In vscode, highlight the version name and select `rename symbol` from the context menu to rename the version throughout your spec
+  - In any editor, search and replace the latest preview version in the spec with the new preview version
+- Update the version value of this version to match the new api-version string, for example:
+
+  ~~v2025_12_01_preview: "2025-12-01-preview",~~
+  v2026_01_01_preview: "2026-01-01-preview",
```
qiaozha:
```suggestion
\```diff
  -v2025_12_01_preview: "2025-12-01-preview",
  +2026_01_01_preview: "2026-01-01-preview",
\```
```

website/src/content/docs/docs/howtos/ARM/versioning/02-preview-after-preview.md
```diff
@@ -0,0 +1,62 @@
...
+
+- The new preview version should already be the _last version_ of the versions enum, also ensure it is decorated with `@previewVersion`
+- Update any version documentation to use the new version
+- Change the name of the `examples` version folder for the latest preview to match the new preview version
```
qiaozha: Should we also add an entry to update the actual api-version in those swagger examples?

website/src/content/docs/docs/howtos/ARM/versioning/02-preview-after-preview.md
```diff
@@ -0,0 +1,62 @@
...
+- Make changes to the API description based on how the API has changed
+  - If any type that was introduced in the latest preview is _not_ in the new preview, simply delete the type
+  - If any other types are removed in this preview (unlikely) mark these with an `@removed` decorator referencing the new version
+  - If any types are added, renamed, or otherwise modified in the new version, mark them with the appropriate versioning decorator
```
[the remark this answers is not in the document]
qiaozha: If we just have vlatestPreview enum key to point to the latest preview version, they actually don't need to use versioning decorators for changes happened between preview after preview? Everything in latest preview version should be decorated as `@added(Versions.vlatestPreview)` as the previous preview version information will be lost from typespec anyway, why we need to care if it's added/renamed or otherwise modified in the new preview?

website/src/content/docs/docs/howtos/Versioning/ARM/02-preview-after-preview.md
```diff
@@ -0,0 +1,62 @@
...
+  - In vscode, highlight the version name and select `rename symbol` from the context menu to rename the version throughout your spec
+  - In any editor, search and replace the latest preview version in the spec with the new preview version
+- Update the version value of this version to match the new api-version string, for example:
+
+  ~~v2025_12_01_preview: "2025-12-01-preview",~~
+  v2026_01_01_preview: "2026-01-01-preview",
+
+- The new preview version should already be the _last version_ of the versions enum, also ensure it is decorated with `@previewVersion`
+- Update any version documentation to use the new version
+- Change the name of the `examples` version folder for the latest preview to match the new preview version
```
qiaozha: Oh I see that from the last entry, might be better to put all examples changes together?

website/src/content/docs/docs/howtos/Versioning/ARM/03-stable-after-preview.md
```diff
@@ -0,0 +1,125 @@
...
+  - Update the versioning decorators for those changes to reference the new stable version
+
+    ```diff lang=tsp
+    - @added(Versions.`2025-10-01-preview`)
+    + @added(Versions.`2026-01-01`)
+      remainingProperty?: string;
+    ```
+
+  - For changes in the latest preview (p) that are _not_ in the new stable version
+    - For any type with an `@added(p)` decorator, delete the type
```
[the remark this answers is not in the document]
qiaozha: In fact, I don't think we should delete, as typespec should still keep latest preview in track.

website/src/content/docs/docs/howtos/Versioning/ARM/03-stable-after-preview.md
```diff
@@ -0,0 +1,125 @@
...
+      - model Foo {}
+      ```
+
+    - For any property or parameter with a `@typeChangedFrom(p, Type)` decorator, replace the property type with the `Type` argument, and then remove the decorator, for example
+
+      ```diff lang=tsp
+      - @typeChangedFrom(Versions.`2025-12-01-preview`, string)
+      - property: int32;
+      + property: string;
+      ```
```
qiaozha: If I understand correctly, this property is string type in stable, but int32 in latest preview? so it actually has two versioning decorators on this property `@added(Versions.SomeOlderStableVersion)` and `@typeChangedFrom(Versions.vlatestPreview)`?  
I wonder if typespec can limit the versioning changes to preview and not extend it to stable version newer than the latest preview for a versioning decorator if the version parameter of that decorator is a preview version?

website/src/content/docs/docs/howtos/Versioning/ARM/04-preview-after-stable.md
```diff
@@ -0,0 +1,79 @@
...
+  ```diff lang=tsp
+  enum Versions {
+    `2025-10-01`,
+  + @Azure.Core.previewVersion
+    `2026-01-01-preview`,
+  }
+  ```
+
+- Make changes to the API description based on how the API has changed
+  - If any types are removed in the new preview (unlikely) mark these with an `@removed` decorator referencing the new version
```
qiaozha: I feel like this is not that unusual? as service could have some new features they would like to retire but they want to remove it from preview first to get some earlier feedback?

qiaozha (commented):
Haven't read all of them yet.

website/src/content/docs/docs/howtos/Versioning/ARM/01-about-versioning.md
```diff
@@ -0,0 +1,30 @@
...
+- [How to add a new stable version when the last version was preview](./03-stable-after-preview.md)
+- [How to add a new preview version when the last version was stable](./04-preview-after-stable.md)
+- [How to add a new stable version when the last version was stable](./05-stable-after-stable.md)
+
+This document also describes how to move an existing multi-api typespec spec with multiple previews:
+
+- [How to convert a spec with multiple preview versions into a spec with a single active preview](./06-converting-specs.md)
+
+Additionally, there are some services that may always have features that remain in preview after a stable version is released. This can happen, for example, if there are multiple independent teams that own different resources in a service and operate on their own schedule. The recommended way to handle
+this scenario is to model your ResourceProvider as multiple services, so each version and the corresponding SDKs can version independently. For some older services, this is not an option, so there is specialized guidance on how to maintain preview features over stable api releases:
```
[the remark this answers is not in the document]
markcowl: You are right about taking the customer perspective.  The reason for this advice is that the customer is served best by being easily able to get the latest and latest stable for features that they use.  If a service is really made up of independent teams that version independently, our unit of versioning is at the sdk level, so having separate SDKs makes sense in this case.

There are some cases where this is not appropriate, for example, if the break would cause too much customer pain, or, in a case like Kubernetes where versioning is frequent, but different parts of the service cannot easily be separated.

I think adding a caveat here about having a discussion with the SDK team before deciding to split up their service (or not) makes a lot of sense.

website/src/content/docs/docs/howtos/Versioning/ARM/06-converting-specs.md
```diff
@@ -0,0 +1,187 @@
+---
+title: Converting Existing Specs with Multiple Previews
+llmstxt: true
+---
+
+Converting a spec with multiple preview versions into a spec with a single, latest preview version is complex because the changes that occur in any
+preview version are carried over into the next version by default. This means that each versioning decorator referencing a preview version must be dealt with _in order_ to end with the correct decoration.
+
+For each type decorated with a versioning decorator that references a preview version, follow these steps. If the type is decorated with _more than one instance_ of a versioning decorator referencing different preview versions, each of these must be processed _in version order (earliest version first)_
```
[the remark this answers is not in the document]
markcowl: If we could create a PR that added both a new stable and a new preview version, this would be the advice  (and I think, in the medium term, we will have the capability to create a PR based on only *some versions* of the spec,  which means that the user could just add both api-versions and use a tool to create spearate PRs for each api-version).

The mechanism here is spoecifically to create those two PRs with the minimum number of steps.

timotheeguerin (approved):

PR #3827, opened 2026-01-22
[tcgc] add `@clientOption` decorator

### LINKED ISSUE #2228: [TCGC] Add a clientOption(name, value, scope) decorator

We want to be able to pass experimental flags fast to emitter, without having to reship TCGC. For instance:

```
@clientOption("enableFeatureFoo", true, "python")
model MyModel {}
```

Target can be anything, name is a string, value can be anything (emitter's job to cast), and scope as usual.

We should have a permanent linter rule applied to that decorator, that's it's required to suppress. This is intended for experiment, or quick escape hatch and shouldn't be used outside of that context, or needs to be justified in the suppression.

Options of that decorator are NOT expected to be documented publicly, for the reason given before.

I expect this to be exposed in the TCGC YAML as something like a list:
```yaml
clientOption:
- enableFeatureFoo: true
```

And any emitter can check if "clientOption" for something they support is there or not.

### Checklist

iscai-msft:
fixes #2228


packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -402,6 +402,14 @@ export function getTypeDecorators(
             getDecoratorArgValue(context, decorator.args[i].jsValue, type, decoratorName),
           );
         }
+
```
iscai-msft: @tadelesh is there a reason we didn't filter raw decorators by scope before? I think we should because we only want actionable decorator values for a language emitter

packages/typespec-client-generator-core/src/public-utils.ts
```diff
@@ -897,3 +900,32 @@ export function getNamespaceFromType(
...
+ */
+export function getClientOptions(decorators: DecoratorInfo[]): SdkClientOption[] {
+  return decorators
+    .filter((d) => d.name === CLIENT_OPTION_DECORATOR_NAME)
+    .map((d) => ({
+      name: d.arguments.name as string,
+      value: d.arguments.value as Value,
+      scope: d.arguments.scope as string | undefined,
+    }));
+}
```
tadelesh: May be `getClientOptions<T extends DecoratedType>(type: T, key): unknown` is much useful?

joheredi (approved):
Looks good! My only question is process-wise, what is the expected lifecycle of an experimental decorator. Once it is validated and proven helpful do we expect the emitter to expose an actual decorator to replace the @@clientOption?

Issue #3868, opened 2026-01-30
[TCGC] Doc auto-update agent


tadelesh:
We have several parts of docs for TCGC:
1. User doc under https://github.com/Azure/typespec-azure/tree/main/website/src/content/docs/docs/howtos/Generate%20client%20libraries which tell TypeSpec user how a spec will be generated to a client code and how to customize the generation.
2. Doc for emitter dev: https://github.com/Azure/typespec-azure/blob/main/website/src/content/docs/docs/libraries/typespec-client-generator-core/guideline.md which show the export types and meanings for TCGC.
3. Design docs: https://github.com/Azure/typespec-azure/tree/main/packages/typespec-client-generator-core/design-docs which are detailed designs for TCGC feature

We have Spector samples to show the functionality of TCGC's feature: https://github.com/Azure/typespec-azure/tree/main/packages/azure-http-specs/specs.

And, TCGC has many code based document and test case which strongly align with the TCGC's features.

For all above, current satiation is they are all not so complete and update to date.

We want to have an agent to gradually check them and help to update them.

PR #3888, opened 2026-02-04
[TCGC] Action to auto update TCGC doc

tadelesh:
Resolve: https://github.com/Azure/typespec-azure/issues/3868

### DIFF
--- a/tsconfig.eng.json
+++ b/tsconfig.eng.json
@@ -4,5 +4,5 @@
     "allowJs": true
   },
   "include": ["./eng", "./core/eng"],
-  "exclude": ["./core/eng/vitest.config.ts"]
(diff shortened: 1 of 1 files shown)


website/src/content/docs/docs/howtos/Generate client libraries/08types.mdx
```diff
@@ -654,6 +667,81 @@ model Config {
...
+
+</ClientTabItem>
+<ClientTabItem lang="java" >
+
+```java
+// When the service returns "", the value will be deserialized as null
+public final class MyModel {
+    public String getDisplayName() { ... }
+    public String getDescription() { ... }
+}
```
weidongxu-microsoft: Java does not support this `@deserializeEmptyStringAsNull`.

PR #3931, opened 2026-02-12
Add hotfix release skill

tadelesh:


tadelesh:
I only tested it locally. I could not ensure it could work well directly from Github copilot agent before the skill is in.

timotheeguerin (approved):

PR #3963, opened 2026-02-23
Add documentation missing for linting rules

timotheeguerin:
Generated documentation for the linting rules with copilot cli. 
Seems to have done a pretty good job. I don't know fully the purpose of all of them so need a 2nd set of eyes


website/src/content/docs/docs/libraries/azure-core/rules/rpc-operation-request-body.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+DELETE operation with no body:
+
+```tsp
+@delete
+op deleteWidget is RpcOperation<{}, Widget>;
+```
+
+GET/DELETE with query parameters using `@bodyIgnore`:
```
markcowl: Also add a valid example for an RPC POST with body

Issue #3782, opened 2026-01-15
[TCGC]: @clientInitialization Decorator Should Allow Making a SubClient a Root Client


jorgerangel-msft:
See [playground link](https://azure.github.io/typespec-azure/playground/?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fdata-plane%22%5D%7D%7D&e=%40azure-tools%2Ftypespec-client-generator-core&vs=%7B%7D&c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsKyTbIKi92ZXJzaW9uaW5n1R9odHRwIjsKCnVzaW5nIEHEbC5DxV1HyFwuQ29yZTvHIlbJT8gSSHR0cDsKCi8qCiAqIEZpcnN0IHNlcnZpY2UgZGVmaW5pdGlvbiBpbiBhIG11bHRpLcgecGFja2FnZSB3aXRoIOoApwogKi8KQMcQZWQox3JzQSkKbmFtZXNwYWNlIFPGREEgewogIGVudW3oAJlzxhMgIGF2MSzHCTLECX0KCiAgQHJvdXRlKCJmb28iKQogIGludGVyZsRRRm9vxznIJC90ZXN0xSYgIMQLKEBxdWVyeSgiYXBpLecAnyIpIGFwacd7OuoAhCk6IHZvaWQ7xXh95AEd5QEeU2Vjb25k%2FwEf%2FwEf%2BAEfQvMBH0LyAR%2FGEyAgYugBH2LzAR9iYXLvAR9CYXL%2FAR%2F%2FAR%2FsAR9C8AEfQOYCzyjEYucA%2FzogW%2BgB5CzpAM9dCn3zAOYuTeQBOMgNQ29tYmluZWQge8RZx1pJ5AFnYWxpemHkAW4oyFouRm9vLCB7IOUBhsQfZWRCeTogZ2xvYmFsLvoDDS7JUMQxLmluZGl2aWR1YWxseSB9KTvfd0IuQmFy33ffd9V3). As part of the multiservice client support for date plane clients, some of the .NET architect feedback was to also support a library shape without sub-clients where the service clients are all root clients. Currently, having a sub-client/interface annotated with only `initializedBy: InitializedBy.individually` results in an error.

[...]

PR #3997, opened 2026-03-03
[TCGC] Client related new features and refactor

tadelesh:
1. `SdkOperationGroup` consolidated into `SdkClient`
- `SdkOperationGroup` → `SdkClient`
- `subOperationGroups` → `subClients`
- `groupPath` → `clientPath`
- `SdkClient.service` → `SdkClient.services`
- `getOperationGroup()` → `getClient()`
- `listOperationGroups()` → `listSubClients()`
- `listOperationsInOperationGroup()` → `listOperationsInClient()`
- `isOperationGroup()` / `getOperationGroup()` → removed
2. `@operationGroup` deprecated
- `@operationGroup` now delegates to `@client` and will be removed in a future release.
3. Multi-service client support
- Use `autoMergeService` property to auto create client hierarchy for multiple service scenarios.
- Support full customization of client hierarchy with multiple services.
4. Simplify some client logic
- Root explicit `@client` must have `service` config.
- Nested interface or namespace without `@client` under other `@client` will be ignored.
- Only `autoMergeService` will merge the service's inner assets. Otherwise, the client will keep empty.
- For multiple service case, remove the use of `@useDependency` to decare each service's API version, but use the latest version instead. The Azure Spector test for old version is deleted.

Resolve: https://github.com/Azure/typespec-azure/issues/3782

### DIFF
--- a/packages/typespec-client-generator-core/CHANGELOG.md
+++ b/packages/typespec-client-generator-core/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log - @azure-tools/typespec-client-generator-core
 
+## 0.66.1
+
+### Bug Fixes
+
+- [#4020](https://github.com/Azure/typespec-azure/pull/4020) Optimize `findMappingWithPath` in `getMethodParameterSegments` for better performance with deeply nested models. Replace O(n) `Array.shift()` with O(1) index-based dequeue, and use parent pointer map instead of O(depth) path copying per node.
+- [#4024](https://github.com/Azure/typespec-azure/pull/4024) Refine orphan model logic to reduce the efforts for model calculations
+
+
 ## 0.66.0
 
 ### Features
--- a/packages/typespec-client-generator-core/package.json
+++ b/packages/typespec-client-generator-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/typespec-client-generator-core",
-  "version": "0.66.0",
+  "version": "0.66.1",
   "author": "Microsoft Corporation",
   "description": "TypeSpec Data Plane Generation library",
   "homepage": "https://azure.github.io/typespec-azure",
--- a/packages/typespec-client-generator-core/src/cache.ts
+++ b/packages/typespec-client-generator-core/src/cache.ts
@@ -68,8 +68,9 @@ export function prepareClientAndOperationCache(context: TCGCContext): void {
 
   // iterate all clients and build a map of operations
   const queue: SdkClient[] = [...clients];
-  while (queue.length > 0) {
-    const client = queue.shift()!;
+  let queueIdx = 0;
+  while (queueIdx < queue.length) {
+    const client = queue[queueIdx++];
 
     // operations directly under the client
     const operations = [];
--- a/packages/typespec-client-generator-core/src/decorators.ts
+++ b/packages/typespec-client-generator-core/src/decorators.ts
@@ -82,12 +82,14 @@ import {
   getScopedDecoratorData,
   isSameAuth,
   isSameServers,
+  legacyHierarchyBuildingKey,
   listAllUserDefinedNamespaces,
   negationScopesKey,
   omitOperation,
   overrideKey,
   parseScopes,
   scopeKey,
+  usageKey,
 } from "./internal-utils.js";
 import { createStateSymbol, reportDiagnostic } from "./lib.js";
 import { getSdkEnum, getSdkModel, getSdkUnion } from "./types.js";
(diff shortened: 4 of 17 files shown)


weidongxu-microsoft (commented):
If emitter is already using `listSubClientsUnderClient` etc. (no call on`listOperationGroups` and `listOperationsInOperationGroup`), does it mean there would be no change to behavior (on existing client config)?

weidongxu-microsoft (commented):
I've asked agent to analyze the merge case

```
> What happens to the second sub-client on "The second's services and subClients are appended", in "Multi-service merging" case? Do you think the handle of it is correct in PR?
```
```
  Here's the timeline:

  Step 1: createSubClient creates sc BEFORE the merge check

   buildSubClientHierarchy(ServiceB, ...)
     └─ createSubClient(ServiceB.Operations, ...)
          ├─ Creates SdkClient `sc` with type=ServiceB.Operations, parent=rootClient
          ├─ Recursively builds sc.subClients (grandchildren)
          ├─ __rawClientsCache[ServiceB.Operations] = sc       ← CACHED
          ├─ __clientToOperationsCache[sc] = []                ← CACHED
          └─ returns sc

  Step 2: handleMultipleServicesSubClientNameConflict merges into the first

   existingSc (from ServiceA) absorbs:
     ├─ existingSc.services.push(sc.services[0])       — gains ServiceB
     ├─ existingSc.subClients.push(...sc.subClients)    — steals sc's children
     ├─ existingSc.type = undefined                     — no single backing type
     └─ mergedSubClientTypes[existingSc] = [ServiceA.Operations, ServiceB.Operations]
   Returns true → sc is NOT added to client.subClients

  What happens to sc after this?

  Nothing. It becomes an orphan — excluded from the hierarchy but never cleaned up from caches. I see three issues:

  -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  Issue 1: Stale __rawClientsCache entry (potential operation loss)

  sc was registered in createSubClient as __rawClientsCache[ServiceB.Operations] = sc. This is never removed.

  During the operation-routing BFS in prepareClientAndOperationCache:

   const clientLocation = getClientLocation(context, op);
   if (clientLocation) {
     if (context.__rawClientsCache.has(clientLocation)) {
       pushClient = context.__rawClientsCache.get(clientLocation)!;  // ← gets orphaned sc!
     }
   }
   context.__clientToOperationsCache.get(pushClient)!.push(op);

  If any operation has @clientLocation(ServiceB.Operations), it would be assigned to the orphaned sc, which is not reachable from the hierarchy. The operation is effectively lost — it won't appear in any
  client's output.

  The same orphaned entry also means needKeep() never visits sc (it only traverses client.subClients), so the stale cache entry is never cleaned up.

  Fix: After the merge, delete the orphaned entries:

   context.__rawClientsCache!.delete(sc.type!);
   context.__clientToOperationsCache!.delete(sc);

  -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  Issue 2: Moved children have wrong parent pointer

  sc's children were created by createSubClient with parent: sc. When moved to existingSc.subClients, their parent still points to the orphaned sc:

   existingSc.subClients = [
     ...existingSc's own children (parent → existingSc ✓),
     ...sc's children (parent → sc ✗ — orphaned!)
   ]

  Currently this doesn't cause immediate bugs because createSdkClientType passes parent via the recursive call parameter rather than reading SdkClient.parent. But SdkClient.parent is read by
  createSdkClientInitializationType (const isRootClient = !client.parent) and could be used by future code traversing the parent chain. It's architecturally wrong.

  Fix: Re-parent the moved children:

   for (const child of sc.subClients) {
     child.parent = existingSc;
   }

  -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  Issue 3: No recursive merge of grandchildren

  If both ServiceA.Operations and ServiceB.Operations each have a child named SubOps, after the merge existingSc.subClients would contain two SubOps instances. There's no recursive merge —
  handleMultipleServicesSubClientNameConflict only operates at one level. This could produce duplicate-named children in the generated SDK.

  This is probably an edge case for now but could surface with deeper multi-service hierarchies.

  -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  Summary

  ┌────────────────────────────────────────────────────────┬─────────────────────────────────────────┐
  │ Aspect                                                 │ Status                                  │
  ├────────────────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ Operations from merged types collected correctly       │ ✅ (via mergedSubClientTypes map)       │
  ├────────────────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ BFS traversal of moved grandchildren                   │ ✅ (they're in existingSc.subClients)   │
  ├────────────────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ Orphaned sc cleaned from __rawClientsCache             │ ❌ leaks                                │
  ├────────────────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ Orphaned sc cleaned from __clientToOperationsCache     │ ❌ leaks                                │
  ├────────────────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ Children parent pointer updated                        │ ❌ points to orphan                     │
  ├────────────────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ Recursive grandchild merge                             │ ❌ not handled                          │
  └────────────────────────────────────────────────────────┴─────────────────────────────────────────┘

  The most impactful bug is Issue 1 — the stale cache entry can cause operation loss when @clientLocation targets a type from the second service. Issues 2 and 3 are correctness concerns that may not manifest
  today but represent latent bugs.
```

PR #4001, opened 2026-03-03
Add missing documentation for linting rules

timotheeguerin:


website/src/content/docs/docs/libraries/azure-resource-manager/rules/arm-resource-operation.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "arm-resource-operation"
+---
+
+```text title="Full name"
+@azure-tools/typespec-azure-resource-manager/arm-resource-operation
+```
+
+Validate ARM Resource operations. All resource operations must be defined inside an interface, include an `api-version` parameter, and use the correct decorator for the HTTP verb.
```
markcowl:
```suggestion
Validate that all ARM Resource operations are defined inside an interface, include an `api-version` parameter, and use the correct decorator for the HTTP verb.
```

website/src/content/docs/docs/libraries/azure-resource-manager/rules/empty-updateable-properties.md
```diff
@@ -0,0 +1,33 @@
+---
+title: "empty-updateable-properties"
+---
+
+```text title="Full name"
+@azure-tools/typespec-azure-resource-manager/empty-updateable-properties
+```
+
+Resources with update operations should have updateable properties. The RP-specific properties of the resource (as defined in the `properties` property) should have at least one updateable property. Properties are updateable if they do not have a `@visibility` decorator, or if they include `update` in the `@visibility` decorator arguments.
```
markcowl:
```suggestion
Resources with update operations should have updateable properties. The RP-specific properties of the resource (as defined in the `properties` property) should have at least one updateable property. Properties are updateable if they do not have a `@visibility` decorator, or if they include `Lifecycle.Update` in the `@visibility` decorator arguments.
```

PR #4033, opened 2026-03-11
Namespace revisit

tadelesh:
When Go and Rust deal with the `namespace` in the emitter, they are asking how to determine a root namespace that could let language know where a package should start from. Current TCGC's namespace concept is just to represent TypeSpec namespace or user specific namespace. We do not have root namespace concept. So, we want to know if all languages have such requirement and does TCGC need a `rootNamespace` config for emitter, and provide `getRootNamespaces` with calculating longest common prefix for all root clients' namespaces?


iscai-msft (approved):
love the doc

PR #4063, opened 2026-03-18
[tcgc] add `.crossLanguageVersioning` hash

iscai-msft:
This hash is computed by collecting the`.crossLanguageDefinitionId`, `optionality`, `type`, `verb`, and enum values for all api-affecting types (operations, params, models, properties, enums). This is following the design of @AlitzelMendez described in her gist [here](https://gist.github.com/AlitzelMendez/febfa059ba81093c16fdff3249e22599)

### DIFF
--- a/packages/typespec-client-generator-core/src/package.ts
+++ b/packages/typespec-client-generator-core/src/package.ts
@@ -76,8 +76,7 @@ export function createSdkPackage<TServiceOperation extends SdkServiceOperation>(
   organizeNamespaces(context, sdkPackage);
 
   // Compute cross-language version hash from source files
-  (sdkPackage as { crossLanguageVersion: string }).crossLanguageVersion =
-    computeCrossLanguageVersion(context);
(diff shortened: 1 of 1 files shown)


packages/typespec-client-generator-core/src/package.ts
```diff
@@ -166,3 +173,112 @@ function populateApiVersionInformation(context: TCGCContext): void {
...
+ * The hash is a SHA256 digest truncated to 12 hex characters.
+ *
+ * Creates a normalized API snapshot capturing:
+ * - Clients, methods, and parameters (with optionality and types)
+ * - Models and properties (with optionality and types)
+ * - Enums and their values
+ * - Unions
+ * - HTTP operation details (verb, path, parameter locations)
+ */
+function computeCrossLanguageVersion<TServiceOperation extends SdkServiceOperation>(
```
timotheeguerin: wouldn't it be simpler to make a has from the source files used. this feels like it probably not covering everything(and are we sure nothing here is affected by client scopes?) or is that on purpose?

PR #4187, opened 2026-04-02
Samples unify playground

timotheeguerin:
Use the samples from packages/samples

image


markcowl (approved):
A few of the samples fail:

- Arm Library (fundamentally these need to be separate projects for this w to work)
- Polymorphic Resource (import/using/namespace not filtered out when files combined)

markcowl (commented):
Arm Library (fundamentally these need to be separate projects for this w to work)
Polymorphic Resource (import/using/namespace not filtered out when files combined)

markcowl (changes requested):
A few of the samples fail:

- Arm Library (fundamentally these need to be separate projects for this w to work)
- Polymorphic Resource (import/using/namespace not filtered out when files combined)

markcowl (changes requested):
Nice!!!!!!!

The following samples should not be in the list:

Legacy Operations
Rename Operations

Also, the current data plane and resource manager samples (Widget Manager,  Tracked Resource) should be *first*

markcowl (approved):

PR #4211, opened 2026-04-08
Node 22 minimum

timotheeguerin:


packages/azure-http-specs/package.json
```diff
@@ -23,7 +23,7 @@
     "test": "echo \"Error: no test specified\""
   },
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=22.0.0"
```
markcowl: Should we separate the minimum engine version from the ci changes?

markcowl (approved):

PR #4231, opened 2026-04-09
Add Python Emitter to Azure Playground

JennyPng:
title

related to https://github.com/microsoft/typespec/pull/10203


timotheeguerin:
I would maybe hold on this one as there is the plan to add the emitters directly into this repo in the workspace. In that case we don't need any special handling

PR #4263, opened 2026-04-14
[python] move `@azure-tools/typespec-python` to typespec-azure repo

iscai-msft:


packages/typespec-python/.prettierrc.yaml
timotheeguerin, line 1: remove this, use the top level one

packages/typespec-python/eslint.config.js
timotheeguerin, line 1: remove this, use the top level one

packages/typespec-python/package.json
```diff
@@ -0,0 +1,109 @@
...
+    "tsx": "catalog:"
+  },
+  "devDependencies": {
+    "@azure-tools/azure-http-specs": "workspace:^",
+    "@azure-tools/typespec-autorest": "workspace:^",
+    "@azure-tools/typespec-azure-core": "workspace:^",
+    "@azure-tools/typespec-azure-resource-manager": "workspace:^",
+    "@azure-tools/typespec-azure-rulesets": "workspace:^",
+    "@azure-tools/typespec-client-generator-core": "workspace:^",
+    "@eslint/js": "catalog:",
```
timotheeguerin: shouldn't be needed

packages/typespec-python/package.json
```diff
@@ -0,0 +1,109 @@
...
+    "@types/fs-extra": "11.0.4",
```
timotheeguerin: use catalog for everything, I also would like to double check all those new deps are needed

packages/typespec-python/package.json
```diff
@@ -0,0 +1,109 @@
...
+    "@types/js-yaml": "~4.0.5",
```
timotheeguerin: like do we need this, we use `yaml` in this repo instead with comes with types

.github/workflows/ci-python.yml
```diff
@@ -0,0 +1,211 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: python-generated
+          path: packages/typespec-python
+
+      - name: Pylint
+        run: npm run test:e2e -- --env lint
+        working-directory: packages/typespec-python
+
+      - name: Lint (extra)
```
timotheeguerin, line 163: i was gonna add that to the global format and lint job but I guess that works too at least for now

mise.toml
timotheeguerin, line 1: I was thinking we'd have that at the top of the repo so we can add java, etc. and locally you can just run once `mise install` to setup everything

packages/typespec-python/package.json
```diff
@@ -0,0 +1,103 @@
...
+  "main": "dist/src/index.js",
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
+    }
+  },
+  "tspMain": "dist/src/index.js",
+  "engines": {
+    "node": ">=20.0.0"
```
timotheeguerin:
```suggestion
    "node": ">=22.0.0"
```

.github/actions/setup-python/action.yml
```diff
@@ -0,0 +1,9 @@
+name: Setup Python
+description: Setup Python for typespec-python package development and CI
+
+runs:
+  using: composite
+
+  steps:
+    - name: Setup mise
+      uses: jdx/mise-action@v2
```
timotheeguerin: use the latest looks like 4 https://github.com/jdx/mise-action/releases and use the sha as its a third party action

.github/workflows/ci-python.yml
```diff
@@ -0,0 +1,205 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: "Build & Regenerate"
+    runs-on: ubuntu-latest
+    env:
```
timotheeguerin: I don't think you need this

.github/workflows/ci-python.yml
```diff
@@ -0,0 +1,205 @@
...
+
+      - uses: ./.github/actions/setup
+
+      - uses: ./.github/actions/setup-python
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build
+        run: pnpm --filter "@azure-tools/typespec-python..." run build
```
timotheeguerin:
```suggestion
        run: pnpm turbo run --filter "@azure-tools/typespec-python..." build
```

same for the others

alternative is set this command under a `build:to` script in http-python(I think do that either way for local convinience) and call that

packages/typespec-python/package.json
```diff
@@ -0,0 +1,103 @@
...
+    "@typespec/rest": "workspace:^",
+    "@typespec/spec-api": "workspace:^",
+    "@typespec/spector": "workspace:^",
+    "@typespec/sse": "workspace:^",
+    "@typespec/streams": "workspace:^",
+    "@typespec/tspd": "workspace:^",
+    "@typespec/versioning": "workspace:^",
+    "@typespec/xml": "workspace:^",
+    "c8": "catalog:",
+    "eslint": "catalog:",
```
timotheeguerin: should be able to remove 
- eslint
- prettier
- typescript-eslint

packages/typespec-python/package.json
```diff
@@ -0,0 +1,103 @@
...
+  "description": "TypeSpec emitter for Python SDKs",
+  "homepage": "https://azure.github.io/typespec-azure",
+  "readme": "https://github.com/Azure/typespec-azure/blob/main/packages/typespec-python/README.md",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Azure/typespec-azure.git"
+  },
+  "bugs": {
+    "url": "https://github.com/Azure/typespec-azure/issues"
```
timotheeguerin, line 14: should add your emitters to labels.ts and area.ts so issues and pr can be labeled correctly

timotheeguerin (approved):

.github/actions/setup-python/action.yml
```diff
@@ -0,0 +1,9 @@
+name: Setup Python
+description: Setup Python for typespec-python package development and CI
+
+runs:
+  using: composite
+
+  steps:
```
timotheeguerin, line 7: actually also this technically will setup everything as other language are added to the mise config.
Which also feels overkill for this action, wondering if there is a way to say use that mise config but only install certain package at the version from the config

iscai-msft: yep, you can use install_args, so just putting in python and uv. there's a little bit of duplicated code bc you specify once in the mise file what your deps are and once again here, but I think that's fine

timotheeguerin: yeah thats good as long as we don't duplicate which version to use

PR #4271, opened 2026-04-15
[tcgc] fix interface flattening under client

iscai-msft:


iscai-msft:
closed as expected behavior @tadelesh

PR #4314, opened 2026-04-23
[test] generated sdk upon https://github.com/microsoft/typespec/pull/10466

weidongxu-microsoft:
generated upon https://github.com/microsoft/typespec/pull/10466/changes

---

**Generated client API surface per language for `SameBodySingleOperation.getAvatar(@path format: string): PngImage | JpegImage`:**

| Language | Method signature | Return type |
|---|---|---|
| **Python** | `get_avatar(self, format: str, **kwargs) -> Iterator[bytes]` | Streaming bytes |
| **C#** | `GetAvatar(string format, RequestContext)` / `GetAvatarAsync(...)` | `Response` (raw) |
| **TypeScript** | `getAvatar(context, format: string, options?) => Promise<SameBodySingleOperationGetAvatarResponse>` | `{ blobBody?, readableStreamBody? }` |
| **Java** | `getAvatar(String format, GetAvatarContentType accept)` / `getAvatarWithResponse(...)` | `BinaryData` / `Response<BinaryData>` |
| **Go** | `GetAvatar(ctx, formatParam string, accept GetAvatarContentType, options*) (Response, error)` | `{ Body io.ReadCloser, ContentType *string }` |

Notable: Java and Go both synthesized a typed `GetAvatarContentType` enum (`image/png` / `image/jpeg`) from the union return type to use as the `Accept` header parameter. Python, C#, and TypeScript treat it as a raw binary response with no typed accept selection.


.tmp/same-body-single-op/tsp-output/@azure-tools/typespec-java/src/main/java/contentnegotiation/models/GetAvatarContentType.java
weidongxu-microsoft, line 1: The model name in Java is generated by TCGC, and is not configurable in client.tsp

And it is a fixed enum, which is generally discouraged.

.tmp/same-body-single-op/tsp-output/@azure-typespec/http-client-csharp/src/Generated/SameBodySingleOperation.RestClient.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            RawRequestUriBuilder uri = new RawRequestUriBuilder();
+            uri.Reset(_endpoint);
+            uri.AppendPath("/content-negotiation/same-body-single-operation/", false);
+            uri.AppendPath(format, true);
+            HttpMessage message = Pipeline.CreateMessage(context, PipelineMessageClassifier200);
+            Request request = message.Request;
+            request.Uri = uri;
+            request.Method = RequestMethod.Get;
+            request.Headers.SetValue("Accept", "image/jpeg, image/png");
```
weidongxu-microsoft, line 30: .NET seems doing special handling on `accept` header

PR #4320, opened 2026-04-23
Setup benchmark testing

timotheeguerin:
image


packages/benchmark/specs/azure-arm-resource-manager/main.tsp
```diff
@@ -0,0 +1,206 @@
+import "@typespec/rest";
```
markcowl, line 1: These are a good start, but I wonder if we should include a larger spec, like network in the benchmark.

markcowl: not a blocking issue, just a comment.  The idea is that high orders of growth in bad performing rules will be more obvious in a larger spec.

packages/benchmark/src/run.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Specific spec names to run (if empty, runs all). */
+  specs?: string[];
+  /** Git commit SHA to record. */
+  commit?: string;
+}
+
+/** Discover benchmark spec directories under the given path. */
+async function discoverSpecs(specsDir: string, filter?: string[]): Promise<string[]> {
+  const entries = await readdir(specsDir, { withFileTypes: true });
+  const dirs = entries.filter((e) => e.isDirectory()).map((e) => e.name);
```
markcowl, line 34: I wonder if eventually we should have a config for these, but simple discovery is good for now

markcowl (approved):
A few comments.  It feels like we need some mechanism to adjust for the performance of individual workers.  Something like a time / complexity measurement,  or just a baseline measurement of the machine performance not related to what we're measuring.

packages/benchmark/specs/azure-arm-resource-manager/main.tsp
timotheeguerin: yeah been wondering exactly that too, I think want to start with this for now too see how it works then we can tweak the spec(s). What I am unsure how other benchmark tool do is how do you reconcyle values across various different compilation where in some cases the values might be large and other not. Average seems like it might just hide some problems

packages/benchmark/src/run.ts
timotheeguerin: yeah for now its just some basic specs. related to other comment need to find a good way to compile various specs with different purpose and make sense of it in the data  then we might want a config to describe that. IT would be extra useful as we move this tool to microsoft/typespec as a generic typespec benchmark engine

PR #4332, opened 2026-04-27
feat(tcgc): enhance @hierarchyBuilding to support arbitrary inheritance replacement

### LINKED ISSUE #3737: [TCGC] improve the `hierarchyBuilding` decorator to support multiple layer inheritance replacement
Now the decorator implementation of this `hierarchyBuilding` decorator is checking properties very strictly.

For instance, in the following case, the decorator will report warning and would not apply.
```
@Azure.ClientGenerator.Core.Legacy.hierarchyBuilding(C)
model A extends B {
a?: string;
}
model B extends C {
b?: string;
}
model C {
c?: string;
}
```
We need to implement this so that in the output of TCGC, we have model A with base type of C, and A should have property `a` (defined on its own) and `b` (from model B) on it, model C has property `c`.

ArcturusZhang:
Resolves #3737.

## Summary

Enhances `@Azure.ClientGenerator.Core.Legacy.hierarchyBuilding` so it can rebase a model to **any** other model — not just to a property-superset sibling. This unblocks scenarios like the ARM Compute migration where a patch model needs to switch its C# base from `UpdateResourceDefinition`/`SubResourceReadOnly` to `Azure.ResourceManager.CommonTypes.Resource`, even though the patch model does not redeclare the inherited `id`/`name`/`type`/`systemData` properties.

## Changes

- **Decorator (`src/decorators.ts`)** — removed the property-superset pre-validation. `$legacyHierarchyBuilding` now always records the requested rebase.
- **Property lifting (`src/types.ts`)** — after the rebase is applied in `handleLegacyHierarchyBuilding`, properties contributed by **removed intermediate ancestors** (parents that the rebase walked past) are lifted onto the target so the SDK model preserves its observable property set. Lifts honour the same filters as `addPropertiesToModelType` (status code / never / void / visibility / scope).
- **Diagnostics (`src/lib.ts`)** — the existing `legacy-hierarchy-building-conflict` code is kept as a **post-apply warning** with a single message id:
  - `property-type-mismatch` — a same-named property exists on the new base chain with a different (and not assignable in either direction) type. The property is dropped to keep the rebase consistent and the warning informs the spec author so the types can be aligned.

  Compatible overlaps (literal vs scalar, sub-scalar vs scalar, identical types) are deduplicated silently — no diagnostic is emitted.
- **Docs** — updated `lib/legacy.tsp` JSDoc with the new behaviour and additional rebase examples. Regenerated `generated-defs/...`, `README.md`, and the website reference page.
- **Tests** — added new cases (issue repro, multi-level lift, rebase to unrelated model, silent compatible-overlap dedupe, `property-type-mismatch`, scope honouring, and a nested `@hierarchyBuilding` chain), and replaced the obsolete `conflicting inheritance` pre-apply diagnostic test with a positive rebase assertion.

## Behaviour preserved

- `scope` parameter still routes to a specific emitter.
- Existing discriminated-multi-level-inheritance cases keep working.
- `enableLegacyHierarchyBuilding` flag still gates the whole feature.

---
🤖 arcturus-copilot


packages/typespec-client-generator-core/lib/legacy.tsp
```diff
@@ -3,13 +3,45 @@ using Reflection;
...
+ *
+ * Common real-world applications:
+ *
+ * - **Multi-level discriminated inheritance**: when discriminated subtypes
+ *   need to inherit from a sibling rather than the discriminator root
+ *   (e.g. `SportsCar` inheriting from `Car` instead of from `Vehicle`).
+ * - **Brownfield base-class alignment**: when a client SDK needs to keep
+ *   API compatibility with a previously-generated SDK that used a
+ *   different base — typically rebasing onto a richer Azure resource base
+ *   such as `TrackedResource` instead of plain `Resource`.
```
tadelesh, line 21: Can you put these comments in the examples?

packages/typespec-client-generator-core/lib/legacy.tsp
```diff
@@ -3,13 +3,45 @@ using Reflection;
...
+ *    chain (the new parent **and** all of its ancestors), since the new
+ *    base will provide them via inheritance.
+ * 3. Whatever remains becomes the rebased model's own properties.
+ *
+ * When step 2 drops a property whose type is not assignable to (and not
+ * assignable from) the same-named property on the new base chain — i.e.
+ * the two types are unrelated, like `int32` vs `string` — TCGC reports a
+ * `legacy-hierarchy-building-conflict` warning so the spec author can
+ * align the types. Compatible refinements (a string literal vs `string`,
+ * a sub-scalar like `azureLocation` vs `string`, etc.) stay silent.
```
tadelesh: Is it possible to make these comments shorter? I don't want to put too many impl detail in the doc, but just describe the results.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -2093,6 +2100,184 @@ function handleLegacyHierarchyBuilding(context: TCGCContext): [void, readonly Di
   return diagnostics.wrap(undefined);
 }
 
+/**
+ * Reconcile properties on a rebased model after `@hierarchyBuilding(target, newBase)`.
```
tadelesh: Can we rename `target` to `oldBase`? The following doc is a little bit mis-leading.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -2093,6 +2100,184 @@ function handleLegacyHierarchyBuilding(context: TCGCContext): [void, readonly Di
...
+ */
+function buildValueChain(context: TCGCContext, value: Model): Model[] {
+  const chain: Model[] = [];
+  const seen = new Set<Model>();
+  let current: Model | undefined = value;
+  while (current && !seen.has(current)) {
+    seen.add(current);
+    chain.push(current);
+    const overridden = getLegacyHierarchyBuilding(context, current);
+    current = overridden ?? current.baseModel;
```
tadelesh: This nested case has problem with this logic. No need to check hierarchy change, just use the original one.

tadelesh: And such case needs a test.

packages/typespec-client-generator-core/src/types.ts
```diff
@@ -2093,6 +2100,184 @@ function handleLegacyHierarchyBuilding(context: TCGCContext): [void, readonly Di
...
+  seen.add(target);
+  let current: Model | undefined = target.baseModel;
+  while (current && !seen.has(current)) {
+    if (valueChain.has(current)) break;
+    seen.add(current);
+    removed.push(current);
+    current = current.baseModel;
+  }
+  return removed;
+}
```
tadelesh: This function could be merged to `buildValueChain` since no need to check type exist in value chain. The following add back property logic should have no impact.

tadelesh (approved):

PR #4336, opened 2026-04-27
[specs] etag with non-standard wire name

iscai-msft:


packages/azure-http-specs/specs/azure/special-headers/conditional-request/main.tsp
```diff
@@ -0,0 +1,51 @@
...
+@scenarioService("/azure/special-headers/conditional-request")
+namespace Azure.SpecialHeaders.ConditionalRequest;
+
+@scenario
+@doc("Check when If-Match and If-None-Match use the standard header names with eTag type.")
+@scenarioDoc("""
+  Check when If-Match and If-None-Match use the standard header names with eTag type.
+  Expected header parameters:
+  - if-match="valid"
+  - if-none-match="invalid"
```
msyyc: Usually it does not make sense to set both `if-match` and `if-none-match` header at same http request. I advise to expect only one in one request. And you may refer to https://github.com/microsoft/typespec/blob/5b4d83729beff779490a28ae0633662516c6bccc/packages/http-specs/specs/special-headers/conditional-request/main.tsp

timotheeguerin (approved):

PR #4382, opened 2026-05-05
Specs/add common package ref test

iscai-msft:


tadelesh (commented):
Not all languages have supported external `alternateType`. Is this feature a must for all lanauges for now?

PR #4390, opened 2026-05-06
Cleanup tests

timotheeguerin:
Had copilot go through our test and try to remove the uncessary code to keep them as simple as possible.

This mostly right now remove `@doc` or doc comments but also include things like using `ResourceName` instead of name with all 4 decorators and removing extra operation/models that are not relevant to many tests


packages/typespec-azure-resource-manager/test/lro-metadata.test.ts
```diff
@@ -35,8 +35,6 @@ it("Returns correct metadata for Async CreateOrUpdate", async () => {
      }
 
       model WidgetProperties {
-        simpleArmId: Azure.Core.armResourceIdentifier;
```
markcowl, line 38: why is it removing these?

packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
```diff
@@ -585,8 +585,6 @@ model Employee is TrackedResource<EmployeeProperties> {
 }
 
 model EmployeeProperties {
-  age?: int32;
```
markcowl, line 588: I'm not sure why it's removing these properties

markcowl (commented):
A few odd removals, otherwise, this looks good

markcowl (approved):

PR #4407, opened 2026-05-11
[python] add nightly e2e builds for language emitters

iscai-msft:


.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,202 @@
+name: Nightly E2E Tests
+
+on:
+  # TODO: Remove push trigger before merging — added for testing only
```
timotheeguerin: comment to make sure todo is not forgotten

.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,202 @@
...
+        with:
+          install_args: ${{ matrix.mise-install-args }}
+
+      # --- Build excluded core packages from source ---
+      # Some language client packages (http-client-python, http-client-java, etc.)
+      # are excluded from the pnpm workspace and normally resolve from npm.
+      # For the nightly, we build them from the core submodule and inject them
+      # via pnpm overrides so we test against the latest dev versions.
+
+      - name: Build excluded core packages from source
```
timotheeguerin: can those script not be inline? I also have in the core typespec repo a util called `tpm` that was designed to install/build, etc. the standalone and core packages with `tpm install` `tpm build` I believe. We use that for the preview url

iscai-msft:
Addressed both comments:

1. **TODO push trigger** — Removed.

2. **Inline scripts / tpm** — Merged the two separate steps (build excluded packages + override injection) into a single inline step. Looked into using `tpm` but it operates on all core packages (no package-selection support) and doesn't handle `--pack-destination` or pnpm override injection, so it would add runtime without removing the workflow-specific logic. If `tpm` gains package selection support in the future, we can switch to it.

.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,179 @@
...
+      - name: Regenerate
+        if: matrix.run-regenerate
+        run: pnpm run regenerate
+        working-directory: ${{ matrix.working-directory }}
+
+      # --- Run E2E tests (standardized script: test:<lang>:e2e) ---
+
+      - name: Run E2E Tests
+        run: pnpm run test:${{ matrix.lang }}:e2e
+        working-directory: ${{ matrix.working-directory }}
```
msyyc, line 179: Overall looks good to me. One more thought: since we will archive autorest.python repo soon, how about we upload spector coverage in this nightly build pipeline like we did in [https://github.com/Azure/autorest.python/blob/26ed19cc4adf5e3142ffe9f0f23a220ceb88586e/eng/pipeline…](https://github.com/Azure/autorest.python/blob/26ed19cc4adf5e3142ffe9f0f23a220ceb88586e/eng/pipelines/ci.yml#L186-L192) so that dashboard could show latest coverage status of python for spector case: [azure.github.io/typespec-azure/can-i-use/http/](https://azure.github.io/typespec-azure/can-i-use/http/)

timotheeguerin (approved):

PR #4508, opened 2026-05-27
[TCGC] Fix orphan type ordering to restore stable naming for anonymous model variants

iscai-msft:
## Problem

PR #4440 rewrote `listOrphanTypes` to iterate via `listScopedDecoratorData(context, usageKey)`, which follows **decorator application order**. The old code walked namespaces and iterated **models first, then enums, then unions**.

This ordering matters because anonymous types (e.g. anonymous model variants inside unions) get their generated name from whichever context processes them first. When `@@usage` on a union appears before `@@usage` on a model that references it, the anonymous variant gets named from the union context (e.g. `RecursiveNullableType1`) instead of the model property context (e.g. `OuterWithNullableValue`).

Reported via https://github.com/Azure/typespec-rust/pull/969 — upgrading from TCGC 0.68.0 to 0.68.2 renamed `OuterWithNullableValue` → `RecursiveNullableType1`.

## Fix

Collect orphan types into three separate buckets (models, enums, unions) and concatenate them at the end: `[...models, ...enums, ...unions]`. This restores the models-first ordering while preserving the #4440 improvement of discovering types from imported libraries.

## Testing

Added a regression test that reproduces the exact scenario from the typespec-rust spec.

### DIFF
--- a/.chronus/changes/fix-orphan-type-naming-order-2026-5-27-14-24-0.md
+++ b/.chronus/changes/fix-orphan-type-naming-order-2026-5-27-14-24-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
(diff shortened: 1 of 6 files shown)


tadelesh (approved):

tadelesh:
@timotheeguerin The E2E Tests stuck, thus the PR could not merge. I tried rerun but not work. Could you help?

PR #4532, opened 2026-06-01
Add Foundations V4 encryption types and no-deprecated-common-types linter rule for CustomerManagedKeyEncryption

markcowl:
## Summary
- Add `Azure.ResourceManager.Foundations` encryption types with V4 suffix (`CustomerManagedKeyEncryptionV4`, `KeyEncryptionKeyIdentityV4`, `KeyEncryptionKeyIdentityTypeV4`) that duplicate `CustomerManagedKeyEncryption` and its dependencies
- Retain v4/v5/v6 `@@armCommonDefinition` refs for `CustomerManagedKeyEncryption` in common-types (required because `Encryption` uses it as a property)
- Add new private decorator `@armCommonDefinitionExcluded(replacementTypeName)` and apply it to `CustomerManagedKeyEncryption` with `"Azure.ResourceManager.CommonTypes.Encryption"` as the replacement type to mark it as internal to the ARM library
- Add new linter rule `no-deprecated-common-types` that emits a warning when a type marked with `@armCommonDefinitionExcluded` is used directly in service specifications outside the `Azure.ResourceManager` namespace, with the diagnostic message suggesting the replacement type
- Add test validating that the diagnostic is thrown when `CustomerManagedKeyEncryption` is used in a user spec
- Add rule documentation


markcowl (commented):
@copilot We should change the implementation plan.  We should *not* remove the versioning decorators from CustomerManagedKeyEncryption, because the type needs to be used by the Encryption property.

We should also rename the types we added in Foundations, by adding a V4 suffix to each type (and updating references inside foundations)

Instead, we should apply a new decorator to CustomerManagedKeyEncryption that will emit a diagnostic if the type is used in typespec outside the azure-resource-manager library.

You will also need to rebuild the documentation for the resource-manager library after all changes,  add a test that validates that the new diagnostic is thrown if used in a user spec.

markcowl:
@copilot we also need to add a changeset

packages/typespec-azure-resource-manager/lib/common-types/customer-managed-keys.tsp
```diff
@@ -48,7 +48,9 @@ model KeyEncryptionKeyIdentity {
 }
 
 /** Customer-managed key encryption properties for the resource. */
+#deprecated "Use Azure.ResourceManager.CommonTypes.Encryption instead"
```
markcowl: remove this deprecation

packages/typespec-azure-resource-manager/lib/common-types/customer-managed-keys.tsp
```diff
@@ -65,6 +67,7 @@ model Encryption {
   infrastructureEncryption?: InfrastructureEncryption;
 
   /** All Customer-managed key encryption properties for the resource. */
+  #suppress "deprecated" "Encryption wraps CustomerManagedKeyEncryption by design"
```
markcowl: This suppression should not be necessary after the other seprecation is removed.

packages/typespec-azure-resource-manager/lib/backcompat.tsp
```diff
@@ -6,6 +6,7 @@ namespace Azure.ResourceManager;
 alias InfrastructureEncryption = CommonTypes.InfrastructureEncryption;
 alias KeyEncryptionIdentity = CommonTypes.KeyEncryptionKeyIdentity;
 alias KeyEncryptionKeyIdentity = CommonTypes.KeyEncryptionKeyIdentity;
+#suppress "deprecated" "Need for back compatibility in common-types"
```
markcowl: Remove this suppression

packages/typespec-azure-resource-manager/src/rules/arm-common-definition-excluded.ts
```diff
@@ -0,0 +1,42 @@
+import { Model, ModelProperty, createRule, getNamespaceFullName, paramMessage } from "@typespec/compiler";
+import { isArmCommonDefinitionExcluded } from "../private.decorators.js";
+
+/**
+ * Rule that prevents direct usage of ARM common types marked with @armCommonDefinitionExcluded
+ * in service specifications outside the Azure.ResourceManager library.
+ */
+export const armCommonDefinitionExcludedRule = createRule({
```
markcowl: rename this rule no-deprecated-common-types, this will mean renaming the method and renaming the rule everywhere.

packages/typespec-azure-resource-manager/src/lib.ts
```diff
@@ -131,6 +131,12 @@ export const $lib = createTypeSpecLibrary({
         default: paramMessage`The specified common-types version '${"version"}'  is not valid for ${"resourceName"} resources. Please use version ${"requiredVersion"} or later of common-types.`,
       },
     },
+    "arm-common-definition-excluded": {
```
markcowl: rename this to match the rule as above.

packages/typespec-azure-resource-manager/src/lib.ts
```diff
@@ -131,6 +131,12 @@ export const $lib = createTypeSpecLibrary({
...
+      severity: "warning",
+      messages: {
+        default: paramMessage`The type '${"typeName"}' is an internal Azure Resource Manager common type and should not be used directly in service specifications. Use the equivalent type from Azure.ResourceManager.Foundations instead.`,
```
markcowl:
```suggestion
        default: paramMessage`The type '${"typeName"}' is an internal Azure Resource Manager common type and should not be used directly in service specifications. Use the type '${"replacementType"}' instead..`,
```

packages/typespec-azure-resource-manager/generated-defs/Azure.ResourceManager.Private.ts
```diff
@@ -334,6 +334,16 @@ export type GenericResourceInternalDecorator = (
...
 ) => DecoratorValidatorCallbacks | void;
 
+/**
+ * Marks an Azure Resource Manager common type as excluded from direct use in
+ * service specifications. Types decorated with this decorator should not be
+ * referenced directly in service specs; use the equivalent Foundations type instead.
+ */
+export type ArmCommonDefinitionExcludedDecorator = (
+  context: DecoratorContext,
+  target: Model,
```
markcowl, line 345: This signature should take a Type as the target and should add a string value parameter for replacementTypeName.  Update the implementation to match - you should be able to get the replacementTypeName from the js api.

packages/typespec-azure-resource-manager/lib/common-types/customer-managed-keys.tsp
```diff
@@ -48,7 +48,9 @@ model KeyEncryptionKeyIdentity {
 }
 
 /** Customer-managed key encryption properties for the resource. */
+#deprecated "Use Azure.ResourceManager.CommonTypes.Encryption instead"
 @added(Versions.v4)
+@Azure.ResourceManager.Private.armCommonDefinitionExcluded
```
markcowl:
```suggestion
@Azure.ResourceManager.Private.armCommonDefinitionExcluded("Azure.ResourceManager.Foundations.Encryption")
```

packages/typespec-azure-resource-manager/src/rules/no-deprecated-common-types.ts
```diff
@@ -0,0 +1,42 @@
+import { Model, ModelProperty, createRule, getNamespaceFullName, paramMessage } from "@typespec/compiler";
+import { isArmCommonDefinitionExcluded } from "../private.decorators.js";
+
+/**
+ * Rule that prevents direct usage of ARM common types marked with @armCommonDefinitionExcluded
+ * in service specifications outside the Azure.ResourceManager library.
```
markcowl, line 6: You should have a test for this linting rule and you will need to include it in the typespec-azure-resource-manager ruleset as well.  See the doc on implementing linting rules for all of the requirements: https://azure.github.io/typespec-azure/docs/howtos/contributing/creating-linter-rules/

markcowl (commented):
@copilot Please poull the latest from main, fix the comments,  rebuild documentation,  make sure all tests pass and sampl,e tests pass, ensure that all chaged files are correclty formatted, and verify that only the changes associated with this pr are in the pr diff

packages/typespec-azure-resource-manager/lib/common-types/customer-managed-keys.tsp
```diff
@@ -49,6 +49,7 @@ model KeyEncryptionKeyIdentity {
 
 /** Customer-managed key encryption properties for the resource. */
 @added(Versions.v4)
+@Azure.ResourceManager.Private.armCommonDefinitionExcluded("Azure.ResourceManager.Foundations.Encryption")
```
markcowl:
```suggestion
@Azure.ResourceManager.Private.armCommonDefinitionExcluded("Azure.ResourceManager.CommonTypes.Encryption")
```

markcowl (commented):
@copilot Please pull the latest, make this change, ensure all changed files are correctly formatted

.chronus/changes/fix-remove-cmk-v5v6-refs-2026-6-2-0-31-57.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add Foundations V4 encryption types (CustomerManagedKeyEncryptionV4, KeyEncryptionKeyIdentityV4, KeyEncryptionKeyIdentityTypeV4) to replace deprecated CustomerManagedKeyEncryption. Add new @armCommonDefinitionExcluded decorator and arm-common-definition-excluded linter rule to warn when CustomerManagedKeyEncryption is used directly in service specifications outside the Azure.ResourceManager namespace. Deprecate CustomerManagedKeyEncryption in favor of the Encryption type.
```
timotheeguerin, line 7:
```suggestion
Add Foundations V4 encryption types (CustomerManagedKeyEncryptionV4, KeyEncryptionKeyIdentityV4, KeyEncryptionKeyIdentityTypeV4) to replace deprecated CustomerManagedKeyEncryption. Add new `@armCommonDefinitionExcluded` decorator and arm-common-definition-excluded linter rule to warn when CustomerManagedKeyEncryption is used directly in service specifications outside the Azure.ResourceManager namespace. Deprecate CustomerManagedKeyEncryption in favor of the Encryption type.
```

packages/typespec-azure-resource-manager/generated-defs/Azure.ResourceManager.Private.ts
```diff
@@ -334,6 +334,18 @@ export type GenericResourceInternalDecorator = (
   target: Model,
 ) => DecoratorValidatorCallbacks | void;
 
+/**
+ * Marks an Azure Resource Manager common type as excluded from direct use in
+ * service specifications. Types decorated with this decorator should not be
+ * referenced directly in service specs; use the equivalent Foundations type instead.
```
timotheeguerin, line 340: can we not use `internal`?

markcowl: I had forgotten we could do that.  I think we should phase this in, with emitting the diagnostic in this version and making the type internal in next version.  This makes a bit better transition if some spec is in development that uses this type. thoughts?

timotheeguerin: Ah yeah didn't pay attention it was added toe existing type. What about deprecation to simplify

markcowl:
This issue was resolved by simply removing the common-types reference

PR #4541, opened 2026-06-03
Add no-url-suffix linter rule

### LINKED ISSUE #4561: Linter: property suffix Url must be renamed to Uri
Parent: #4442

## Convention

**DO NOT** end property names with `Url` — use `Uri` suffix instead.

> Source: [Azure SDK for .NET Mgmt Naming Conventions](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/packages/http-client-csharp-mgmt/generator/Azure.Generator.Management/src/Visitors/NameVisitor.cs) — `TryTransformUrlToUri` method

## Observed today

The .NET SDK generator already auto-renames `Url` to `Uri` via `TryTransformUrlToUri` in `NameVisitor.cs`. For example, in the Workloads SDK:

- `prometheusUrl` is renamed to `PrometheusUri` (verified in [generated SDK](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/workloads/Azure.ResourceManager.Workloads/src/Generated/Models/PrometheusOSProviderInstanceProperties.cs))

The External Integration check against `azure-rest-api-specs` shows **172 specs** with existing violations (123 management plane, 49 data plane).

## Proposed linter

**Detection**

Flag any model property whose C#-resolved name (via `getLibraryName`) ends with `Url` (case-sensitive PascalCase).

**Auto-fix**

Suggest replacing `Url` suffix with `Uri` via `@@clientName(<Model>.<property>, "<name>Uri", "csharp")` in `client.tsp`.

**Scope**

The rule is csharp-scoped — it checks the .NET-resolved name, respecting `@clientName` overrides. The fix writes to `client.tsp` since this is a .NET-specific convention.

## Implementation

Draft PR: #4541

- Package: `@azure-tools/typespec-client-generator-core`
- Rule name: `dotnet-no-url-suffix`
- Registered in both `resource-manager` and `data-plane` rulesets
- 20 tests covering detection, `@clientName` resolution, spread/is/extends, codefix

## Acceptance criteria

- [x] New lint rule registered in TCGC
- [x] Doc page under `website/src/content/docs/docs/libraries/.../rules/`
- [x] Unit tests covering positive / negative cases and the auto-fix
- [x] A `.chronus/changes/` entry
- [ ] External Integration check passes (spec fixes or suppressions needed for 172 specs)

haiyuazhang:
Closes #4561

Add `dotnet-no-url-suffix` linter rule that flags model properties ending with `Url` and suggests using `Uri` suffix instead, to follow .NET SDK naming conventions.


haiyuazhang:
The **External Integration / Azure REST API Specs** check failure is expected — it detects 172 existing specs (123 management plane, 49 data plane) with properties ending in `Url` that violate this new rule. These are existing violations, not regressions. They will be resolved in a separate PR to `Azure/azure-rest-api-specs` by either adding `@@clientName` overrides in `client.tsp` or adding `#suppress` directives where appropriate.

haiyuazhang:
@markcowl @tiotheeguerin @m-nash Could you please review this PR?

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Add `dotnet-no-url-suffix` linter rule that flags model properties ending with `Url` and suggests using `Uri` suffix instead, following .NET SDK naming conventions. Includes auto-fix to add `@@clientName` decorator in client.tsp.
```
timotheeguerin: why `dotnet`? feels like not inline with the rest of naming

packages/typespec-client-generator-core/src/rules/no-url-suffix.ts
```diff
@@ -0,0 +1,144 @@
...
+function createRenameUrlToUriCodeFix(
+  property: ModelProperty,
+  csharpName: string,
+  host: CompilerHost,
+  projectRoot: string,
+) {
+  const newName = csharpName.slice(0, -1) + "i"; // Url → Uri
+  const codeFix: CodeFix = {
+    id: "dotnet-rename-url-to-uri",
+    label: `Add @@clientName to client.tsp → "${newName}"`,
```
timotheeguerin: as I assume every single rule will need the same fix with just different value, lets create some helpers:
- createAugmentDecoratorCodeFix()
- createClientTspAugmentDecoratorCodeFix()
similar to how we have a createAddDecoratorCodeFIx() in compiler

packages/typespec-client-generator-core/src/rules/no-url-suffix.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+      edits.push({
+        kind: "insert-text",
+        pos: existingText.length,
+        text: `@@clientName(${fqn}, "${newName}", "csharp");\n`,
+        file: clientFile,
+      });
+
+      return edits;
+    }) as CodeFix["fix"],
```
timotheeguerin: why do you need to cast here, this means we lost the whole typing for this function

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
```diff
@@ -0,0 +1,290 @@
...
+  return diagnostics;
+}
+
+describe("codefix", () => {
+  it("generates @@clientName in client.tsp with correct path and content", async () => {
+    const diagnostics = await compileAndGetRuleDiagnostics(
+      runner,
+      `model Foo { imageUrl: string; }`,
+    );
+    expect(diagnostics.length).toBeGreaterThanOrEqual(1);
```
timotheeguerin: those tests are super verbose, can you simplify that, please look how the doc describes how to test codefix and make sure it looks like that. If there is some custom thing that needs to be done for client.tsp please write an helper (so it looks similar to the standard tester))

timotheeguerin (changes requested):
We also cannot merge any of this until the client.tsp import has been resolved in the spec repo. Please also have the PR ready to merge in spec repo and spec repo -pr against typespec-next (or main if you prefere)

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
haiyuazhang: Renamed to no-url-suffix.

packages/typespec-client-generator-core/src/rules/no-url-suffix.ts
haiyuazhang: Extracted createAugmentDecoratorCodeFix() and createClientTspAugmentDecoratorCodeFix() into a shared codefix-helpers.ts, following the same API pattern as the compiler's createAddDecoratorCodeFix().

packages/typespec-client-generator-core/src/rules/no-url-suffix.ts
haiyuazhang: Without the cast, we get a build error because CodeFix.fix doesn't allow Promise&lt;CodeFixEdit[]&gt; — only Promise&lt;void&gt;. Our codefix needs async because it reads client.tsp from disk via await host.readFile().

A one-line change in the compiler's types.ts would fix this — extending the fix return type to also allow Promise&lt;CodeFixEdit | CodeFixEdit[]&gt;.

Is this the right approach?

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
haiyuazhang: Simplified codefix tests with a shared applyClientTspCodeFix() helper in test-codefix-helpers.ts. Reusable by any rule that writes to client.tsp.

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
timotheeguerin: i mean more csharp if we want to keep those for c# only but I think this is pending the final descirion on how we want to structure all of this

packages/typespec-client-generator-core/src/rules/no-url-suffix.ts
timotheeguerin: if that is the case and it works in the compiler, we should update the compiler types first. this is making an assumption that will NOT be backward compatible if the compiler decide to change.

But you also most likely  do not need to read the client.tsp async, It should already be loaded in program.sourceFiles. If not this means that the codefix anyway won't fix the issue

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
timotheeguerin: I was hopping we could do something closer `RuleTester.applyCodeFix().toEqual({
  "client.tsp": "Abc"
})`

but maybe we should move that to the compiler

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
```diff
@@ -0,0 +1,180 @@
...
+ * @param args The decorator arguments as literal strings.
+ */
+export function createClientTspAugmentDecoratorCodeFix(
+  target: Model | ModelProperty,
+  decoratorName: string,
+  host: CompilerHost,
+  projectRoot: string,
+  args?: string[],
+): CodeFix {
+  const fqn = buildFqn(target);
```
timotheeguerin: I think it would be nice to have this be a bit smarter(if there is already using) to not repeat the service namespace.

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      model Foo extends Base {}`,
+    )
+    .toEmitDiagnostics([{ code: "@azure-tools/typespec-client-generator-core/no-url-suffix" }]);
+});
+
+it("does not flag non-model-property types", async () => {
+  await tester.expect(`scalar ImageUrl extends string;`).toBeValid();
+});
+
+it("does not flag properties from library types even if they end with Url", async () => {
```
timotheeguerin: i don't think we need this test, this is just standard linter engine behavior

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  await tester.expect(`model Foo { @clientName("imageUri") imageUrl: string; }`).toBeValid();
+});
+
+it("does not flag inherited properties", async () => {
+  await tester
+    .expect(
+      `model Base { imageUrl: string; }
+      model Foo extends Base {}`,
+    )
+    .toEmitDiagnostics([{ code: "@azure-tools/typespec-client-generator-core/no-url-suffix" }]);
```
timotheeguerin: a better test here would be to add suppression on the base property and have .toBeValid()

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
timotheeguerin: I created this PR https://github.com/microsoft/typespec/pull/10897 please try using this version of the submododule and see if that helps here

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
haiyuazhang: removed

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
haiyuazhang: Fixed. The codefix now adds using <namespace> and uses short refs (e.g, Foo.imageUrl).

packages/typespec-client-generator-core/src/rules/no-url-suffix.ts
haiyuazhang: Switched to using program.sourceFiles which is synchronous, so the as CodeFix["fix"] cast is no longer needed.   The one requirement is that client.tsp must be imported (in tspconfig.yaml add imports: - ./client.tsp)

packages/typespec-client-generator-core/test/rules/no-url-suffix.test.ts
haiyuazhang: Done! Updated the core submodule to include this PR and migrated the codefix tests to use the standard multi-file applyCodeFix().toEqual({ "client.tsp": ... }) pattern. Removed the custom test-codefix-helpers

haiyuazhang:
Spec-fix PR is ready: Azure/azure-rest-api-specs#43818. It prepares existing specs for this no-url-suffix rule using isolated C# @@clientName customizations where safe and targeted suppressions where importing existing client.tsp or targeting inline/auth metadata would risk unrelated changes.

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
```diff
@@ -0,0 +1,217 @@
...
+        }
+      }
+
+      const relativePath = computeRelativeImportPath(target, projectRoot);
+      const clientFile = createSourceFile(existingText, clientTspPath);
+      const edits: InsertTextCodeFixEdit[] = [];
+
+      // Build imports/using to insert at the TOP of the file
+      const tcgcImport = `import "@azure-tools/typespec-client-generator-core";`;
+      const modelImport = `import "${relativePath}";`;
```
timotheeguerin: why do we need to import this?

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
```diff
@@ -0,0 +1,217 @@
...
+    id: `add-${decoratorName}-in-client-tsp`,
+    label: `Add \`${decoratorText}\` in client.tsp`,
+    fix: (fixContext) => {
+      if (target.node === undefined) return [];
+      const clientTspPath = resolvePath(projectRoot, "client.tsp");
+
+      // Look up client.tsp in program.sourceFiles (sync).
+      // Assumes client.tsp is imported via tspconfig.yaml imports.
+      let existingText = "";
+      for (const [path, script] of program.sourceFiles) {
```
timotheeguerin: i think you should still be able to get this with a sourceFiles.has instead of a loop, this is quite costly here for every model property to run.

This is also something we could do once in the rule initializer instead

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
```diff
@@ -0,0 +1,217 @@
...
+      const relativePath = computeRelativeImportPath(target, projectRoot);
+      const clientFile = createSourceFile(existingText, clientTspPath);
+      const edits: InsertTextCodeFixEdit[] = [];
+
+      // Build imports/using to insert at the TOP of the file
+      const tcgcImport = `import "@azure-tools/typespec-client-generator-core";`;
+      const modelImport = `import "${relativePath}";`;
+      const usingTcgc = `using Azure.ClientGenerator.Core;`;
+
+      let headerToInsert = "";
```
timotheeguerin: would using an array that join with `\n` at the end not be a bit clener?

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
haiyuazhang: This imports main.tsp so that the augment decorators in client.tsp can resolve the model types defined in the spec.

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
haiyuazhang: a bit confused here - should we rename the rule to csharp-no-url-suffix? If not, what would the final structure look like for language-specific rules?

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
haiyuazhang: updated

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
haiyuazhang: updated

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
timotheeguerin: but this should never happen, this means client.tsp wasn't imported which means adding this will not make the error disapear

haiyuazhang: You're right — tested locally and confirmed that when client.tsp is imported in tspconfig.yaml,  we don't need to import main.tsp in client.tsp. Removed the model import logic. Regarding the work to import client.tsp - should I hold on until that's done, or is there anything I can help with?

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
timotheeguerin: yes that's what I meant but I think we need to discuss with everyone if thats the way we want to do it. Specially around the rules that have complete opposite requirement across languages

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
timotheeguerin: it is a requirement to get those rules added. As soon as they are added the spec repo need to be fixed so if we don't have the client.tsp imported its won't be possible and break typespec upgrade.
It is tracked here https://github.com/Azure/typespec-azure/issues/4564 we should hopefully get 95% of the services migrated soon and a few ones might need extra work

haiyuazhang:
https://github.com/Azure/azure-rest-api-specs/pull/43818 is the PR fixing the existing violations  in current specs.

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
haiyuazhang: Renamed to csharp-no-url-suffix for now

jsquire:
> We also cannot merge any of this until the client.tsp import has been resolved in the spec repo. Please also have the PR ready to merge in spec repo and spec repo -pr against typespec-next (or main if you prefere)

Please help me understand why this is blocking rather than a future enhancement.   This tool will be impactful in helping partners be successful in authoring Azure SDK packages and remove a non-trivial amount of friction and delay in the process.  I'm not well-versed in this space, but my understanding is that the concern here involves the need to emit a line to the TSP config that we would prefer not be present but which also causes no issues.    

Unless there's a real-world impact that I'm overlooking, we should prioritize moving this forward.   If I'm missing something, please help me understand what.  


//cc: @lmazuel

timotheeguerin:
client.tsp not being imported means it is not part of the main compilation(neither during tsp compile . or in the IDE) which means the rule can never be successful.

Not sure what line in the tspconfig you are refering to.

jsquire:
> client.tsp not being imported means it is not part of the main compilation(neither during tsp compile . or in the IDE) which means the rule can never be successful.
> 
> Not sure what line in the tspconfig you are refering to.

It's working today in prototype form.   So, yes, it can be successful.

m-nash:
If the fixer adds an import line to the tspconfig.yaml as well as the @@clientName in client.tsp then it works fine.  Eventually if we ever do this dependency invert we can always remove that part of the fixer but it will be no harm / redundant.

timotheeguerin:
importing that line in the tspconfig.y aml is exactaly the same as importing client.tsp from main.tsp.... 
I do not want to keep doing workaround that we keep regresting and have to cleanup(we are the ones that keep having to do that) when we can just do the fix here correctly at the same cost

timotheeguerin:
and the "workaround" doesn't even work, it works for the demo but as soon as this rule is enabled and merged every service needs to pass, which means every service needs to add the suppression or read the client.tsp.

I have  4 prs migrating all but 6 services for which importing client.tsp cause more issues and the tspconfig import still will cause those issues

jsquire:
> importing that line in the tspconfig.y aml is exactaly the same as importing client.tsp from main.tsp.... I do not want to keep doing workaround that we keep regresting and have to cleanup(we are the ones that keep having to do that) when we can just do the fix here correctly at the same cost

I'm good with whatever path this takes, so long as we've got prioritization and the timelines are equivalent.   This tool solves a very real problem and if the fix that we'd prefer is going to be a significant timeline delay, then I'd like to push on a phased approach to get this in now and do the fix after.

timotheeguerin:
if that simplify this PR cannot merge until this check pass 
image
and to make it pass we need to do that change first then the suppression for the grandfathered names need to be added

timotheeguerin:
so yes this is why i have split https://github.com/Azure/typespec-azure/issues/4564 into multiple phases. THe last 6 services will require more change and we might want to disable the rules for them until we have them migrated but this would allow the many others from benfiting from it from next release

m-nash:
Thanks for the clarification.  I do think if we need to push forward with those 6 specs opted out in the short term lets do that as this will be a pretty big win for our partner teams to get the signal with an auto fixer at the earliest stage possible.

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
timotheeguerin, line 1: most services have been migrated to import client.tsp, there is still a few with more complex issues including 4 that we'll have to decide what to do but we might want to disable the rules for them. 

Can you update the PR in the spec repo to reflect the changes. Use the pkg links in the comment here to set the versions in that pr and tsv-all should be green

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
timotheeguerin, line 1: There is a bug when client.tsp already has import, the using is added in the wrong place

Image

haiyuazhang: fixed

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
```diff
@@ -65,6 +65,58 @@ function buildFqn(target: Model | ModelProperty): string {
...
+  let pos = start;
+  while (pos < text.length) {
+    const lineEnd = getLineEnd(text, pos);
+    if (text.slice(pos, lineEnd).trim() !== "") break;
+    pos = lineEnd;
+  }
+  return pos;
+}
+
+function findUsingInsertPos(text: string, importInsertPos: number): number {
```
timotheeguerin: You might be able to get a more reliable solution here by scanning the ast instead. Look either for the first using statement or the last import statement and insert after?

packages/typespec-azure-rulesets/src/rulesets/resource-manager.ts
```diff
@@ -99,5 +99,6 @@ export default {
     "@azure-tools/typespec-client-generator-core/require-client-suffix": true,
     "@azure-tools/typespec-client-generator-core/property-name-conflict": true,
     "@azure-tools/typespec-client-generator-core/no-unnamed-types": false, // Too bad performance https://github.com/Azure/typespec-azure/issues/2803
+    "@azure-tools/typespec-client-generator-core/csharp-no-url-suffix": true,
```
timotheeguerin: writing comment here too, we might actually not want those rules as part of the same rulesets as many services might not want to release SDK. Wondering if that then should be enabled by default. We can talk about this tomorrow

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
timotheeguerin: do you have a PR for RPSaaSMaster branch as well, we also need to keep that one green

packages/typespec-client-generator-core/src/rules/codefix-helpers.ts
haiyuazhang: updated

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
haiyuazhang: Created the RPSaaSMaster PR here: https://github.com/Azure/azure-rest-api-specs-pr/pull/29078

praveenkuttappan:
Liftr service team has mentioned that CI failures in generated .NET SDK PRs are a key pain point in their release process, highlighting the need for earlier validation.
To address this, Liftr has requested a feature to detect .NET naming convention issues early—prior to SDK generation—so that problems can be identified during the TypeSpec authoring or validation stage. This would enable teams to correct issues in the TypeSpec itself, preventing errors from surfacing later in the generated SDK PR. TypeSpec authoring extension would help scale Liftr service onboarding and streamline releases.
//cc: @m-nash, @jsquire

packages/typespec-azure-rulesets/src/rulesets/resource-manager.ts
timotheeguerin: issue about it https://github.com/Azure/typespec-azure/issues/4697

I think we should have that under a `client-sdk` ruleset(we could split into csharp-client-sdk`, `java-client-sdk` in the future if needed)

Specs repo prs means also should enable those rulesets

.chronus/changes/add-no-url-suffix-rule-2026-6-3-15-20-0.md
timotheeguerin, line 1: I think this looks good just need to move this to a new ruleset as per the other comment and getting someone else to also review the specs repo pr, I am not really sure there should be that many clientName added

timotheeguerin: [PR](https://github.com/Azure/azure-rest-api-specs/pull/44277) adding the tsv validaiton to enforce that new ruleset to be enabled when a service emit clients

timotheeguerin (approved):

PR #4553, opened 2026-06-03
[typespec-ts] Use monorepo lint

timovv:
Adopting the typespec-azure lint conventions


packages/typespec-ts/package.json
timotheeguerin: I think you can remove the eslint plugins here now

timotheeguerin: actually is `eslint-plugin-require-extensions` something you want to keep? if so we should move it up, I don't think it is really needed as i feel like typescript already fail if you do that but maybe im missing something

timovv: Good catch, yeah we should be able to drop it

timotheeguerin (approved):

PR #4565, opened 2026-06-04
feat(typespec-autorest): add @scope TCGC decorator support

timotheeguerin:
## Summary

Add support for the `@scope` TCGC decorator in the autorest emitter. Operations, model properties, and parameters that are scoped out of autorest are now omitted from the generated swagger output.

## Changes

- Import `isInScope` from `@azure-tools/typespec-client-generator-core`
- Filter operations by scope before emission (scoped-out operations omitted entirely)
- Filter model properties by scope in schema emission
- Filter endpoint parameters by scope
- Added 10 test cases covering all filtering scenarios

## How it works

The `@scope` decorator uses short language names (e.g., `"autorest"`, `"python"`, `"csharp"`). When an operation or property is decorated with `@scope("python")`, it will be excluded from the autorest swagger output. When decorated with `@scope("autorest")`, it will only appear in autorest output.

### DIFF
--- a/.chronus/changes/autorest-scope-decorator-support-2026-6-4-15-27-0.md
+++ b/.chronus/changes/autorest-scope-decorator-support-2026-6-4-15-27-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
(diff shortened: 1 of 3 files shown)


markcowl (commented):
Fix looks good, need to validate the failures

timotheeguerin:
Failures were unrelated all green now, I also changed PR to point to release brancht o make this a hotfix

iscai-msft (approved):

PR #4567, opened 2026-06-05
fix(tcgc): move EXACT_NAME_PREFIX to internal-utils and remove from public exports

### LINKED ISSUE #4519: [Bug]: npm run build failure due to export of internal/private member
### Describe the bug

`functions.js` is exporting private member `EXACT_NAME_PREFIX` which is being stripped from `functions.d.ts`, but `index.d.ts` tries to re-export it which causes npm run build error.

### Reproduction

Upgrade TCGC to version 0.68.2 and run npm run build, make sure with `skipLibCheck: false`

### Checklist

iscai-msft:
## Problem

`EXACT_NAME_PREFIX` in `functions.ts` is marked `@internal`, which causes it to be stripped from `functions.d.ts`. However, `index.ts` re-exported it without the `@internal` marker, so `index.d.ts` still referenced a symbol that no longer exists in `functions.d.ts`. This breaks builds with `skipLibCheck: false`.

## Fix

- Move `EXACT_NAME_PREFIX` to `internal-utils.ts` (where other internal helpers live)
- Import it in `functions.ts` from `internal-utils.ts`
- Remove the re-export from `index.ts` entirely
- `hasExactNameMarker()` and `normalizeExactName()` remain public exports

Fixes #4519

### DIFF
--- a/.chronus/changes/fix-tcgc-internal-export-2026-6-5-11-54-0.md
+++ b/.chronus/changes/fix-tcgc-internal-export-2026-6-5-11-54-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Move internal `EXACT_NAME_PREFIX` constant out of public exports to fix build failures when `skipLibCheck` is disabled.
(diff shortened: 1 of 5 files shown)


.chronus/changes/fix-tcgc-internal-export-2026-6-5-11-54-0.md
timotheeguerin, line 1: now that I see the PR, wondering why it needs to be public? what uses it that cause it to be needed in the definition?

iscai-msft: it doesn't need to be public, it's purely internal usage, that's why I moved it to internal-utils.ts, that doesn't get exported outwardly

timotheeguerin (approved):

PR #4569, opened 2026-06-05
Remove duplicate tests

timotheeguerin:
## Summary

Remove tests that are exact duplicates (identical test bodies) found across the test suite.

## Removed tests

| File | Duplicate removed | Reason |
|------|------------------|--------|
| `typespec-autorest/test/models.test.ts` | "errors on duplicate model names" | Already exists in `openapi-output.test.ts` |
| `typespec-azure-core/test/operations.test.ts` | "Gets Lro for non-standard polling in custom put operations without finalLocation" | Identical body to "custom createOrReplace operations without finalLocation" (same decorator used) |
| `typespec-client-generator-core/test/types/model.test.ts` | "no models don't filter core" | Identical to "no models filter core" |
| `typespec-client-generator-core/test/rules/property-name-conflict.test.ts` | "don't emit " (truncated name) | Broken copy — claims to test no-emit but has identical body asserting diagnostic emission |
| `typespec-client-generator-core/test/clients/structure.test.ts` | "one client from multiple services without version dependency" | Identical body to "one client from multiple services" |
| `typespec-client-generator-core/test/examples/load.test.ts` | "example default config" + "load example without version" | Both identical to "example config" |

## Testing

All tests pass:
- `typespec-autorest`: 477 tests ✅
- `typespec-azure-core`: 328 tests ✅
- `typespec-client-generator-core`: 1322 tests ✅


packages/typespec-azure-core/test/operations.test.ts
```diff
@@ -2062,66 +2062,6 @@ JobLroResponse) | Azure.Core.Foundations.ErrorResponse;
     deepStrictEqual(metadata.pollingInfo.terminationStatus.succeededState, ["Succeeded"]);
   });
 
-  it("Gets Lro for non-standard polling in custom put operations without finalLocation", async () => {
```
markcowl, line 2065: Is there an identical test elsewhere?

timotheeguerin: its the same exact body as the previous test `Gets Lro for non-standard polling in custom createOrReplace operations without finalLocation` if you feel like we should remove the other one, im good too or if you think one of those test is actually not testing what it was meant to let me know

markcowl (approved):

PR #4579, opened 2026-06-08
fix spec coverage upload for azure sceanrios

iscai-msft:


eng/pipelines/jobs/build-for-publish.yml
```diff
@@ -36,6 +36,11 @@ jobs:
         displayName: "Run Python e2e tests"
         workingDirectory: packages/typespec-python
 
+      - script: |
+          curl -s -X POST http://localhost:3000/.admin/stop || true
```
timotheeguerin, line 40: I thought we had a cli command that did that

timotheeguerin (changes requested):

eng/pipelines/jobs/build-for-publish.yml
msyyc: After deep thinking, I think we could optimize conftest.py instead of adding additional step in pipeline.
@iscai-msft pls review https://github.com/Azure/typespec-azure/pull/4584

PR #4580, opened 2026-06-08
RFC: TypeSpec Breaking Change Detection Tool - Design Overview

markcowl:
## Summary

Design overview RFC for \@azure-tools/typespec-breaking-change\, a TypeSpec-native API diff and breaking change detection tool for Azure ARM APIs.

This document covers:
- Architecture (two-layer: diff engine + policy engine)
- Comparison phases (Phase A: same-version regression, Phase B: cross-version breaking changes)
- Rule summary tables (request/response/model breaking changes)
- Suppression mechanism (two decorators: \@approvedBreakingChange\, \@approvedUnversionedChange\)
- CI integration and output format

## Supporting Documents

Additional design detail documents (taxonomy, classification, rules, OAD correlation, validation strategy) are available on the [\
fc/breaking-changes\ branch](https://github.com/markcowl/typespec-azure/tree/rfc/breaking-changes/rfcs/breaking-changes) for reference.


rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+- Integrate with GitHub CI so findings can surface as diagnostics, machine-readable output, and PR feedback.
+- Replace OpenAPI-based breaking change tools for TypeSpec-authored specifications.
+- Coexist with OpenAPI-based breaking change tools for hand-authored OpenAPI specifications that are outside this tool's scope.
+
+### Non-Goals
+
+The tool is explicitly not trying to solve every compatibility problem around a service. The following are out of scope:
+
+- SDK-facing concerns such as generated client names or client-shaping behavior.
+- `operationId`, because it is not part of the HTTP wire contract and is primarily an SDK/client concern.
```
mikekistler: Schema names probably also fall into this category.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+base@2024-05-01-preview  vs  head@2024-05-01-preview
+```
+
+If the same version exists on both sides, it is compared directly to itself across branches. This is the primary way the tool detects regressions in published or previously-declared versions.
+
+### Phase B: cross-version evolution
+
+After Phase A, the tool evaluates versions that are new on the head branch.
+
+For every new version on head, the tool compares that new version against the **previous stable** version.
```
mikekistler: I think this should be done for any new _or changed_ version on head.  Yes, the changes will be flagged in Phase A, but we need to know if any are actually breaking vs the previous stable to know if we should permit them to be suppressed.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+| `RequestEncodingChanged`                                                                     | Serialization/encoding change          | Error    |
+| `RequestConstraintStrengthened`                                                              | Tighter validation constraint          | Error    |
+| `RequestParameterMadeRequired`                                                               | Optional→required parameter            | Error    |
+| `RequestPropertyMadeRequired`                                                                | Optional→required body property        | Error    |
+| `RequestParameterLocationChanged`                                                            | Moving between header/query/path/body  | Error    |
+| `RequestParameterDefaultChanged`                                                             | Changing a parameter default value     | Error    |
+| `RequestPropertyDefaultChanged`                                                              | Changing a property default value      | Error    |
+| `RequestContentTypeRemoved`                                                                  | Removing a request content type        | Error    |
```
mikekistler, line 335: Curious about changes to readOnly -- are these flagged anywhere?

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+- removed property
+- removed operation
+- removed enum member
+- removed response header
+- removed response status code
+- inline request/response metadata changes with no dedicated declaration node
+
+#### Removed property
+
+When a property is removed, the approval is placed on the model that still survives:
```
mikekistler, line 481: Seems like this could still go on the property, since it is retained but just has a @removed attribute on it.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+- **Format changes** are always breaking (Error) regardless of direction.
+
+### Service-Level Rules
+
+| DiffKind                     | Rule                                    | Severity |
+| ---------------------------- | --------------------------------------- | -------- |
+| `ApiVersionRemoved`          | Removing a stable api-version           | Error    |
+| `AuthSchemeRemoved`          | Removing an authentication scheme       | Error    |
+| `AuthSchemeAdded` (required) | Adding a required authentication scheme | Error    |
+| `OAuthScopeAdded`            | Narrowing OAuth scopes                  | Error    |
```
mikekistler, line 304: What about changes to the version of ARM common types? These get flagged quite often by openapi-diff.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
mikekistler: I see this is answered below.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+
+- decorator kind
+- `DiffKind`
+- `path`
+- `since`
+- `reason`
+
+The CI semantics are intentionally conservative:
+
+- **new approvals** require reviewer sign-off
```
mikekistler: Is this "sign-off" an approval on the PR? If so, I'm concerned there may be confusion about what approvals mean. An approval by a breaking change reviewer only means "the new/changed suppressions are ok", but is not approving the API as a whole.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+- the reporting layer explains the finding
+- the suggestion shows how to approve it
+- the approval-diff logic determines whether that approval is new or existing
+- the PR label enforces final reviewer sign-off when needed
+
+### 6.5 Stale Approvals
+
+Approvals can outlive the changes they were meant to justify.
+A property may be restored, a constraint may be relaxed again, or a declaration may be refactored so the old approval no longer matches any finding.
+When that happens, the approval becomes stale.
```
mikekistler, line 732: Seems like @approvedVersionedChange will always be stale immediately after merge, right?

mikekistler (commented):
This looks awesome!

I left a few comments on things to think about / discuss, but overall this seems well thought out.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+
+```text
+Base TypeSpec ── compile ── version mutators ── HttpCanonicalizer ──┐
+                                                                    ├── graph walk ── rules ── findings
+Head TypeSpec ── compile ── version mutators ── HttpCanonicalizer ──┘
+                                                                                      │
+                                                                                      └── suppressions / CI output
+```
+
+For the complete taxonomy of detectable diffs, see `diff-taxonomy.md`. For classification policies, see `breaking-change-classification.md`. For detailed rule definitions and type transition tables, see `typespec-breaking-change-rules.md`.
```
haolingdong-msft, line 84: Seems I did not see the three markdowns: `diff-taxonomy.md`, `breaking-change-classification.md` and  `typespec-breaking-change-rules.md`, do I miss anything?

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+
+---
+
+## 5. Breaking Change Rules (Summary)
+
+The `@azure-tools/typespec-breaking-change` tool evaluates wire compatibility by comparing a version to the **previous stable version**.
+For versions that already existed before the change, the same rule semantics still apply when the tool detects a structural regression.
+
+Rules are evaluated against canonical HTTP metadata.
+That means the tool cares about the observable contract on the wire: operation identity, parameters, payload shapes, status codes, headers, content types, authentication, and encoded value sets.
```
haolingdong-msft, line 274: Just to confirm my understanding: change a fixed enum(`enum`) to extensible enum(`union`) will not be detected as a breaking change, though type changed from `enum` to `union` in TypeSpec, right?

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+For detailed rule descriptions with examples, see `typespec-breaking-change-rules.md`.
+
+### Phase B Classification Principle
+
+In Phase B, the classification follows a simple directional principle:
+
+- **Request narrowing** is always breaking (Error) — the server accepts fewer inputs.
+- **Response widening** is always breaking (Error) — the server returns more outputs.
+- **Request widening** is not breaking (Ignore) — the server accepts more inputs.
+- **Response narrowing** is not breaking (Ignore) — the server returns fewer outputs.
```
haolingdong-msft, line 285: Does the response narrowing mean remove a few properties from the response model? May I know why do we ignore those? (treated as non-breaking)

haolingdong-msft:
Thanks @markcowl for the PR! The design looks awesome. I really love the idea of creating canonicalization on TypeSpec versioning.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+
+### Version enum integrity
+
+Before comparison begins, the tool validates the version model itself.
+
+The following integrity rules are part of the design:
+
+- Stable versions cannot be removed.
+- Versions must be monotonically increasing.
+- Preview replacement is allowed.
```
mikeharder, line 193: let's add "preview and stable must use distinct dates".  specs shouldn't have both "2026-01-01" and "2026-01-01-preview" with the same date.

this might be covered by "must be monotonically increasing".

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
```diff
@@ -0,0 +1,842 @@
...
+
+## 6. Suppression Mechanism
+
+Every approved breaking change lives inline in TypeSpec source via the `@approvedBreakingChange` decorator (for Phase B breaks) or `@approvedUnversionedChange` decorator (for Phase A same-version changes).
+Approvals are version-controlled, reviewable, and co-located with the declarations they affect.
+That keeps the approval record in the same PR as the API change, instead of splitting intent across external manifests or CI-only state.
+
+The design uses two suppression decorators:
+
+- `@approvedBreakingChange` suppresses Phase B breaking changes. New or modified instances add the `BreakingChangeReviewRequired` label.
```
timotheeguerin, line 405: I wonder if that is actually the best, could we maybe not instead rely on the suppression mechanism replacing lint diff? Require a  suppression for breaking changes but we highlight those newly added/scan for those codes

markcowl: I started with #suppress, and I am kind of torn on this.  I would really like all the tools to be similar.

The reason I decided to go with a decorator was that:
- We need to be able to include the path in some suppressions, because the suppressed type is removed and we need to be explicit about what we suppress, also, we have a similar problem for encoding 'since' when breakign changes are repeated.  #suppress doesn't provide any structured way to allow us to do this, but using a decorator we have some constraints over provided data and can implement validation.
- The standard suppression logic doesn't work if we somehow encode the 'since' value and path in one of the suppression arguments, so we would have to either create a new suppression directive, allow a suppression to suppress *any breaking change* under a particular scope, or have custom suppression  logic.  Using getters over metadata seemed cleaner to implement

markcowl: But if I could implement anything,  I would have a structured suppression type that allowed encoding "since" when needed and "path" when needed, and new suppress functionality to allow more easily detecting matching suppressions.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: Montonically increasing is meant to disallow versions with the same number regardless of the `-preview` suffix.  I don't think this should be enforced by this tool, though, I think the newly-added linting tule for this is the right way to do this (we currently *have* to deal with specs that have a preview and stable with the same date)

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: No, narrowing is about allowed values for scalar types.  For example string-> url is narrowing (because both are strings, but fewer values are allowed by url).

Array types can be narrowed by narrowing the item type.

Record types can be narrowed by narrowing the value type

Keyed types are not really thought of as narrowing or widening,  the rules are recursively applied to each keyed type.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: Changing from fixed to extensible would be a breaking change (widening) in responses, but non-breaking in a request.

Similarly extensible->fixed would be a breaking change in a request (narrowing) but non-breaking in a response.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: If you want to see thesew, they are here: https://github.com/markcowl/typespec-azure/tree/rfc/breaking-changes/rfcs/breaking-changes

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: Good catch!  I am not sure if this means these decorators should be removed as part of check-in (or in the next PR), or if they shouldn't exist.  The reason to keep them would be to register a reason for the change for the reviewer.

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: Will clarify

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: As discussed, these will be handled by comparing the underlying schemas (which are available in TypeSpec for all arm common-types)

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: Should be flagged by having extra / missing properties in requests

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: So, the plan is to develop this using the decorator, once we are sure the tool is working and the data format is sufficient, we can look at how to move from a decorator to a directive.

timotheeguerin: I see, though I am not sure I get the issue with the removed cases, it would still have `@removed` and a node that could be annotated no?

I do see the issue with repeated approval though, my worry is that basically we just for every breaking decorator and this kinda just feels like we duplicating the verbosity but not sure how that helps that much the review

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
haolingdong-msft: Thanks for the explain! Got it, make sense!

rfcs/breaking-changes/typespec-breaking-change-design-overview.md
markcowl: Not necessarily.  A node could be deleted without versioning (which would show up in phase A)

markcowl: Also, there are still a few places where suppressions cannot be placed, for example, anonymous union variants, so that, if we are going to be exact, we need some extra data.

PR #4583, opened 2026-06-09
fix: write spector coverage by gracefully stopping server in conftest

msyyc:
## Summary

Stop the tsp-spector mock server gracefully from the Python e2e test run so that the spector coverage file (`spec-coverage.json`) is written **without** needing an extra step in the publish pipeline.

## Background

`tsp-spector` only persists `spec-coverage.json` from its process `exit` handler, which is triggered by `POST /.admin/stop` (the admin route then calls `process.exit(0)`). Today the server is hard-killed when the parent process exits, which skips that handler — so no coverage file is produced. That is why #4579 adds a manual `curl -X POST .../.admin/stop` step in `build-for-publish.yml`.

## Change

Add a `pytest_unconfigure` hook in `packages/typespec-python/tests/conftest.py` that gracefully stops the server via the admin endpoint:

- Runs only in the xdist **controller** (or the non-xdist main process), i.e. after all workers finish, so the shared server is not killed prematurely while other workers still need it. This preserves the existing `testserver` fixture behavior that intentionally avoids stopping the server in per-worker teardown.
- Waits until the server is no longer reachable, guaranteeing the synchronous coverage write has been flushed to disk before the upload step runs.
- Is a safe no-op if the server is already stopped / was never started (mirrors the unconditional `curl ... || true` behavior it replaces).

With this in place, the extra pipeline step from #4579 is no longer needed.

## Testing

Will be validated in the publish pipeline.

### DIFF
--- a/.chronus/changes/python-spector-coverage-graceful-stop-2026-6-9-11-30-0.md
+++ b/.chronus/changes/python-spector-coverage-graceful-stop-2026-6-9-11-30-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@azure-tools/typespec-python"
+---
+
+Gracefully stop the spector mock server at the end of the test session so it writes its coverage file, removing the need for an extra pipeline step.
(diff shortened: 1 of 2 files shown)


msyyc:
Superseded by #4584, which uses a branch on the origin repo so the publish pipeline can run against it.

PR #4585, opened 2026-06-09
[typespec-ts] fix platform import issue for customization

### LINKED ISSUE #4581: [Bug]: typespec-ts: multipartHelpers.ts and get-binary-stream-response.ts compilation error in non-standard generation output path
### Describe the bug

my generation output is `src/generated` because of existing storage-blob package.  After code generation with typespec-ts 0.54.0 the two files comtain compilation errors. Likely due to them being static files and their hard-coded `import { NodeReadableStream } from "#platform/static-helpers/platform-types";`

Image.

In my package.json, `"#platform"` is mapped to `src/`

The correct path is included in the generated `index.ts`:

Image

1) I noticed that src\generated\static-helpers\serialization\get-binary-stream-response-browser.mts doesn't use `#platform` import: `import { NodeReadableStream } from "../platform-types-browser.mjs";`  For static files I guess it probably should work if we just import directory from the file, i.e., src\generated\static-helpers\serialization\get-binary-stream-response.ts version would have `import { NodeReadableStream } from "../platform-types.js";`. 

2) src\generated\static-helpers\multipartHelpers.ts needs proper `#platform` imports though.  It probably will need to be dynamic as the path in `"#platform/static-helpers/platform-types";` could vary

### Reproduction

run `tsp-client update` under `sdk/storage/storage-blob` using the https://github.com/Azure/azure-sdk-for-js/tree/storage/blob-typespec-based branch

### Checklist

kazrael2119:
fixes https://github.com/Azure/typespec-azure/issues/4581


jeremymeng:
@kazrael2119 maybe as part of this PR, we could get rid of the `platform-types` source files, and import from `@azure/core-rest-pipeline`? The type is available now in latest v1.24.0 https://github.com/Azure/azure-sdk-for-js/blob/%40azure/core-rest-pipeline_1.24.0/sdk/core/core-rest-pipeline/review/core-rest-pipeline-node.api.md?plain=1#L239

JialinHuang803:
Maybe we can remove `#platform` in this PR after https://github.com/Azure/azure-sdk-for-js/pull/38909.

packages/typespec-ts/src/modular/static-helpers-metadata.ts
```diff
@@ -59,11 +61,7 @@ export const SerializationHelpers = {
     name: "getBinaryResponse",
     location: "serialization/get-binary-response.ts",
   },
-  getBinaryStreamResponse: {
-    kind: "function",
-    name: "getBinaryStreamResponse",
-    location: "serialization/get-binary-stream-response.ts",
-  },
+  getBinaryStreamResponse: AzureCoreDependencies["getBinaryStreamResponse"],
```
JialinHuang803: Why do we still need to add `getBinaryStreamResponse` and `NodeReadableStream` to the static helpers instead of importing from the core packages directly?

packages/typespec-ts/src/modular/static-helpers-metadata.ts
```diff
@@ -151,11 +149,7 @@ export const MultipartHelpers = {
 } as const;
 
 export const PlatformTypeHelpers = {
```
JialinHuang803: Can we remove the PlatformTypeHelpers since we no longer use the platform things.

packages/typespec-ts/src/rlc-common/metadata/build-package-file.ts
```diff
@@ -146,6 +143,12 @@ export function updatePackageFile(
     };
   }
 
+  packageInfo.dependencies = {
```
JialinHuang803, line 120: I am not quite certain about this. Did we do batch update on the existing sdks to bump these dependency versions? If so, do we still need to update here?

packages/typespec-ts/src/metadata/build-package-file.ts
JialinHuang803: Do we need to remove the code here?

packages/typespec-ts/test-next/unit/metadata/package-json.test.ts
```diff
@@ -138,20 +138,9 @@ describe("Package file generation", () => {
       expect(packageFile).to.have.property("module", "./dist/esm/index.js");
       expect(packageFile).to.have.property("types", "./dist/commonjs/index.d.ts");
       expect(packageFile).to.have.property("browser", "./dist/browser/index.js");
-      // Default: no react-native entrypoint
```
JialinHuang803, line 120: Why do we need to remove this assertion about reacti-native?

packages/typespec-ts/test-next/unit/metadata/package-json.test.ts
```diff
@@ -597,34 +578,24 @@ describe("Package file generation", () => {
...
-      expect(packageFile).to.have.property("imports");
-      expect(packageFile.imports).toEqual({
-        "#platform/*": {
-          browser: "./src/*-browser.mts",
-          default: "./src/*.ts",
-        },
-      });
+      expect(packageFile.dependencies).to.have.property("@azure-rest/core-client", "^2.7.0");
+      expect(packageFile.dependencies).to.have.property("@azure/core-rest-pipeline", "^1.24.0");
+      expect(packageFile.dependencies).to.have.property("tslib", "^2.8.1");
```
JialinHuang803, line 474: I remember that tslib is already changed to be imported from catalog. Should we change this as well?

packages/typespec-ts/test-next/unit/metadata/package-json.test.ts
```diff
@@ -342,22 +324,21 @@ describe("Package file generation", () => {
       const packageFile = JSON.parse(packageFileContent?.content ?? "{}");
 
       expect(packageFile).to.have.property("browser", "./dist/browser/index.js");
-      // Default: no react-native entrypoint
-      expect(packageFile).not.toHaveProperty("react-native");
+      expect(packageFile.exports["."]).to.have.property("browser");
     });
 
-    it("should include react-native entrypoint when generateReactNativeTarget is true", () => {
```
JialinHuang803, line 310: I found that many tests about `generateReactNativeTarget` are changed, but seems we are still using it. Could you help confirm?

packages/typespec-ts/test-next/unit/metadata/package-json.test.ts
kazrael2119: this is a test case to test updatePackageFile with customized packageinfo, see 
https://github.com/kazrael2119/typespec-azure/blob/fix-platfor-import/packages/typespec-ts/test-next/unit/metadata/package-json.test.ts#L566-L581

jeremymeng (approved):
LGTM. Thanks for the fix!

JialinHuang803 (approved):

PR #4613, opened 2026-06-11
Batch ts-morph mutations across modular emit phases

JialinHuang803:
## Summary

Generalise the ts-morph batching pattern shipped in autorest.typescript [#4004](https://github.com/Azure/autorest.typescript/pull/4004) (which targeted `resolveReferences`) into a reusable `framework/source-file-batch.ts` module, and apply it to the modular emit phases that still walked the `add*` hotpath one statement at a time.

In ts-morph 23, every individual `sourceFile.addX(...)` call re-parses the entire source file via `insertChildText` -> `doManipulation`. Loops that call `addExportDeclaration` / `addInterface` / etc. against the same file therefore grow as O(N x file_size). Collecting the statements and issuing a single `addStatements([...])` per file collapses N re-parses into 1.

Ported from [Azure/autorest.typescript#4015](https://github.com/Azure/autorest.typescript/pull/4015).

## Design

A single module-level reference counter (`batchDepth`) tracks how many `beginSourceFileBatch` scopes are open, with one shared `Map<SourceFile, StatementStructures[]>` collecting pending writes. Each `enqueueStatement` appends to the queue while `batchDepth > 0` and writes through to ts-morph immediately otherwise -- so call sites like `addDeclaration` can use it unconditionally.

Only the outermost `flushSourceFileBatch` drains the queue; inner flushes just decrement the counter. That keeps nested begin/flush pairs composable: an inner caller can open its own batch without prematurely flushing what an outer caller queued. On drain, all queued structures for a file are written in insertion order with one `sourceFile.addStatements([...])` call, so declaration order matches the order callers emitted in and the cost is one re-parse per file.

**Why the binder is not migrated.** The binder already collects imports per file in `this.imports` and issues a single `file.addImportDeclarations(arr)` per file at the end of `resolveReferences`. That is the same end-state the new framework produces (one re-parse per file), so routing it through `enqueueStatement` would add per-call overhead without changing the flush cost, and would push binder-specific ordering policy (sort by module specifier) into a generic module. The framework targets the other pattern: many independent callers each writing one statement to a shared file, or a single caller that interleaves writes with AST reads.

## What changed

- **New** `packages/typespec-ts/src/framework/source-file-batch.ts`
  - `beginSourceFileBatch` / `flushSourceFileBatch` -- reference-counted scope; nested begin/flush pairs compose. The flush is wrapped in `try/finally` so a thrown writer never leaks queue state into the next emit.
  - `enqueueStatement(file, structure)` -- queues while a batch is open, writes immediately otherwise (single entry point for both modes).
  - `getEffectiveExportedNames` / `getQueuedExportNames` -- batch-aware views of the export set, so dedup against still-pending writes stays correct.
  - On flush, all queued structures for a file are written in insertion order via a single `sourceFile.addStatements([...])`.

- `framework/declaration.ts` -- drops its own ad-hoc batch and routes `addDeclaration` through `enqueueStatement`. Single code path for batched/unbatched.

- `modular/emit-models.ts` -- wraps the `emitTypes` main loop with `beginSourceFileBatch` / `flushSourceFileBatch` (try/finally). The generated models file banner is written as a plain `/* ... */` block instead of `/** ... */`; it is an eslint-disable file banner, not a doc comment, and a plain block keeps `addStatements` on its fast path.

- `modular/build-subpath-index.ts` -- wraps `buildSubpathIndexFile`; `partitionAndEmitExports` enqueues both the type-only and value exports.

- `modular/build-root-index.ts` -- wraps `buildRootIndex` and `buildSubClientIndexFile`; all four `addExportDeclaration` call sites use `enqueueStatement`. `getExistingExports` swapped to `getEffectiveExportedNames` (one-line body change); `getExportedDeclarations().keys()` dedup sets union with `getQueuedExportNames`.

Because structures are written in insertion order, the emitted output matches the per-statement ordering the callers previously produced.

## Performance results

Measured (on the autorest.typescript repo) by `tsp compile` of the Network and Compute ARM specs, against the post-#4004 baseline.

### Network (`client.tsp`, total onEmit `7:27.367` -> `2:20.474`, **-69%**)

| phase | before | after | delta |
| --- | --- | --- | --- |
| emit models - emitTypes | 4:32.239 | 1.588s | -99.4% |
| build api subpath index | 58.813s | 50.054s | -15% |
| build classic subpath index | 17.015s | 0.452s | -97% |
| buildRootIndex | 27.414s | 0.372s | -98.6% |

### Compute (`client.tsp`, total onEmit `1:47.209` -> `0:56.485`, **-47%**)

| phase | before | after | delta |
| --- | --- | --- | --- |
| emit models - emitTypes | (slow) | 0.66s | huge |
| build api subpath index | 17.482s | 13.500s | -23% |
| build classic subpath index | 3.856s | 0.291s | -92% |
| buildRootIndex | 6.204s | 0.168s | -97% |


packages/typespec-ts/src/framework/source-file-batch.ts
```diff
@@ -0,0 +1,237 @@
...
+        exports.push(s);
+        break;
+      case StructureKind.ImportDeclaration:
+        imports.push(s);
+        break;
+      default:
+        throw new Error(`Unsupported pending statement kind ${(s as any).kind}`);
+    }
+  }
+  if (imports.length > 0) {
```
JialinHuang803: All modifications in the test files are just changes in the order, as statements of the same type are generated in batches now in the sequence of interface, functions, enums...

maorleger:
> orks around a ts-morph 23 RangeParentHandler bug in mixed-kind addStatements

it would be great if we can use sourceFile.addStatements(allStructures) rather than splitting per kind, just to avoid the snapshot churn due to reordering...

I couldn't reproduce the RangeParentHandler bug on 23.0.0 across a handful of realistic batches (happy to share the repro), and if it holds up it'd let you drop the per-kind grouping and keep statements in insertion order. 

longer-term, for the fresh model files specifically, the documented `createSourceFile(path, { statements })` [structures-first pattern](https://ts-morph.com/manipulation/performance#performance-tip-work-with-structures-instead) is the one `ts-morph` actively recommends for codegen and would beat even the bulk adds, though I'd keep that as a separate follow-up since it's a real refactor. Either way the deferred-queue plumbing itself doesn't have a built-in twin, so that part's fair to keep. 

I try to bias towards using the built-in ts-morph tooling where we can but totally understand if its not doable in this case 👍

JialinHuang803:
> > orks around a ts-morph 23 RangeParentHandler bug in mixed-kind addStatements
> 
> it would be great if we can use sourceFile.addStatements(allStructures) rather than splitting per kind, just to avoid the snapshot churn due to reordering...
> 
> I couldn't reproduce the RangeParentHandler bug on 23.0.0 across a handful of realistic batches (happy to share the repro), and if it holds up it'd let you drop the per-kind grouping and keep statements in insertion order.
> 
> longer-term, for the fresh model files specifically, the documented `createSourceFile(path, { statements })` [structures-first pattern](https://ts-morph.com/manipulation/performance#performance-tip-work-with-structures-instead) is the one `ts-morph` actively recommends for codegen and would beat even the bulk adds, though I'd keep that as a separate follow-up since it's a real refactor. Either way the deferred-queue plumbing itself doesn't have a built-in twin, so that part's fair to keep.
> 
> I try to bias towards using the built-in ts-morph tooling where we can but totally understand if its not doable in this case 👍

@maorleger I did a deeper investigation on it and found that `addStatements(...)` throws the error because model.ts has a single leading JSDoc-style `/** … */` comment at the top with no actual statements after it. Therefore, it's **not** caused by mixing declaration kinds in one `addStatements` call. The actual trigger is the `/**` banner.

The reason why splitting into `addInterfaces / addFunctions` can fix it is that they don't touch the RangeParentHandler.

packages/typespec-ts/src/modular/emit-models.ts
```diff
@@ -117,32 +118,37 @@ export function emitTypes(context: SdkContext, { sourceRoot }: { sourceRoot: str
...
-    sourceFile = outputProject.getSourceFile(filepath);
-    if (!sourceFile) {
-      sourceFile = outputProject.createSourceFile(filepath);
-      sourceFile.addStatements(`/**
+      const namespaces = getModelNamespaces(context, type);
+      const filepath = getModelsPath(sourceRoot, namespaces);
+      sourceFile = outputProject.getSourceFile(filepath);
+      if (!sourceFile) {
+        sourceFile = outputProject.createSourceFile(filepath);
+        sourceFile.addStatements(`/*
```
JialinHuang803, line 133: The `/**` is changed to `/*` because the single leading JSDoc-style comment makes the `addStatements` call fail.

packages/typespec-ts/test/modular-unit/scenarios/api-operations/azureCoreOperations.md
```diff
@@ -122,14 +123,14 @@ export async function getWidgetOperationStatus(
...
-import { ErrorModel } from "@azure-rest/core-client";
-
-/**
+/*
  * This file contains only generated model types and their (de)serializers.
  * Disable the following rules for internal models with '_' prefix and deserializers which require 'any' for raw JSON input.
  */
 /* eslint-disable @typescript-eslint/naming-convention */
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import { ErrorModel } from "@azure-rest/core-client";
```
JialinHuang803, line 130: The single character change ( /**  →  /* ) reclassifies the banner from "JSDoc owned by the first statement" to "detached file header," and that shifts the insertion point of imports by ts-morph.

qiaozha (approved):

PR #4617, opened 2026-06-11
feat(typespec-ts): Use CompilerHost for filesystem access (playground compatibility)

timotheeguerin:
## Summary

This PR upgrades the `@azure-tools/typespec-ts` emitter to use the TypeSpec `CompilerHost` (from `program.host`) instead of Node.js `fs` APIs, enabling the emitter to run in the TypeSpec playground (browser environment).

### Remove Node.js `fs` dependencies
- **`src/utils/resolve-project-root.ts`** — Static URL-based resolution using `import.meta.url`, no filesystem access needed
- **`src/utils/file-system-utils.ts`** — `pathExists`, `emptyDir`, `clearDirectory` now take `CompilerHost` as first param
- **`src/framework/load-static-helpers.ts`** — Uses `host.readDir/readFile/stat` instead of `fs/promises`
- **`src/modular/emit-tests.ts`** — Uses `host.stat/rm` instead of `existsSync/rmSync`
- **`src/rlc-common/metadata/build-readme-file.ts`** — `hasClientNameChanged` and `updateReadmeFile` accept content strings instead of reading from disk
- **`src/index.ts`** — Threads `host` through all callees; derives emitter package root from `program.emitters` metadata

### Browser/playground fixes
- **All `new Project()` calls** — Added `{ useInMemoryFileSystem: true }` for ts-morph to work without real filesystem
- **`src/utils/emit-util.ts`** — Explicitly passes prettier plugins (`typescript`, `estree`) since browser bundles don't auto-discover them
- **`src/lib.ts`** — Changed `file-formatting-error` severity to `warning` so formatting failures don't block file emission

image

### DIFF
--- a/packages/typespec-ts/src/utils/emit-util.ts
+++ b/packages/typespec-ts/src/utils/emit-util.ts
@@ -1,5 +1,6 @@
 import { CompilerHost, getDirectoryPath, joinPaths, NoTarget, Program } from "@typespec/compiler";
 import { format } from "prettier";
+import prettierPluginBabel from "prettier/plugins/babel";
 import prettierPluginEstree from "prettier/plugins/estree";
 import prettierPluginTypescript from "prettier/plugins/typescript";
 import { prettierJSONOptions, prettierTypeScriptOptions, reportDiagnostic } from "../lib.js";
@@ -74,7 +75,7 @@ async function emitFile(
     try {
       prettierFileContent = await format(prettierFileContent, {
         ...(isJson ? prettierJSONOptions : prettierTypeScriptOptions),
-        plugins: [prettierPluginTypescript, prettierPluginEstree],
+        plugins: [prettierPluginTypescript, prettierPluginEstree, prettierPluginBabel],
       });
     } catch (e) {
       reportDiagnostic(program, {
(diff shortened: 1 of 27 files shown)


packages/typespec-ts/src/utils/emit-util.ts
```diff
@@ -70,10 +72,10 @@ async function emitFile(
   // Format the contents if necessary
   if (isJson || isSourceCode) {
     try {
-      prettierFileContent = await format(
-        prettierFileContent,
-        isJson ? prettierJSONOptions : prettierTypeScriptOptions,
-      );
+      prettierFileContent = await format(prettierFileContent, {
+        ...(isJson ? prettierJSONOptions : prettierTypeScriptOptions),
```
maorleger, line 77: When isJson is true, do we still need prettierPluginTypescript & prettierPluginEstree ?

And from the other side, are there any plugins that we need for json support?

maorleger (approved):
Had one question about json formatting but otherwise LGTM

packages/typespec-ts/src/utils/emit-util.ts
timotheeguerin: yep good point, need the json plugin too

maorleger (approved):

PR #4628, opened 2026-06-15
Migrate @azure-tools/typespec-go from autorest.go into this repo

tadelesh:
Migrates `@azure-tools/typespec-go` from `Azure/autorest.go` into this monorepo. See [Azure/autorest.go#1982](https://github.com/Azure/autorest.go/issues/1982) for full background and plan.

### Source code (Phase 3 of the migration plan)
- New `packages/typespec-go/` containing the emitter and its three previously-private workspace dependencies inlined:
  - `codegen.go` → `src/codegen/`
  - `codemodel.go` → `src/codemodel/`
  - `naming.go` → `src/naming/`
- Static assets: `README.md`, `LICENSE`, `CHANGELOG.md`, `vitest.config.ts`, `tsconfig.json` (mirrors `typespec-python`).
- `.chronus` change entry.

### Tests
The emitter is tested with three tiers, in order of preference (documented in `docs/development.md`):

1. **Emitter unit tests (scenarios)** — the default. Markdown snapshot tests under `test/unittest/scenarios/*.md`: each `.md` compiles a small TypeSpec in memory and snapshots the generated Go, with no Go toolchain or mock server. ~29 focused scenarios cover naming, models/serialization, client & options shapes, polymorphism, client customizations (`@@alternateType` / `@@clientName`), containing-module, sample generation (example JSON input → `*_example_test.go`), and the former `Regressions` code-gen cases.
   - `scenario-runner.ts` — in-memory compile + snapshot harness (`SCENARIOS_UPDATE=true` regenerates expected blocks; ARM specs auto-routed to an ARM tester).
   - `gen-scenario-suites.mjs` — generates one vitest suite per `.md`; `scenario-suites.guard.test.ts` fails if they drift.
2. **External service specs** — all Azure RM / data-plane service specs (17 modules) are sourced from `Azure/azure-rest-api-specs` at a pinned commit via a sparse, blobless, shallow clone:
   - `.scripts/azure-rest-api-specs.json` — pinned commit + sparse paths
   - `.scripts/sync-azure-rest-api-specs.js` — idempotent sync (skipped when cache marker matches)
   - `.scripts/tspconfig.yaml` — stub config that suppresses upstream tspconfig bleed (the per-language `@azure-tools/typespec-go` options that target `azure-sdk-for-go`)
3. **Local spec tests** — kept only where the behavior can *only* be verified by executing generated Go at runtime. Just two remain:
   - `gogenerate` (`test/tsp/GoGenerate` → `test/local/gogenerate`) — the post-generation `go generate` hook.
   - `fakeserver` (`test/tsp/FakeServer` → `test/local/fakeserver`) — fake-server runtime behavior (context cancellation, dispatch races, path/query decoding); a minimal 2-operation spec.

Removed local specs (migrated into scenarios): `ClientOption`, `Internal.Pager`, `ModelsOnlyWithBaseTypes`, `NoOptionalBody`, `PageableLROs`, `Random.Management`, `RawJson`, and the `Test.Management` ARM provider. The old `Regressions` grab-bag was split into scenarios plus the slimmed `fakeserver` runtime test. The local `ExactName` test was dropped (duplicated by an azure-http-specs scenario).

### Build / regenerate
- `.scripts/tspcompile.js` — Always invoked with `--config=.scripts/tspconfig.yaml` so upstream tspconfigs don't bleed.
- `.scripts/semaphore.js` — 8-way parallel compile primitive.
- `.scripts/spector.js` — reused start/stop wrapper around `npx tsp-spector`.

### CI
- `.github/workflows/ci-go.yml` — jobs: `build`, `test` (Go matrix `1.26.1`/`1.25.8` with regenerate + `lint:go` + spector-backed `test:go:e2e`), `unit` (vitest), `docs` (`regen-docs`).
- `.github/actions/setup-go/action.yml` — composite wrapping `actions/setup-go@v5`.
- `eng/config/labels.ts` — adds `emitter:go` area label + path mapping.
- `eng/pipelines/jobs/build-for-publish.yml` — adds Go regen + e2e + `upload-spector-coverage.yml` template call (`GeneratorName=@azure-tools/typespec-go`) after the Python block.
- `tsconfig.ws.json` includes the new package.
- Root `mise.toml` pins `go = "1.26.1"` for local development; in CI Go is installed via `actions/setup-go` (GH Actions) and `GoTool@0` (ADO release pipeline).
- `.gitignore` ignores generated `zz_*.go` / `go.mod` / `go.sum` / `LICENSE.txt` / `testdata` under `packages/typespec-go/test/`.

### Skipped per discussion
- No nightly pipeline — all TypeSpec deps live in this repo now.
- `test/local/` stays in this repo — not relocated to `azure-sdk-for-go`.

## Validation (local)
- `pnpm build` ✅
- `pnpm test` (vitest unit) ✅ 47/47 (29 scenario suites + guard + emitter tests)
- `pnpm regenerate` ✅ all tsp compiles
- `pnpm format:check` ✅ clean for all `packages/typespec-go/**` files
- `pnpm lint:go` ✅ `0 issues` (golangci-lint v2.11.4 + shadow, every generated module)
- `pnpm test:go:e2e` ✅ spector start → hand-written test modules (azblob / azkeys / gogenerate / fakeserver) PASS → spector stop

## Follow-ups (out of scope; tracked in [autorest.go#1982](https://github.com/Azure/autorest.go/issues/1982))
- Migrate remaining swagger-based data-plane SDKs to the latest autorest.go (Phase 1, in flight).
- Stop maintaining `codegen.go` / `codemodel.go` / `naming.go` in `autorest.go` once Phase 4 lands.
- Announce deprecation of `autorest.go` (after Phase 4).


.chronus/changes/typespec-go-move-into-typespec-azure-2026-06-11-14-30-00.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
```
timotheeguerin:
```suggestion
changeKind: internal
```

.github/actions/setup-go/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Setup Go
```
timotheeguerin, line 1: setup go version in mise.toml too for local developement

packages/typespec-go/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "LICENSE"
+  ],
+  "peerDependencies": {
+    "@azure-tools/typespec-client-generator-core": "workspace:^",
+    "@typespec/compiler": "workspace:^",
+    "@typespec/http": "workspace:^"
+  },
+  "dependencies": {
+    "semver": "catalog:",
+    "source-map-support": "catalog:"
```
timotheeguerin: I don't think this should be a dependency
1. I believe source-map-support is not needed anymore, node can follow sourcemaps 
2. If it is probably be better to have it as a dev like we have in typespec packages and conditionally try to import(though as point 1. we should cleanup there too)

packages/typespec-go/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "dist/**",
+    "!dist/test/**",
+    "README.md",
+    "CHANGELOG.md",
+    "LICENSE"
+  ],
+  "peerDependencies": {
+    "@azure-tools/typespec-client-generator-core": "workspace:^",
+    "@typespec/compiler": "workspace:^",
+    "@typespec/http": "workspace:^"
```
timotheeguerin: is it really only those 3 peers?

.github/workflows/ci-go.yml
```diff
@@ -0,0 +1,129 @@
+name: "go / test e2e"
```
timotheeguerin, line 1: are we settled on the inclusion of generated code?(this pr looks good but don't want to merge that if there is unresolved concern about the generated output)

.chronus/changes/typespec-go-move-into-typespec-azure-2026-06-11-14-30-00.md
timotheeguerin, line 1: doesn't need to be part of this PR if the emitter wasn't built to handle it nicely but did you try to add it to the playground as well?

.github/workflows/ci-go.yml
jhendrixMSFT: I don't believe a consensus has been reached but I think it's leaning towards _not_ including it.  However, no alternative mechanism has been identified/implemented.

Also, it seems a bit weird that we commit the client-side Spector tests without the generated content.  Is that the pattern followed by other languages?  Shouldn't they live together?  Each half is pretty useless on its own.

packages/typespec-go/.scripts/azure-rest-api-specs.json
jhendrixMSFT, line 1: Do I need to be added as a code-owner for this directory, or how does that work?

.github/workflows/ci-go.yml
timotheeguerin: I believe that is the case, well the files are there as soon as you run the build

jhendrixMSFT: Oh so the generated content is dumped into where the tests live.  Ok I think this makes sense.

jhendrixMSFT:
Are there plans to get more granular WRT what runs in CI?  At present it looks like quite a few things ran that's not germane to this emitter.

packages/typespec-go/.scripts/azure-rest-api-specs.json
tadelesh: The script helps to sparse checkout the spec from the configured commit id and spec folder. And the regen script use it instead of what we maintain locally in the previous autorest.go.

jhendrixMSFT: To clarify, my question was about the `packages/typespec-go` directory and not the file (sorry for the confusion).

tadelesh:
> Are there plans to get more granular WRT what runs in CI? At present it looks like quite a few things ran that's not germane to this emitter.

`go / test e2e` is specific for Go. And it will build, regenerate, lint and test just like previous ci process. Other CIs are common for all typespec-azure packages.

packages/typespec-go/.scripts/go-test.js
```diff
@@ -0,0 +1,49 @@
...
+// use `pnpm spector --start`/`--stop` (see spector.js) to manage it.
+
+import { spawnSync } from "child_process";
+import { existsSync, readdirSync, statSync } from "fs";
+import { dirname, resolve } from "path";
+import { fileURLToPath } from "url";
+
+const scriptDir = dirname(fileURLToPath(import.meta.url));
+const pkgRoot = resolve(scriptDir, "..");
+const testLocal = resolve(pkgRoot, "test", "local");
```
jhendrixMSFT: This is limiting to the local tests.  Is this temporary for initial testing?

jhendrixMSFT:
> Other CIs are common for all typespec-azure packages.

Are there plans to improve this (probably out of scope for this PR) so that emitter PRs don't have to be gated by unrelated tests?

tadelesh:
> > Other CIs are common for all typespec-azure packages.
> 
> Are there plans to improve this (probably out of scope for this PR) so that emitter PRs don't have to be gated by unrelated tests?

@timotheeguerin Do you think we could skip some of the CIs that not related to emitters?

.chronus/changes/typespec-go-move-into-typespec-azure-2026-06-11-14-30-00.md
tadelesh: do you mean let playground could review go emitted code? that's not the goal for this pr. we just want to do the migration first before deprecate autorest.go.

packages/typespec-go/.scripts/azure-rest-api-specs.json
tadelesh: Currently, we do not have package level ownership for typespec-azure repo. Anyone have write access could approve the PR. @timotheeguerin Can we add that? Also, it seems the approval will not be steal if new commit is added after an approval.

packages/typespec-go/.scripts/go-test.js
tadelesh: Changed to all the tests.

packages/typespec-go/test/unittest/scenarios/client-option-monomorphic-response-field.md
```diff
@@ -0,0 +1,36 @@
...
+@route("/list")
+op list(): Array<Person>;
+
+#suppress "@azure-tools/typespec-client-generator-core/client-option" "response has custom field name Users"
+@@clientOption(list, "monomorphicResponseFieldName", "Users", "go");
+```
+
+## The monomorphic response field is named `Users` instead of the default
+
+```go responses
```
jhendrixMSFT, line 24: Since this file only has a `response` section, does it mean that the other generated files are ignored for this test (if so it makes sense as they aren't interesting for this test case)?

jhendrixMSFT: Maybe a better question is, for these markdown-style tests, do we only need to include the portion(s) of generated code that's interesting for these tests?

packages/typespec-go/test/unittest/scenario-suites/body-root.test.ts
jhendrixMSFT, line 1: So each test suite has a 1:1 mapping of `.ts` file and `.md` file yes?

packages/typespec-go/test/unittest/scenarios/client-option-monomorphic-response-field.md
tadelesh: Yes. We should only add the generated files that we are care about (we want to check for the test scenario).

packages/typespec-go/test/unittest/scenario-suites/body-root.test.ts
tadelesh: Yes. The ts file is auto-generated. The reason is to use such way to split them into multiple group, so the test results can be clear.

packages/typespec-go/docs/development.md
```diff
@@ -0,0 +1,207 @@
...
+From the repo root, install all dependencies:
+
+```terminal
+pnpm install
+```
+
+Then build the emitter with Turbo. Building through Turbo (rather than calling `tsc` directly) ensures the package's workspace dependencies are built first and lets Turbo cache unchanged builds:
+
+```terminal
+npx turbo run build --filter=@azure-tools/typespec-go
```
jhendrixMSFT: I assume we can add a script to `package.json` to simplify this yes?

packages/typespec-go/package.json
```diff
@@ -0,0 +1,103 @@
...
+    "build": "tsc -p .",
+    "watch": "tsc -p . --watch",
+    "test": "vitest run --pass-with-no-tests",
+    "test:update": "node .scripts/test-update.js",
+    "gen:scenario-suites": "node test/unittest/gen-scenario-suites.mjs",
+    "test:ci": "vitest run --pass-with-no-tests --coverage",
+    "test:go:e2e": "pnpm spector --start && pnpm go-test && pnpm spector --stop",
+    "go-test": "node .scripts/go-test.js",
+    "lint:go": "node .scripts/lint-go.js",
+    "regenerate": "node .scripts/tspcompile.js",
```
jhendrixMSFT: Why was this renamed to `regenerate`?  At any rate, if this is the preferred name (e.g. aligns with what other emitter call it) it's fine but then we should rename the script to match.

packages/typespec-go/docs/development.md
```diff
@@ -0,0 +1,207 @@
...
+
+To regenerate a single test, pass a `--filter` matching the test name (the value is treated as a regular expression):
+
+```terminal
+pnpm regenerate --filter=TestName
+```
+
+The first run also syncs the Azure REST API specs and the generated-code baseline (used by [Review the generated-code diff](#review-the-generated-code-diff)). To skip the baseline sync (for example when working offline), set `TYPESPEC_GO_SKIP_BASELINE=1`.
+
+### Review the generated-code diff
```
jhendrixMSFT: I tried this after building two projects (one Spector, the other local) but I see no diff (includes a test codegen change).

packages/typespec-go/docs/development.md
```diff
@@ -0,0 +1,207 @@
...
+- ` ```yaml ` — emitter options for the compile, as kebab-case keys (for example `generate-samples: true`, `containing-module: foo/bar`, `file-prefix: zz_`).
+- ` ```json <path> ` — an extra input file written verbatim into the in-memory file system at `<path>`. This is how additional compiler inputs are supplied by filename; for example ` ```json examples/Foo.json ` provides the example JSON that drives sample generation.
+- ` ```go <name> ` — a snapshot of a generated Go file, matched by base name. It supports subdirectories (`fake/widget_server`) and sample files (`widgets_client_example_test`). Example (`*_example_test.go`) output is checked with the same `go <name>` block as any other generated file.
+
+Order a scenario input-first so it reads top-to-bottom: `yaml` config, then `tsp`, then any `json` inputs, then the `go` output blocks.
+
+To (re)generate the expected output blocks from the current emitter, run the update command and review the produced snapshots as part of your change:
+
+```terminal
+pnpm test:update
```
jhendrixMSFT, line 100: I'm not able to get this to work.  I added a new `testing.md` file with a `tsp` block (simple operation that returns a model).  After executing `pnpm test:update` the tests run/pass but I don't see any Go code in the `.md` file.

packages/typespec-go/docs/development.md
tadelesh: This is because I have not merged the initial generated code version. Since we are going to switch to Libba's solution, I removed this solution code.

packages/typespec-go/docs/development.md
tadelesh: Did you add empty block with which generated Go file you want to check? I added a warning for empty Go block case.

packages/typespec-go/package.json
tadelesh: Added.

packages/typespec-go/docs/development.md
jhendrixMSFT: It could be the case that I misspelled the file.  I have it working now (erroring on no blocks is a nice addition and makes it clear the author did something wrong).  But this makes me think that maybe we need some tweaks here.
- if no Go blocks are specified, it would be great if the failure output could list the possible block names
- it would also be good to have a simple way to say "all blocks".  maybe something like `go all`, or even just an unnamed Go block

jhendrixMSFT (approved):

packages/typespec-go/test/unittest/scenarios/discriminated-base-no-subtypes.md
```diff
@@ -0,0 +1,77 @@
...
+```go models
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) Go Code Generator. DO NOT EDIT.
+
+package testmodule
+
+// DiscriminatedBaseNoSubTypes - A discriminated base type with no subtypes.
+type DiscriminatedBaseNoSubTypes struct {
+// REQUIRED; Some property.
```
jhendrixMSFT: Nit: indentation is missing

tadelesh: Fixed

packages/typespec-go/docs/development.md
tadelesh: I added possible block name to the warning message. But I do not prefer to add a all-in-one block, which makes the md too long.

PR #4632, opened 2026-06-15
[specs] add missing ARM scenarios

### LINKED ISSUE #4609: [specs] Push to add more coverage for ARM scenarios

iscai-msft:
fixes #4609

1. `getLro` with `@markAsLro`
2. `postActionPaging`
3. `@markAsPageable`
4. `postPaginglroWithBody`
5. legacy template usages


.chronus/changes/specs-addArmTests-2026-5-15-14-14-34.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/azure-http-specs"
+---
+
+Add tests for ARM `getLro`, `postActionPaging`, `markAsPageable`, and `postPagingLroWithBody`
```
XiaofeiCao: Add legacy cases to changelog as well?

IMHO, from reviewer's perspective, better to separate these cases into separate PRs.

packages/azure-http-specs/specs/azure/resource-manager/operation-templates/mockapi.ts
```diff
@@ -894,3 +894,411 @@ Scenarios.Azure_ResourceManager_OperationTemplates_LroPaging_postPagingLro = pas
...
+const validConfigurationWithBody = {
+  ...validConfiguration,
+  properties: {
+    configValue: "custom-value",
+    provisioningState: "Succeeded",
+  },
+};
+
+Scenarios.Azure_ResourceManager_OperationTemplates_Legacy_createOrReplaceOptionalBody =
+  passOnSuccess([
```
XiaofeiCao: For same method on the identical uri, consider using `withServiceKeys`. Example: https://github.com/Azure/typespec-azure/blob/ca5bc9e0036e71b616702f7d38c275992fed4316/packages/azure-http-specs/specs/azure/resource-manager/operation-templates/mockapi.ts#L621-L624

The `passOnSuccess` for the same uri of the same HTTP method would probably end up registering only one of them.

.chronus/changes/specs-addArmTests-2026-5-15-14-14-34.md
iscai-msft: Updated the changelog to include the legacy cases (routedGet and createOrReplaceOptionalBody). Re splitting: these are all related ARM resource-manager    operation template scenarios, so I'd prefer to keep them in one PR and it's easier to keep track of. hope that's ok, thanks!

packages/azure-http-specs/specs/azure/resource-manager/operation-templates/mockapi.ts
iscai-msft: Good catch, fixed! Switched to `withServiceKeys(["EmptyBody", "WithBody"]).pass({...})` with a handler that checks for body content, matching the existing `OptionalBody_patch` pattern.

packages/azure-http-specs/specs/azure/resource-manager/operation-templates/mockapi.ts
```diff
@@ -894,3 +894,413 @@ Scenarios.Azure_ResourceManager_OperationTemplates_LroPaging_postPagingLro = pas
...
+        operationId: "report1",
+      },
+      query: {
+        "api-version": "2023-12-01-preview",
+      },
+    },
+    response: {
+      status: 202,
+      headers: {
+        location: dyn`${dynItem("baseUrl")}/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/providers/Azure.ResourceManager.OperationTemplates/locations/eastus/operationResults/get_lro_location`,
```
XiaofeiCao: nit, service example and my local testing recording both indicate that the `Location` header shares the same URL with initial request:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cost-management/resource-manager/Microsoft.CostManagement/CostManagement/examples/2025-03-01/CostDetailsOperationResultsBySubscriptionScope.json#L41

Now sure if this is the guideline to follow (I'm not able to find any guidelines for GET LRO).

packages/azure-http-specs/specs/azure/resource-manager/operation-templates/lropaging.tsp
```diff
@@ -93,4 +98,97 @@ interface LroPaging {
     """)
   @list
   postPagingLro is ArmResourceActionAsync<Product, void, ProductListResult>;
```
XiaofeiCao, line 100: Sorry, seems we already have LRO pageable case..

iscai-msft: we do, but the existing `postPagingLro` uses `ArmResourceActionAsync` with no body, while the new one I'm adding uses a request body, so it is a different template pattern. I do think it's worth to have both

packages/azure-http-specs/specs/azure/resource-manager/operation-templates/mockapi.ts
iscai-msft: ok will follow the service example, might as well since this is a very specific test

packages/azure-http-specs/specs/azure/resource-manager/operation-templates/legacy.tsp
```diff
@@ -0,0 +1,155 @@
...
+
+// --- Models for legacy template scenarios ---
+
+@resource("configurations")
+model Configuration is TrackedResource<ConfigurationProperties> {
+  ...ResourceNameParameter<Configuration>;
+}
+
+model ConfigurationProperties {
+  @doc("The configuration value.")
```
timotheeguerin, line 22: try to use doc ccomments instead of `@doc` when possible

timotheeguerin (approved):

XiaofeiCao (approved):

PR #4646, opened 2026-06-17
fix(typespec-ts): keep paged model public when used as a property

### LINKED ISSUE #4645: Paged result model used as a property incorrectly gets '_' prefix
## Bug Description

When a model decorated with `@pageItems`/`@nextLink` (i.e., a paged result model) is used **both** as a paging operation response **and** as a property on another model, the emitter incorrectly prefixes it with `_` in the generated TypeScript code.

For example, in the Azure Kusto spec, `LanguageExtensionsList` is:
1. The response type of `listLanguageExtensions` (a paging operation)
2. A property on `ClusterProperties`: `languageExtensions?: LanguageExtensionsList`

The generated code produces:

```typescript
languageExtensions?: _LanguageExtensionsList;
```

The `_` prefix is meant to hide paging envelope models that users interact with via `PagedAsyncIterableIterator`. But when the same model is also used as a regular property, it's part of the public API surface and should **not** have the `_` prefix.

## Root Cause

In `packages/typespec-ts/src/modular/emit-models.ts`, the `normalizeModelName` function applies `_` when:

```typescript
isPagedResultModel(context, type) && !pagedModelsUsedInNonPagingOps.has(type)
```

The `trackPagedModelInNonPagingMethod` function only tracks paged models used as **direct method response types** in non-paging operations. It does not account for paged models used as **property types** on other models.

## Proposed Fix

Add tracking for paged models used as model properties in `visitType`, so they are added to the public set and keep their name without the `_` prefix.

## Reproduction

Minimal TypeSpec spec with a `Cluster` resource that has:
- `ClusterProperties.languageExtensions?: LanguageExtensionsList`
- A paging operation: `listLanguageExtensions is ArmResourceActionSync<Cluster, void, ArmResponse<LanguageExtensionsList>>`

## Related

- https://github.com/Azure/azure-sdk-for-js/pull/38977#discussion_r3417330382

JialinHuang803:
## Problem

Paged result models decorated with `@pageItems`/`@nextLink` were incorrectly prefixed with `_` even when used as a model property (e.g. `ClusterProperties.languageExtensions?: _LanguageExtensionsList`).

The `_` prefix should only apply when the model is exclusively used as a paging operation response envelope. When the same model is also referenced as a property on another model, users interact with it directly — it must remain public.

## Root Cause

`trackPagedModelInNonPagingMethod` only tracked paged models used as **direct method response types** in non-paging operations. It did not account for paged models used as **property types** on other models.

## Fix

- Added `trackPagedModelUsedAsProperty()` — when visiting model properties in `visitType`, if a property's type is a paged result model, add it to the public set so it keeps its name without the `_` prefix.
- Renamed `pagedModelsUsedInNonPagingOps` → `pagedModelsKeptPublic` for clarity, since the set now covers both non-paging method responses and property usage.

## Test

Added a new scenario test case in `pagedModelKeptPublic.md` that reproduces the bug: a `LanguageExtensionsList` model used both as a paging operation response and as a property on `TestProperties`.

Fixes #4645

Related: https://github.com/Azure/azure-sdk-for-js/pull/38977#discussion_r3417330382


kazrael2119 (approved):

JialinHuang803:
The spector test failure is already fixed in the main branch.

PR #4647, opened 2026-06-17
fix(typespec-ts): preserve $DO_NOT_NORMALIZE$ operation-group names in classical client

### LINKED ISSUE #4604: [typespec-ts] Keep do-not-normalize operation-group names in classic client and resolve references
related pr: https://github.com/Azure/azure-rest-api-specs/pull/43684
When generating the classical client, the do-not-normalize operation-group is unexpectedly normalized a second time, causing a case mismatch between the referenced symbol name and the declared symbol name, which leads to placeholder resolution failure. During classical client generation, the original (unnormalized) input should be consumed to avoid re-normalization.
typespec: 
```
model RelayResource {
  name?: string;
}

@route("/items")
interface WcfRelaysOps {
  @get
  get(@query id: string): RelayResource;
}

@@clientLocation(WcfRelaysOps.get, "$DO_NOT_NORMALIZE$wCFRelays", "javascript");
```

generate code: `public readonly wCFRelays: __PLACEHOLDER_sWCFRelaysOperations_s0_sclassicOperations__;`

v-jiaodi:
When `@@clientLocation` assigns an operation to a `$DO_NOT_NORMALIZE$`-prefixed group (e.g. `"$DO_NOT_NORMALIZE$wCFRelays"`), the classical client generator was double-normalizing the name: the first pass strips the prefix → `"wCFRelays"`, then a second `normalizeName(..., NameType.OperationGroup)` call converts it to `"WCFRelays"` via PascalCase reconstruction. This produced a refkey mismatch with the names registered by `buildClassicOperationFiles` (which uses single normalization), leaving unresolved placeholders in the output:

```ts
// Before fix
public readonly wCFRelays: __PLACEHOLDER_sWCFRelaysOperations_s0_sclassicOperations__;

// After fix
public readonly wCFRelays: wCFRelaysOperations;
```
fixes #4604


JialinHuang803 (approved):

JialinHuang803:
The spector tests are already resolved in the main branch.

PR #4650, opened 2026-06-17
feat(typespec-client-generator-core): support per-service api-version map for multi-service packages

### LINKED ISSUE #4009: How to set the `api-version` option for multi-service package in tspconfig.yaml
Currently tspconfig.yaml allows an `api-version` option as a string, which is also used in multi-service scenarios and leads to problems in the generated code. Should we enable setting multiple api-versions here, or should we disable this field when dealing with multiple services?

If we decide to support this option for multiple services, does it function the same way as `useDependency` in client.tsp?

tadelesh:
Fixes #4009

## Summary
Extends the TCGC `api-version` emitter option to accept either a `string` or a `Record<string, string>` (a map from each service namespace's full name to its desired version), enabling per-service version selection for multi-service packages.

## Behavior
- **`string`** (e.g. `"2024-10-01"`): applies to single-service packages only. Ignored for multi-service.
- **`"latest"`**: global keyword, applies regardless of service count.
- **`"all"`**: single-service only. **Multi-service packages do not support `all`** (in either string or map form) — it is ignored and each service falls back to its latest version.
- **`Record<string, string>`** (e.g. `{ "ServiceA": "av1", "ServiceB": "bv2" }`): maps each service namespace full name to a version. Services not listed default to their latest version.

## Implementation notes
- Added `resolveApiVersionForService` as the single source of truth for resolving the version that applies to a given service.
- `getServiceNamespaces` is memoized on the context (`__serviceNamespaces`) since it is consulted on the per-type hot path via `getAvailableApiVersions`.
- `handleVersioningMutationForGlobalNamespace` now mutates each service independently using its resolved version.

## Validation
- Added multi-service map tests (per-service versions, fallback to latest, `all` ignored).
- Full TCGC suite: 1330 passed / 2 skipped. TypeScript build and ESLint clean.


weidongxu-microsoft (commented):
agent:
```
 1. Docs are now inconsistent with the feature and behavior.
 packages/typespec-client-generator-core/README.md*#COLON|*63-67 and website/src/content/docs/docs/libraries/typespec-client-generator-core/reference/emitter.md*#COLON|*57-61 still say api-version is Type: string, but this PR changes the option schema to also accept a per-service map (packages/typespec-client-generator-core/src/lib.ts*#COLON|*12-27, src/interfaces.ts*#COLON|*44). The docs also say all is accepted without clarifying that multi-service packages intentionally do not support all and fall back to latest. As written, users will configure the option incorrectly.
```

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -364,6 +364,15 @@ export function filterApiVersionsWithDecorators(
...
   apiVersions: string[],
 ): string[] {
+  // The service namespace is only needed to resolve a per-service version map.
+  const serviceNamespace =
+    typeof context.apiVersion === "object" ? getServiceNamespaceForType(context, type) : undefined;
+  const apiVersion = resolveApiVersionForService(context, serviceNamespace);
+  // index of the explicitly specified version in the list; -1 means latest / not found
+  const apiVersionIndex =
+    apiVersion === undefined || apiVersion === "latest" || apiVersion === "all"
+      ? -1
```
weidongxu-microsoft, line 375: nit, `-1` seems a bit hacky (it's fine in C, but probably not good in typescript)

fine to leave as is

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -1020,6 +998,97 @@ export function handleVersioningMutationForGlobalNamespace(context: TCGCContext)
...
+ * - When the option is a record, the version is looked up by the service
+ *   namespace's full name. Services that are not listed return `undefined`
+ *   (meaning "use the latest version").
+ *
+ * Multi-service packages do not support the special `all` value (in either the
+ * string or the record form); it is ignored and treated as `undefined` (use the
+ * latest version of each service).
+ *
+ * The returned value can be a specific version, the special values `latest` /
+ * `all`, or `undefined`.
```
weidongxu-microsoft, line 1036: maybe we can drop `latest` as well as `all`, in future PR. that could make the logic simpler.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -1020,6 +998,97 @@ export function handleVersioningMutationForGlobalNamespace(context: TCGCContext)
...
+    // `latest` is a global keyword that applies regardless of how many services
+    // the package targets.
+    if (config === "latest") return "latest";
+    // `all` and specific version strings only apply to the single service case;
+    // multi-service packages do not support `all`.
+    return isMultiService ? undefined : config;
+  }
+  // Record case: map each service namespace's full name to a version.
+  if (serviceNamespace === undefined) return undefined;
+  const version = config[getNamespaceFullName(serviceNamespace)];
```
weidongxu-microsoft, line 1055: so, if the `api-version` dict does not have a full mapping from serviceNamespace to apiVersion (e.g. miss a `Microsoft.Disk`, will it still be correct for the missed serviceNamespace (does it throw -- "no such key", or get `undefined` and work as if `latest`)?

weidongxu-microsoft (commented):
another one from agent that is too complex for me to understand :-)

therefore just throw it out for expert

```
 1. Per-service version resolution ignores the wrapper, so shared/common types can get the wrong apiVersions.
 getAvailableApiVersions still accepts a wrapper specifically so version filtering can be contextual (src/internal-utils.ts:409-426), but filterApiVersionsWithDecorators now resolves the selected service only from type (src/internal-utils.ts:362-370). That works when the type lives under a service namespace, but it breaks for shared models/enums/unions that are outside the service namespace tree. In those cases getServiceNamespaceForType(context, type) returns undefined, the per-service map is ignored, and the type can keep a broader apiVersions set than the selected service version should allow. You can already see affected call sites that pass type.namespace instead of an operation/service wrapper, e.g. model and enum creation in src/types.ts:876 and src/types.ts:1059.
 Suggested fix: carry the wrapper through into filterApiVersionsWithDecorators, and resolve the service from the wrapper first, then fall back to the type.
```

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: Yes. I'm willing to remove them all. I'm not sure if any language uses it. I did not see any in the spec repo.

packages/typespec-client-generator-core/src/internal-utils.ts
tadelesh: It will fall back to latest version.

weidongxu-microsoft (approved):

PR #4660, opened 2026-06-18
feat(typespec-autorest): add type-name-strategy option to remove namespace prefix from names

### LINKED ISSUE #2055: [typespec-autorest]: suggest removing namespace from the definition name in swagger

The most common case occurs with Liftr-related services. The typespec-autorest tool prepends the namespace to model names when generating definition names in Swagger. For [example](https://github.com/Azure/azure-rest-api-specs/blob/a5e15ea76b123c1fed61c2b3464688edb8d84788/specification/liftrpinecone/resource-manager/Pinecone.VectorDb/preview/2024-10-22-preview/openapi.json#L469):

    "LiftrBase.MarketplaceDetails": {
 

However, for client emitters (such as C#, PowerShell, etc.), the namespace is ignored, for the example above, **MarketplaceDetails** will be used as the generated class name. This inconsistency will create significant challenges in the future when we transition from Swagger-based generation to typespec-based generation, as changes in definition (class) names will result in breaking changes.

I understand that the purpose of prepending the namespace is to avoid name conflicts. However, I propose the following approach:

Remove the namespace from the definition name and only prepend it when a name conflict is detected.

### Checklist

timotheeguerin:
Fixes #2055

## Problem

`@azure-tools/typespec-autorest` prepends the namespace to model names when generating Swagger definition names (e.g. `LiftrBase.MarketplaceDetails`). Client emitters (C#, PowerShell, etc.) ignore the namespace and use just the type name (`MarketplaceDetails`). This inconsistency causes breaking changes when migrating from swagger-based codegen to TypeSpec-native codegen.

## Change

Adds a new emitter option `type-name-strategy`:

- `"namespaced"` (default): current behavior — names include the namespace prefix for types outside the service namespace (e.g. `LiftrBase.Foo`). The service (and root `TypeSpec`) namespace is always stripped.
- `"name-only"`: names use only the type name without any namespace prefix (e.g. `Foo`), matching what client emitters produce.

```yaml
options:
  "@azure-tools/typespec-autorest":
    type-name-strategy: "name-only"
```

The strategy applies globally to all OpenAPI names derived from TypeSpec types (definition/schema names, parameter keys, inline names, `x-typespec-name`) via the shared `namespaceFilter`, keeping everything consistent.

When two types from different namespaces collapse to the same name, the existing `@typespec/openapi/duplicate-type-name` error is reported (no auto fallback to the qualified name). `@clientName` and `@friendlyName` continue to take precedence.

## Backward compatibility

Default is `"namespaced"`, so existing behavior is unchanged unless the option is explicitly set to `"name-only"`.

## Tests

Added `packages/typespec-autorest/test/type-name-strategy.test.ts` covering:
- default / explicit `namespaced` keeps the prefix, and same-named types in different namespaces coexist
- `name-only` removes the prefix from definitions
- `name-only` collision across namespaces errors with `duplicate-type-name`
- `@friendlyName` still wins under `name-only`

### DIFF
--- a/.chronus/changes/autorest-type-name-strategy-2026-6-18-12-15-0.md
+++ b/.chronus/changes/autorest-type-name-strategy-2026-6-18-12-15-0.md
@@ -1,0 +1,13 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+Add `type-name-strategy` emitter option to control how OpenAPI names are derived from TypeSpec types. The new `"name-only"` strategy removes the namespace prefix from names (e.g. `Foo` instead of `LiftrBase.Foo`), matching the names used by client emitters. The default `"namespaced"` keeps the current behavior. When two types collapse to the same name, a `duplicate-type-name` error is reported.
+
+```yaml
+options:
+  "@azure-tools/typespec-autorest":
+    type-name-strategy: "name-only"
+```
--- a/packages/typespec-autorest/README.md
+++ b/packages/typespec-autorest/README.md
@@ -196,6 +196,14 @@ Determines whether output should be split into multiple files. The only supporte
 
 When enabled, the emitter will not copy example files to the output directory. Instead, it will reference the source example files using relative file paths.
 
+### `type-name-strategy`
+
+**Type:** `"namespaced" | "name-only"`
+
+**Default:** `"namespaced"`
+
+Strategy for naming the OpenAPI names derived from TypeSpec types. "namespaced" (default) includes the namespace prefix for types outside the service namespace (e.g. `LiftrBase.Foo`). "name-only" uses only the type name without any namespace prefix (e.g. `Foo`), reporting an error when two types collapse to the same name.
+
 ## Decorators
 
 ### Autorest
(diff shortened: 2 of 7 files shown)


.chronus/changes/autorest-type-name-strategy-2026-6-18-12-15-0.md
```diff
@@ -0,0 +1,13 @@
...
+packages:
+  - "@azure-tools/typespec-autorest"
+---
+
+Add `type-name-strategy` emitter option to control how OpenAPI names are derived from TypeSpec types. The new `"bare"` strategy removes the namespace prefix from names (e.g. `Foo` instead of `LiftrBase.Foo`), matching the names used by client emitters. The default `"namespaced"` keeps the current behavior. When two types collapse to the same bare name, a `duplicate-type-name` error is reported.
+
+```yaml
+options:
+  "@azure-tools/typespec-autorest":
+    type-name-strategy: "bare"
```
markcowl: I think we should call this option "plain" or "sparse" or "simple" or "name-only"

markcowl (changes requested):

.chronus/changes/autorest-type-name-strategy-2026-6-18-12-15-0.md
timotheeguerin: name-only sounds good

markcowl (approved):

PR #4664, opened 2026-06-19
Add @featureFile decorators in Azure.ResourceManager and fix template suppression propagation

### LINKED ISSUE #4435: [design]: Remove need to use legacy and private decorators in Network spec
To prevent a name clash with a custom `ProxyResource`, network specs cannot use standard resource templates for new resources.  We should create a mechanism where private dcorators and suppressions are not needed for new ntwork resources.  This could be:

- Fixes to SDK emitters to allow the standard and custom ProxyResource in network
- Allowing suppression in the base type to apply to new resources
- Adding new templates that only require suppression in the base type and provide any private decorators
example: https://github.com/Azure/azure-rest-api-specs/pull/42717/changes#diff-ec5e52cdc9d434e2bde0ebd57ae1230d36275f3f638e3577065fcd9fda30d16e

markcowl:
### New decorators in Azure.ResourceManager namespace
Adds new decorators as alternatives to the Legacy versions:
- \@featureFiles\ (replaces \@Azure.ResourceManager.Legacy.features\)
- \@featureFile\ (replaces \@Azure.ResourceManager.Legacy.feature\)
- \@featureFileOptions\ (replaces \@Azure.ResourceManager.Legacy.featureOptions\)

The Legacy decorators remain unchanged for backward compatibility.

### Fix @customAzureResource suppression propagation
Adjusts the \rm-custom-resource-usage-discourage\ linter rule so that if a \@@suppress\ directive is applied to a model template with \@customAzureResource\, the suppression propagates to all instantiations of that template.

### Files changed
- \lib/decorators.tsp\ - New decorator declarations
- \generated-defs/Azure.ResourceManager.ts\ - New type definitions
- \src/resource.ts\ - New exports delegating to existing implementations
- \src/tsp-index.ts\ - New decorator registrations
- \src/rules/arm-custom-resource-usage-discourage.ts\ - Suppression propagation logic
- Tests, README, and reference docs updated


.chronus/changes/markcowl-move-feature-fix-suppression-2026-5-18-16-41-6.md
timotheeguerin, line 1: That is technically a breaking change right?

.chronus/changes/markcowl-move-feature-fix-suppression-2026-5-18-16-41-6.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+  - "@azure-tools/typespec-autorest"
+---
+
+Move `@feature`, `@features`, and `@featureOptions` decorators from `Azure.ResourceManager.Legacy` to `Azure.ResourceManager` namespace. Propagate `arm-custom-resource-usage-discourage` suppressions from custom resource templates to their instantiations.
```
timotheeguerin: So now there will be no warning at all preventing usage of this functionality?

I guess if we migrated everybody now there is no reason one would want to use that so we don't have to worry about it too much?

markcowl: I have been thinking about this and I think we have 3 options:
1. Move `@feature` and `@featureOptions` out of legacy, which means any new usage (which requires `@features`) would emit a diagnostic, but normal api evolution would not.
2. Move all 3 decorators out of legacy and add a linting rule that emits a warning diagnostic specific to this feature when a namespace uses this functionality
3. Move all 3 decorators out of legacy and emit no warnings

markcowl: I am kind of favoring 2.  I think having the decorators in different namespace is confusing, and it makes sense to have a review for any APIs that suddenly move to using this feature.

.chronus/changes/markcowl-move-feature-fix-suppression-2026-5-18-16-41-6.md
markcowl: It is, indeed.  I wonder if we need to leave the decorators in the legacy namespace for a release or two and deprecate them (or, at least, deprecate one of them)

markcowl:
Note: will follow up with deprecation of the old decorators next sprint, and removal afterward.

timotheeguerin (approved):

Issue #4564, opened 2026-06-04
Phase out client.tsp separate entrypoint


timotheeguerin:
Phase 1: Import client.tsp from main.tsp
- a. First services that need no changes apart from adding `import "./client.tsp"`
    - https://github.com/Azure/azure-rest-api-specs/pull/43749
    - https://github.com/Azure/azure-rest-api-specs-pr/pull/28819
- b. Second services that might need simple tweaking (adding !autorest to scope)
    - https://github.com/Azure/azure-rest-api-specs/pull/43751
    - https://github.com/Azure/azure-rest-api-specs-pr/pull/28820
    - https://github.com/Azure/azure-rest-api-specs/pull/44009
- c. more complex services
   - https://github.com/Azure/azure-rest-api-specs/pull/44027
```
specification/applink/AppLink.Management
specification/edge/Microsoft.Edge.DisconnectedOperations.Management
specification/frontdoor/resource-manager/Microsoft.Network/FrontDoor
specification/migrate/resource-manager/Microsoft.Migrate/Waves
specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents
specification/napster/Napster.CompanionAPI.Management
specification/programenrollment/resource-manager/Microsoft.ProgramEnrollment/ProgramEnrollment
specification/purviewdatagovernance/data-plane/DataAccess
specification/relationships/Relationships.Management
specification/resources/resource-manager/Microsoft.Authorization/policy
specification/resources/resource-manager/Microsoft.Resources/deploymentStacks
specification/search/resource-manager/Microsoft.Search/Search
specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache
```

Phase 2: Enforce client.tsp import
- Add tsv validation to make sure client.tsp is imported https://github.com/Azure/azure-rest-api-specs/pull/43762/

Phase 3: Update tooling
- Linting rule to prevent bad pattern in client.tsp https://github.com/Azure/typespec-azure/issues/4652
- Update tsp client and logic in emitters 
  - https://github.com/Azure/azure-sdk-tools/issues/15955
- cleanup/fix test in typespec-azure reproing the structure
- Documentation
  - https://github.com/Azure/typespec-azure/pull/4677
  - Should we relax restriction about tcgc needing to be imported in client.tsp only?? 

Phase 4: Cleanup
- Remove import main.tsp from client.tsp

markcowl:
complete

timotheeguerin:
[...]

PR #4677, opened 2026-06-22
docs: reference new spec structure where main.tsp imports client.tsp

timotheeguerin:
Part of the migration in #4564 (Phase 3 — documentation/tooling).

Updates scaffold templates and documentation to reflect the new spec repo service structure where **`main.tsp` is the single entrypoint that imports `client.tsp`** (transitional Phase 1: `client.tsp` still imports `main.tsp`).

## Scaffold templates (`eng/feeds/`)
- `data-plane/main.tsp` now adds `import "./client.tsp";`.
- Regenerated snapshots (`__snapshots__/azure-core/main.tsp`, `azure-core_stand_alone/main.tsp`) via the e2e test (6/6 pass, including `tsp compile .` validation).
- ARM templates left untouched — they have no existing `client.tsp` (only updating where `client.tsp` already exists).

## Docs
- `migrate-arm-tips.md` — reversed "❌ DON'T reference client.tsp in main.tsp" → "✅ DO reference client.tsp in main.tsp".
- `migrate-dp-tips.md` — added "✅ DO reference client.tsp in main.tsp".
- `frequentquestions.md` — updated ARM file-layout to note `main.tsp` imports `client.tsp`.
- `01setup.mdx` — added the `main.tsp` import example; changed `tsp compile client.tsp` → `tsp compile .`.
- `03client.mdx` — clarified the `main.tsp` entrypoint wording.

## Out of scope
- Phase 3 structure test fixtures under `packages/azure-http-specs/specs/client/structure/*` (tracked separately in #4564).

### DIFF
--- a/website/src/content/docs/docs/howtos/Generate client libraries/01setup.mdx
+++ b/website/src/content/docs/docs/howtos/Generate client libraries/01setup.mdx
@@ -23,7 +23,13 @@ Ensure that your `package.json` includes a link to the customization library, `t
 }
 ```
 
-Customization should be implemented in a file named `client.tsp`, which should be used alongside with the `main.tsp`.
+Customization should be implemented in a file named `client.tsp`, which should be used alongside with the `main.tsp`. The `main.tsp` imports `client.tsp` so that the whole specification, including client customizations, compiles from the single `main.tsp` entrypoint.
+
+```typespec
+// main.tsp
+import "./client.tsp";
+// ... rest of the service definition
+```
 
 ```typespec
 // client.tsp
@@ -35,8 +41,8 @@ using Azure.ClientGenerator.Core;
 // Add your customizations here
 ```
 
-After you've created your customization file, use the command `tsp compile client.tsp` to generate output that includes your customizations:
+After you've created your customization file, compile the project from the `main.tsp` entrypoint to generate output that includes your customizations:
 
 ```shell
-tsp compile client.tsp
+tsp compile .
 ```
--- a/website/src/content/docs/docs/howtos/Generate client libraries/01setup.mdx
+++ b/website/src/content/docs/docs/howtos/Generate client libraries/01setup.mdx
@@ -23,7 +23,7 @@ Ensure that your `package.json` includes a link to the customization library, `t
 }
 ```
 
-Customization should be implemented in a file named `client.tsp`, which should be used alongside with the `main.tsp`. The `main.tsp` imports `client.tsp` so that the whole specification, including client customizations, compiles from the single `main.tsp` entrypoint.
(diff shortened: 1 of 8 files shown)


website/src/content/docs/docs/howtos/Generate client libraries/01setup.mdx
```diff
@@ -23,7 +23,13 @@ Ensure that your `package.json` includes a link to the customization library, `t
 }
 ```
 
-Customization should be implemented in a file named `client.tsp`, which should be used alongside with the `main.tsp`.
+Customization should be implemented in a file named `client.tsp`, which should be used alongside with the `main.tsp`. The `main.tsp` imports `client.tsp` so that the whole specification, including client customizations, compiles from the single `main.tsp` entrypoint.
```
markcowl:
```suggestion
Customization should be implemented in a file named `client.tsp`, which should be used alongside the `main.tsp`. The `main.tsp` imports `client.tsp` so that the whole specification, including client customizations, compiles from the single `main.tsp` entrypoint.
```

website/src/content/docs/docs/migrate-swagger/checklists/migrate-dp-tips.md
```diff
@@ -87,6 +87,8 @@ union WidgetColor {
 
 ✅ **DO** make client customizations in a `client.tsp` file
 
+✅ **DO** reference `client.tsp` in `main.tsp` (add `import "./client.tsp";` so `main.tsp` is the single entrypoint)
+
 ❌ **DON'T** import or use `@azure-tools/typespec-client-generator-core` in other files aside from client.tsp.
```
markcowl, line 92: I wonder about this one.  There are some cases that come up, such as making a client-specific decoration change in a new version on a type that are made much easier if this is a should not rather than a must not.  Given that we now have a single entrypoint, it seems that this should be allowed in special cases.

markcowl: I think we can make this a second pr, though, I am fine with not changing this here.

markcowl (approved):

website/src/content/docs/docs/migrate-swagger/checklists/migrate-dp-tips.md
timotheeguerin: hhm yeah, I think it is still a good guidance for organization but yeah doesn't really matter now

timotheeguerin: added note in parent issue to resolve

PR #4682, opened 2026-06-23
Remove dead typespec-ts emitter options (module-kind, azure-sdk-for-js, source-from)

xirzec:
## Why

The `@azure-tools/typespec-ts` emitter carried three options that only ever supported a single value in practice, leaving large amounts of unreachable code to maintain:

- `module-kind` — output was always ESM; the CommonJS path was never exercised.
- `azure-sdk-for-js` — output always targeted the Azure SDK for JS monorepo layout; the standalone-package path was only used to make generated packages runnable in the integration tests.
- `source-from` — this internal option was never read. `sourceFrom` was hardcoded to `"TypeSpec"`, so every `=== "Swagger"` branch was statically unreachable.

This PR removes all three options and collapses the code to the surviving defaults, deleting the complexity we won't have to maintain anymore. The emitter now has a single generation mode.

## What changed

- **`module-kind` / `azure-sdk-for-js`**: Generation now always emits ESM into the Azure monorepo layout. Removed the option declarations, the CommonJS-only and standalone-package generation paths, and the now-dead helper files (`build-rollup-config.ts`, `build-azure-standalone-package.ts`, `package-util.ts`).
- **`source-from`**: Removed the option, the `sourceFrom`/`specSource` config plumbing, and all Swagger-only generation branches (logger path resolution, discriminator typing, sample/operation name concatenation, monorepo package metadata, and the dead `buildClientPathParameters`). Also threaded out the now-unused `model` parameter that those branches were the only consumers of.
- **Spector test harness**: Previously the integration tests relied on `azure-sdk-for-js: false` to make each generated package self-contained so vite/oxc could transform it. With that option gone, the emitter always produces the monorepo layout, whose `tsconfig.json` references `config/*.json` files that `extends` the azure-sdk-for-js repo's shared `eng/tsconfigs/*` (which don't exist in this repo). The generated `#platform/*` subpath imports and dependencies already resolve under vitest, so the only missing piece was a loadable tsconfig. `test/commands/run.ts` now writes a self-contained `tsconfig.json` into each generated package after emit. `tsconfig.json` isn't part of the committed baseline, so this only affects local/CI test runs and keeps the package-generation concern out of the emitter.
- **Test snapshots / baselines**: Stripped `azure-sdk-for-js: false` from the 208 integration `tspconfig.yaml` files and regenerated baselines. Six `index.d.ts` rollups changed: platform-types specs now emit an external `#platform/...` import for `NodeReadableStream` instead of an inlined `declare type`.
- **Docs/tests**: Updated `README.md` and `CONTRIBUTING.md`, removed the dead "source is swagger" test and its mock-helper config, regenerated the website emitter reference (`emitter.md`), and documented the collapsed defaults in `.github/instructions/typespec-ts.instructions.md` so the options aren't reintroduced.

## Notes for reviewers

- Two `breaking` changesets are included since public emitter options were removed.
- The self-contained tsconfig that the test harness writes lives only in `test/commands/run.ts` (not in committed baselines), so it relocates the "make the package testable" responsibility into the test infra rather than keeping a test-only generation mode in the emitter.
- Verified locally: build is green; `test-next` (245) and `unit-test:rlc` (316) pass. `unit-test:modular` passes except for one pre-existing, unrelated tcgc failure (`overrideReservedkeywords.md` / `override-parameters-mismatch`) that also fails on a clean `main`. Full Spector e2e suites pass locally: **azure-modular 101 files / 934 tests**, **azure-rlc 82 files**.


xirzec:
The failing unit test is covered by this PR: https://github.com/Azure/typespec-azure/pull/4684

timotheeguerin (approved):

PR #4701, opened 2026-06-25
Speed up @azure-tools/typespec-ts tests

### LINKED ISSUE #2411: How to express a model inherits from multiple parent models
I tried something like this 
```
    model Creature {
      life: int64;
    }
    model Animal {
      name: string;
    }
    model Pet extends Animal, Creature {
      weight?: float32;
    }
    model Cat extends Pet {
      kind: "cat";
      meow: int32;
    }
```

but it doesn't seem to be able to parse it. [playground link](https://cadlplayground.z22.web.core.windows.net/?c=ICAgIG1vZGVsIENyZWF0dXJlIHsNCsQWICBsaWZlOiBpbnQ2NDvGFH3GB8YxQW5pbWFsyi9uYW1lOiBzdHJpbmfUMFBldCBleHRlbmRzxzws03V3ZWlnaHQ%2FOiBmbG9hdDMy1EpDYcpKxFbJPWtpbmQ6ICJjYXQixzggIG1lb3flAMbKTA%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

xirzec:
## Summary

A measurement-first effort to cut the runtime of the `@azure-tools/typespec-ts` test suites **without lowering coverage**. Work landed in low-churn phases, each validated against a baseline before committing. All test counts are preserved (identical tests run; coverage unchanged), and the e2e baseline tree (`src/index.d.ts` per spec) remains byte-for-byte identical.

### `unit-modular` suite (the headline win)

| Stage | Wall clock | Heap / worker |
| --- | --- | --- |
| Baseline (monolithic, effectively serial) | **~1822 s (~30 min)** | 8 GB |
| After all changes (split + cache + trimmed imports + `createTester`) | **~180 s (~3 min)** | 1 GB |

**≈ 10× faster wall clock and 8× less memory per worker**, with the same tests.

What got us there:

- **Shared compiler `createTester` (loads libraries once per worker).** The old `createTestHost` re-globbed and re-read all ~10 TypeSpec libraries (`.tsp`/`.json`) from disk on **every** compile. The tester loads them once per worker and clones an in-memory filesystem per compile, so the dominant library-load cost is amortized across the distinct compiles (complementary to the per-scenario cache, which only dedupes *identical* recompiles). Measured with a clean same-environment A/B (only the test host differs): the isolated `operations` suite dropped **~160 s → ~81 s (~49%)** and the full `unit-modular` suite **~320 s → ~180 s (~44%, ≈1.8×)**. Suggested by @timotheeguerin in review.
- **Parallelized the suite.** The single monolithic `scenarios.test.ts` was split into **39 per-directory suites**, so vitest's `forks` pool can actually run them concurrently (the old single file pinned everything to ~1 worker). This was the dominant win of the original phases (~5.25× on its own).
- **Compile-once-per-scenario cache.** Many `it` blocks recompiled identical TypeSpec inputs; a per-scenario content-keyed compile cache eliminates **~20–25% of compiles** (measured 38.7% of compiles were exact duplicates). Stats are gated behind `COMPILE_CACHE_STATS` (off by default → zero overhead).
- **Slashed the per-worker heap from 8 GB → 1 GB.** The 8 GB `--max-old-space-size` was vestigial (dated from when the suite was one giant file). With the split, no worker needs more than ~1 GB, and lower heap lets more workers run concurrently. Two module-level emitter caches were also converted to `WeakMap` to reduce cross-call retention.
- **Trimmed per-test library imports.** `rlcEmitterFor` / `compileTypeSpecFor` loaded ~10 TypeSpec libraries per test plus a heavy prefix; narrowing those to what each test actually uses makes every compile cheaper. (Unused `import`+`using` can't change emit output, so coverage is unaffected.)
- **Removed two stale skipped tests** (a `void` failure that no longer occurs, and a test for the closed/won't-fix microsoft/typespec#2411) so the suite reflects current behavior.

A guard test was added so a new scenario directory can never be silently dropped from the split (it fails CI prompting a regen).

### Integration (spector) e2e path

- **Overlap declaration regen with the test run.** Generation is split into a `client` phase (the `src/*.ts` sources the tests import) and a `declarations` phase (the tracked `src/index.d.ts` baseline via api-extractor). Since only `check:tree` consumes the declarations, the e2e pipeline now runs the vitest suite **in parallel** with declaration regen instead of serially — moving the slow api-extractor work off the test critical path.
- **Dropped the `npx tsp` boot tax.** Each spec paid a full `npx` re-resolution (~4 s) on top of compile; switching to a once-resolved direct `node <@typespec/compiler>/cmd/tsp.js compile` cuts a small-spec client compile from **~9.3 s → ~6.1 s** while keeping one fresh subprocess per compile (no leak regression).
- Estimated e2e generation+test wall clock: **~1854 s → ~1284 s (~31%)** on a 12-core box (overlap-bound; realized savings vary with core/memory pressure).

## Developer-experience / cross-platform cleanup

- **Cross-platform `copy:typespec`.** Replaced the Unix-only `rm`/`cp` shell script with a `fs/promises`-based Node TS script (`test/commands/copy-typespec.ts`) — runs identically on Windows/macOS/Linux. Removes the prior Windows manual-workaround gotcha.
- **Removed the `tsx` dependency.** All `test/commands/*` scripts now run directly on `node` (native TypeScript type stripping, Node ≥ 22.18). Required explicit `.ts` import specifiers and `import type` for type-only CJS imports.
- **Removed dead plumbing.** The `--tag`/`mode` switch in `gen-spector.js` (only `azure-modular` was ever valid) was deleted.
- **Updated package instructions** (`.github/instructions/typespec-ts.instructions.md`) to document the two-phase generation, the overlap pipeline, the node-based scripts, the baseline guard, and the modular unit-test compile conventions (shared `createTester`, import trimming, compile cache).

## Validation

- Baseline tree (`src/index.d.ts` per spec) regenerates **byte-for-byte identical** to `main`.
- Typecheck, lint, and prettier are clean.
- Test counts unchanged (652 tests across 42 files) → no coverage loss.

> Note: wall-clock figures are ±~50 s noisy due to local contention; treat them as directional. The `createTester` A/B above is reported as a **same-environment** comparison (both arms measured back-to-back on the same machine and core build, differing only in the test host) because a `main` merge mid-PR bumped vitest (4.1.7 → 4.1.9), which shifted absolute baselines. A `chronus` changeset (`internal`) is included.


packages/typespec-ts/test/util/test-util.ts
```diff
@@ -95,21 +182,30 @@ export async function rlcEmitterFor(
...
+  const needVersioning = withVersionedApiVersion || VERSIONING_USAGE.test(code);
+  const needXml = XML_USAGE.test(code);
   const content = withRawContent
     ? code
     : `
 import "@typespec/http";
-import "@typespec/rest";
-import "@typespec/versioning";
-import "@typespec/xml";
+${needRest ? 'import "@typespec/rest";' : ""}
```
timotheeguerin, line 215: is the change that there is only now a single compilation per worker? if not you might get some additional performance gain by using the new(not so new anymore) Tester framework which will cache a lot of the loaded files.

xirzec: To answer your question, in Copilot's words: 

> Good question — no, it's not one compile per worker. This PR does two things: splits the monolithic `scenarios.test.ts` into 39 per-directory suites so vitest parallelizes across worker processes, and adds a per-scenario `compileCache` that only dedupes *identical* recompiles within a scenario (~23% of compiles). Each worker still runs many *distinct* compiles, and each currently goes through `createTestHost`, which re-globs and re-reads all ~10 libraries from disk on every call.

But yes, the Tester framework does seem to help, we tried it with one suite and got a ~34% speedup. Copilot really wants to do this in a follow-up, but I think I can persuade it to merge the work into this PR. 😉

xirzec: @timotheeguerin this change has been pushed up, looks like we are sitting at a 10x speedup now

timotheeguerin (approved):

PR #4714, opened 2026-06-25
Parallelize typespec-ts scenarios via vitest projects (no generated test files)

timotheeguerin:
## Summary

Alternative to the **generated scenario test files** introduced in #4701, built on top of
that branch. It keeps the same parallelism win but **without generating/committing test
code**.

#4701 splits the monolithic `scenarios.test.ts` into 39 generated thin `.test.ts` files so
vitest's `forks` pool can parallelize at file granularity. To keep them honest it also adds
a generator (`gen-scenario-suites.ts` / `npm run gen:scenario-suites`) and a drift **guard
test** that fails CI when the generated files get out of sync with the scenario tree.

This PR replaces that mechanism with **config-computed vitest projects**: `vitest.config.ts`
is real code, so it walks the scenario directory tree at config-load time and emits **one
project per leaf scenario directory**. Each project includes a single shared
`scenario-suite.entry.ts` and injects which directory it owns via `provide` / `inject`.

## What changes

- **`vitest.config.ts`** — derive one `unit-modular:<dir>` project per leaf scenario
  directory (dir that directly contains `.md`) from the tree, plus a `unit-modular` project
  for the remaining modular-unit tests (`diagnostics`, `static/url-template`).
- **`test/modular-unit/scenario-suite.entry.ts`** (new) — reads `inject("scenarioDir")` and
  calls `describeScenarioDir(...)`. No `.test.` in the name, so the general glob never
  collects it; only the per-dir projects include it explicitly.
- **Removed**: the 39 generated `scenario-suites/*.test.ts`, `scenario-suites.guard.test.ts`,
  `test/commands/gen-scenario-suites.ts`, the `gen:scenario-suites` npm script, and the
  now-unused `getLeafScenarioDirs` export in `scenario-runner.ts`.
- **`unit-test:modular`** now runs `--project 'unit-modular*'` to match the per-dir projects.
- Migrated the pool config to the **Vitest 4** top-level `execArgv` option (`poolOptions` was
  removed in Vitest 4; the previous config emitted a deprecation warning).
- Updated the typespec-ts instructions + the changeset wording.

## Why

- **No generated test code to commit or regenerate**, and **no guard test** — the project
  list is derived live from the directory tree, so it can't drift out of sync.
- Same `forks`-pool, file-level parallelism as #4701 (each project = one schedulable unit).

## Validation

- `vitest run --project 'unit-modular*'`: **41 files / 651 tests pass**; wall clock ~2 min vs
  ~6m18s user time → parallelism confirmed.
- `eslint` clean, `prettier --check` clean, `tsc` build clean.

## Trade-offs / open questions

- 39 projects vs 39 files: per-project startup + reporter grouping differ slightly (suite
  output is grouped per project). Worth a side-by-side wall-clock check on a CI-sized box.
- All scenario projects include the *same* physical entry file (tasks are keyed by
  project+file) — unusual but works and is validated above.

> Targets the #4701 branch so the diff is purely the generation→projects swap. Draft for
> discussion.


timotheeguerin:
## Measured performance (`pnpm unit-test:modular`)

Same machine (12 logical cores), warm runs, `vitest run` wall clock. Numbers are noisy
(±~15 s, shared machine) so treat as directional.

| State | What | Wall clock | Parallelism (user/real) | Tests |
| --- | --- | --- | --- | --- |
| 1. `main` | monolithic `scenarios.test.ts`, 8 GB heap | **~190–217 s** | ~1.2× (serial-bound) | 652 (+2 skipped) |
| 2. #4701 | 39 generated `.test.ts` files, 1 GB heap | **~46–66 s** (median ~65 s) | ~5–6× | 652 (incl. 1 guard test) |
| 3. this PR | per-leaf-dir vitest **projects**, 1 GB heap | **~81–107 s** (median ~90 s) | ~4–4.5× | 651 |

(Test parity holds: this PR's 651 = #4701's 652 minus the removed guard test.)

### Honest trade-off: this approach is ~1.4× slower than #4701

Both are far faster than `main` (#4701 ~3.1×, this PR ~2.2×), but the **projects** approach
is measurably slower than the **generated-files** approach. Root cause is in the run
breakdown: `import` time jumps from **~55–78 s** (#4701) to **~275–332 s** (this PR), and
`transform` roughly 2–3×.

Why: with 39 files in **one** project, the heavy emitter module graph
(`scenario-runner` → `emit-util` → emitter `src/**`) is transformed and evaluated once per
worker and **amortised across all files in that project**. With one **project per
directory**, every project is isolated, so each one re-transforms and re-evaluates that
whole module graph from cold — ~40× duplicated module evaluation.

So this PR buys "no generated test code / no guard test / can't drift" at the cost of ~1.4×
wall clock vs #4701. Possible mitigations worth exploring if we want both: fewer/coarser
project groupings (e.g. one project per top-level scenario group instead of per leaf dir),
or sharing the emitter module graph across projects.

PR #4718, opened 2026-06-26
[typespec-ts] Fix sample generation for grouped and nested method parameters

### LINKED ISSUE #4649: JS emitter sample generator does not respect @@override parameter groups — generated samples fail to compile
### Description

The TypeSpec JS emitter (`@azure-tools/typespec-ts`) generates samples that do not respect `@@override` parameter grouping. The **type generator** correctly produces `OptionalParams` types with a nested `queryOptions` property (from the `@@override` operation), but the **sample generator** puts the query parameters directly on the `OptionalParams` object, causing TS2353 compilation errors.

### Reproduction

This is reproducible from the **main branch** of `azure-rest-api-specs` with the `specification/policyinsights` TypeSpec spec. No PR changes are needed to trigger it - any PR touching the policyinsights spec triggers SDK regeneration and exposes the bug.

**Steps:**
1. Generate the JS SDK from `specification/policyinsights/resource-manager/Microsoft.PolicyInsights/PolicyInsights/tspconfig.yaml` using `@azure-tools/typespec-ts`
2. Build the generated package (succeeds)
3. Build the generated samples with `tsc -p config/tsconfig.samples.json` (fails)

### Root Cause

The spec uses `@@override` to group query parameters into a `queryOptions` model. For example, in `client.tsp`:

```typespec
op policyEventsListQueryResultsForSubscription(
  ...ApiVersionParameter,
  ...SubscriptionIdParameter,
  @bodyRoot body: void,
  @path @key @segment("subscription") name: string,
  ...ProviderNamespace<SubscriptionActionScope>,
  @path @segment("policyEvents") policyEventsResource: PolicyEventsResourceType,
  queryOptions?: PolicyEventsListQueryResultsForSubscriptionQueryOptions,
): PolicyEventsQueryResults | QueryFailure;

@@override(
  PolicyEventsOperationGroup.listQueryResultsForSubscription,
  policyEventsListQueryResultsForSubscription,
  "!java,!python"
);
```

The **type generator** correctly produces:
```typescript
export interface PolicyEventsListQueryResultsForSubscriptionOptionalParams extends OperationOptions {
  queryOptions?: PolicyEventsListQueryResultsForSubscriptionQueryOptions;
}
```

But the **sample generator** ignores the override and produces code using the raw (pre-override) parameters:
```typescript
// Generated sample - fails with TS2353
const options = {
  fromParam: new Date("2018-02-05T18:00:00Z"),   // does not exist on OptionalParams
  filter: "PolicyDefinitionAction eq 'deny'",     // does not exist on OptionalParams
};
```

The correct sample should be:
```typescript
// Expected
const options = {
  queryOptions: {
    fromParam: new Date("2018-02-05T18:00:00Z"),
    filter: "PolicyDefinitionAction eq 'deny'",
  },
};
```

### Error Output

All operations with `queryOptions` parameter groups are affected (~40 errors). Example errors:

```
samples-dev/policyEventsListQueryResultsForSubscriptionSample.ts(21,7): error TS2353: 'fromParam' does not exist in type 'PolicyEventsListQueryResultsForSubscriptionOptionalParams'.
samples-dev/policyStatesListQueryResultsForResourceSample.ts(112,7): error TS2353: 'expand' does not exist in type 'PolicyStatesListQueryResultsForResourceOptionalParams'.
samples-dev/attestationsListForResourceGroupSample.ts(37,5): error TS2353: 'top' does not exist in type 'AttestationsListForResourceGroupOptionalParams'.
```

### Impact

This blocks **any** spec PR touching `specification/policyinsights` from passing the `SDK Validation - JS` CI check. See: https://github.com/Azure/azure-rest-api-specs/pull/43715/checks?check_run_id=81354360613

### Environment

- Spec repo: `azure-rest-api-specs` (main branch)
- Spec path: `specification/policyinsights/resource-manager/Microsoft.PolicyInsights/PolicyInsights`
- Emitter: `@azure-tools/typespec-ts`

JialinHuang803:
## Description

Fixes sample generation for grouped and nested method parameters.

When a method parameter is model-typed (e.g. an `@@override` that groups query params into `queryOptions?: QueryOptions`, a directly-defined grouped model parameter, or body-extracted query/header params), the sample generator previously emitted those parameters **flat**, producing TypeScript compile errors in the generated `*Sample.ts` files.

## Fix

`prepareExampleParameters` now resolves every HTTP parameter to its method parameter through `methodParameterSegments` — the same source of truth the operation generator uses — and places each example value at its path. This yields exactly one argument per method parameter, regardless of whether the parameter is flat, nested, or grouped, without special-casing any of those shapes:

- `path.length === 1` -> scalar positional argument
- longer path -> value nested under its root (e.g. `{ queryOptions: { top, filter, orderBy } }`)
- the root segment's `optional` flag decides positional vs. options-bag placement

A `bodyRootName` guard skips the root that the dedicated body section already emits, preventing double-emission when a query/header parameter shares the body's model root.

## Tests

Added `test/modular-unit/scenarios/samples/parameters/groupedParameter.md` covering three structural outcomes:
1. flat + optional grouped together -> `client.list("books", { queryOptions: { ... } })`
2. optional grouped only -> `client.list({ queryOptions: { ... } })`
3. required grouped only -> `client.list({ top, filter, orderBy })`

All 59 samples scenarios pass.

Fixes #4649

### DIFF
--- a/.chronus/changes/fix-sample-grouped-parameters-2026-6-26-9-40-28.md
+++ b/.chronus/changes/fix-sample-grouped-parameters-2026-6-26-9-40-28.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-ts"
+---
+
+[typespec-ts] Fix sample generation for grouped and nested method parameters by resolving every HTTP parameter through `methodParameterSegments`, so flat, nested, and grouped parameters are emitted correctly without special casing.
--- a/packages/typespec-ts/src/modular/emit-samples.ts
+++ b/packages/typespec-ts/src/modular/emit-samples.ts
@@ -316,12 +316,36 @@ function prepareExampleParameters(
 
   let subscriptionIdValue = `"00000000-0000-0000-0000-000000000000"`;
   let isSubscriptionIdAdded = false;
-  // required parameters
+
+  // Map each HTTP parameter to its method parameter via `methodParameterSegments`
+  // and place its example value at that path, yielding one argument per method
+  // parameter regardless of flat, nested, or grouped shape. Client- and
+  // body-carried roots are emitted by their own sections and skipped here.
+  const bodyParam = method.operation.bodyParam;
+  let bodyRootName: string | undefined;
+  if (
+    bodyParam &&
+    !isSpreadBodyParameter(bodyParam) &&
+    bodyParam.methodParameterSegments.length === 1 &&
+    bodyParam.methodParameterSegments[0] !== undefined &&
+    bodyParam.methodParameterSegments[0].length >= 1
+  ) {
+    bodyRootName = bodyParam.methodParameterSegments[0][0]!.name;
+  }
+
+  const methodArguments = new Map<
+    string,
+    { name: string; isOptional: boolean; value: ValueTreeNode | string }
+  >();
   for (const param of method.operation.parameters) {
-    if (param.optional === true || param.type.kind === "constant" || param.clientDefaultValue) {
+    if (param.type.kind === "constant" || param.clientDefaultValue) {
       continue;
     }
-
+    const path = param.methodParameterSegments[0];
+    if (path === undefined || path.length === 0) {
+      continue;
+    }
+    const root = path[0]!;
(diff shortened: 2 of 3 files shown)


qiaozha (approved):

qiaozha:
wonder why the benchmark result increases a lot?

PR #4591, opened 2026-06-09
[Java] Move `@azure-tools/typespec-java` to `typespec-azure` repo 

srnagar:

PR #4721, opened 2026-06-26
[typespec-java] flatten typespec-java, cleanup autorest related code

XiaofeiCao:
## Summary

Builds on top of #4591 (which moved `@azure-tools/typespec-java` into this repo). This PR overrides that structure with the source-copy approach (only `options.ts` is authored here; the rest is copied from `core/` at build time), flattens the package layout, and removes autorest.java-specific tooling.

## Layout (`packages/typespec-java/`)

- **(root)** — the published `@azure-tools/typespec-java` package. Only `src/options.ts` is committed; the rest of the emitter TS is copied from `core/` at build time (`Copy-Sources.ps1`). `Build-TypeSpec.ps1` applies `core.patch` (Azure customization), builds `emitter.jar` from core, then packs.
- **`emitter-tests/`** — standalone npm + Maven Spector suite (excluded from the pnpm workspace; consumes the packed `.tgz`). Hand-written JUnit + `tsp/` specs synced from core via `SyncTests.ps1`; `customization/` is hand-maintained.
- **`docs/`** — developer/user docs ported from #4591.

## Key points

- **Workspace + catalog:** the emitter is a normal workspace package using `workspace:^` (in-repo/core deps) and `catalog:` (external), like the other emitters.
- **Java 11:** the emitter only requires JDK 11; `mise.toml` uses `microsoft-11` + `maven 3.9.16`.
- **CI:** `.github/workflows/ci-java.yml` builds + runs unit and Spector tests; `nightly-e2e.yml` adds a Java matrix entry; ADO `build-for-publish.yml` wires Spector coverage upload.
- **SDK regeneration pipeline removed** — `post-publish-sdk.yaml` / `sync_sdk.py` / patches will live in the SDK repo instead.


eng/pipelines/jobs/build-for-publish.yml
```diff
@@ -41,6 +41,31 @@ jobs:
...
 
+      # Java e2e tests and spector coverage upload. The emitter-tests project is a
+      # standalone npm+Maven project (not in the pnpm workspace). Generate.ps1
+      # builds the emitter jar + tgz (Build-TypeSpec.ps1), npm installs it and
+      # regenerates the SDK from the spector specs; Spector-Tests.ps1 starts the
+      # mock server, runs the JUnit tests and produces tsp-spector-coverage-java.json.
+      - script: pnpm turbo run --filter "@azure-tools/typespec-java^..." build
+        displayName: "Build typespec-java dependencies"
+
+      - pwsh: ./Generate.ps1
```
XiaofeiCao: To speedup release process, we probably only need to do `SyncTests.ps1`.

mise.toml
```diff
@@ -1,3 +1,5 @@
 [tools]
 python = "3.12"
 uv = "latest"
+java = "microsoft-11"
+maven = "latest"
```
timotheeguerin: i know uv use latest, but we should try to pick specific versions for security and repeatibility

packages/typespec-java/package.json
```diff
@@ -0,0 +1,92 @@
...
+      "default": "./dist/src/index.js"
+    }
+  },
+  "engines": {
+    "node": ">=20.0.0"
+  },
+  "scripts": {
+    "clean": "rimraf ./dist ./temp ./generator",
+    "build": "pwsh -NoProfile -File eng/scripts/Copy-Sources.ps1 && tsc -p .",
+    "build:full": "pwsh -NoProfile -File ./Build-TypeSpec.ps1",
```
timotheeguerin: I think ideally build should be doing the full build, but we can optimize and reorganize later too when all emitters are there and we found some pain points

XiaofeiCao: My initial concern is that the full build would require `mvn` and `java` present in environment. When doing full build of the typespec-azure repo, non-java dev would fail. Hence the "partial" build without compilation.

Though since we are now mostly doing turbo build, guess it's safe?

tsconfig.ws.json
```diff
@@ -10,8 +10,7 @@
     { "path": "packages/azure-http-specs/tsconfig.build.json" },
     { "path": "packages/typespec-metadata/tsconfig.build.json" },
     { "path": "packages/samples/tsconfig.build.json" },
-    { "path": "packages/typespec-python/tsconfig.json" },
-    { "path": "packages/typespec-java/typespec-extension/tsconfig.json" }
+    { "path": "packages/typespec-python/tsconfig.json" }
```
XiaofeiCao: This seem not needed in tsconfig.ws.json. Since the source is copied from core, there's no need to watch the source.

packages/typespec-java/package.json
timotheeguerin: this is why we have mise setup so anyone can just run one time `mise setup` and it will install all the tools required. What I do not want is it being a wack a mole to fix all packages because every single one has it sown special way of being built. 
If you want to do that as a follow up that is fine but Ideally i want to run `pnpm build` and everything in the repo is built. Turborepo will cache output and we can optimize to make the experience fast when working on a specific area

eng/config/labels.ts
timotheeguerin, line 1: think you can revert those files now, just reordering things

pnpm-workspace.yaml
```diff
@@ -34,13 +32,12 @@ catalog:
...
   "@astrojs/check": ^0.9.8
-  "@autorest/codemodel": ~4.20.1
   "@astrojs/react": ^6.0.0
   "@astrojs/starlight": ^0.40.0
+  "@autorest/codemodel": ~4.20.1
   "@azure-rest/core-client": ^2.3.1
   "@azure-tools/codegen": ~2.10.1
-  "@azure-tools/openai-typespec": "1.11.0"
-  "@azure-tools/typespec-liftr-base": "0.14.0"
+  "@azure-tools/typespec-liftr-base": ">=0.14.0 <1.0.0"
```
timotheeguerin, line 41: is this one also necessary?

pnpm-workspace.yaml
```diff
@@ -157,7 +152,7 @@ catalog:
   happy-dom: ^20.9.0
   hast-util-to-html: ^9.0.5
   is-unicode-supported: ^2.1.0
-  js-yaml: ~4.1.1
+  js-yaml: ~4.2.0
```
timotheeguerin: upgrade of this expected in this pr?

timotheeguerin: ideally should migrate to `yaml` package we use across all other packages

.chronus/changes/add-typespec-java-2026-6-25-0-0-0.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: internal
+packages:
+  - "@azure-tools/typespec-java"
+---
+
+[Java] Flatten the `typespec-java` emitter folder structure and clean up autorest.java-specific tooling.
```
timotheeguerin, line 7: doesn't matter because its internal but you shouldn't add the package in the scope here or it will get included in the changelog for that package

.github/workflows/ci-java.yml
```diff
@@ -8,35 +8,82 @@ on:
...
+        run: pnpm run test
+        working-directory: packages/typespec-java
+
+  e2e:
+    name: "Regen & Spector Tests"
     needs: build
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
```
timotheeguerin: I think v7 is the latest now

packages/typespec-java/emitter-tests/.vscode/launch.json
weidongxu-microsoft, line 1: we should not include this folder?

packages/typespec-java/CONTRIBUTING.md
weidongxu-microsoft, line 1: align the common part of this file to other emitter modules

if other emitter does not have this fine, we probably only need to extract part of it for agent instructions.

packages/typespec-java/LICENSE
weidongxu-microsoft, line 1: Do we need this license file?

packages/typespec-java/CHANGELOG.md
weidongxu-microsoft, line 1: I assume later we will use the `pnpm change` as the changelogs?

We can still keep it, but probably no longer need to add.

tsconfig.ws.json
weidongxu-microsoft: Maybe useful when we debug at this repo?

mise.toml
XiaofeiCao: Srikanta updated to use 3.9.6.

packages/typespec-java/package.json
XiaofeiCao: Agree. Updated `build` to build the full emitter. I disabled cache for typespec-java turbo build, since we rely on the compiled jar from core(typespec).

tsconfig.ws.json
XiaofeiCao: Sure, added.

eng/config/labels.ts
XiaofeiCao: done

pnpm-workspace.yaml
XiaofeiCao: currently yes. when we do `tsp-client generate-config-files` in SDK repository, it takes emitter's package.json as argument, hence we need its(along with `openai-typespec`) appearance.

quick question: for versions in catalog, who's able to update them?

pnpm-workspace.yaml
XiaofeiCao: yeah, it's to align with the one used in http-client-java: https://github.com/microsoft/typespec/blob/main/packages/http-client-java/package.json#L68

I'll do a follow-up clean-up for this dependency, along with `lodash`, there.

.github/workflows/ci-java.yml
XiaofeiCao: thanks, updated!

packages/typespec-java/emitter-tests/.vscode/launch.json
XiaofeiCao: removed

packages/typespec-java/CONTRIBUTING.md
XiaofeiCao: python and ts have this file. I aligned it with python's. Also updated the release part be TODO.

packages/typespec-java/LICENSE
XiaofeiCao: emm, no, removed

packages/typespec-java/CHANGELOG.md
XiaofeiCao: yes, no need for manual update. keep it here for past changelogs.

.chronus/changes/add-typespec-java-2026-6-25-0-0-0.md
XiaofeiCao: could you elaborate? I've tried to remove the package, CI will complain: https://github.com/Azure/typespec-azure/actions/runs/28504252500/job/84488633473?pr=4721

or other than adding new one here, I should update existing one Srikanta added instead?

packages/typespec-java/package.json
```diff
@@ -65,7 +65,8 @@
     "@autorest/codemodel": "catalog:",
     "@azure-tools/codegen": "catalog:",
     "js-yaml": "catalog:",
-    "lodash": "catalog:"
+    "lodash": "catalog:",
+    "yaml": "catalog:"
```
XiaofeiCao, line 69: This is preparation for future microsoft/typespec switch of js-yaml.

Otherwise, typespec-azure integration would fail there: https://github.com/microsoft/typespec/actions/runs/28558818690/job/84672023268?pr=11134

pnpm-workspace.yaml
```diff
@@ -157,7 +153,7 @@ catalog:
   happy-dom: ^20.9.0
   hast-util-to-html: ^9.0.5
   is-unicode-supported: ^2.1.0
-  js-yaml: ~4.1.1
+  js-yaml: ^4.2.0
```
timotheeguerin, line 145: do we still need that now that you migrated?

eng/pipelines/jobs/build-for-publish.yml
```diff
@@ -41,6 +41,30 @@ jobs:
           PackagePath: packages/typespec-python
           GeneratorName: "@azure-tools/typespec-python"
 
+      # Java e2e tests and spector coverage upload. The emitter-tests project is a
+      # standalone npm+Maven project (not in the pnpm workspace). Generate.ps1
+      # builds the emitter jar + tgz (Build-TypeSpec.ps1), npm installs it and
+      # regenerates the SDK from the spector specs; Spector-Tests.ps1 starts the
+      # mock server, runs the JUnit tests and produces tsp-spector-coverage-java.json.
+      - script: pnpm turbo run --filter "@azure-tools/typespec-java^..." build
```
timotheeguerin: isn't it already built above?

pnpm-workspace.yaml
XiaofeiCao: we still need js-yaml before we actually replaced `js-yaml` in core(microsoft/typespec): https://github.com/microsoft/typespec/pull/11134
also, that PR needs the new `yaml` be in main, otherwise would fail typespec-azure integration CI..

our `http-client-java` on main uses `^4.2.0`. align it here for now.

eng/pipelines/jobs/build-for-publish.yml
XiaofeiCao: Nice catch. `pnpm build` should already build everything. Removed it and added java install in install.yaml.

pnpm-workspace.yaml
XiaofeiCao: One note: our Java generator depends on `snakeyaml`, which conforms to YAML-1.1 (same as original `js-yaml`) and handles serialization/deserialization slightly differently than YAML-1.2. I force using 1.1 syntax there: https://github.com/microsoft/typespec/pull/11134/changes#diff-9b551a4702889299c21829d1d75dd091223f39e6a5ae14e3bda4b4fcc887bbf6R83

Alternative be using `snakeyaml-engine` which is YAML-1.2 compatible. But it may involve way more code change.

timotheeguerin: is it because you are coping the sources?

>also, that PR needs the new yaml be in main, otherwise would fail typespec-azure integration CI..
that is fine, that check can fail if you are ready to fix the upstream part quickly after, it is not required and is more of a oh you are breaking typespec-azure make sure you have a pr ready to fix it

timotheeguerin: fine anyway for now if you want to merge like that

timotheeguerin (approved):

PR #4494, opened 2026-05-26
Improve documentation for ARM library decorators, interfaces, and models

tadelesh:
## Summary

Adds clear documentation to decorators, template interfaces, and models in the ARM library that were missing docs or had unclear/incorrect documentation.

### Changes

**lib/decorators.tsp**
- Fixed typo in @armLibraryNamespace
- Added descriptions to @armResourceAction, @armResourceCreateOrUpdate, @armResourceRead, @armResourceUpdate, @armResourceDelete, @armResourceList
- Expanded @armResourceCollectionAction description
- Fixed @armProviderNameValue wrong description
- Removed ghost @param tags from @armProviderNamespace and @armVirtualResource
- Fixed @resourceBaseType param name to match implementation

**lib/interfaces.tsp**
- Fixed misleading doc on ResourceDeleteWithoutOkAsync

**lib/models.tsp**
- Improved docs for ResourceListCustomResult

**lib/legacy-types/**
- Fixed empty @doc

PR #4788, opened 2026-06-30
Run typespec-ts unit tests in the repo-wide vitest run

xirzec:
## What

Removes the `typespec-ts` exclusion from the root `vitest.config.ts` / `vitest.config.fast.ts` so the repo-wide vitest run now includes the typespec-ts **unit** tests (`test-next` + `unit-modular`), while keeping the end-to-end **Spector** integration project (`integration-azure-modular`) out of that run.

## Why it wasn't this simple

Two things had to be solved:

1. **Vitest flattens but ignores nested projects.** When the root config references `packages/typespec-ts/vitest.config.ts` as a project, vitest 4 ignores that file's nested `test.projects` and instead uses its top-level `test.include`. So I added a top-level `include` listing only the unit globs, plus the `unit-modular` pool/timeout settings (`testTimeout: 0`, forked pool, 1 GB heap) at the top level. The three named projects remain for local `--project` selection (`pnpm test-next`, `pnpm unit-test`, `pnpm integration-test:alone`).

2. **The real reason for the original exclusion:** the generated `modular-unit` scenario suites resolved scenario directories relative to `process.cwd()`, so they only worked when cwd was `packages/typespec-ts`. They now resolve against the package root (via `import.meta.url`) in `scenario-runner.ts` and `scenario-suites.guard.test.ts`, making them cwd-independent.

## Validation

- Repo-wide-style run: 57 files / 909 unit tests pass, **zero** integration tests collected.
- Real `vitest.config.fast.ts` runs the typespec-ts modular-unit suite (657 tests) — all pass.
- Local `pnpm test-next` (252) and `pnpm unit-test` (657) still pass; the integration project still resolves for `pnpm integration-test:alone`.
- `gen:scenario-suites` output unchanged; lint, `tsc`, prettier, and chronus verify all pass.

The dedicated `ci-typescript.yml` workflow is intentionally left as-is (fast path-filtered unit feedback + the e2e Spector job).


.chronus/changes/typespec-ts-unit-tests-repo-wide-2026-06-30-16-50-00.md
timotheeguerin, line 1: Hhm it does seem like this unfortunately adds quite a bit of time to the CI no?

Before

Image


This pr

Image


which tests are those?

I do 100% want to find a shared solution across all our emitters here so we can run everything with common commands and its not a scavenger hunt when you touch multiple packages

xirzec: Ah yeah, looks like this is still pretty heavy. As an alternative, I'm going to add only our "test-next" tests (which take 14s locally to run) and as a compromise to avoid breaks from TCGC expand the TS-emitter workflow to run on tcgc changes specifically

xirzec: of course, a different PR (#4494) managed to break our tests... addressing that in a separate PR

xirzec: seems better now? 
image

timotheeguerin (approved):

PR #4793, opened 2026-07-01
Document resolveArmResources design proposal

ArcturusZhang:
## Summary
- Add a proposal for a version-aware `resolveArmResources` design
- Add linter candidates needed to support trusting registered ARM operation kinds

## Current implementation mismatches

| Area | Current implementation | Proposed design mismatch |
| --- | --- | --- |
| Version-aware output | Returns a single `Provider` and caches by `Namespace -> Provider`. | Proposal expects explicit provider/version dimensions with one result per service version. |
| Provider discovery | `resolveProviderNamespace` returns the first namespace with ARM provider state. | Proposal shape allows multiple providers; current API effectively resolves one. |
| Resource detection source | Any registered ARM operation can create a `ResolvedResource`. | Proposal says only `Read` and `CreateOrUpdate` paths should create concrete resources. |
| List/action can create resources | `list` and `action` operations can initialize a new resource group. | Proposal says list/action should only append to already detected resources. |
| No separate append pass | Operations are grouped while resources are being discovered. | Proposal separates detection from appending other lifecycle/list/action operations. |
| Unmatched registered operations | Registered ARM operations that cannot resolve resource info are skipped and also excluded from `providerOperations`. | Proposal says operations that cannot be associated with detected resources should remain available for provider-level/non-resource handling. |
| Associated operations still exist | `ResolvedResource` includes `associatedOperations`, and `other` operations are attached there. | Proposal removed `associatedOperations`. |
| Path identity is operation-derived | Resource identity is derived from each operation path. | Proposal says canonical resource identity should be derived from grouped `Read`/`CreateOrUpdate` instance paths. |
| Path parser is permissive for non-instance paths | List operations can synthesize instance paths by appending `{name}`. | Proposal wants candidate paths validated/parsed as ARM resource instance paths for detection. |
| Normalized comparison incomplete | Resource matching normalizes some provider/type/scope casing, but parent matching is case-sensitive and full path variable names are not consistently ignored. | Proposal requires normalized comparison whenever comparing resource paths/identities. |
| Parent handling | Missing parents are synthesized as empty `ResolvedResource` records. | Proposal says skip missing parents instead of synthesizing empty parent resources. |
| Scope handling | Scope may be `ExternalResource` or a synthesized `ResolvedResource`. | Proposal removed `ExternalResource` and wants structured references only when concrete. |
| Parent/scope reference shape | `parent` and resource-valued `scope` are full `ResolvedResource` objects. | Proposal uses lightweight `ResolvedResourceReference`. |
| Singleton detection | Singleton metadata comes from `@singleton` state and model key-property union. | Proposal says singleton metadata should be path-based from literal or closed-literal name segments. |
| Lifecycle arrays optional/undefined | Lifecycle buckets are initialized with `undefined` values. | Proposal shape uses arrays for every lifecycle bucket. |
| Resource model detail naming remains current | `listArmResources` returns `ArmResourceDetails`, deduped by qualified TypeSpec model name. | Proposal notes this is really candidate resource model detail; implementation still uses the old model and may collapse version/projected distinctions. |
| Protocol/response linters not part of implementation | Current decorators register operation kinds but resolver does not enforce the proposed linter contracts. | Proposal relies on new linters for lifecycle verb consistency, read/create response shape, and list item shape. |

## Related issue: customAzureResource-based converted specs

Opened Azure/typespec-azure#4798 for a real-world mismatch found while comparing legacy/Swagger-shaped resource detection with `resolveArmResources` output.

Network is the main case: the .NET management emitter comparison found 140 legacy-detected resources, but `resolveArmResources` returns 0 resources and 776 non-resource methods for `specification/network/resource-manager/Microsoft.Network/Network`. The spec uses legacy/custom resource bases marked with `@Azure.ResourceManager.Legacy.customAzureResource(#{ isAzureResource: true })`, which means the current TypeSpec-based detection path can intentionally diverge from Swagger-shaped legacy detection.

This needs either a spec-side resolution, if Network can be migrated to the supported `Azure.ResourceManager.Legacy.CustomAzureResource<true>` pattern and future converted specs can be prevented from using the unsupported form, or a library-side resolution in `resolveArmResources` if this custom-resource shape must continue to be recognized.
## Related Azure SDK analysis PR

The generated comparison reports and manual root-cause notes are preserved in Azure/azure-sdk-for-net#60620.
## Validation
- Documentation-only change

### DIFF
--- a/packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
+++ b/packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
@@ -46,13 +46,13 @@ This is important because the proposed resolver can use `CreateOrUpdate` as a re
 
 Every registered lifecycle operation kind should use the expected HTTP verb:
 
-| Lifecycle kind | Expected HTTP verb |
-| --- | --- |
-| `read` | `GET` |
-| `createOrUpdate` | `PUT` |
-| `update` | `PATCH` |
-| `delete` | `DELETE` |
-| `checkExistence` | `HEAD` |
+| Lifecycle kind   | Expected HTTP verb |
+| ---------------- | ------------------ |
+| `read`           | `GET`              |
+| `createOrUpdate` | `PUT`              |
+| `update`         | `PATCH`            |
+| `delete`         | `DELETE`           |
+| `checkExistence` | `HEAD`             |
 
 This lets `resolveArmResources` attach lifecycle operations by their registered ARM operation kind without re-validating protocol shape during resource graph construction.
 
--- a/packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
+++ b/packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
@@ -1,0 +1,80 @@
+# Linter candidates for `resolveArmResources`
+
+## Why these linters are needed
+
+The proposed `resolveArmResources` design should use registered ARM operation kinds as the source of truth for resource detection. In particular, `Read` and `CreateOrUpdate` operations should define candidate resource instance paths.
+
+That design is simpler and more stable than re-inferring operation meaning from HTTP verb and response shape during resource detection. However, it requires the TypeSpec program to be internally consistent: if an operation is registered as a resource `Read`, it must actually behave like a resource read operation; if it is registered as `CreateOrUpdate`, it must actually behave like a resource create-or-update operation.
+
+These consistency checks should be enforced by linters rather than by `resolveArmResources` itself. Resource resolution should focus on building the resource graph. Linters should tell spec authors when an operation is mislabeled or malformed.
+
+## Candidate: resource read operation shape
+
+A resource `Read` operation should satisfy the resource-read contract:
+
+```ts
+@armResourceRead(ResourceModel)
+op get(...): ResourceModel;
+```
+
+Expected checks:
+
+1. The operation must use HTTP `GET`.
+2. The operation must return the candidate resource model.
+3. The operation path must be a valid ARM resource instance path.
+
+Without this linter, `resolveArmResources` would need to defensively verify every registered `Read` operation before trusting its path. That would mix validation logic into the resolver and make the resolver less deterministic.
+
+## Candidate: resource create-or-update operation shape
+
+A resource `CreateOrUpdate` operation should satisfy the resource create-or-update contract:
+
+```ts
+@armResourceCreateOrUpdate(ResourceModel)
+op createOrUpdate(...): ResourceModel;
+```
+
+Expected checks:
+
+1. The operation must use HTTP `PUT`.
+2. The operation must target the candidate resource model.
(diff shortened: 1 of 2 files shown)


packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
```diff
@@ -0,0 +1,235 @@
...
+}
+
+interface ResolvedProviderVersion {
+  version?: string;
+  resources: ResolvedResource[];
+  providerOperations: ArmResourceOperation[];
+}
+
+interface ResolvedResource {
+  type: Model;
```
RodgeFu, line 53: what' if the resource has different Model for post/get/put/...? Will the API already have helped merged them? we will need the API to expose these different models if not so that the consumers can further process them.

ArcturusZhang: this `type` here (type is not a good property name, I will change it later) represents the "resource model" of this resource. theoretically, you could consider this as the "entity model" of this ARM resource on the service side.
This is not showing any request or response types.
But some operations indeed have some rules, which I have put in the linter candidates.
For instance, read operation (get) should always return the same model as the resource model. createOrUpdate operation (put) should always return the same model as the resource model as well. In real world, createOrUpdate operation can use a different model as request.
action operations (usually post) do not have constraint on request/response type.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
```diff
@@ -0,0 +1,80 @@
+# Linter candidates for `resolveArmResources`
+
+## Why these linters are needed
```
markcowl, line 3: Note that this part of the spec is addressed in this PR: https://github.com/Azure/typespec-azure/pull/4808

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
```diff
@@ -0,0 +1,235 @@
+# Design for `resolveArmResources`
+
```
markcowl, line 2: I think this mechanism, using get as the main ,echanism for determining the resource is a good way to do this.

The only thing that's missing is a mechanism for opting into another grouping - this is currently provided by the 'ResourceName' parameter in the operation decorators.  This handles cases in which the same Resource model type is used in multiple resources (e.g. Site and SiteDeployment in Microsoft.Web)  or the pths of certain operations are nonstandard.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
```diff
@@ -0,0 +1,235 @@
...
+## Expected output
+
+`resolveArmResources(program)` should produce a complete, deterministic, version-aware description of the ARM provider shape represented by the TypeSpec program.
```
markcowl, line 5: Per discussion, I believe we decided that version should be external to this api - we should be able to pass in the projected program for a version to get the results for that version

markcowl (changes requested):

PR #4796, opened 2026-07-01
Fix stale typespec-ts modular-unit snapshot after ARM doc changes (#4795)

### LINKED ISSUE #4795: [Bug]: typespec-ts modular-unit snapshots stale after ARM doc changes in #4494 — CI red on main
### Describe the bug

`packages/typespec-ts` `unit-modular` scenario snapshots are stale on `main`, causing the **typespec-ts / CI → Run unit tests** job to fail for any PR that touches `packages/typespec-ts/**` (the `pull_request` build merges `main`, so the stale snapshots are exercised against the current ARM library).

Root cause: #4494 ("Improve documentation for ARM library decorators, interfaces, and models") changed the ARM operation templates in `packages/typespec-azure-resource-manager/lib/*.tsp` — the human-facing operation docs (e.g. `A long-running resource action.`) were moved to `@dev` and the public doc set to `@doc("")`. As a result ARM LRO operations now emit **no** public doc comment. The checked-in typespec-ts `modular-unit` scenario snapshots still expect the old doc text, so the generated output no longer matches.

Because #4494 did not touch `packages/typespec-ts/**`, `ci-typescript.yml` never ran on it, so the snapshot drift landed on `main` unnoticed. The `merge_group` runs for pr-4494 and pr-4547 show the same failure on the queue.

Example diffs from the failing job:

```
- /** A long-running resource action. */ export function backup(
+ export function backup(
```

```
- /** * This sample demonstrates how to a long-running resource action. * * @summary a long-running resource action. ...
+ /** * This sample demonstrates how to * * @summary * ...
```

Affected scenarios include (at least) the ARM samples/operations that use `ArmResourceActionAsync`-style templates, e.g. `test/modular-unit/scenarios/samples/parameters/bodyOptionalParameterName.md` and `test/modular-unit/scenarios/operations/lroPaging.md`.

Suggested fix: regenerate the affected `modular-unit` snapshots against current `main` by running the unit tests with the update flag, then commit the updated `.md` scenario files:

```
cd packages/typespec-ts
SCENARIOS_UPDATE=true pnpm unit-test
```

Broader follow-up to consider: ARM library (`typespec-azure-resource-manager`) changes can alter typespec-ts emitted output but do not currently trigger `ci-typescript.yml`, so this class of drift can recur. Consider widening the `ci-typescript.yml` `paths` triggers to include the emitter's TypeSpec library dependencies (e.g. `typespec-azure-resource-manager`, `typespec-client-generator-core`) and/or reconsidering which suites run in the repo-wide vitest run.

Context: found while working on #4788 (enabling typespec-ts unit tests in the repo-wide run). That PR is currently red solely due to this pre-existing drift, not its own changes.

### Reproduction

Not a spec/playground bug — this is a repo CI / snapshot drift issue reproducible in the repo itself:

1. Check out current `main` (which includes #4494).
2. `pnpm install` and build typespec-ts and its dependencies.
3. `cd packages/typespec-ts && pnpm unit-test`.
4. Observe `AssertionError` snapshot mismatches for ARM LRO operations/samples where the expected output contains doc comments (e.g. `A long-running resource action.`) that the current ARM library no longer emits.

### Checklist

- [x] Follow our Code of Conduct
- [x] Check that this issue is about the Azure libraries for typespec
- [x] The provided reproduction is a minimal reproducible example of the bug.

xirzec:
## Why

PR #4494 ("Improve documentation for ARM library decorators, interfaces, and models") added `@doc("")` to the ARM `ArmResourceActionAsync` operation template, so those operations no longer emit a public doc comment. Because #4494 did not touch `packages/typespec-ts/**`, `ci-typescript.yml` never ran on it, and the stale `modular-unit` scenario snapshot landed on `main` unnoticed. This made the **typespec-ts / CI -> Run unit tests** job red on `main` and on every typespec-ts PR (including #4788).

## What changed

- Updated the `bodyOptionalParameterName.md` modular-unit scenario to drop the now-empty ARM operation doc comment (`/** A long-running resource action. */`) and the corresponding sample summary text, matching what the current ARM library emits.
- Added an `internal` chronus changeset for `@azure-tools/typespec-ts`.

## Notes for reviewers

- The drift was narrower than the issue implied. Only `bodyOptionalParameterName.md` actually failed: its `backup` op uses `ArmResourceActionAsync` (which got `@doc("")`). `lroPaging.md` still passes because its `suspend` op uses `ArmResourceActionAsyncBase`, which #4494 left with a public doc, so it correctly still emits `A long-running resource action.`
- I did not commit the full `SCENARIOS_UPDATE=true` output. Write mode reformats every scenario through prettier and produced 46 changed files (+1039/-763), but 44 of those were pure formatting noise: the runner's assertion is whitespace-normalized (`ignoreWeirdLine`), so those never actually failed. I discarded the noise and hand-applied only the 3 genuine doc removals to keep the diff minimal and reviewable.

## Verification

- `pnpm unit-test` -> 657 passed
- `pnpm test-next` -> 252 passed
- `pnpm chronus verify` -> clean
- prettier `--check` on changed files -> clean

## Follow-up (out of scope)

ARM library (`typespec-azure-resource-manager`) changes can alter typespec-ts emitted output but do not trigger `ci-typescript.yml`, so this drift class can recur. Consider widening the workflow `paths` triggers to include the emitter's TypeSpec library dependencies. Tracked separately.

Fixes: #4795

### DIFF
--- a/packages/typespec-ts/test/modular-unit/scenarios/samples/parameters/bodyOptionalParameterName.md
+++ b/packages/typespec-ts/test/modular-unit/scenarios/samples/parameters/bodyOptionalParameterName.md
@@ -191,7 +191,6 @@ export async function _backupDeserialize(result: PathUncheckedResponse): Promise
   return backupResultDeserializer(result.body);
 }
 
-/** A long-running resource action. */
 export function backup(
   context: Client,
   resourceGroupName: string,
@@ -218,9 +217,9 @@ import { HardwareSecurityModulesClient } from "@azure/internal-test";
 import { DefaultAzureCredential } from "@azure/identity";
 
 /**
- * This sample demonstrates how to a long-running resource action.
+ * This sample demonstrates how to
  *
- * @summary a long-running resource action.
+ * @summary
  * x-ms-original-file: 2021-10-01-preview/json_for_CloudHsmClusters_backup.json
  */
 async function cloudHsmClustersBackup(): Promise<void> {
--- a/.chronus/changes/fix-typespec-ts-stale-modular-snapshots-2026-07-01-15-20-0.md
+++ b/.chronus/changes/fix-typespec-ts-stale-modular-snapshots-2026-07-01-15-20-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@azure-tools/typespec-ts"
+---
+
+Regenerate stale `modular-unit` scenario snapshot to match the ARM library doc changes from #4494, which moved the `ArmResourceActionAsync` operation doc to `@dev` (`@doc("")`) so those operations no longer emit a public doc comment.


packages/typespec-ts/test/modular-unit/scenarios/samples/parameters/bodyOptionalParameterName.md
```diff
@@ -218,9 +217,9 @@ import { HardwareSecurityModulesClient } from "@azure/internal-test";
 import { DefaultAzureCredential } from "@azure/identity";
 
 /**
- * This sample demonstrates how to a long-running resource action.
+ * This sample demonstrates how to
```
JialinHuang803, line 220: Our sample uses the doc of the method to generate the description. Maybe we can improve the logic of this part in the JS emitter.   https://github.com/Azure/typespec-azure/blob/cd9c4224dbd2558d2aef39d713ced9213d8b3629/packages/typespec-ts/src/modular/emit-samples.ts#L221.

JialinHuang803 (approved):

PR #4817, opened 2026-07-06
fix(typespec-ts): simplify SDK User-Agent telemetry prefix

lirenhe:
## Fixes #4774

Simplify the generated SDK `User-Agent` telemetry prefix to follow the [Azure Core telemetry policy](https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy).

### Problem
Generated modular clients emit a stacked User-Agent such as:

```
azsdk-js-client azsdk-js-api azsdk-js-ai-projects/2.0.0-beta.4 core-rest-pipeline/1.22.2 Node/22.14.0 (...)
```

The `azsdk-js-client` and `azsdk-js-api` tokens were added in [autorest.typescript#2616](https://github.com/Azure/autorest.typescript/pull/2616) to distinguish the classic-client / modular-API / REST layers. They are not real packages, carry no `/version`, and do not match the policy format `azsdk-<lang>-<package>/<version>` — confusing downstream service teams (#4774).

### Change
- `client-helpers.ts` `buildUserAgentOptions` now builds the prefix from `azsdk-js-<package>/<version>` only (plus any caller-supplied application id); dropped the static token and the `.includes("api")` gate. Returns no override when package name/version is unavailable.
- `build-classical-client.ts` forwards `options` untouched, so the application id still flows down and is prepended exactly once.

Resulting User-Agent:

```
[<application_id> ]azsdk-js-<package>/<version> core-rest-pipeline/x Node/y (...)
```

### Validation
- `pnpm build` clean; modular unit suite passing.
- New `test/modular-unit/userAgent.test.ts` covers the `azsdk-js-<pkg>/<version>` output, application-id forwarding, and the no-package fallback.
- Updated 10 modular-unit scenario snapshots.

### Note
The `@azure/ai-projects` team already ships a `scripts/post-emitter.mjs` that strips these tokens as a workaround; this emitter fix makes such workarounds redundant.


dargilco:
Thank you @lirenhe for doing this work! Looking forward to see this PR completed. FYI @glharper @bobogogo1990

packages/typespec-ts/src/modular/helpers/client-helpers.ts
```diff
@@ -313,10 +313,19 @@ function buildLoggingOptions(): string | undefined {
...
+/**
+ * Builds the `userAgentOptions` for the generated client, producing a telemetry
+ * prefix of the form `[<application_id> ]azsdk-js-<package>/<version>` per the Azure
+ * Core telemetry policy: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy
+ *
+ * Historically (autorest.typescript PR #2616) the emitter also prepended per-layer
+ * tokens `azsdk-js-client` and `azsdk-js-api` to distinguish the classic-client,
+ * modular-API and REST layers. Those tokens have no `/version`, are not real packages,
+ * and confused service teams, so they were removed (see Azure/typespec-azure#4774).
+ */
```
JialinHuang803: As this is already captured in the PR description, I suggest that we remove the historical content from the source comment.

JialinHuang803 (approved):
Overall the code looks good to me, and I left a small comment. I'm not sure whether `azsdk-js-api` and `azsdk-js-client` were added in https://github.com/Azure/autorest.typescript/pull/2616 because we wanted to track traffic from the API layer and client layer separately. I may need @qiaozha to help confirm if this change impacts anything else.

lirenhe:
@qiaozha   @xirzec,  could you help to review this PR?  thanks

qiaozha:
@dargilco could you explain why you need this userAgent change? Otherwise, I don't think this change makes sense. Because it has nothing to do with the RLC but the modular's own subpath export concept And currently Classical Client layer/API layer are still exposed in the Modular client, and people who have special bundle size requirement can still use API layer only, they might want to filter out their traffic from other users who uses classical client within their organization. /cc @xirzec

dargilco:
@qiaozha the description at the top of this GitHub issue explains is very well. I expect User Agent format to be simple and follow Azure SDK guidelines and be similar to other languages, so our many telemetry queries do not need to continue special casing JS.

@bobogogo1990 @glharper FYI

xirzec (approved):
I'm fine with these changes to align with UA across languages. Since we have less complexity in our emit now (with the deprecation of RLC) this feels less valuable than it was originally.

As @qiaozha has pointed out, we are losing the ability to see if requests are made by the client class or the api methods directly, but I can't say that we have been looking closely at this data and making decisions with it today. 

Anecdotally we've had signal that the api methods are useful for those concerned with bundle size, and I'm fine with that being sufficient to justify continuing to support them without needing to validate using non-standard telemetry.

lirenhe:
Thanks all for the suggestion.  I will merge this PR if there is no further comments.

PR #4842, opened 2026-07-06
Added Linter rule to disallow reserved property presence in resource's property bag.

vayada:
Added Linter rule to disallow reserved property presence in resource's property bag.
One such example is "BillingData" property which is reserved for Billing Feature through RP Platform and is being standardized through Common Types.


packages/typespec-azure-resource-manager/src/rules/no-billing-data-in-properties-bag.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const noBillingDataInPropertiesBagRule = createRule({
+  name: "no-billing-data-in-properties-bag",
+  description:
+    "A property named 'BillingData' must not be present in a resource's property bag. The name is reserved for platform billing integration.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-billing-data-in-properties-bag",
+  messages: {
+    default: paramMessage`Property "${"propertyName"}" is not allowed in the resource property bag. The "BillingData" property name is reserved for platform billing integration.`,
+  },
```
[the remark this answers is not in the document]
vayada: Updated the Title and description.

.chronus/changes/add-no-billing-data-in-properties-bag-rule-2026-7-6-12-40-0.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Add `no-billing-data-in-properties-bag` linter rule that flags a `BillingData` property (case-insensitive) present in an ARM resource's property bag, since the name is reserved for platform billing integration.
```
timotheeguerin: will there be other properties that will/should be banned from resource propeties?
`no-reserved-resource-property` might be a better name accomdating that, if not any shorter name we could use?

packages/typespec-azure-resource-manager/src/rules/no-billing-data-in-properties-bag.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { getArmResource } from "../resource.js";
+import { getProperties } from "./utils.js";
+
+const restrictedPropertyName = "billingdata";
+
+export const noBillingDataInPropertiesBagRule = createRule({
+  name: "no-billing-data-in-properties-bag",
+  description:
+    "A property named 'BillingData' must not be present in a resource's property bag. The name is reserved for platform billing integration.",
```
timotheeguerin:
```suggestion
    "A property named 'billingData' must not be present in a resource's property bag. The name is reserved for platform billing integration.",
```
properties are camelcase usually right? (same fix below)
might be also good to mention this is case insensitive

packages/typespec-azure-resource-manager/src/rules/no-billing-data-in-properties-bag.ts
```diff
@@ -0,0 +1,41 @@
...
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-billing-data-in-properties-bag",
+  messages: {
+    default: paramMessage`Property "${"propertyName"}" is not allowed in the resource property bag. The "BillingData" property name is reserved for platform billing integration.`,
```
markcowl: We probably want to parameterize this message to allow for additional reserved properties / reasons for reservation

packages/typespec-azure-resource-manager/src/rules/no-billing-data-in-properties-bag.ts
vayada: properties are camel case and kept this check case insensitive to not allow this named property at all in any form. As per Mark's comment, I have also updated this rule to check the list (extensible) of reserved properties to check for in property bag.

packages/typespec-azure-resource-manager/src/rules/no-reserved-resource-property.ts
```diff
@@ -0,0 +1,49 @@
...
+import { getArmResource } from "../resource.js";
+import { getProperties } from "./utils.js";
+
+/**
+ * Property names that are reserved and must not appear in a resource's property bag, mapped to
+ * the reason each name is reserved. Names are compared case-insensitively, so keys must be
+ * lowercase. Add a new entry here to reserve an additional resource property name.
+ */
+const reservedProperties = new Map<string, string>([
+  ["billingdata", "platform billing integration"],
```
markcowl:
```suggestion
  ["billingData", "platform billing integration"],
```

packages/typespec-azure-resource-manager/src/rules/no-billing-data-in-properties-bag.ts
vayada: Made this list and parameterized for reserved property names.

packages/typespec-azure-resource-manager/src/rules/no-reserved-resource-property.ts
```diff
@@ -0,0 +1,49 @@
...
+          return;
+        }
+
+        const resourceProperties = getProperty(model, "properties")?.type;
+        if (resourceProperties === undefined || resourceProperties.kind !== "Model") {
+          return;
+        }
+
+        for (const property of getProperties(resourceProperties)) {
+          const reason = reservedProperties.get(property.name.toLowerCase());
```
markcowl: The move to mixed case in the reservedProperties seems to have broken this check.

I would suggest processing the list into a set of lowercase keynames, and then checking that set for has.  This will allow you to uyse the key name in the message with the appropriate casing and be explicit about lower casing both the source and target property name.

markcowl (changes requested):

packages/typespec-azure-resource-manager/src/rules/no-reserved-resource-property.ts
vayada: Thanks for checking this. With the current change message should look like this. Does it look fine?

 Property 'billingData' is reserved. Reserved for Billing Through RP platform.

markcowl (approved):

Issue #4830, opened 2026-07-06
Remove readme.md


timotheeguerin:
**Target: January 2027**

Part of #4825.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/service-yaml.md

**Remove `readme.md`** from the repo once `service.yaml` is validated (#4828) and contributor docs are updated (#4829). Generate on demand only if a legacy consumer still needs it.

Depends on #4828 and #4829.

Issue #4829, opened 2026-07-06
Contributor docs: author API versions via service.yaml


timotheeguerin:
Part of #4825.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/service-yaml.md

Update contributor documentation so authors **add/bump API versions in `service.yaml`** (instead of editing `readme.md` tags). Cover ARM and data-plane, TypeSpec and swagger versions.

Depends on #4826.

Issue #4828, opened 2026-07-06
TSV validation: require + sync-check service.yaml


timotheeguerin:
Part of #4825.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/service-yaml.md

Add a **TypeSpec validation (TSV)** rule for `service.yaml`:
- **present** for every service
- **schema-valid** against the published JSON Schema
- **in sync** — the version set matches the `@versioned` enum (TypeSpec) and the `readme.md` version set during coexistence

Depends on #4826.

Issue #4827, opened 2026-07-06
typespec-autorest: inject API versions from TypeSpec


timotheeguerin:
Part of #4825.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/service-yaml.md

Add `@azure-tools/typespec-autorest` support to **derive/inject the API version list from the TypeSpec `@versioned` enum** for TypeSpec services, so versions don't have to be hand-maintained in `service.yaml`.

- TypeSpec services: versions come from `@versioned`.
- Swagger versions stay explicit via `swagger-files`.
- Resolves the RFC's version auto-detection open question toward auto-detection for TypeSpec.

Depends on #4826 (format/schema).

PR #4846, opened 2026-07-07
feat(typespec-autorest): emit service.yaml manifest from @versioned

timotheeguerin:
Part of #4825 — implements #4827 (and the schema half of #4826).

Adds `service.yaml` emission to `@azure-tools/typespec-autorest`, deriving the ordered API version list from the `@versioned` enum.

## What

- **New `service-yaml` emitter option** (`"auto"` | `"always"` | `"never"`, default `"auto"`):
  - `auto`: emit/update `service.yaml` only if the file already exists at the project root (opt-in during rollout; will be forced later).
  - `always`: always emit.
  - `never`: disable emission.
- **Output**: `service.yaml` is written at the **project root, next to `tspconfig.yaml`**:
  ```yaml
  versions:
    - version: 2023-11-01
      source: typespec
      swagger-files:
        - resource-manager/Contoso/stable/2023-11-01/openapi.json
    - version: 2024-06-01-preview
      source: typespec
      swagger-files:
        - resource-manager/Contoso/preview/2024-06-01-preview/openapi.json
  ```
  - `versions` are ordered oldest→newest from the `@versioned` enum.
  - `source` differentiates TypeSpec versions from legacy `swagger` versions (autorest always emits `typespec`).
  - `swagger-files` are paths to the emitted openapi file(s), relative to `service.yaml`.
- **Schema defined in TypeSpec**: `schema/service-yaml.tsp` (`@jsonSchema`), with the JSON Schema generated via `@typespec/json-schema` (new `regen-service-yaml-schema` build step, shipped through `package.json` `files`).
- Multiple services in one project → a warning diagnostic; only the first service is included.

## Scope

This PR intentionally covers only autorest emission + the schema. Deferred to sibling issues: the `readme.md → service.yaml` conversion tool and repo-wide migration (rest of #4826), TSV sync-check (#4828), contributor docs (#4829), and `readme.md` removal (#4830).

## Real-world rollout validation

Validated end-to-end against **azure-rest-api-specs** by rolling this emitter out to every TypeSpec project that emits `@azure-tools/typespec-autorest`: [Azure/azure-rest-api-specs#44506](https://github.com/Azure/azure-rest-api-specs/pull/44506) (draft, depends on this PR).

- Seeded an empty `service.yaml` at each project root and recompiled with this PR's build (via `pkg.pr.new`); `auto` mode populated **558 services**.
- Cross-checked every generated `service.yaml` against the existing `readme.md` `input-file` mappings across all 579 readmes: **1360/1365 (99.6%) of swagger references match exactly**; the manifests are a faithful subset (TypeSpec versions only, legacy `source: swagger` versions excluded as expected).
- The rollout surfaced that multi-service aggregate/wrapper projects (e.g. ResourceGraph, Microsoft.Insights, RadiologyInsights) don't map to a single project-root output — real per-service manifests come from their sub-projects instead. Reinforces the "only first service" multi-service caveat below.

## Open questions

- Option name/shape (`service-yaml` enum vs a boolean).
- Whether unversioned services should emit a `service.yaml`.
- Whether to embed a `$schema` / yaml-language-server reference in the emitted file.

## Validation

- Full package build (schema regen + tsc + library lint) ✅
- `oxlint` clean ✅
- 501/501 package tests pass, including 6 new `test/service-yaml.test.ts` cases ✅

### DIFF
--- a/packages/typespec-autorest/test/service-yaml.test.ts
+++ b/packages/typespec-autorest/test/service-yaml.test.ts
@@ -95,6 +95,48 @@ describe("emission trigger option", () => {
     );
     expect(raw).toEqual("versions: []\n");
   });
+
+  it("preserves comments and unrelated keys when updating an existing file", async () => {
+    const existing = [
+      "# Manifest for MyService",
+      "versions:",
+      "  # first stable version",
+      '  - version: "2023-01-01"',
+      "    source: typespec",
+      "    swagger-files:",
+      "      - old/path.json",
+      "  # to be removed",
+      '  - version: "2020-01-01"',
+      "    source: swagger",
+      "    swagger-files:",
+      "      - legacy/openapi.json",
+      "# trailing note",
+      "custom-key: keep-me",
+      "",
+    ].join("\n");
+
+    const { raw, manifest } = await emitServiceYaml(
+      { "main.tsp": versionedService, "service.yaml": existing },
+      { "service-yaml": "auto" },
+    );
+
+    assert(raw);
+    // File-level comments and unrelated keys are preserved.
+    expect(raw).toContain("# Manifest for MyService");
+    expect(raw).toContain("# trailing note");
+    expect(raw).toContain("custom-key: keep-me");
+    // Per-version comment on a retained version is preserved.
+    expect(raw).toContain("# first stable version");
+    // Removed version and its comment are gone.
+    expect(raw).not.toContain("# to be removed");
+    expect(raw).not.toContain("2020-01-01");
+    // Generated data reflects the current @versioned enum.
+    assert(manifest);
+    expect(manifest.versions.map((v) => v.version)).toEqual(["2023-01-01", "2024-01-01-preview"]);
+    expect(manifest.versions[0]["swagger-files"]).toEqual([
+      "tsp-output/@azure-tools/typespec-autorest/stable/2023-01-01/openapi.json",
+    ]);
+  });
 });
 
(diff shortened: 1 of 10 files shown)


markcowl:
@timotheeguerin Here is an example of the readme.md needed by RPaaS: https://github.com/Azure/azure-rest-api-specs-pr/tree/RPSaaSMaster/specification/windowsesu/resource-manager

It looks like this is just boilerplate, so we should be able to get away with just leaving it in place without updates.

RPSaaSMaster also has mostly v1 specs, so these would need to be converted to v2 and have the boilerplate readme.md under `resource-manager`

There may be some need for this in live validation as well, will see if I cna find an instance.

packages/typespec-autorest/test/service-yaml.test.ts
```diff
@@ -0,0 +1,169 @@
...
+import { AutorestEmitterOptions } from "../src/lib.js";
+import { ServiceYaml } from "../src/types.js";
+import { Tester } from "./test-host.js";
+
+interface EmitServiceYamlResult {
+  /** Parsed service.yaml, or undefined if the file was not written. */
+  readonly manifest: ServiceYaml | undefined;
+  readonly raw: string | undefined;
+  readonly diagnostics: readonly Diagnostic[];
+}
```
markcowl, line 14: It feels like we need a test where we are emitting multiple files per service.

markcowl (approved):

PR #4851, opened 2026-07-08
Anchor ARM resource identity to read and create operations

ArcturusZhang:
## Summary
- Change ARM resource resolution so concrete resource identities are seeded only from registered read/createOrUpdate operations
- Replace permissive operation-path resource ID inference with strict ARM resource instance path parsing
- Append lifecycle/list/action operations only when they match an already detected resource identity
- Add regression coverage for list-only/action-only operations not creating resources and create-only operations still creating resources
- Include the `resolveArmResources` design proposal and related linter candidates

## Designed resource detection flow

1. **Collect candidate resource model details** from `listArmResources(program)`. These are registered TypeSpec resource models, not yet concrete ARM resources.
   - Current implementation largely already does this.

2. **Find identity operations** for each candidate model: `read` and `createOrUpdate`.
   - **Required change:** previous implementation considered all registered ARM operations equally during resource creation, including `list`, `action`, `update`, `delete`, and `checkExistence`.
   - **Covered by this PR:** resources are now seeded only from `read` and `createOrUpdate`.

3. **Use identity operation paths as candidate resource instance paths.** The resolver should trust the registered ARM operation kind; HTTP verb/response-shape correctness should be enforced by linters.
   - **Required change:** previous implementation derived resource identity from every operation path.
   - **Covered by this PR:** only `read`/`createOrUpdate` paths can establish resource identity.

4. **Parse and validate each candidate path** as an ARM resource instance path.
   - **Required change:** previous parser synthesized resource IDs from non-instance paths, especially list/action paths, and fabricated providerless resource identities beyond known exceptions.
   - **Covered by this PR:** `getResourcePathElements` was removed and replaced with strict `parseArmResourceInstancePath`, which requires a literal `/providers/{namespace}` segment except for known providerless exceptions such as tenant, subscription, resource group, and explicit generic resource-id parameter paths.

5. **Group candidates by normalized resource identity**: provider namespace, resource type segments, literal path segments case-insensitively, and variable path segments by position while ignoring variable names.
   - **Required change:** previous grouping did not consistently compare full normalized instance paths and could merge across paths/scopes when explicit resource names matched.
   - **Covered by this PR:** resource identity grouping now uses normalized instance-path comparison for identity operations.

6. **Create one concrete `ResolvedResource` per identity group.** `read` + `createOrUpdate` on the same normalized path become one resource; different valid paths become separate resources; `createOrUpdate` alone can create a resource; `list`/`action` alone cannot.
   - **Required change:** previous `list` and `action` operations could create a `ResolvedResource`.
   - **Covered by this PR:** added tests for list-only/action-only not creating resources and create-only creating a resource.

7. **Derive stable resource facts from the canonical instance path**: `resourceInstancePath`, `resourceType`, provider namespace, scope, resource name parameter or singleton key, and parent candidate information.
   - **Required change:** previous implementation derived these facts independently from each operation path; singleton metadata is model/decorator-based rather than path-based.
   - **Partially covered by this PR:** `resourceInstancePath`/`resourceType` now come from strict identity paths. Path-based singleton cleanup remains future work.

8. **Append remaining operations** only to already-detected resources: lifecycle operations by matching normalized instance path, list operations by detected-resource-ID prefix, and action operations under the detected resource instance path.
   - **Required change:** previous detection and operation grouping happened in one phase, allowing non-identity operations to define resources. Previous list matching also parsed list paths to reconstruct possible resources.
   - **Covered by this PR:** non-identity operations are appended only after detected identities exist. List operations now attach only to the closest detected resource whose normalized resource ID has the list request path as a prefix. Non-prefix list mappings are intentionally left as a TODO and do not create resource identities.

9. **Leave unmatched operations for provider/non-resource handling.** They should not invent new resources.
   - **Required change:** previous registered ARM operations that failed association could be skipped and also excluded from `providerOperations`.
   - Not covered by this PR.

10. **Resolve parent and scope relationships** after the full resource set is known, using normalized path comparison. Missing parents should be skipped rather than synthesized as empty resources.
    - **Required change:** previous implementation can synthesize empty parent/scope resources and return `ExternalResource` for ambiguous external scopes.
    - **TODO:** clarify scope representation. Concrete external scopes can currently be synthesized as `ResolvedResource` entries, while ambiguous/generic external scopes use `"ExternalResource"` as an unresolved-external sentinel. We should decide whether to keep that sentinel, replace it with `undefined` or a structured unresolved scope, and remove the unused `parent`/`scope` fields from the internal `ResolvedResourceOperations` staging type if they remain unused.
    - Not covered by this PR.

## Validation
- `pnpm --filter @azure-tools/typespec-azure-resource-manager exec tsc -p tsconfig.build.json --noEmit`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager exec vitest run test/resource-resolution.test.ts --testNamePattern "parseArmResourceInstancePath"`
- Diagnostic-tolerant resolver behavior script verified list-only=0, create-only=1, full resource prefix-list/action attachment=1/1, resource group exception, and generic resource-id exception
- Full `pnpm --filter @azure-tools/typespec-azure-resource-manager exec vitest run test/resource-resolution.test.ts` is currently blocked by baseline `experimental-feature` warnings from compiler/agent TypeSpec files


packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
```diff
@@ -647,10 +549,7 @@ interface Employees {
           ],
         },
         actions: [{ operationGroup: "Employees", name: "move", kind: "action" }],
-        lists: [
-          { operationGroup: "Employees", name: "listBySubscription", kind: "list" },
-          { operationGroup: "Employees", name: "listByResourceGroup", kind: "list" },
-        ],
+        lists: [{ operationGroup: "Employees", name: "listByResourceGroup", kind: "list" }],
```
ArcturusZhang: not sure if this is an issue.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
```diff
@@ -0,0 +1,92 @@
+# Linter candidates for `resolveArmResources`
```
markcowl, line 1: I think we should probably make this into an issue  Most of this validation is being added as part of the new linters for typespec suppression, but it would be nice to have this to track

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
```diff
@@ -0,0 +1,238 @@
+# Design for `resolveArmResources`
+
+## Expected output
+
+`resolveArmResources(program)` should produce a complete, deterministic, version-aware description of the ARM provider shape represented by the TypeSpec program.
+
+The API should make the version dimension explicit. For an unversioned TypeSpec program, it should return one provider result. For a versioned TypeSpec program, it should return provider results for each declared service version so downstream emitters and analyzers can reason about the resource shape that exists in each API version.
```
markcowl: I am skeptical of this requiremntt for two reasons:
- It assumes that every consumer wants a version-projection for each version in the program
- it assumes that every consumer does not want a version-independent canonical view of the resources (this was the original reason for this API, in fact).

Instead, I would suggest that either a versioned program is passed in (for which helpers could be provided) or there is a mechanism that allows projecting the result.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
```diff
@@ -0,0 +1,238 @@
...
+  provider: string;
+  types: string[];
+}
+
+type ArmResourceScope =
+  | "Tenant"
+  | "Subscription"
+  | "ResourceGroup"
+  | "ManagementGroup"
+  | "Scope"
```
markcowl, line 70: We should add "ServiceGroup"

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
```diff
@@ -0,0 +1,238 @@
...
+| `operations`           | Resource lifecycle, list, and action operations grouped by operation kind. These are the operations that directly operate on this resource path.                                                                   |
+| `parent`               | A reference to the parent resource when this resource is a child resource. The parent should be resolved from the path after the full resource set is known.                                                       |
+| `scope`                | The scope where this resource exists. This can be a well-known ARM scope string or a structured reference to another resource used as the extension target.                                                        |
+| `singleton`            | Singleton metadata when the resource has a fixed name segment, such as `default`, or a fixed set of allowed singleton names.                                                                                       |
+
+## High-level flow
+
+1. Find the ARM provider namespace with `resolveProviderNamespace(program)`.
```
markcowl, line 138: versioning doesn't seem to be mentioned here, how will we handle versions?

packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -705,173 +706,198 @@ function normalizePathForScopeComparison(path: string): string {
...
-          instanceSegments.push("{name}");
-          break;
-        default:
-          break;
-      }
-      break;
-    }
+  for (let i = 0; i < resourceSegments.length; i += 2) {
+    const typeSegment = resourceSegments[i];
+    const nameSegment = resourceSegments[i + 1];
```
markcowl, line 776: what if the name segment isn't a variable segment?

packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
markcowl: seems like we need both

markcowl (changes requested):
I need a little more time to go over the test changes, but will add in the external integration ci job as well.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
ArcturusZhang: sure I could create an issue to track these candidates. do you prefer one per one candidate or we have a overall issue to contain all of these?

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
ArcturusZhang: yeah I agree. but per my investigation in [this issue](https://github.com/Azure/typespec-azure/issues/4858), **currently** passing in a version mutated program does not really change the result, just FYI.

packages/typespec-azure-resource-manager/src/resource.ts
ArcturusZhang: then it should be a singleton resource. the handling of singleton resource is not covered (or not fully covered) in this PR yet.
there are two kinds of singleton resources:
1. the resource's name is a constant
2. the resource's name is not a constant but an enum - we should expect fixed enum.
last time we did a change in the resource result schema to let this api to support both - the singletonresourcename can take a single value or an array of names.

FYI I have to stop the modification in some point to prevent this PR to contain every change we would like, because it might make the PR very hard to review. What do you think?
I was trying to split the changes we would like to have when making the change, but those changes kind of are linked with each other, therefore the PR's size just grew and grew

packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
ArcturusZhang: yeah I am thinking the same thing right now.
I will see where it goes wrong and get it back.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
ArcturusZhang: I think our design is:
this API just uses the program which has been (or not) mutated by versioning. therefore this API does not do anything about versioning explicitly.
is this understanding correct?
if so, we do not need to mention versioning here at all.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
ArcturusZhang: Addressed in caf191eb by removing the per-version result requirement from the proposal and documenting that 
esolveArmResources uses the view represented by the input program; version-specific analysis should be layered on top with version-projected programs/helpers.

---
🤖 arcturus-copilot

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
ArcturusZhang: Addressed in caf191eb: added ServiceGroup to the proposal's ArmResourceScope union and added code-side ArmResourceScope typing including ServiceGroup.

---
🤖 arcturus-copilot

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources.md
ArcturusZhang: Addressed in caf191eb. The high-level flow remains focused on resource resolution, and versioning is now described earlier as an input-program/projection concern rather than as a resolver step.

---
🤖 arcturus-copilot

packages/typespec-azure-resource-manager/test/resource-resolution.test.ts
ArcturusZhang: Agreed. The strict prefix rule was too narrow for list operations at wider scopes. I opened https://github.com/Azure/typespec-azure/issues/5080 to track the complete matching-rule design, and added an issue-linked TODO in the matcher.

In this PR I am applying the focused case we know is valid: subscription-list operations can attach to resource-group-scoped resources when the concrete provider/type path is the same. That brings this test back to expecting both `listBySubscription` and `listByResourceGroup`, without letting list operations create resource identities.

---
🤖 arcturus-copilot

ArcturusZhang: Follow-up: refined in `9ef00922` to avoid a scope-specific exception. The matcher now prefers direct list-path prefix matches first; if none exists, it falls back to matching by concrete provider/type path. The resource side reuses `parseArmResourceInstancePath`, while list/collection paths use a small collection-path parser.

---
🤖 arcturus-copilot

ArcturusZhang: Follow-up in `553cf2bc`: `getListMatchDistance` now trims the ARM scope prefix before comparing path lengths, so prefix-match ranking is based on the resource-type tail rather than subscription/resource-group/tenant scope length.

---
🤖 arcturus-copilot

markcowl (commented):
The only ask is that there is a mechanism to force a particular grouping for non-standard APIs  (for example, APIs in which there is a non-standard ath template, such as resource operations with multiple additional path segments).

In the old mechanism, you could always override the default grouping of operations using a 'resourceName' property: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/decorators.tsp#L164 , and we should be able to add override the default grouping of an operation in the same way.  I think it is *fine* to not call something a resource if it does not have a get (single) operation.

packages/typespec-azure-resource-manager/proposal/resolve-arm-resources-linter-candidates.md
markcowl: We are tracking the work here: https://github.com/Azure/typespec-azure/issues/5005

Please add any additional needed linting rules.

markcowl (approved):

PR #4859, opened 2026-07-10
Read typespec-ts client user-agent version dynamically from package.json

xirzec:
## Why

The `@azure-tools/typespec-ts` emitter burned the `package-details.version` emitter option into the generated client's user-agent string. Once a package is generated, that version is really owned by the package's own `package.json`, so hardcoding it forced a whole machinery (the `//metadata.constantPaths` `userAgentInfo` entries) to exist purely so `dev-tool` could rewrite the burned-in constant on every release. This PR removes that coupling and simplifies the version handling along the way.

## What changed

**1. Stop defaulting `package-details.version` in the datamodel; default at the point of use.**
The version was previously defaulted to `1.0.0-beta.1` during option transform, and a separate `isVersionUserProvided` flag was carried around so the README logic could tell "user set it" from "we defaulted it." That flag existed for exactly one consumer. Now `version` stays `undefined` when unset and each consumer (package.json, CHANGELOG, README, user-agent) applies its own default locally, so `isVersionUserProvided` is gone entirely.

**2. Emit code that reads the version dynamically at runtime.**
Instead of interpolating a literal version into the generated source, `buildUserAgentOptions` now emits a self-referencing JSON import and reads `pkgJson.version`:

```ts
import pkgJson from "@azure/foo/package.json" with { type: "json" };
...
const userAgentInfo = `azsdk-js-foo/${pkgJson.version}`;
```

The self-reference (`<packageName>/package.json`) relies on the `"./package.json"` export the emitted package already declares, so the specifier resolves identically from `src` and from the built `dist` output regardless of directory depth. The import is deduped per file and only emitted for the api context (the classical-client prefix never surfaced `userAgentInfo`).

**3. Remove the now-obsolete `constantPaths` plumbing.**
With no literal version in the source, there is nothing for `dev-tool` to bump, so the `//metadata.constantPaths` `userAgentInfo` generation and the `clientContextPaths` option threaded through `buildPackageFile` / `updatePackageFile` / `buildAzureMonorepoPackage` (plus the `getRelativeContextPaths` helper) are dead. All removed.

## Notes for reviewers

- Emitter type-checks; `test-next` (252) and `unit-modular` (657) all pass. The modular snapshot scenarios don't set `package-details`, so there is zero snapshot churn.
- One thing worth a careful look / a spector run before merge: I verified the emitted import shape and emitter-side behavior, but did not run a full downstream `azure-sdk-for-js` build (tshy + api-extractor) to confirm the JSON import compiles end to end in a real generated package. The syntax and module resolution are correct, but that integration path is out of scope for this environment.
- ts-morph gotcha handled: the import attribute value is passed unquoted (`value: "json"`) because the writer auto-quotes it.

### DIFF
--- a/.chronus/changes/dynamic-user-agent-version-2026-7-9-20-55-0.md
+++ b/.chronus/changes/dynamic-user-agent-version-2026-7-9-20-55-0.md
@@ -4,4 +4,4 @@ packages:
   - "@azure-tools/typespec-ts"
 ---
 
-Read the client user-agent version dynamically from the generated `package.json` at runtime (via the package's own `./package.json` export) instead of hardcoding the emitter's `package-details.version` into the generated source.
+Read the client user-agent version dynamically from the generated `package.json` at runtime (via the package's own `./package.json` export) instead of hardcoding the emitter's `package-details.version` into the generated source. The generated `config/tsconfig.src.*.json` files now set `resolveJsonModule: true` so the JSON import type-checks across compiler versions.
--- a/.chronus/changes/dynamic-user-agent-version-2026-7-9-20-55-0.md
+++ b/.chronus/changes/dynamic-user-agent-version-2026-7-9-20-55-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@azure-tools/typespec-ts"
+---
+
+Read the client user-agent version dynamically from the generated `package.json` at runtime (via the package's own `./package.json` export) instead of hardcoding the emitter's `package-details.version` into the generated source.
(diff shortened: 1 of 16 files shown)


JialinHuang803:
I tested the generated code in the sdk repo and encountered an error when running the test:
image
This may impact the ci of the SDK PR.

xirzec:
> I tested the generated code in the sdk repo and encountered an error when running the test: <img alt="image" width="2000" 

@JialinHuang803 this is a great find, thank you for the validation. I have a fix upstream in the test config for the JS repo: https://github.com/Azure/azure-sdk-for-js/pull/39280

JialinHuang803 (approved):

Issue #4801, opened 2026-07-02
Benchmark: support external specs (e.g. azure-rest-api-specs) for evaluating TCGC/library perf impact


tadelesh:
## Summary

Today the benchmark tool (`packages/benchmark`) can only run the built-in specs that live under `packages/benchmark/specs/`. We want to be able to benchmark **external specs** — in particular real specs from [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs) — so we can evaluate the performance impact of changes to our libraries against realistic, large inputs. This covers both:

- **TCGC** (`@azure-tools/typespec-client-generator-core`), the shared client-generation core, and
- the **client emitters** in this repo (e.g. Python `@azure-tools/typespec-python`, Java `@azure-tools/typespec-java`, and others), whose end-to-end generation cost we also want to measure.

## Motivation

- The built-in specs are synthetic and may not exercise the code paths that a given change touches.
- TCGC is a library consumed by client emitters, not a stand-alone emitter in the built-in specs, so its cost is not measured today. We want to measure the TCGC pipeline (`createSdkContext` / `getAllModels`) in isolation on real specs.
- We also want to measure the **client emitters themselves** (Python, Java, etc.) end-to-end on real specs, since a change may affect emitter performance independently of TCGC.
- When making a change to TCGC or a client emitter, we want a repeatable way to compare "before vs after" on the same external spec.

## Requirements

1. The benchmark tool can run against external specs sourced from another repository (e.g. azure-rest-api-specs), not only the built-in specs.
2. External spec sources are not committed into this repository.
3. Benchmarking an external spec reflects the current local state of our libraries, so a change under development can be measured before it ships.
4. It is possible to choose which part of the pipeline is measured for an external spec — e.g. TCGC alone, or a specific client emitter such as Python or Java.
5. Both TCGC and each client emitter (Python, Java, etc.) are reported individually with their own timing.
6. Built-in and external specs can be run and compared together.
7. There is a repeatable "before vs after" workflow to evaluate the performance impact of a change to TCGC or a client emitter.

tadelesh:
## Design

### Overview

Add an **external-spec ingestion** flow to the benchmark tool. External specs are pulled in on demand, compiled against the current workspace so results reflect local library changes, and measured through the same runner and reporting as the built-in specs. The emitters run for an external spec are configurable, so the pipeline of interest (TCGC, or a client emitter such as Python/Java) can be isolated.

### Configuration

[...]

PR #4862, opened 2026-07-10
Benchmark external specs (azure-rest-api-specs) with a separate dashboard

tadelesh:
Implements external-spec benchmarking, designed in #4801.

## What this adds

Benchmark real specs from another repository (e.g. [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs)) as a separate, opt-in group alongside the built-in specs, with their own dashboard view — kept out of the main baseline.

### Runner (generic)
A benchmark spec is anything the runner can compile: a **local spec** (a directory with `main.tsp`) or an **external spec** (a directory with `spec.json`). `--specs-dir` selects which set to run; both are treated uniformly and produce one results file. There is no built-in/external split in the tool itself — the distinction is made by CI (which directory to run) and the dashboard (which dataset to show).

External specs are sparse-checked-out into `packages/benchmark/.external/` (git-ignored) and compiled against **this workspace's** packages, so results reflect local library changes (TCGC, client emitters, linters). `ref` tracks latest (no pinning). Each external spec directory carries a `tspconfig.yaml` (copied into the checkout) that selects the emitters and linter measured — e.g. TCGC + the ARM ruleset — exactly like a local spec.

### Storage
`store-results --results-dir <dir>` stores each group under its own directory on `benchmark-data` (`results/` for the baseline, `external-results/` for external). The push rebases-and-retries so the two workflows writing the branch concurrently don't fail.

### Dashboard
The benchmarks page gains a **Main baseline / External specs** switch. The External view reads `external-results/history.json` and shows a single chart switchable between two modes:
- **By emitter (cross-spec)** — one line per spec for a chosen emitter.
- **By spec (cross-emitter)** — one line per emitter for a chosen spec.

### PR comments
The external group posts its own sticky PR comment ("⚡ External Spec Benchmark Results"), separate from the main one, comparing against the `external-results` rolling baseline.

### Workflows (reusable)
`benchmark-run.yml` and `benchmark-comment-run.yml` are reusable templates holding the run/comment logic. `benchmark.yml` (main) and `benchmark-external.yml` instantiate the run template with their own parameters; `benchmark-comment.yml` and `benchmark-comment-external.yml` instantiate the comment template. Adding a group is a small instance file, not duplicated logic.

## Included external specs
`compute`, `web` (AppService), and `network` — large ARM specs from azure-rest-api-specs, measuring TCGC + the ARM ruleset.

## Notes
- Both `@azure-tools/typespec-benchmark` and the website package are private, so no changeset is required.
- The external `external-results` history only exists after the external workflow has run on `main`; until then the external dashboard/PR comment shows "no baseline".
- Design discussion: #4801.

### DIFF
--- a/.github/workflows/benchmark-comment-external.yml
+++ b/.github/workflows/benchmark-comment-external.yml
@@ -1,8 +1,8 @@
-name: Benchmark Comment
+name: Benchmark Comment External
 
 on:
   workflow_run:
-    workflows: ["Benchmark"]
+    workflows: ["Benchmark External"]
     types:
       - completed
 
@@ -18,8 +18,8 @@ jobs:
     uses: ./.github/workflows/benchmark-comment-run.yml
     with:
       run-id: ${{ github.event.workflow_run.id }}
-      header: benchmark-results
-      comment-file: benchmark-comment.md
+      header: benchmark-results-external
+      comment-file: benchmark-comment.external.md
     permissions:
       actions: read
       pull-requests: write
--- a/.github/workflows/benchmark-comment-run.yml
+++ b/.github/workflows/benchmark-comment-run.yml
@@ -1,0 +1,46 @@
+name: Benchmark Comment Run
+
+# Reusable workflow that posts one sticky benchmark comment from a Benchmark run.
+on:
+  workflow_call:
+    inputs:
+      run-id:
+        description: "The Benchmark workflow run id to download the comment artifact from"
+        required: true
+        type: string
+      header:
+        description: "Sticky comment header (identifies the comment to update)"
+        required: true
+        type: string
+      comment-file:
+        description: "Comment markdown file name inside the artifact"
+        required: true
+        type: string
+
+permissions:
+  actions: read
+  pull-requests: write
+
+jobs:
+  comment:
+    name: Post Benchmark Comment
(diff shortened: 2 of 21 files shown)


tadelesh:
@timotheeguerin this is ready for review — implements the external-spec benchmarking we discussed in #4801.

Summary of the approach:
- External specs are just another kind of benchmark spec (a directory with `spec.json` + `tspconfig.yaml`); `--specs-dir` selects which set to run. The built-in/external distinction lives in CI and the dashboard, not in the tool.
- External specs track `main` (no pinning), compile against the workspace packages, and each carries its own `tspconfig.yaml` (emitters + linter to measure).
- Results are stored in a separate `external-results/` directory and shown in a dedicated **External specs** dashboard view (by-emitter / by-spec modes), kept out of the main baseline.
- The run/comment workflows are reusable templates instantiated per group (main + external), so no duplicated CI logic.

Included external specs: `compute`, `web` (AppService), `network` (ARM), measuring TCGC + the ARM ruleset. Happy to adjust the set or the per-emitter/graph layout.

timotheeguerin (approved):
Approving, didn't check too much the code, will check what the benchmark looks like after it runs and we'll tweak as needed

PR #4866, opened 2026-07-10
[typespec-java] Remove js-yaml and lodash

XiaofeiCao:
## Summary

Adds `@azure-tools/typespec-azure-portal-core` as a workspace dependency of `@azure-tools/typespec-java` (both `peerDependencies` and `devDependencies`).

## Changes

- Add `@azure-tools/typespec-azure-portal-core` (`workspace:^` peer / `workspace:*` dev) to `packages/typespec-java/package.json`.
- Pin all `devDependencies` workspace entries to exact versions (`workspace:*`).
- Improve `Build` section of contributing.md.
- Document `core-commit.json` in the Build section of `CONTRIBUTING.md` and drop the redundant `build:emitter` snippet.
- Add a `dependencies` changelog entry.


packages/typespec-java/package.json
```diff
@@ -69,26 +70,27 @@
...
-    "@typespec/events": "workspace:^",
-    "@typespec/http": "workspace:^",
-    "@typespec/openapi": "workspace:^",
-    "@typespec/rest": "workspace:^",
-    "@typespec/sse": "workspace:^",
-    "@typespec/streams": "workspace:^",
-    "@typespec/tspd": "workspace:^",
-    "@typespec/versioning": "workspace:^",
-    "@typespec/xml": "workspace:^",
+    "@typespec/compiler": "workspace:*",
```
timotheeguerin: why changing to * here?

weidongxu-microsoft: Agree. why touch other lines?

---

PS:
I am fine if we'd prefer to update the workflow in downstream to add such libs (include liftr-base and openai-typespec), to avoid adding the lib here.

However, the continuous adding of lib (like this `typespec-azure-portal-core`) could be a problem for emitters.
Particularly this lib (compared to openai-typespec that include models), that should not have any effect on SDK, but every downstream now need to include the lib, otherwise source just won't compile.

XiaofeiCao: The changing of "^" to "*" is for `tsp-client generate-config-files` in SDK repository. 

Context is that we are pinning to specific version there: https://github.com/Azure/azure-sdk-for-java/blob/main/eng/emitter-package.json

"^" will result in caret version in generated emitter-package.json. With "workspace:^", published packages will be e.g. "^0.69.2"

`--use-npm-pinning` doesn't seem to solve this either, as it uses published package's `devDependencies`:
https://github.com/Azure/azure-sdk-tools/pull/15329/changes#diff-26808cdda9f5eeea17dae5ef5b094cefc04ea58f37e96af8a55ec21be4e7cea3

` npm view @azure-tools/typespec-python@0.63.2 devDependencies --json` result:
```json
{
  "@types/node": "^25.5.2",
  "@types/semver": "^7.7.1",
  "@types/yargs": "^17.0.35",
  "c8": "^11.0.0",
  "picocolors": "^1.1.1",
  "rimraf": "^6.1.3",
  "typescript": "~6.0.2",
  "vitest": "^4.1.3",
  "@azure-tools/azure-http-specs": "^0.1.0-alpha.42",
  "@azure-tools/typespec-azure-core": "^0.69.0",
  "@azure-tools/typespec-azure-resource-manager": "^0.69.1",
  "@azure-tools/typespec-azure-rulesets": "^0.69.1",
  "@azure-tools/typespec-autorest": "^0.69.1",
  "@azure-tools/typespec-client-generator-core": "^0.69.0",
  "@typespec/events": "^0.83.0",
  "@typespec/compiler": "^1.13.0",
  "@typespec/http-specs": "^0.1.0-alpha.38",
  "@typespec/openapi": "^1.13.0",
  "@typespec/http": "^1.13.0",
  "@typespec/rest": "^0.83.0",
  "@typespec/spec-api": "^0.1.0-alpha.14",
  "@typespec/sse": "^0.83.0",
  "@typespec/streams": "^0.83.0",
  "@typespec/spector": "^0.1.0-alpha.25",
  "@typespec/tspd": "^0.75.0",
  "@typespec/versioning": "^0.83.0",
  "@typespec/xml": "^0.83.0"
}
```

weidongxu-microsoft: Should we check with the tsp-client CLI instead? (how does it map `^` to `^0.69.2`?)

XiaofeiCao: tsp-client CLI won't map `workspace` dependencies in mono repo. It's solved by introducing `--use-npm-pinning` to use published packages' `devDependencies`:
https://github.com/Azure/azure-sdk-tools/issues/15277

weidongxu-microsoft: What I mean is whether e.g. tsp-client would then add another flag like `--use-exact-version`?

We know python no longer use tsp-client to update emitter-package. How does JS do?

XiaofeiCao: seems we are the only one using `generate-config-files`..
JS is using agent to bump emitter-package.json. From log, agent is fetching dependencies from npm registry: https://github.com/Azure/azure-sdk-for-js/tasks/24eb7685-0177-404d-9c23-8b057dc1ca53?q=is%3Aclosed

GO is using `ncu -u`: https://github.com/Azure/azure-sdk-for-go/blob/main/.github/skills/upgrade-emitter-version/SKILL.md

Rust is using custom script: https://github.com/Azure/azure-sdk-for-rust/pull/4359/changes#diff-350195b603eda666558fd7e67104842578a9eb3bdefe221ad3ae786b3a877414

there seems no `emitter-package.json` for cpp.

@timotheeguerin would like your opinion here. is `generate-config-files` the recommended way to update SDK emitter-package.json after moving emitters to typespec-azure repo?

timotheeguerin: My opinion is that this whole setup stinks, i never really understood why we needed to go through this whole magic to pin dependencies. IMO it would be much easier if we just stopped this, synced deps from spec repo and not have to redefine every dependency here. Yes, there might be some rare cases where a dep of a dep changes output slightly between the emitter test here and the uprade in sdk repo but this feels worth it IMO to simplify. (Every other project works that way and it works just fine)
For workspace deps I think @catalinaperalta worked on an alternative that pulled the version from what was published instead of the source in the repo (which makes a lot more sense). I don't remember the flag/command though.

weidongxu-microsoft: Xiaofei, let's see if we can avoid adding the lib here.

We probably still use `generate-config-files` to create the emitter-package.json (we don't want to use `ncu -u` because this likely get shut down after CFS -- `npm view` is better), then use script to add a few more designated lib from package.json of specs repo (e.g. openai-typespec and liftr-base and typespec-azure-portal-core, probably just use the exact semver from there -- there might be a possibility that the package.json hadn't updated at specs repo at this time, but I guess these lib are not that sensitive to versions -- and we usually can wait after it updated), then do a `generate-lock-file` again.

There could be problem we didn't expect...

weidongxu-microsoft (approved):

Issue #4077, opened 2026-03-23
Document all supported emitter CLI options across languages


lirenhe:
## Problem statement

While discussing with partners about the emitter support, I noticed that api-version is only explicitly documented for Python, even though other languages appear to support this option as well.

https://azure.github.io/typespec-azure/docs/emitters/clients/typespec-python/reference/emitter/#api-version

## Ask
We should explicitly document all supported emitter options that can be configured via command line.
Please help to compare the docs current emitter implementations to identify:

Supported CLI options that we allow customers/partners to use and update them into the spec.

msyyc:
Pls correct me if wrong:
1. Python/Java/JS already supports `regen-docs` which could update related docs
2. There is no CI check whether `regen-docs` is executed so we may need (1) add CI check to make sure each PR to contain latest docs or (2) a nightly build pipeline to update for all docs include all emitters.
NOTE: pipeline does not have access to create PR for now so I propose (1)

timotheeguerin:
1. yes
2. no it is enforced by ci via the website build job which will run regen-docs and check there is no diff. 

So to summarize, now that emitters are in this repo as long as they have a `regen-docs` script this I think can be closed

JialinHuang803:
Documented 3 more options (api-version, examples-dir, and namespace) that are applicable for JS: https://github.com/Azure/typespec-azure/pull/4873

weidongxu-microsoft:
java also updated https://github.com/Azure/typespec-azure/pull/4863 via regen-docs

timotheeguerin:
@weidongxu-microsoft confused here how this change was not caught before for java

timotheeguerin:
hhm I see diff for python too when running regen-docs, must be something wrong with CI then

timotheeguerin:
ah you need to add the sdk emitter as deps of the website.

PR #4873, opened 2026-07-13
[TS] Document api-version, examples-dir, namespace emitter options; remove unused should-use-pnpm-dep

JialinHuang803:
Addresses part of #4077 (document all supported emitter CLI options).

## What
- Document three TCGC-forwarded emitter options that were supported by `typespec-ts` but undocumented:
  - `api-version` – generate the SDK for a specific version (or `latest`/`all`, or a per-service map)
  - `examples-dir` – directory the emitter looks in for example files
  - `namespace` – override namespace for spec types
- Remove the unused internal `should-use-pnpm-dep` option (it was declared in the schema/interface but never read anywhere, and was flagged with a `// TODO should remove` comment).

## Why
These options are forwarded via `context.options` into TCGC's `createSdkContext` (`api-version` → `apiVersion`, `examples-dir` → `examplesDir`, `namespace` → `namespaceFlag`) and affect generation, but were missing from the emitter options schema, so they never showed up in the generated reference docs. `api-version` in particular was only documented for Python (per the issue).

## Changes
- `packages/typespec-ts/src/lib.ts` – add the three options to `EmitterOptions` + `EmitterOptionsSchema`; remove `should-use-pnpm-dep`.
- `packages/typespec-ts/README.md` – matching doc updates.
- `website/.../typespec-ts/reference/emitter.md` – regenerated via `npm run regen-docs`.
- Changeset (`internal`).

Note: options with `additionalProperties: true` already worked at runtime; this change makes them discoverable in docs and adds validation/typing.

### DIFF
--- a/packages/typespec-ts/src/lib.ts
+++ b/packages/typespec-ts/src/lib.ts
@@ -7,6 +7,21 @@ import { PackageDetails } from "./interfaces.js";
 
 export interface EmitterOptions {
   /**
+   * Use this flag if you would like to generate the sdk only for a specific version.
+   * Also accepts values `latest` and `all`, or a map from service namespace full name to version for multi-service packages.
+   */
+  "api-version"?: string | Record<string, string>;
+  /**
+   * Specifies the directory where the emitter will look for example files.
+   * Defaults to an `examples` directory located at the project root.
+   */
+  "examples-dir"?: string;
+  /**
+   * Specifies the namespace you want to override for namespaces set in the spec.
+   * With this config, all namespace for the spec types will default to it.
+   */
+  namespace?: string;
+  /**
    * Indicates whether to include response headers in the generated response type for modular operations.
    * When set to true, modular operation responses with model or void bodies will have their headers
    * represented as properties on the response type. Other SDK styles and response shapes are not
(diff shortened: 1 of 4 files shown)


lirenhe (approved):

packages/typespec-ts/src/lib.ts
```diff
@@ -50,8 +65,6 @@ export interface EmitterOptions {
   "ignore-property-name-normalize"?: boolean;
   "typespec-title-map"?: Record<string, string>;
   "ignore-enum-member-name-normalize"?: boolean;
-  //TODO should remove this after finish the release tool test
-  "should-use-pnpm-dep"?: boolean;
```
qiaozha, line 54: This option is a transit option for backward compatibility when we migrate the SDK repo from rush into pnpm to manage the packages.

qiaozha (approved):

packages/typespec-ts/src/lib.ts
JialinHuang803: Yes, I've checked and confirmed that this option is no longer used in this package, so it's safe to remove it.

Issue #4860, opened 2026-07-10
Trigger downstream language emitter CI checks when upstream dependencies change


msyyc:
## Context

More language emitters have moved into the `typespec-azure` repo. As a result, a change to an upstream dependency (e.g. the compiler / core libraries) can now cause a regression in a downstream language emitter that lives in the same repo.

Today there is **no mechanism** in CI that triggers the checks of the language emitters when a change is detected in one of their upstream dependencies. This means regressions in downstream emitters can slip through unnoticed until much later.

## Proposal

To make `typespec-azure` behave like a true mono-repo, we should optimize the CI system so that:

- When CI detects a change in an upstream dependency, it automatically triggers the checks of the affected downstream language emitters.

This will guard against upstream changes silently breaking downstream emitters.

### Trade-off & mitigation

Triggering downstream emitter checks on every upstream compiler change will make the upstream compiler CI heavier. To reduce unnecessary checks, we can add **skip logic** (e.g. a label): when a specific label is applied to a PR, the emitter checks can be skipped.

## Open questions

- Which dependency graph / affected-project detection mechanism should we rely on (e.g. existing tooling in the repo)?
- What should the exact skip label(s) be named, and who is allowed to apply them?
- Should skipped checks be reported as neutral/skipped or omitted entirely?

Feedback and questions welcome.

iscai-msft:
I think one thing here is that we wouldn't want spector e2e tests run on every single pr, that will be too costly. This is one reason @timotheeguerin and I have been pushing for more unit tests, that we could run more easily on each PR

msyyc:
We could rule a time limit for each emitter CI then each emitter could decide which checks to run in `ci-*` within the timelimit.

After https://github.com/Azure/typespec-azure/pull/4874 merged, we could make more optimization. For example: for PR, detected-change job pass a special string like "PR" to `ci-*` so that emitter ci could pick a small set of test to run; for nightly build, detected-change job pass "all" then emitter ci to run full test, With such mechanism, we could balance cost and coverage in different scenario.

> I think one thing here is that we wouldn't want spector e2e tests run on every single pr, that will be too costly. This is one reason [@timotheeguerin](https://github.com/timotheeguerin) and I have been pushing for more unit tests, that we could run more easily on each PR

PR #4874, opened 2026-07-13
Upstream-aware downstream CI (#4860)

msyyc:
Implements Azure/typespec-azure#4860.

## What
Adds **upstream-aware downstream CI**: when a PR/merge-group changes an upstream dependency (compiler `core/` submodule or in-repo libs like `typespec-client-generator-core`), the affected downstream emitter checks now run automatically. Closes the gap where an upstream-only change could merge with no emitter check.

## How
- Upstream **package** dependencies are derived from the real **pnpm workspace graph** (`pnpm --filter "...<pkg>"`), not a hand-maintained list — adding a new upstream lib needs zero config changes and cannot drift/under-trigger.
- `eng/scripts/detect-affected.ts` (zero external deps, run directly via `node`; `node --test` unit/contract tests) — diffs the event's base..head (event-aware for `pull_request` vs `merge_group`), self-filters ignored files (`**/test/**`, `**/tests/**`, `**/*.md`), maps changes to workspace packages, expands to graph dependents, and emits one boolean output per target.
- The small residual config the graph can't express is an inlined typed `CONFIG` in `detect-affected.ts`: per-target package name, CI-infra `extra` paths, and a `coreSubmodule` flag (for `core/` submodule bumps).
- `.github/workflows/ci-downstream.yml` — always-on entrypoint: `detect-affected` -> conditional per-target reusable-workflow jobs -> `downstream-ci-gate`.
- `ci-python/java/typescript.yml` converted to `on: workflow_call` (invoked only by the entrypoint -> no double-runs).

## Follow-up required before this gates merges (Strategy A migration)
A repo admin must add `downstream / CI / Downstream CI Gate` to required status checks (PR + merge queue) and remove any obsolete standalone emitter check names in the same window.


eng/scripts/detect-affected.js
timotheeguerin, line 1: lets write those in typescript we can run directly without tsx now.

eng/scripts/detect-affected.js
```diff
@@ -0,0 +1,124 @@
...
+  return /** @type {Config} */ (load(readFileSync(path, "utf8")));
+}
+
+/**
+ * @param {string} base
+ * @param {string} head
+ * @returns {string[]}
+ */
+function getChangedFiles(base, head) {
+  const out = execFileSync("git", ["diff", "--name-only", base, head], {
```
timotheeguerin: lets use simple-git if we include deps for this workflow

eng/ci/downstream-deps.yml
```diff
@@ -0,0 +1,53 @@
...
+    - "packages/typespec-azure-core/**"
+    - "packages/typespec-azure-resource-manager/**"
+  # Shared CI infrastructure every target's jobs consume.
+  shared-ci:
+    - ".github/actions/setup/**"
+
+# Downstream targets. Key = short expression-safe id used for the job output.
+targets:
+  python:
+    self: "packages/typespec-python/**"
```
timotheeguerin: should this not just use the list of deps from package.json instead of having to keep this in sync?

timotheeguerin: and use pnpm api

eng/scripts/detect-affected.js
msyyc: Done — now a .ts run directly via 
ode (native type-stripping), no tsx, and zero external deps.

eng/scripts/detect-affected.js
msyyc: updated the script.

eng/ci/downstream-deps.yml
msyyc: Done — targets are now derived from the pnpm workspace graph (pnpm --filter \...<pkg>\), no synced list. Also fixes under-triggering: the old map missed autorest/rulesets; the graph covers them for free. Only 3 things stay explicit (graph can't express them): core submodule bumps, CI-infra paths, and http-client-python.

Now only keep a simple configuration in script `eng/scripts/detect-affected.ts` and there is no need to add any static dependency map anymore.

eng/scripts/detect-affected.ts
```diff
@@ -0,0 +1,221 @@
...
+// Downstream CI target configuration.
+//
+// Upstream *package* dependencies are NOT listed here: they are derived from the
+// real pnpm workspace graph (`getDependents`). This config only declares what the
+// graph cannot express.
+//
+// TO ADD A NEW TARGET:
+//   1. Add an entry to `TARGETS` below (key = short, expression-safe id).
+//   2. Add a reusable `.github/workflows/ci-<id>.yml` (on: workflow_call).
+//   3. Add a job + gate `needs:` entry in `.github/workflows/ci-downstream.yml`.
```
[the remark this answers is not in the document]
msyyc: Fixed — comment now says \CONFIG.targets\ (also matched by autofix).

eng/scripts/detect-affected.ts
```diff
@@ -0,0 +1,221 @@
...
+  /** Globs whose sole change should NOT trigger anything (matched per file). */
+  ignore: string[];
+  /** Paths that trigger every target (shared CI infrastructure). */
+  sharedExtra: string[];
+  targets: Record<string, Target>;
+}
+
+export const CONFIG: Config = {
+  ignore: ["**/test/**", "**/*.md"],
+  sharedExtra: [".github/actions/setup/**"],
```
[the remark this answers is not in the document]
msyyc: Good catch — added \**/tests/**\ alongside \**/test/**\ (only typespec-python uses the plural dir today). Added a regression test.

eng/scripts/detect-affected.ts
```diff
@@ -0,0 +1,221 @@
+import { execFileSync } from "node:child_process";
+import { appendFileSync } from "node:fs";
+import { relative } from "node:path";
+import { fileURLToPath } from "node:url";
+
+// -----------------------------------------------------------------------------
+// Downstream CI target configuration.
+//
+// Upstream *package* dependencies are NOT listed here: they are derived from the
+// real pnpm workspace graph (`getDependents`). This config only declares what the
```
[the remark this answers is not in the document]
msyyc: Intentional: the refactor replaced the YAML map with an inlined typed \CONFIG\ (upstream deps now come from the pnpm graph). I'll update the PR description to match.

eng/scripts/package.json
```diff
@@ -2,6 +2,7 @@
   "type": "module",
   "dependencies": {
     "autorest": "^3.4.2",
-    "js-yaml": "^4.1.0"
+    "js-yaml": "^4.1.0",
+    "minimatch": "^9.0.5"
   }
```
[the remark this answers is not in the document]
msyyc: Fixed — removed \minimatch\ (replaced by the built-in glob matcher during the pnpm-graph refactor). Its lockfile is gitignored, so no committed lockfile change.

eng/scripts/detect-affected.test.ts
```diff
@@ -0,0 +1,150 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import {
+  CONFIG,
+  computeAffected,
```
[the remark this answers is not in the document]
msyyc: Fixed — the detector tests now run in CI via \
ode --test\ in the detect-affected job, plus a \	est:eng\ root script for local parity.

.github/workflows/ci-downstream.yml
```diff
@@ -0,0 +1,82 @@
...
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup
+
+      - name: Detect affected targets
+        id: detect
+        env:
+          BASE_SHA: ${{ github.event_name == 'merge_group' && github.event.merge_group.base_sha || github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.sha }}
+        run: node eng/scripts/detect-affected.ts
```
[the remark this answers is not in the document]
msyyc: Verified this isn't the case: pnpm resolves the workspace graph from \pnpm-workspace.yaml\ + each \package.json\, not from \
ode_modules\. I tested a 2-package workspace with **zero** node_modules — both \pnpm list -r --json\ and \pnpm --filter \...<pkg>\ list --json\ returned the full/correct graph. The setup action installs the pnpm binary, which is all that's needed, so no \pnpm install\ step is required.

eng/scripts/detect-affected.ts
```diff
@@ -0,0 +1,221 @@
...
+  for (const name of changedPackages) args.push("--filter", `...${name}`);
+  args.push("list", "--depth", "-1", "--json");
+  const out = runPnpm(args);
+  const parsed = JSON.parse(out) as Array<{ name?: string }>;
+  return new Set(parsed.map((p) => p.name).filter((n): n is string => Boolean(n)));
+}
+
+// CLI entry: run only when executed directly (not when imported by tests).
+if (process.argv[1] && fileURLToPath(import.meta.url) === process.argv[1]) {
+  const base = process.env.BASE_SHA;
```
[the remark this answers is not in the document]
msyyc: Node resolves `process.argv[1]` to an absolute path before running (verified: relative invocation yields an absolute `argv[1]` identical to `fileURLToPath(import.meta.url)`), so the guard already fired correctly. Regardless, switched to the canonical URL comparison `import.meta.url === pathToFileURL(process.argv[1]).href` so any separator/format difference is a non-issue. Verified: executes when run directly, stays inert when imported by the tests.

eng/scripts/detect-affected.ts
```diff
@@ -0,0 +1,180 @@
...
+interface Config {
+  /** Globs whose sole change should NOT trigger anything; passed to pnpm via `--changed-files-ignore-pattern`. */
+  ignore: string[];
+  /** Paths that trigger every target (shared CI infrastructure). */
+  sharedExtra: string[];
+  targets: Record<string, Target>;
+}
+
+export const CONFIG: Config = {
+  ignore: ["**/test/**", "**/tests/**", "**/*.md"],
```
timotheeguerin: test also live side by side, we should also ignore `*.test.ts `

eng/scripts/detect-affected.ts
```diff
@@ -0,0 +1,180 @@
...
+    },
+    java: {
+      package: "@azure-tools/typespec-java",
+      extra: [".github/workflows/ci-java.yml", ".github/actions/setup-java/**"],
+      coreSubmodule: true,
+    },
+    typescript: {
+      package: "@azure-tools/typespec-ts",
+      extra: [".github/workflows/ci-typescript.yml"],
+      coreSubmodule: true,
```
timotheeguerin: do we need this? isn't always going to be true?

eng/scripts/detect-affected.ts
msyyc: Good catch

eng/scripts/detect-affected.ts
msyyc: Make sense. `core` module is actually dependency of all downstream modules.

msyyc:
@timotheeguerin updated with your review comment and pls help review again.

.github/workflows/ci-java.yml
```diff
@@ -1,23 +1,11 @@
...
-  pull_request:
-    branches:
-      - main
-      - release/*
-    paths:
-      - "packages/typespec-java/**"
-      - ".github/workflows/ci-java.yml"
-      - ".github/actions/setup-java/**"
-  merge_group:
+  workflow_call:
```
weidongxu-microsoft, line 4: nit, So, this ci-java would now only be triggered by "downstream / CI"? Would it cause CI name change?

CC @XiaofeiCao

msyyc: image

It seems so. After this PR merged, I will create a fake PR to trigger emitter CI independently to make sure it also work for typespec-java-only PR

.github/workflows/ci-downstream.yml
```diff
@@ -0,0 +1,85 @@
+name: "downstream / CI"
```
timotheeguerin:
```suggestion
name: "Integration"
```

suggest this as the name, downstream CI is kinda just an implementation detail

.github/workflows/ci-downstream.yml
```diff
@@ -0,0 +1,85 @@
...
+    uses: ./.github/workflows/ci-java.yml
+
+  typescript:
+    name: "TypeScript"
+    needs: detect-affected
+    if: needs.detect-affected.outputs.typescript == 'true'
+    uses: ./.github/workflows/ci-typescript.yml
+
+  downstream-ci-gate:
+    name: "Downstream CI Gate"
```
timotheeguerin:
```suggestion
    name: "Gate"
```
don't need to repeat makes the final name too long

.github/workflows/ci-downstream.yml
```diff
@@ -0,0 +1,85 @@
...
+    needs: [detect-affected, python, java, typescript]
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate downstream results
```
timotheeguerin, line 67:
```
  - name: Validate CI results
        run: |
          if [[ "${{ needs.core.result }}" == "failure" || "${{ needs.core.result }}" == "cancelled" ]]; then
            echo "Core CI failed or was cancelled"
            exit 1
          fi
          echo "All CI checks passed or were appropriately skipped"

```

this is the code I have in typespec core. might be more flexible if we add more stuff

eng/scripts/detect-affected.test.ts
timotheeguerin, line 1: are those test run? Need to make sure they are included in a vitest project.

Also need to use vitest api not node test

eng/scripts/detect-affected.test.ts
```diff
@@ -0,0 +1,98 @@
...
+import test from "node:test";
+import { CONFIG, computeAffected, matchesAny } from "./detect-affected.ts";
+
+const NONE = new Set<string>();
+
+// --- matchesAny (exact path or `dir/**` prefix) ----------------------------
+
+test("matchesAny: `dir/**` matches the dir and anything under it", () => {
+  const g = ".github/actions/setup/**";
+  assert.equal(matchesAny(".github/actions/setup", [g]), true);
```
timotheeguerin: use vitest `expect`

eng/scripts/detect-affected.test.ts
timotheeguerin: we have a similar need in typespec core repo if you want to copy the setup

eng/scripts/detect-affected.test.ts
```diff
@@ -0,0 +1,98 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import { CONFIG, computeAffected, matchesAny } from "./detect-affected.ts";
+
+const NONE = new Set<string>();
+
+// --- matchesAny (exact path or `dir/**` prefix) ----------------------------
+
+test("matchesAny: `dir/**` matches the dir and anything under it", () => {
```
timotheeguerin: can we also mock console.log so it doesn't affect the test UI

eng/scripts/detect-affected.test.ts
msyyc: Done -- the test now uses `describe`/`it`/`expect` from vitest instead of `node:test` + `node:assert/strict`.

eng/scripts/detect-affected.test.ts
msyyc: Migrated the test to the vitest API and registered `eng/` as a vitest project (new `eng/vitest.config.ts` added to the `projects` list in `vitest.config.fast.ts` + `vitest.config.ts`), mirroring the eng test setup in typespec core. It now runs as part of the normal `pnpm test` suite in `ci.yml` on every PR, so I dropped the redundant `node --test` step from `ci-downstream.yml`. Verified locally: `vitest run --project eng` -> 11/11 pass.

eng/scripts/detect-affected.test.ts
msyyc: Added a `console.log` spy (`vi.spyOn(console, 'log')` in `beforeEach`, restored in `afterEach`). Note the module's CLI/logging block is guarded by an `import.meta.url === pathToFileURL(process.argv[1]).href` check so it never runs on import (the tests only import the pure functions), but mocking keeps the test UI clean defensively.

.github/workflows/ci-downstream.yml
msyyc: Made the gate generic: it validates results via `toJSON(needs)` + `jq`, so no job names are hardcoded here. Adding an emitter (e.g. `go`) then only needs a `go` entry in `detect-affected.config.json`, a `ci-go.yml`, and a `go` job block + `go` in this gate's `needs:`. The `needs:` entry and job block are unavoidable (GitHub requires a static `uses:` and explicit `needs:`), but the gate body never changes.

msyyc:
python ci failure is caused by https://github.com/Azure/typespec-azure/pull/4909

msyyc:
### ⚠️ Follow-up for repo admins: make the downstream CI actually gate merges

This PR adds the downstream emitter CI (`Integration` workflow) with a single aggregating **`Gate`** job (`if: always()`), but **the check is not yet enforced**. The `main` ruleset [**"CI"** (id 598502)](https://github.com/Azure/typespec-azure/rules/598502) currently requires only:

`Build (ubuntu-latest, Node 22.x)`, `Build (ubuntu-latest, Node 24.x)`, `Build (windows-latest, Node 24.x)`, `Check Changes`, `E2E Tests`, `Format`, `Lint`, `Spell check`, `Website`, `license/cla`

None of the downstream contexts (`Gate`, `Python / …`, `Java / …`, `TypeScript / …`, `Go / …`) are in that list, so **a failing emitter CI will not block merge** until the ruleset is updated.

**Action (admin, one-time):** add the single context **`Gate`** to ruleset 598502's `required_status_checks`.

- Require **only `Gate`** — it aggregates every emitter's result and fails if any downstream job failed/was cancelled. This is future-proof: adding a new emitter needs **no** ruleset change.
- Do **not** add the individual emitter jobs (`Go / Mock API Tests`, `Python / …`, etc.) as required. They run conditionally via `if: fromJSON(...).<target>`, so they're skipped on unaffected PRs; requiring them directly leads to fragile skipped/pending-check behavior. `Gate` runs with `if: always()` and handles skips correctly.

Rulesets are GitHub server-side settings (Settings → Rules → Rulesets), not a file in this repo, so this cannot be done via a PR.

======================
Up words are posted by AI but I am not sure whether it is right solution to make the new `Integration / Gate` a required item like other required CI items. @timotheeguerin, do you know how to make our new github action a required check?

timotheeguerin (approved):

.github/workflows/ci-java.yml
XiaofeiCao: It's triggering java ci for go changes: https://github.com/Azure/typespec-azure/pull/4985
image


But it's not triggering for java changes: https://github.com/Azure/typespec-azure/pull/4960
image

PR #4880, opened 2026-07-13
add back no unnamed types

### LINKED ISSUE #4724: Update no-unnamed-unions to all expressions
As nameless expression is coming for all data types we'll want to validate all have a name in azure.
 
Proposing to rename the existing rule to keep it simple

iscai-msft:
fixes #4724 

Replace the `no-unnamed-union` linter rule with `no-unnamed-types` in `@azure-tools/typespec-azure-core`. The new rule flags anonymous models in addition to unnamed unions, walking the type graph from operations to detect anonymous models on the client surface. The `no-unnamed-types` rule has been removed from `@azure-tools/typespec-client-generator-core`.

in `type spec-next`, 94 existing suppressions renamed, 28 net new suppressions added

## Companion PR
- Azure REST API Specs suppressions: https://github.com/Azure/azure-rest-api-specs/pull/44736


packages/typespec-azure-core/src/rules/no-unnamed-types.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    // Anonymous models reachable from the client surface that should be reported.
+    const flaggedModels = new Set<Model>();
+    const visited = new Set<Type>();
+
+    // Unions: flag all unnamed unions (like old no-unnamed-union rule), with exclusions.
+    const excludedUnions = new Set<Union>();
+    const invalidUnions = new Set<Union>();
+
+    function enqueueAndWalk(type: Type | undefined): void {
```
timotheeguerin: why do we need to do a custom walk, can we do it like it was done in union case where we keep track of what to include and exclude and do the diff at the end?

packages/samples/specs/data-plane/widget-manager/main.tsp
```diff
@@ -18,6 +18,7 @@ namespace Contoso.WidgetManager;
 model WidgetManagerApiKeyAuth is ApiKeyAuth<ApiKeyLocation.header, "X-Key">;
 
 /** Authenticate requests with OAuth2 using the widget service scope. */
+#suppress "@azure-tools/typespec-azure-core/no-unnamed-types" "OAuth2 flow configuration"
```
timotheeguerin: I think we should detect that one

timotheeguerin: (it should not flag it)

packages/typespec-azure-core/src/rules/no-unnamed-types.ts
```diff
@@ -0,0 +1,125 @@
...
+          if (!excludedModels.has(model)) {
+            context.reportDiagnostic({ target: model, format: { type: "model" } });
+          }
+        }
+      },
+    };
+  },
+});
+
+function isStandardLibraryType(type: Model | Union): boolean {
```
timotheeguerin: do we actually need this? the linter shouldn'tg report warning outside of the user project

packages/typespec-azure-core/src/rules/no-unnamed-types.ts
```diff
@@ -0,0 +1,125 @@
...
+          }
+        }
+      },
+      operation: (operation) => {
+        // Exclude the top-level return type union (response envelope).
+        if (operation.returnType.kind === "Union") {
+          excludedUnions.add(operation.returnType);
+        }
+      },
+      union: (union) => {
```
timotheeguerin, line 84: we should also add `enum` and `scalar` for the upcoming expression declaration

packages/typespec-azure-core/src/rules/no-unnamed-types.ts
iscai-msft: Done - refactored to use the visitor pattern (include/exclude approach). No more custom walk.

packages/samples/specs/data-plane/widget-manager/main.tsp
iscai-msft: Done - added `isInsideTemplateArgument()` which walks up the AST node tree to detect template argument positions. This correctly excludes OAuth2Auth, ApiKeyAuth, and similar patterns without needing a suppression.

packages/typespec-azure-core/src/rules/no-unnamed-types.ts
iscai-msft: Removed - the linter infrastructure already scopes diagnostics to the user's project.

packages/typespec-azure-core/src/rules/no-unnamed-types.ts
iscai-msft: Done - added enum and scalar detection in the `modelProperty` visitor. They use the same `isInsideTemplateArgument` exclusion.

timotheeguerin (approved):

PR #4884, opened 2026-07-13
fix(arm): refine experimental Agent base type conversation/response items

vayada:
## Summary

Refines the experimental Agent base type (`Azure.ResourceManager.BaseTypes.Agents`) conversation and response item models based on review feedback.

### Changes
- **Remove `conversationId` and `responseId`** from `ConversationProperties` / `ResponseProperties` — these are inferred from the resource's top-level `name` property.
- **Model the message author `role` as an enum** (`MessageRole`) with values `Developer`, `User`, `Assistant`, `Tool` (previously a free-form `string`).
- **Add the `FunctionCall`, `FunctionCallOutput`, and `Compaction` item variants** (in addition to `Message`) to `ConversationItem` and `ResponseItem` via a new `ItemType` discriminator. Variant-specific fields (`role`/`content`, `callId`/`name`/`arguments`, `output`, `summary`) are all **optional** so no single field is forced mandatory across variants.
- The `type` discriminator on `ConversationItem` is **required with no default**, so callers must explicitly declare each item's variant (a default would silently mislabel non-message items as `Message`).

`InputMessage`/`InputTypeProperty`/`ConversationOutput`/`ResponseOutputItem` are consolidated into the new `ConversationItem` and `ResponseItem` models.

### Generated artifacts
- Regenerated the sample OpenAPI snapshot (`packages/samples/.../agent/.../openapi.json`).
- Regenerated the ARM reference docs (`data-types.md`, `index.mdx`).
- Added a Chronus changeset (`fix`).

### Validation
- `lint-typespec-library` compiles clean with `--warn-as-error`.
- All 69 ARM samples regenerate and pass.
- All Agent base-type unit tests pass.


packages/typespec-azure-resource-manager/lib/base-types/agent.tsp
```diff
@@ -269,39 +316,66 @@ model ResponseProperties {
...
-  content?: string;
+  callId?: string;
+
+  /** The name of the function (tool) invoked. Applies to `FunctionCall` items. Read-only. */
+  @visibility(Lifecycle.Read)
+  name?: string;
+
+  /** JSON-encoded arguments passed to the function (tool). Applies to `FunctionCall` items. Read-only. */
+  @visibility(Lifecycle.Read)
+  arguments?: string;
```
markcowl: We shouldn't have json-encoded string values.  This should likely be an arry type or a model type

markcowl (commented):
Also, any needed changes to the linting rules?

packages/typespec-azure-resource-manager/lib/base-types/agent.tsp
vayada: Updated it to the keyed collection of unknown types. Please let me know if it looks good now.

vayada:
> Also, any needed changes to the linting rules?

There are no changes to the linting rules for this. Please let me know if there is need of any.

markcowl (approved):

PR #4887, opened 2026-07-14
fix(benchmark): compare PR-comment baseline at the flat-label level

tadelesh:
## Problem

The benchmark PR comments show phantom emitters `emit/@azure-tools` and `emit/@typespec` that don't exist, and the real emitters' baseline is wrong.

Root cause: to build the rolling baseline, the PR-comment path reconstructed a `RuntimeStats` from the flat metrics stored on the `benchmark-data` branch (`expandRuntimeMetrics`). It split each `emit/...` label on every `/` and used the **second segment** as the emitter name — but a scoped package name contains a slash:

| Flat history label | Parsed as (before) |
|---|---|
| `emit/@azure-tools/typespec-client-generator-core` | emitter `@azure-tools`, step `typespec-client-generator-core` |
| `emit/@typespec/openapi3` | emitter `@typespec`, step `openapi3` |

So it produced a fake `@azure-tools` / `@typespec` emitter and left the real emitter with no baseline.

This is pre-existing (introduced with the rolling-baseline reconstruction), now also visible on the external-spec benchmark comment.

## Fix

The `flatten → reconstruct structure → flatten again` round-trip was the only place that had to *parse* labels, and parsing a name that itself contains `/` is inherently ambiguous. Instead, compare the current run against the baseline **directly at the flat-label level**:

- Labels are opaque keys (`compareFlatMetrics`) — scoped names like `emit/@azure-tools/pkg` are compared as-is, never split.
- The current run's runtime is flattened with `flattenRuntime` (the same function that produces the stored history), so both sides use identical keys.
- `expandRuntimeMetrics` is removed entirely.

The `benchmark-data` storage format (`history.json` / `latest.json`) is **unchanged**, so existing data and the website dashboard are unaffected. The `compare` CLI keeps its structural path (it compares two real runs and never had this bug).

## Validation

Ran the fixed `upload-pr-comment` against the real `benchmark-data` history for both datasets:
- **external-results**: comment shows only `emit/@azure-tools/typespec-client-generator-core` — no phantom `@azure-tools`.
- **results (main)**: comment shows all real emitters (`@azure-tools/typespec-autorest`, `@azure-tools/typespec-python`, `@typespec/http-client-js`, `@typespec/openapi3` + steps) — no phantom `@azure-tools`/`@typespec` rows.

## Tests

Added `test/compare-flat.test.ts`: scoped labels are compared as opaque keys (no phantom emitters), missing values default to 0, and `flattenRuntime` keeps scoped emitter names intact.

`@azure-tools/typespec-benchmark` is private, so no changeset is required.

### DIFF
--- a/packages/benchmark/src/compare.ts
+++ b/packages/benchmark/src/compare.ts
@@ -22,6 +22,25 @@ export function isNotableMetricChange(
   return Math.abs(metric.percentChange) >= threshold && Math.abs(metric.change) >= minChangeMs;
 }
 
+/**
+ * Compare two flat metric maps (label → ms). Labels are treated as opaque keys,
+ * so scoped names like `emit/@azure-tools/pkg` are compared as-is without being
+ * parsed into emitter/step segments. Ordering follows `current` (the logical
+ * order produced by flattenRuntime), with baseline-only labels appended.
+ */
+export function compareFlatMetrics(
+  baseline: Record<string, number>,
+  current: Record<string, number>,
+): MetricComparison[] {
+  const labels = [...Object.keys(current)];
+  for (const label of Object.keys(baseline)) {
+    if (!(label in current)) {
+      labels.push(label);
+    }
+  }
+  return labels.map((label) => createMetric(label, baseline[label] ?? 0, current[label] ?? 0));
+}
+
 function extractRuntimeMetrics(
   baselineRuntime: RuntimeStats,
   currentRuntime: RuntimeStats,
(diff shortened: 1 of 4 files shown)


tadelesh:
Superseded: per discussion with @timotheeguerin, the benchmark PR comments are being removed entirely (we only monitor changes on main via the dashboard). Since this PR fixed the PR-comment baseline logic that is now being deleted, it's no longer needed. Follow-up PR removes the comment feature.

PR #4891, opened 2026-07-14
Add built-in service group extension support

alexchro93:
Service groups are an ARM grouping construct owned by Microsoft.Management (somewhat similar to management groups), but TypeSpec authors previously had to model them manually with `Extension.ExternalResource` to build service-group extension resources. This adds a first-class `Extension.ServiceGroup` target so service-group scoped resources can be modeled the same way as management-group scoped resources.

This change adds the built-in service-group target, teaches ARM resource resolution to classify service-group extension resources with `scope: "ServiceGroup"`, and extends ARM resource identifier metadata to accept `ServiceGroup` as a deployment scope. It also updates the existing specific-extension sample, adds azure-http-specs coverage and mock APIs for service-group extension operations, and refreshes docs/spec summaries so the new surface is discoverable.

Notable review point: this intentionally does not add `ServiceGroupNameParameter` to ARM common types because that parameter does not exist in upstream ARM common-types JSON.


website/src/content/docs/docs/howtos/ARM/resource-type.mdx
```diff
@@ -153,6 +157,50 @@ interface ManagementGroups extends EmplOps<Extension.ManagementGroup> {}
 | `Extension.Subscription`    | Subscription-level scope                  |
 | `Extension.ResourceGroup`   | Resource group-level scope                |
 | `Extension.ManagementGroup` | Management group-level scope              |
+| `Extension.ServiceGroup`    | Service group-level scope                 |
+
+##### Note: Service Group Scope
+
+Service groups are a grouping mechanism allowing you to create flexible, custom groupings of Azure resources. For more information on service groups, including when to use service groups instead of, for example, management groups, please see the official Azure documentation: [https://aka.ms/servicegroups](https://aka.ms/servicegroups).
```
markcowl:
```suggestion
Service groups are a grouping mechanism allowing you to create flexible, custom groupings of Azure resources. For more information on service groups, including when to use service groups instead of management groups, please see the official Azure documentation: [https://aka.ms/servicegroups](https://aka.ms/servicegroups).
```

markcowl (approved):
One nit, otherwise, this looks great

PR #4892, opened 2026-07-14
Add Relationship Base Type

alexchro93:
This PR introduces a base type for namespace Microsoft.Relationships. Relationships are extension resources used to establish a connection between a source ARM resource and target ARM resource.

Notably, two relationship types exist in a public preview state in ARM today: serviceGroupMember and dependencyOf. 

The implementation of the Microsoft.Relationships base type is heavily inspired by the only other existing base type current: agent. 

The relationship type is meant to be a lightweight ARM resources (meaning the resource payload has minimal properties and only really carries references to other ARM resources), however implementations of the relationship base type can new metadata about a relationship through extending the RelationshipMetadata model

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

Copilot-Session: 8a6e07d6-3bd4-4373-b9a2-88ad69127fcf


packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,111 @@
...
+// ============================================================================
+// Resource Templates
+// ============================================================================
+
+/**
+ * Model template for a Relationship extension resource.
+ * Applies the Relationship base type decorator automatically.
+ * @template Properties RP-specific properties for the relationship (must extend RelationshipProperties)
+ */
+@azureBaseType(#{ baseType: "Relationship", version: "2024-06-01" })
```
vayada: Just thinking more about the baseType value and keeping a note for me. Assuming 'Relationship' also becomes the canonical base type, I would consider defining baseType as a constrained enum set having 'Relationship' & 'Agent'. Currently that is modeled as string which allows any string for the baseType. An open-ended string contract may introduce divergence across producers and consumers, making dependency tracking, and long-term schema governance significantly more challenging. I can do this once your PR gets merged.

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,111 @@
...
+  targetTenant: string;
+
+  /** Metadata describing the relationship source and target types. */
+  metadata: Metadata;
+
+  /** Origin information for the relationship. */
+  originInformation?: OriginInformation;
+
+  /** The provisioning state of the relationship. */
+  provisioningState: RelationshipProvisioningState;
```
vayada: If you do not require any resource provider specific provisioning states, consider using the standard ARM provisioningState contract via DefaultProvisioningStateProperty rather than defining a custom provisioning state model with RelationshipProvisioningState.

alexchro93: Updated the relationship base type to use DefaultProvisioningStateProperty. I allowed it to be easily overrideable.

packages/samples/specs/resource-manager/resource-types/relationship/main.tsp
```diff
@@ -0,0 +1,96 @@
...
+  ProvisioningStateProperty = DependencyOfProvisioningStateProperty
+>;
+
+/**
+ * A dependency relationship between a source scope and target entity.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/basetypes-experimental" "Experimental BaseTypes"
+model DependencyOf is Relationship<DependencyOfProperties> {
+  /** The relationship name. */
+  @key("relationshipName")
```
markcowl: use ...ResourceNameParameter  with appropriate properties set to match these values

packages/samples/specs/resource-manager/resource-types/relationship/main.tsp
```diff
@@ -0,0 +1,96 @@
...
+  Provisioning: "Provisioning",
+
+  /** The dependency relationship is being deleted. */
+  Deleting: "Deleting",
+
+  string,
+}
+
+/** Provisioning state property for a Contoso dependency relationship. */
+model DependencyOfProvisioningStateProperty {
```
markcowl: Make this an alias so it doesn't show up in the type graph

markcowl: Actually, I think we should only ask the user to create a ProvisioningState union

packages/samples/specs/resource-manager/resource-types/relationship/main.tsp
```diff
@@ -0,0 +1,96 @@
...
+enum Versions {
+  /** 2024-06-01 version */
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v5)
+  `2024-06-01`,
+}
+
+interface Operations extends Azure.ResourceManager.Operations {}
+
+/** Metadata for a Contoso dependency relationship. */
+model DependencyOfMetadata is RelationshipMetadata {
```
markcowl: This should be an alias, so it doesn't show up in the type graph

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Required properties for a relationship resource.
+ * @template Metadata The relationship metadata model.
+ * @template OriginInformation The relationship origin information model.
+ * @template ProvisioningStateProperty The provisioning state property model. Defaults to DefaultProvisioningStateProperty.
+ */
+model RelationshipProperties<
+  Metadata extends RelationshipMetadata = RelationshipMetadata,
+  OriginInformation extends RelationshipOriginInformation = RelationshipOriginInformation,
+  ProvisioningStateProperty extends TypeSpec.Reflection.Model = DefaultProvisioningStateProperty
```
markcowl: The property name for provisioiingState shouldn't change, not should its optionality or documentation, so we should just ask for the union of states

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
markcowl: One question, is it possible for a resource to be an Agent *and* a Relationship?

The decorator allows this, but we should provide an example if this is a real scenario

packages/typespec-azure-resource-manager/src/rules/use-relationship-required-properties.ts
```diff
@@ -0,0 +1,143 @@
...
+                missing:
+                  "sourceId, sourceTenant, targetId, targetTenant, metadata, provisioningState",
+              },
+              target: relationshipResource.typespecType,
+            });
+            continue;
+          }
+
+          const missing = [
+            "sourceId",
```
markcowl, line 44: I think we should either create templates for these types,  or we should have a type that defines the required properties and a decorator that indicates what is required (property exists, optionality, etc.)  and use this to drive the linting rule

markcowl: Essentially, if we have a `@baseTypeProperty decorator, which took a base type kind and version, we could have an internal-only type (or a set of them)  that defined the required properties.

markcowl (commented):
I also wonder if we want a resource type for this, especially if it is required to be an extension.

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
alexchro93: IMO, I do not think it is a real scenario. A single resource can't be an Agent (tracked type) and Relationship (extension type).

alexchro93:
> I also wonder if we want a resource type for this, especially if it is required to be an extension.

@markcowl, can you please explain what you mean by this?

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Metadata describing the source and target types of a relationship.
+ */
+model RelationshipMetadata {
+  /** The type of the relationship source. */
+  sourceType: string;
+
+  /** The type of the relationship target. */
+  targetType: string;
+}
```
rohitmsft: Do we need source Type and targetType to get present in the base type?

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,84 @@
...
+  sourceId: string;
+
+  /** The tenant identifier of the relationship source. */
+  sourceTenant: string;
+
+  /** The relationship target identifier. */
+  targetId: string;
+
+  /** The tenant identifier of the relationship target. */
+  targetTenant: string;
```
rohitmsft, line 28: Relationship publisher may not know the target resources Tenant Id. Forcing it in baseType may not be helpful for publisher.

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Origin information for a relationship.
+ */
+model RelationshipOriginInformation {
+  /** The origin type. */
+  relationshipOriginType: string;
+
+  /** The discovery engine that found the relationship. */
+  discoveryEngine?: string;
+}
```
rohitmsft: Does RelationshipOriginInformation need to be part of the base type?

I would recommend keeping the base type limited to the minimum set of properties required to define and validate a relationship

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
alexchro93: In other cases in the RP we may not know the source tenant, for example with serviceGroupMember relationships. I think it's okay to keep the four properties and implementation can decide if one is readonly or now.

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
alexchro93: IMO no. We can remove metadata as a whole.

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
alexchro93: IMO no. I can remove origin information as well.

.chronus/changes/add-relationship-base-type-2026-7-10-17-13-0.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Add the experimental Relationship base type for Azure Resource Manager extension resources, including flat relationship properties and lint validation for Relationship base type conformance.
```
markcowl: We should add a short example of usage here - this will go into the release notes and the release blog post

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
```diff
@@ -0,0 +1,46 @@
...
+  baseTypes: BaseTypeInfo[];
+
+  /** The relationship source identifier. */
+  sourceId: string;
+
+  /** The tenant identifier of the relationship source. */
+  sourceTenant: string;
+
+  /** The relationship target identifier. */
+  targetId: string;
```
markcowl, line 25: In the Agent types, we split these into Appliance and UserDefined templates, with the appliance type having readOnly properties.  If we are considering making some of these readOnly, it is easiest to make this part of the Appliance template

packages/typespec-azure-resource-manager/src/rules/arm-relationship-base-type-required-properties.md
```diff
@@ -0,0 +1,41 @@
+---
+title: arm-relationship-base-type-required-properties
```
markcowl: To follow guidelines, we should rename this rule and type to use-relationship-required-properties

packages/typespec-azure-resource-manager/src/rules/arm-relationship-base-type-required-properties.ts
```diff
@@ -0,0 +1,91 @@
+import { Model, Program, createRule, getProperty, paramMessage } from "@typespec/compiler";
+
+import { AzureBaseTypeInfo, getAzureBaseTypes } from "../base-types.js";
+import { getArmResources } from "../resource.js";
+
+export const armRelationshipBaseTypeRequiredPropertiesRule = createRule({
```
markcowl: will need to change this name as well

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
alexchro93: This now uses the baseType enum.

packages/typespec-azure-resource-manager/lib/base-types/relationship.tsp
alexchro93: Some relationship types should mark targetId and targetTenant as readonly while other relationship types mark sourceId and sourceTenant as readonly. For this reason, I didn't specify either set of properties as readonly in this single base class.

Do you think it's better to have a split like agent into a base relationship with readonly source id/tenant and another with readonly target id/tenant?

markcowl: In that case, this makes sense.  In the long term, we may use functions to allow creating the set of types based on up-front options, but for now, this works well.

markcowl (approved):

Issue #4836, opened 2026-07-06
TCGC + consumers adoption of `examples.yaml`


timotheeguerin:
Part of #4831.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/unified-examples-format.md

Adopt the new examples format in the **consumers**.

### Scope
- **TCGC** parses `examples.yaml` and surfaces real example content (all language generators consume examples via TCGC, so TCGC must adopt the format).
- **SDK test generation** consumes resolved examples (via `examples-resolve`) to generate test cases.
- Join Swagger operations to examples via **`x-id`** rather than embedded `$ref`.

Depends on #4834 (resolve) and #4835 (`x-id` emission).

Issue #4833, opened 2026-07-06
examples-migrate tool


timotheeguerin:
Part of #4831.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/unified-examples-format.md

Build **`examples-migrate`**: convert existing `x-ms-examples` JSON into `examples.yaml`.

### Scope
- Read all `x-ms-examples` JSON for a service and emit a single `examples.yaml` (or `examples/<Interface>.yaml` for large services).
- **Dedup across versions** into `since` variants (full restatement, no partial overrides).
- **Normalize embedded version strings to `{api-version}`** in headers and body values (e.g. `Location`, `Azure-AsyncOperation`, `nextLink`) so trivially-versioned variants collapse.
- Emit interface-relative operation keys under a single `$namespace`.

Depends on #4832 (schema).

Issue #4832, opened 2026-07-06
examples.yaml JSON Schema + examples-validate


timotheeguerin:
Part of the Unified Examples Format epic.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/unified-examples-format.md

Publish the **JSON Schema** for `examples.yaml` (at the `$schema` URL) and build the **`examples-validate`** tool.

### Validation rules (RFC §3)
- `$`-prefixed file metadata (`$schema`, `$namespace`); bare top-level keys are always operations.
- Response status codes are **bare integer keys**; `default`/`2XX` ranges are **not allowed**.
- `since` values are **quoted** and **must be a version listed in `service.yaml`**.
- Per lineage: **at most one entry without `since`** (the base) and **unique `since`** values.
- An operation's full lineage lives in **one file**; split large services **by interface** (`examples/<Interface>.yaml`).
- The only supported placeholder is `{api-version}`.

Issue #4831, opened 2026-07-06
Epic: Unified Examples Format


timotheeguerin:
**Epic** tracking the rollout of the **Unified Examples Format** in `Azure/typespec-azure` — a single `examples.yaml` per service that replaces ~282K individual `x-ms-examples` JSON files.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/unified-examples-format.md

Examples are authored once per service in YAML, version-aware via `since`, joined to Swagger operations through an `x-id` extension. Version ordering comes from `service.yaml` (see the [service.yaml epic](https://github.com/Azure/typespec-azure/issues/4825)).

### Early august
- [ ] #4832 — `examples.yaml` JSON Schema + `examples-validate`
- [ ] #4833 — `examples-migrate` tool
- [ ] #4834 — `examples-resolve` tool
- [ ] #4835 — `x-id` emission: `typespec-autorest` + transitional JSON

## August-September
- [ ] #4836 — TCGC + consumers adoption
- [ ] #4837 — Auxiliary tooling: `examples-scaffold` + `examples-diff`

## Early September
- [ ] #4838 — Pilot migration (EventGrid) + CI round-trip
- [ ] #4839 — Migrate all services to `examples.yaml`

## January 2027
- [ ] #4840 — Disable old example tooling
- [ ] #4841 — Delete all `x-ms-examples` JSON files

_Child issues are attached as sub-issues below._

PR #4908, opened 2026-07-15
feat(examples): add examples.yaml JSON schema and examples-validate tool

timotheeguerin:
Part of the **Unified Examples Format** epic (#4831). Implements #4832 — the `examples.yaml` JSON Schema + `examples-validate` tool.

## What

New package **`@azure-tools/typespec-azure-examples`**, the foundational piece of the epic.

- **Schema (TypeSpec → JSON Schema → ajv)** — `schema/examples-yaml.tsp` is the source of truth (RFC §4 model), compiled via `@typespec/json-schema` to `ExamplesYaml.json` and wrapped as `schema.js`, mirroring the existing `service.yaml` schema pattern in `typespec-autorest`. `ajv` (draft 2020-12) validates parsed files against it.
- **`examples-validate` CLI + programmatic API** — discovers `examples.yaml` / `examples/*.yaml` in a service directory, reads the adjacent `service.yaml` for version metadata, reports located diagnostics, and exits non-zero on error (`--warn-as-error` to also fail on warnings).

## Rules enforced (RFC §3)

- Only `$schema`/`$namespace` may be `$`-prefixed; every other bare top-level key is an operation that must be a list of examples.
- Response keys are integer status codes; range keys (`2XX`) and `default` are rejected.
- `since` must be a **quoted** string and a version listed in `service.yaml`.
- Per lineage (entries grouped by `title`; untitled → default lineage): at most one entry without `since`, and `since` values are unique.
- An operation's full example set lives in a single file; each interface appears in exactly one file.
- `{api-version}` is the only supported placeholder, and `api-version` must not appear as a request parameter.

## Notes / scope

- Package name `@azure-tools/typespec-azure-examples` is intended to host the rest of the `examples-*` toolchain (migrate/resolve/scaffold/diff/emit — #4833/#4834/#4835/#4837).
- Deep request/response type-checking against the operation's TypeSpec models is **out of scope** here (→ #4836).
- The top-level schema uses a widened `Record<Example[] | string>` indexer (TypeSpec can't emit `patternProperties`); the tighter `$`-metadata vs operation-key rules are enforced by the semantic layer.
- Publishing the schema to the `$schema` URL in `azure-rest-api-specs` is a follow-up cross-repo step.

## Validation

Build (schema regen + tsc), 24 vitest tests, oxlint, prettier, cspell, and `tsc -b` all pass; CLI verified end-to-end (exit 0 on valid, exit 1 with located diagnostics on invalid).

## Update — `legacyFilename`

The format now carries an optional `legacyFilename` per example so the emitter can re-materialize the original `x-ms-examples` file name during rollout (defaults follow the Azure `<OperationId>.json` convention, so most examples don't need it). A shared naming helper (`defaultLegacyExampleFilename` / `slugify` / `stripJsonExtension`) is exported as the single source of truth reused by the migrate tool and the `typespec-autorest` emitter.


timotheeguerin:
Addressed review feedback:
- **CodeQL** (polynomial regex in `slugify`): the `[^A-Za-z0-9]+` collapse already guarantees no consecutive `_`, so the trailing-trim alternation was linearized to `/^_|_$/` (functionally identical, no backtracking).
- **Robustness**: a present-but-unparseable `service.yaml` used to silently degrade to an empty version list and flood every `since` with `unknown-since-version`. It now emits a single `invalid-service-yaml` warning and skips the membership check. Added `test/discover.test.ts`.

PR #4910, opened 2026-07-16
Add AZC0034 C# type-name conflict linter

haiyuazhang:
## Summary
- add the TCGC `csharp-no-type-name-conflict` linter rule for AZC0034-style reserved C# type-name conflicts
- generate and consume the Azure SDK analyzer reserved-name catalog offline
- add deterministic C# `@@clientName` quick fixes, tests, docs, ruleset registration, and Chronus change entry

## Catalog generation
The checked-in catalog is generated by `eng/scripts/generate-azc0034-catalog.ts` and consumed at lint time from `packages/typespec-client-generator-core/src/rules/azc0034-reserved-type-names.generated.ts`.

The generator fetches the same two source files used by the .NET Azure SDK analyzer from `Azure/azure-sdk-for-net`:

- `sdk/tools/Azure.SdkAnalyzers/src/reserved-type-names.txt`
- `sdk/tools/Azure.SdkAnalyzers/src/reserved-type-qualified-names.txt`

During generation, the script validates that both files have matching line counts, no blank names, sorted and unique simple names, and matching simple-name/qualified-name pairs. It emits a TypeScript map keyed by the reserved C# metadata name, with the conflicting fully qualified type name and package name as diagnostic details. The generated file includes source file paths and SHA256 hashes so future catalog refreshes are traceable.

## Suggested fix generation
When a model, enum, or union has a generated C# name that appears in the catalog, the rule suggests adding a C# `@@clientName` override in `client.tsp`.

The suggested name is built from a contextual prefix plus the conflicting C# name. The prefix comes from the enclosing `@service` namespace when available; otherwise it falls back to the nearest user namespace. The rule uses the C# library name for that namespace, strips a trailing `Client` suffix when present, and normalizes it to PascalCase. For example, `namespace Billing; model Operation {}` suggests `BillingOperation`.

If the first candidate is also reserved, the rule tries stable fallback suffixes in order: no suffix, `Info`, `Data`, then `Type`. This keeps the quick fix deterministic while avoiding suggestions that would still trigger AZC0034.

## Companion spec PRs

- Public specifications: https://github.com/Azure/azure-rest-api-specs/pull/44995
- Private specifications: https://github.com/Azure/azure-rest-api-specs-pr/pull/29699


packages/typespec-client-generator-core/src/rules/azc0034-reserved-type-names.generated.ts
```diff
@@ -0,0 +1,11340 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
timotheeguerin, line 1: that is quite a large file, do we really need every single instance here, are those not namespace that we should ban instead?

packages/typespec-client-generator-core/src/rules/azc0034-reserved-type-names.generated.ts
```diff
@@ -0,0 +1,11340 @@
...
+// Licensed under the MIT License.
+
+// <auto-generated>
```
timotheeguerin, line 4: This is a c# doc comment format, align to JS Doc

eng/scripts/generate-azc0034-catalog.ts
timotheeguerin, line 1: when is this going to be run?

package.json
```diff
@@ -23,6 +23,7 @@
     "format:dir": "prettier --write",
     "format:extra": "pnpm run-all format:extra",
     "format:extra:check": "pnpm run-all format:extra:check",
+    "generate:azc0034-catalog": "tsx eng/scripts/generate-azc0034-catalog.ts",
```
timotheeguerin, line 26: the c# code for is is quite meaningless do not include it in this repo apart from the doc

PR #4921, opened 2026-07-17
Enforce CFSClean for Network Isolation

chidozieononiwu:
This pull request makes a small update to the pipeline template configuration by adjusting the network isolation policy to allow for a more permissive setting.

- Pipeline configuration:
  * [`eng/pipelines/templates/1es-redirect.yml`](diffhunk://#diff-ec46c104f08f2e12cd123355c582e0ec5a62343d3021f69dd7c66e827e127916R31): Added `networkIsolationPolicy: Permissive, CFSClean` to the pipeline settings to modify network access behavior.


nuget.config
timotheeguerin, line 1: I think the recommended casing is `nuget.config` also align with exising ones we have

timotheeguerin: and same comment as for typespec repo, why do we need to force the feed for nuget but not for the other ones?

eng/common/pipelines/templates/steps/auth-dev-feed.yml
timotheeguerin, line 1: does this make sense as the name if its just for python

chidozieononiwu: Yes, this should probably be renamed now that it is used as a shared script. I will do that.

timotheeguerin, line 1: is it not still the same name?

nuget.config
timotheeguerin: still would like to understand why dotnet is the only feed that require to be forced for local developement here

chidozieononiwu: Actually seems like nuget config is not needed.

timotheeguerin (approved):

PR #4972, opened 2026-07-20
Fix typespec-java emitter build hanging in parallel repo build

timotheeguerin:
## Problem

Running a full-repo `pnpm build` intermittently **hangs at the Java emitter build**, even though building directly in `packages/typespec-java` works fine.

## Root cause

`packages/typespec-java/Copy-Sources.ps1` (and `emitter-tests/SyncTests.ps1`) read the emitter/generator sources from the `core/` submodule at the commit pinned in `core-commit.json`. The old `CoreCommit.ps1` did this by **checking that commit out in place** (`Enter-CoreCommit` → `git -C core checkout <pinnedSha>`, `Restore-CoreCommit` → checkout back), mutating the shared `core/` working tree.

During a full `pnpm build`, turbo builds many `core/packages/*` concurrently, and the website's `regen-all-packages-docs` runs `typespec-java`'s `regen-docs` (which re-runs `build:emitter` → `Copy-Sources.ps1`) in parallel with other packages' `regen-docs` that read `core/`. When the pin is newer than the local `core/` checkout, the in-place checkout swaps `core/` files out from under those concurrent readers → intermittent hang. An isolated package build never hits this because nothing else runs concurrently.

## Fix

`CoreCommit.ps1` now reads the pinned sources with `git archive` into a temp directory (`Get-CoreSourceRoot` / `Remove-CoreSourceRoot`) and **never checks out or otherwise touches the `core/` working tree**. When the pin is not newer than the current checkout, it reads the live working tree as before. `git status` on `core/` now always stays clean.

## Verification

- Pin **newer** than `core/` checkout: sources extracted via `git archive`, `core/` left untouched, no temp dirs leaked.
- Pin **older/equal** (normal state): reads live tree, `core/` untouched.
- Full package build (emitter + Maven generator) succeeds, `emitter.jar` produced.

### DIFF
--- a/.chronus/changes/fix-java-emitter-build-core-checkout-race-2026-07-20-14-30-00.md
+++ b/.chronus/changes/fix-java-emitter-build-core-checkout-race-2026-07-20-14-30-00.md
@@ -4,4 +4,4 @@ packages:
   - "@azure-tools/typespec-java"
 ---
 
-Fix the emitter build hanging during a full repo `pnpm build`. `Copy-Sources.ps1` no longer checks out the pinned `core/` commit in place (which mutated the shared submodule working tree and raced with the concurrent monorepo build / `regen-all-packages-docs`); it now reads the pinned sources with `git archive`, leaving the `core/` checkout untouched.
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft (approved):
@XiaofeiCao I approved and will merge soon to unblock.

We can refine later if necessary.

PR #4975, opened 2026-07-20
perf: cache getHttpOperation results across linter rules

iscai-msft:
## Motivation

Feedback from the Shanghai team indicates that linter delays on large ARM spec conversions (e.g., Compute RP with 66 tsp files, 31,831 types) are long enough that engineers sometimes give up waiting for results. As we plan to add more rules, this will worsen.

## Approach

Profiling with `tsp compile --no-emit --stats` on Compute RP shows the linter accounts for ~580ms (42% of total compile time), with the top 5 rules consuming 62% of that:

| Rule | Time |
|------|------|
| response-schema-problem | 91ms |
| no-query-explode | 86ms |
| no-header-explode | 86ms |
| arm-no-path-casing-conflicts | 55ms |
| no-route-parameter-name-mismatch | 43ms |

The root cause: `getHttpOperation()` in `@typespec/http` creates a fresh empty `Map()` as its internal cache on every call (line 37 of operations.ts), so when ~8 rules each call it per-operation, the HTTP operation details (path resolution, parameters, responses, authentication) are recomputed from scratch each time.

This PR adds a module-level `WeakMap<Program, WeakMap<Operation, HttpOperation>>` cache in `packages/typespec-azure-core/src/rules/utils.ts`. The first rule to request an operation's HTTP details pays the computation cost; subsequent rules get the cached result.

Additionally, `createMetadataInfo()` in the response-schema-problem rule is hoisted to be created once per lint run instead of once per operation visit.

## Changes

- **`packages/typespec-azure-core/src/rules/utils.ts`**: Added `getCachedHttpOperation()` utility with program-scoped WeakMap cache
- **8 rule files updated** to use the cached version: no-query-explode, no-header-explode, response-schema-multi-status-code, no-route-parameter-name-mismatch, byos, no-error-status-codes, use-standard-names, rpc-operation-request-body

## Testing

All 26 existing rule tests pass. The cache is transparent to rule behavior -- it only avoids redundant recomputation.

### DIFF
--- a/packages/typespec-azure-core/src/rules/byos.ts
+++ b/packages/typespec-azure-core/src/rules/byos.ts
@@ -1,11 +1,10 @@
 import {
   Operation,
   createRule,
-  ignoreDiagnostics,
   isTemplateInstance,
   paramMessage,
 } from "@typespec/compiler";
-import { getHttpOperation } from "@typespec/http";
+import { getCachedHttpOperation } from "./utils.js";
 
 const binaryContentTypes = new Set(["application/octet-stream", "multipart/form-data"]);
 export const byosRule = createRule({
@@ -23,7 +22,7 @@ export const byosRule = createRule({
           // Operation template instance are just referenced in `op is`. Main operation will be validated.
           return;
         }
-        const httpOperation = ignoreDiagnostics(getHttpOperation(context.program, operation));
+        const httpOperation = getCachedHttpOperation(context.program, operation);
         if (httpOperation.parameters.body !== undefined) {
           for (const contentType of httpOperation.parameters.body.contentTypes) {
             if (binaryContentTypes.has(contentType)) {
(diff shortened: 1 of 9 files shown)


packages/typespec-azure-core/src/rules/byos.ts
timotheeguerin, line 1: for this one I was hoping to fine a more core compiler solution. the Issue with doing this is one rule will be the one taking the hit for performance and it might create confusing benchmark over time if that rule is not always the same.

What I was thinking is that the `program` expose what stage it is in. `getHttpOperation` would then enable the cache if the stage is >= validate/we provide some built in caching system for the compilerthat does that automatically.

timotheeguerin: if we do that the http validator will always run first and be the one taking the hit

iscai-msft:
Closing in favor of a core compiler solution -- will add built-in caching to getHttpOperation in @typespec/http so all consumers benefit automatically without per-rule caching.

packages/typespec-azure-core/src/rules/byos.ts
iscai-msft: yes, I definitely agree, let me shift gears and just make the changes to `getHttpOperation` caching instead of doing it more rule-focused

PR #4976, opened 2026-07-20
perf: optimize ARM linter rules to avoid redundant HTTP operation resolution

iscai-msft:
## Motivation

Follow-up to #4975. After optimizing the azure-core rules (which cached `getHttpOperation()` calls), the ARM-specific rules are now the next set of bottlenecks on large specs like Compute RP.

Profiling with `tsp compile --no-emit --stats` on Compute RP shows these ARM rules in the top contributors:

| Rule | Before | After |
|------|--------|-------|
| arm-common-types-version | 29ms | 0ms |
| arm-no-path-casing-conflicts | 55ms | 30ms |
| arm-resource-operation-response | 24ms | ~same (eliminates redundant getHttpOperation calls) |

### 1. `arm-common-types-version`: Use `listServices()` instead of `getAllHttpServices()`

This rule only needs service namespaces to check for the `@armCommonTypesVersion` decorator. It was calling `getAllHttpServices()` which resolves ALL HTTP operations (paths, parameters, responses, authentication) for every operation in every service -- expensive work that this rule never uses.

Replaced with `listServices()` from `@typespec/compiler` which returns just the service namespace metadata without resolving HTTP operations. This drops the rule from ~29ms to 0ms.

### 2. `arm-no-path-casing-conflicts`: Cache `getAllHttpServices()` result

Added a `getCachedHttpServices()` utility in `utils.ts` using a `WeakMap<Program, HttpService[]>`. This caches the expensive `getAllHttpServices()` result so that if any other rule needs it in the future, the work is not repeated.

### 3. `arm-resource-operation-response`: Use cached ARM resource data

The rule was calling `resolveResourceOperations(program, model)` per ARM resource model, which internally calls `getHttpOperation()` for each lifecycle operation. Since `getArmResources()` already resolves and caches all resource operations (including HTTP operation details), the rule now builds a lookup map from the cached data instead of re-resolving per model.

## Testing

All 202 ARM rule tests pass (42 test files).

### DIFF
--- a/packages/typespec-azure-resource-manager/src/rules/arm-common-types-version.ts
+++ b/packages/typespec-azure-resource-manager/src/rules/arm-common-types-version.ts
@@ -1,5 +1,4 @@
-import { Program, SemanticNodeListener, createRule } from "@typespec/compiler";
-import { getAllHttpServices } from "@typespec/http";
+import { listServices, Program, SemanticNodeListener, createRule } from "@typespec/compiler";
 import { getVersion } from "@typespec/versioning";
 import { getArmProviderNamespace } from "../namespace.js";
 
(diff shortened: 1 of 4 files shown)


iscai-msft:
Closing in favor of a core compiler solution -- will add built-in caching to getHttpOperation in @typespec/http.

PR #4981, opened 2026-07-21
Fix ResponseAsBool_HeadAsBoolean scenario coverage failing

weidongxu-microsoft:
## Problem

The `Azure_ClientGenerator_Core_ResponseAsBool_HeadAsBoolean` scenario fails coverage for **all languages**, even when the client behaves correctly (verified: Java `ResponseAsBoolTests` assertions pass, but the Spector coverage file reports `... _HeadAsBoolean => fail`).

### Root cause

The mockapi used `passOnSuccess` for a two-endpoint scenario where `notExists` intentionally returns **404** (to exercise the `@responseAsBool` false path). `passOnSuccess` sets `passCondition: "response-success"`, which the coverage tracker evaluates as `checkAll(x => x.status >= 200 && x.status < 300)` — so the 404 endpoint can never pass. `passOnCode` can't help either, since it forces a single code for all endpoints while this scenario mixes 204 and 404.

Secondary: `@route("/not-exists")` was nested under the namespace's `@route("/exists")`, resolving to `/exists/not-exists` (the `@scenarioDoc` claimed `/not-exists`).

## Fix

Split into two scenarios:

- `..._HeadAsBoolean_exists` → `passOnSuccess` (HEAD `/exists` -> 204 -> true)
- `..._HeadAsBoolean_notExists` → `passOnCode(404)` (HEAD `/not-exists` -> 404 -> false)

Also corrects the `notExists` route to `/not-exists` and regenerates `spec-summary.md`.

> Note: downstream language repos referencing the old scenario name `Azure_ClientGenerator_Core_ResponseAsBool_HeadAsBoolean` will need to update to the two new names.


packages/azure-http-specs/specs/azure/client-generator-core/response-as-bool/main.tsp
```diff
@@ -22,23 +22,28 @@ model Resource {
...
+  @scenarioDoc("""
+    Test that a HEAD operation decorated with @responseAsBool returns false for a 404 response.
+    
+    Expected call:
+    - HEAD /azure/client-generator-core/response-as-bool/not-exists -> 404 (returns false)
+    """)
+  @route("/not-exists")
   @global.Azure.ClientGenerator.Core.responseAsBool
   @head
-  @route("/not-exists")
```
weidongxu-microsoft, line 42: I will ask agent to revert this path change (this is `/exists/not-exists`).

Taking the chance to see whether downstream CI works (downstream should fail, due to the path change, which should cause generated code change).

msyyc (approved):

PR #4982, opened 2026-07-21
Add reference documentation for TCGC diagnostics

### LINKED ISSUE #4478: Write more detailed document for diagnostics in TCGC
We should provide separated docs for each diagnostic, explaining including but not limited to:
1. why we have this diagnostic
2. how this diagnostic is triggered to report
3. what you should do to "fix" it

By adding this doc, we should make the customers more easier to self-serve

tadelesh:
## Summary

Adds auto-generated reference documentation for all `@azure-tools/typespec-client-generator-core` (TCGC) diagnostics, and cleans up / tightens the diagnostic set along the way.

Each diagnostic has a source markdown file under `packages/typespec-client-generator-core/src/diagnostics/*.md` that `tspd doc` turns into a reference page (mirroring the linter-rule docs). The generated website pages are gitignored; the markdown files are the source of truth.

## Diagnostic reference docs (60 docs)

Every diagnostic doc follows a consistent structure (all `h2` sections; multi-case diagnostics nest their examples at `h3`):

- **Impact** — what the diagnostic affects (SDK surface / tooling, and language where relevant), and what it does not affect.
- **❌ Incorrect Usage** — the triggering TypeSpec (or config), with the offending line marked by an inline comment.
- **Diagnostic Message** — the actual message text from `lib.ts`.
- **✅ How to Fix** — the corrected TypeSpec / config.
- **Suppression** — only for suppressible warnings; warnings that must be fixed (not suppressed) say so explicitly, and errors have no Suppression section.

Docs and examples are grounded in each diagnostic's real report site and, where available, in the corresponding test.

Wiring: a `doc(code)` helper in `lib.ts` sets both `docs` (source markdown) and `url` (the published reference page, so editors can surface it via the diagnostic code's `codeDescription`).

## Diagnostic set cleanup

- Removed diagnostics with no report site: `client-service`, `api-version-not-string`, `duplicate-decorator`.
- Replaced invariant checks with `compilerAssert` (and removed the diagnostics): `discriminator-not-constant`, `discriminator-not-string`, `unexpected-http-param-type`, `unexpected-pageable-operation-return-type`.
- Removed the never-reported `default` message from `client-name-ineffective` (kept the single `override` message).
- Kept `no-emitter-name` and `invalid-client-service-multiple` (both are reachable / tested).

## Severity changes

Raised from `warning` to `error`: `inconsistent-multiple-service-dependency`, `missing-scope`, `mark-as-lro-ineffective`, `client-name-ineffective`, `missing-service-versions`. Specs that previously compiled with only these warnings will now fail.

## Validation

- Build, lint (`oxlint`) and `prettier` pass.
- Full TCGC test suite passes (affected tests updated for the new severities and message change).
- `tspd` regeneration produces one page per diagnostic with no `documentation-missing` warnings: `pnpm --filter @azure-tools/typespec-client-generator-core regen-docs`.

Resolve: https://github.com/Azure/typespec-azure/issues/4478


.gitignore
```diff
@@ -259,3 +259,8 @@ packages/typespec-java/emitter-tests/tsp-spector-coverage-*.json
 # each rule's `docs` field). Regenerated during the website build (build:web), so
 # they are not tracked. The rule source docs live in packages/*/src/rules/*.md.
 website/src/content/docs/docs/libraries/*/rules/
+
+# Generated diagnostic reference pages (produced by `tspd doc` / regen-docs via
+# each diagnostic's `docs` field). Regenerated during the website build (build:web),
+# so they are not tracked. The diagnostic source docs live in packages/*/src/diagnostics/*.md.
+website/src/content/docs/docs/libraries/*/reference/diagnostics/
```
tadelesh: Merge the comment with the linter one.

packages/typespec-client-generator-core/src/diagnostics/add-parameter-duplicate.md
```diff
@@ -0,0 +1,22 @@
...
+
+@service
+namespace MyService {
+  op myOp(name: string): void;
+}
+
+model ExtraParams {
+  name: string;
+}
+#suppress "experimental-feature" "customizing generated parameters"
```
tadelesh:
```suggestion
```

packages/typespec-client-generator-core/src/diagnostics/api-version-not-string.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when an API version value is not a string and is not a member of a string enum.
+
+To fix this issue, define API versions as strings or as enum members with string values.
```
tadelesh: This diagnostic is not used anymore. Remove it.

packages/typespec-client-generator-core/src/diagnostics/nested-client-service-not-subset.md
```diff
@@ -0,0 +1,20 @@
+This diagnostic is issued when a nested `@client` specifies services that are not a subset of the parent client's services.
+
+To fix this issue, restrict the nested client's `service` option to parent services, or omit the option to inherit from the parent.
```
tadelesh: The reason and example is not correct. We only allow sub client's service is a subset of parent client. The example's original service namespace should be named as one & two. and the parent @client should include both services, that's the correct usage.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,25 @@
+This diagnostic is issued when `@clientLocation` requests a move that conflicts with the client model TCGC is building.
```
tadelesh: You should mention all kinds of conflict separately. List them one by one, and if the one could provide example, provide one. Basic case (the example current shows) does not need example.

packages/typespec-client-generator-core/src/diagnostics/client-option.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when `@clientOption` is used.
+
+To fix this issue, remove `@clientOption` unless it is necessary; if necessary, suppress the diagnostic with a clear temporary-workaround justification.
```
tadelesh: Since client option is a temporary thing for language emitter. This diagnostic will always be added. Tell the user to double confirm if this is a language behavior and always add suppression if yes.

packages/typespec-client-generator-core/src/diagnostics/client-service.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when a client declared with `@client` is not inside a namespace decorated with `@service`, so TCGC cannot associate the client with a service.
```
tadelesh: Not used anymore. Remove this diagnostic.

packages/typespec-client-generator-core/src/diagnostics/discriminator-not-constant.md
```diff
@@ -0,0 +1,31 @@
+This diagnostic is issued when a derived model in a discriminated hierarchy defines the discriminator property with a non-constant type. TCGC needs a fixed discriminator value to identify each subtype.
```
tadelesh: This should not happen. Remove `discriminator-not-constant` and `discriminator-not-string` diagnostics and replace the current check logic with `compilerAssert`.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name-warning.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when two generated SDK declarations have the same client name in a scope where TCGC reports a warning instead of an error, such as C# operation overload scenarios.
+
+To fix this issue, prefer unique generated names or explicit `@clientName` values unless the duplicate is intentional for that emitter.
```
tadelesh: This is a bydesign diagnostic for C# since they want to do overload for such cases. Make sure author is intentional to do so and alwasys add suppression.

packages/typespec-client-generator-core/src/diagnostics/duplicate-decorator.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when the same decorator is used more than once on the same declaration with the same scope.
```
tadelesh: Not used anymore, removed.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when a value from an example file cannot be matched to the TypeSpec definition for the operation.
+
+To fix this issue, update the example value or the TypeSpec operation so parameters, request bodies, and responses follow the declared shapes.
```
tadelesh:
```suggestion
To fix this issue, update the example value or the TypeSpec definition so parameters, request bodies, and responses follow the declared shapes.
```

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when a value from an example file cannot be matched to the TypeSpec definition for the operation.
```
tadelesh:
```suggestion
This diagnostic is issued when a value from an example file cannot be matched to the TypeSpec definition.
```

packages/typespec-client-generator-core/src/diagnostics/flatten-polymorphism.md
```diff
@@ -0,0 +1,24 @@
+This diagnostic is issued when `@flattenProperty` is applied to a model property whose type is polymorphic through a discriminator.
+
+To fix this issue, remove `@flattenProperty` from polymorphic properties or change the property to use a non-polymorphic type.
```
tadelesh: Most languages do not support flatten on polymorphic type. So, this is not supported.

packages/typespec-client-generator-core/src/diagnostics/inconsistent-multiple-service-dependency.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when services merged into the same client resolve different versions of a shared dependency namespace.
```
tadelesh, line 1: This needs an example.

packages/typespec-client-generator-core/src/diagnostics/invalid-client-service-multiple.md
```diff
@@ -0,0 +1,17 @@
+This diagnostic is issued when `@client` is applied to an interface with multiple services in the `service` option.
```
tadelesh, line 1: Check why we need this diagnostic. Why interface could not be multiple service?

packages/typespec-client-generator-core/src/diagnostics/no-emitter-name.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when TCGC cannot determine the emitter name from its options. The emitter name is needed to resolve language-specific scopes and naming rules.
```
tadelesh: This should never happen. Change to use compileAssert.

packages/typespec-client-generator-core/src/diagnostics/required-parameter-scoped-out.md
```diff
@@ -0,0 +1,18 @@
+This diagnostic is issued when a required HTTP parameter is excluded from the current emitter by `@scope`.
+
+To fix this issue, keep required parameters in scope, make the parameter optional when appropriate, or provide it through another reliable mechanism such as a custom header policy.
```
tadelesh: This is usually not allowed. Let author confirm it is intentional and always add suppression if intentional.

packages/typespec-client-generator-core/src/diagnostics/unexpected-http-param-type.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when an HTTP parameter is expected to have one kind, such as `body`, `path`, `query`, or `header`, but resolves to a different kind.
```
tadelesh: Can this happen? If not, a compileAssert is more reaonable.

packages/typespec-client-generator-core/src/diagnostics/unexpected-pageable-operation-return-type.md
```diff
@@ -0,0 +1,3 @@
+This diagnostic is issued when an operation marked pageable with `@list` does not return a valid paging model, or paging metadata cannot be resolved from the response.
```
tadelesh: Can this happen? If not, a compileAssert is more reaonable.

packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -142,142 +142,144 @@ const TCGCEmitterOptionsSchema: JSONSchemaType<TCGCEmitterOptions> = {
...
   name: "@azure-tools/typespec-client-generator-core",
   diagnostics: {
-    "client-service": {
-      severity: "warning",
-      messages: {
-        default: paramMessage`Client "${"name"}" is not inside a service namespace. Use @client({service: MyServiceNS})`,
-      },
-    },
     "union-null": {
+      docs: fileRef.fromPackageRoot("src/diagnostics/union-null.md"),
```
timotheeguerin: do you also need to add the url so it can reference it in ide?

 we might want to create an helper so you can just do
```
...doc("union-null")
```

packages/typespec-client-generator-core/src/diagnostics/add-parameter-duplicate.md
tadelesh: Done - removed the `#suppress` line from the example.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
tadelesh: Applied.

packages/typespec-client-generator-core/src/diagnostics/client-service.md
tadelesh: Removed this diagnostic - it had no report site anywhere in the source.

packages/typespec-client-generator-core/src/diagnostics/invalid-client-service-multiple.md
tadelesh: Kept. This fires when `@client({service: [A, B]})` targets an **interface**. Multi-service clients must be namespaces because the merge builds a per-service sub-client hierarchy that an interface cannot host; it also has a dedicated validation test (`invalid-client-service-multiple on interface`). Allowing multi-service interfaces would be a separate design change - let me know if you want that.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Reworked - each conflict kind is now listed separately (string target with multiple root clients, operation-to-operation, model-property-to-string, parameter-name collision with client initialization, same-parameter-different-types). Dropped the basic operation-to-operation example and kept an example only for the parameter-type conflict.

packages/typespec-client-generator-core/src/diagnostics/discriminator-not-constant.md
tadelesh: Done - removed both `discriminator-not-constant` and `discriminator-not-string` and replaced the checks in `types.ts` with `compilerAssert`.

packages/typespec-client-generator-core/src/diagnostics/api-version-not-string.md
tadelesh: Removed this diagnostic - it had no report site anywhere in the source.

packages/typespec-client-generator-core/src/diagnostics/flatten-polymorphism.md
tadelesh: Reworded - the doc now states flattening a polymorphic type is not supported because most languages cannot represent it.

packages/typespec-client-generator-core/src/diagnostics/duplicate-decorator.md
tadelesh: Removed this diagnostic - it had no report site anywhere in the source.

packages/typespec-client-generator-core/src/diagnostics/inconsistent-multiple-service-dependency.md
tadelesh: Added an example (two services merged into one client pinning a shared library to different versions).

packages/typespec-client-generator-core/src/diagnostics/client-option.md
tadelesh: Reworded - the doc now says this is always reported because client options are a temporary language-emitter mechanism, and advises confirming it is intended language behavior and suppressing if so.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
tadelesh: Applied.

packages/typespec-client-generator-core/src/diagnostics/required-parameter-scoped-out.md
tadelesh: Reworded - the doc now says this is usually not allowed and advises confirming the intent and suppressing when the value is supplied another way.

packages/typespec-client-generator-core/src/diagnostics/unexpected-pageable-operation-return-type.md
tadelesh: Converted to `compilerAssert` and removed the diagnostic.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name-warning.md
tadelesh: Reworded - the doc now states this is by design for languages like C# that use the duplicated name for overloads, and advises confirming intent and always suppressing.

packages/typespec-client-generator-core/src/diagnostics/unexpected-http-param-type.md
tadelesh: Converted both call sites to `compilerAssert` and removed the diagnostic.

packages/typespec-client-generator-core/src/diagnostics/nested-client-service-not-subset.md
tadelesh: Fixed - the parent `@client` now includes both services (`service: [ServiceOne, ServiceTwo]`) and the correct sub client uses a subset (`service: ServiceOne`); the incorrect example references `ServiceThree` which the parent does not include.

packages/typespec-client-generator-core/src/diagnostics/no-emitter-name.md
tadelesh: I tried converting this to `compilerAssert`, but it broke `context.test.ts`: `createTCGCContext` is legitimately called without an emitter name (generic/tooling/language-server contexts where `program.emitters[0]?.metadata?.name` is undefined), so it is reachable and must stay a diagnostic. Reverted and kept it.

.gitignore
tadelesh: Done - merged the diagnostic and linter gitignore entries into a single comment block.

packages/typespec-client-generator-core/src/diagnostics/auto-merge-service-conflict.md
```diff
@@ -0,0 +1,38 @@
...
+    name: "ChildClient",
+    service: ServiceA,
+  })
+  namespace Child {
+
+  }
+}
+```
+
+The parent client is auto-merging services while the child declares its own service; omit the child `service` option or remove `autoMergeService`.
```
tadelesh: I don't know if this example is clear. Check the test code to see if there is a better example.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,34 @@
...
+
+alias WithPreviewHeader<T extends FeatureOptInKeys> = {
+  @clientLocation(Default)
+  @header("x-preview-features")
+  previewFeatures: T;
+};
+
+op getInsights(...WithPreviewHeader<FeatureOptInKeys.insights>): void;
+op getSchedules(...WithPreviewHeader<FeatureOptInKeys.schedules>): void;
+```
```
tadelesh, line 149: Put `To fix this issue` and `example` to different cases. not show them in one place.

packages/typespec-client-generator-core/src/diagnostics/client-option.md
```diff
@@ -0,0 +1,19 @@
...
+@service
+namespace MyService;
+
+@clientOption("enableFeatureFoo", true, "python")
+model Test {
+  id: string;
+}
+```
+
+`@clientOption` always reports this diagnostic because it is experimental; remove it or suppress the warning with a temporary-workaround justification.
```
tadelesh: Add suppression in the example.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name-warning.md
```diff
@@ -0,0 +1,23 @@
...
+  @route("/list")
+  list(): void;
+
+  @clientName("list", "csharp")
+  @route("/listByParent")
+  listByParent(parent: string): void;
+}
+```
+
+For the C# emitter, both operations resolve to the client name `list`; choose a distinct `@clientName` unless the overload-style duplicate is intentional.
```
tadelesh: Add suppression in the example.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
```diff
@@ -0,0 +1,14 @@
...
+### Example
+
+```typespec
+@service
+namespace TestClient {
+  op get(): string;
+}
+```
+
+In the test, TCGC is run with `examples-dir` set to `./examples`, but that directory does not exist, so example loading is skipped for the service.
```
tadelesh: This example is meaningless. Remove it.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
```diff
@@ -0,0 +1,22 @@
...
+  @route("/{b}")
+  op parametersDiagnostic(
+    @header a: string,
+    @path b: string,
+    @query c: string,
+    @body d: string,
+  ): void;
+}
+```
+
```
tadelesh, line 24: You need also put the json example here to show the exact problem.

packages/typespec-client-generator-core/src/diagnostics/required-parameter-scoped-out.md
```diff
@@ -0,0 +1,18 @@
...
+using Azure.ClientGenerator.Core;
+
+op getWidget(
+  @header("x-client-id")
+  @scope("!python")
+  clientId: string,
+): void;
+```
+
+When generating for Python, the required `clientId` header is scoped out; keep it in scope, or confirm the intent and suppress the diagnostic if the value is supplied another way.
```
tadelesh: Add suppression to the example.

packages/typespec-client-generator-core/src/diagnostics/unsupported-generic-decorator-arg-type.md
```diff
@@ -0,0 +1,16 @@
...
+The test that exercises this diagnostic compiles a service and asks TCGC to include the generic `TypeSpec.@service` decorator in emitted decorator metadata:
+
+```typespec
+@service
+namespace TestService {
+  op test(): void;
+}
+```
+
+The unsupported argument shape is encountered while parsing that decorator metadata; avoid exposing generic decorator arguments that TCGC cannot convert to SDK decorator metadata.
```
tadelesh: Remove this useless example.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Split them - the per-case fixes stay in the `To fix this issue` section and the `### Example` now only shows the triggering code for the most common (parameter-type-conflict) case.

packages/typespec-client-generator-core/src/diagnostics/auto-merge-service-conflict.md
tadelesh: The example is grounded in the `auto-merge service conflict with nested client specifying services` test (structure.test.ts:2831); added an inline comment marking the conflicting child `service` line to make it clearer.

packages/typespec-client-generator-core/src/diagnostics/unsupported-generic-decorator-arg-type.md
tadelesh: Removed the example - it did not actually demonstrate the trigger.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name-warning.md
tadelesh: Added `#suppress` to the example.

packages/typespec-client-generator-core/src/diagnostics/client-option.md
tadelesh: Added `#suppress` to the example.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
tadelesh: Added the JSON example (the `parameters` object uses `test` which maps to no declared parameter) so the mismatch is explicit.

packages/typespec-client-generator-core/src/diagnostics/required-parameter-scoped-out.md
tadelesh: Added `#suppress` to the example.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
tadelesh: Removed the example.

packages/typespec-client-generator-core/src/diagnostics/add-parameter-duplicate.md
```diff
@@ -0,0 +1,19 @@
...
+```typespec
+@service
+namespace MyService {
+  op myOp(name: string): void;
+}
+
+model ExtraParams {
+  name: string;
+}
+alias Modified = addParameter(MyService.myOp, ExtraParams.name);
```
lirenhe, line 19: Could you add the comments here that this line would cause t he diagnostic error.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,32 @@
...
+}
+
+alias WithPreviewHeader<T extends FeatureOptInKeys> = {
+  @clientLocation(Default)
+  @header("x-preview-features")
+  previewFeatures: T;
+};
+
+op getInsights(...WithPreviewHeader<FeatureOptInKeys.insights>): void;
+op getSchedules(...WithPreviewHeader<FeatureOptInKeys.schedules>): void;
```
lirenhe, line 148: Could you add a comment about which line would cause problem and provide a suggested fix?
The part of 'to fix this issue' is a bit difficult to understand.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,111 @@
...
+- **Moving a model property to a string target** — a model property can only be moved to an interface or namespace, not to a string-named target.
+- **Parameter name already used in client initialization** — the parameter produced by the moved model property collides with an existing client initialization parameter.
+- **Same parameter moved with different types** — the same parameter name is moved to one client with conflicting types, which commonly happens when `@clientLocation` is applied to a templated parameter that is instantiated with different types.
+
+## Impact
+
+- **Area:** Client operation and parameter placement. Generation continues, but the requested `@clientLocation` move cannot be applied safely to the generated client structure.
+- **Not affected:** HTTP routes, parameter wire names, and service operation definitions are unchanged.
+
+### String target with multiple root clients
```
tadelesh: These four cases (string-target-with-multiple-root-clients, operation-to-operation, model-property-to-client-initialization, model-property-to-string) are **Diagnostic Message only** — only the `Moved parameters with conflicting types` case (below) has a grounded `Incorrect Usage` / `How to Fix`, because it is the only one with a test. Should I add examples for the other four cases too, or is message-only acceptable for them?

packages/typespec-client-generator-core/src/diagnostics/invalid-access.md
```diff
@@ -0,0 +1,33 @@
+This diagnostic is issued when the value passed to the `@access` decorator is not one of the allowed access levels.
+
+## Impact
+
+- **Area:** SDK visibility overrides. Blocks the invalid `@access` value from changing generated public/internal visibility.
+- **Not affected:** Model structure, operation signatures, and wire payloads are unchanged.
+
+#### ❌ Incorrect Usage
```
tadelesh: Heads up: this `Incorrect Usage` example is **inferred**, not taken from a test — there is currently no test that triggers `invalid-access`. Please sanity-check that `@access(CustomAccess.private)` is a faithful trigger.

packages/typespec-client-generator-core/src/diagnostics/invalid-usage.md
```diff
@@ -0,0 +1,33 @@
+This diagnostic is issued when the value passed to the `@usage` decorator is not one of the supported usage flags.
+
+## Impact
+
+- **Area:** SDK model usage overrides. Blocks unsupported explicit usage flags from being added to generated model, enum, or union metadata.
+- **Not affected:** Usage inferred from actual operation inputs and outputs is still calculated separately.
+
+#### ❌ Incorrect Usage
```
tadelesh: Same caveat as `invalid-access`: this example is **inferred** (no test triggers `invalid-usage`). Please confirm the trigger shape is accurate.

packages/typespec-client-generator-core/src/diagnostics/empty-client-namespace.md
```diff
@@ -0,0 +1,33 @@
+This diagnostic is issued when `@clientNamespace` is given an empty or whitespace-only value.
+
+## Impact
+
+- **Area:** SDK namespace customization. Generation continues, but the empty `@clientNamespace` override cannot place generated declarations in a custom namespace.
+- **Not affected:** Model names, operation routes, and payload serialization are unchanged.
+
+#### ❌ Incorrect Usage
```
tadelesh: This example is **inferred** — no test triggers `empty-client-namespace`. Please confirm `@clientNamespace("")` (empty value) is the right trigger.

packages/typespec-client-generator-core/src/diagnostics/no-emitter-name.md
```diff
@@ -0,0 +1,26 @@
+This diagnostic is issued when TCGC cannot determine the emitter's language because no usable emitter name is available (for example, the emitter name is missing or does not match the expected `*-<language>` pattern).
+
+## Impact
```
tadelesh, line 3: Note on Impact/Suppression: this diagnostic only fires when TCGC runs **without a real emitter** (tooling / language server / tests where `program.emitters` is empty). It is kept as a warning with a Suppression section on that basis — flag if you'd rather frame it differently (it is not something a normal spec author hits).

packages/typespec-client-generator-core/src/diagnostics/api-version-undefined.md
```diff
@@ -0,0 +1,40 @@
...
+@service(#{ title: "Contoso Widget Manager" })
+@versioned(Contoso.WidgetManager.Versions)
+namespace Contoso.WidgetManager;
+
+enum Versions { // configured api-version `v4` is not defined here
+  v1,
+  v2,
+  v3,
+}
+```
```
tadelesh: need to add the example of the tspconfig.yaml config of api-version

packages/typespec-client-generator-core/src/diagnostics/api-version-undefined.md
```diff
@@ -0,0 +1,40 @@
...
+#### ✅ How to Fix
+
+Set `api-version` to a service version that exists, or use `latest` or `all` when those behaviors are intended.
+
+## Suppression
+
+Suppress this warning only if falling back to the latest service version is the intended behavior for the configured emitter run.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/api-version-undefined" "latest version fallback is intentional"
```
tadelesh: This should not be suppressed. Need to fix the config.

packages/typespec-client-generator-core/src/diagnostics/add-parameter-duplicate.md
```diff
@@ -0,0 +1,32 @@
...
+
+For the declaration above, TCGC reports:
+
+```text
+Parameter "name" already exists in operation "myOp".
+```
+
+#### ✅ How to Fix
+
+Choose a unique parameter name or use `replaceParameter` when the intent is to replace an existing parameter.
```
tadelesh, line 32: Give the correct tsp. Apply for all docs that have code example.

packages/typespec-client-generator-core/src/diagnostics/add-parameter-duplicate.md
tadelesh: Done - the Incorrect Usage example now has an inline comment marking the offending line (the 
ame property that duplicates the existing myOp parameter).

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Done - the Incorrect Usage example now marks the offending @clientLocation line inline, and the How to Fix has been reworded with a concrete corrected example (moving the parameter per-operation) instead of the hard-to-follow prose.

packages/typespec-client-generator-core/src/diagnostics/api-version-undefined.md
tadelesh: Done - reframed the Suppression section: this diagnostic should not be suppressed; the fix is to correct the `api-version` option in tspconfig.yaml (or use `latest`/`all`). Removed the #suppress example.

packages/typespec-client-generator-core/src/diagnostics/api-version-undefined.md
tadelesh: Done - the Incorrect Usage now includes the tspconfig.yaml block showing the offending `api-version: "v4"` option (the actual trigger), and the How to Fix shows the corrected config.

packages/typespec-client-generator-core/src/diagnostics/add-parameter-duplicate.md
tadelesh: Done - added a correct, self-contained tsp snippet to the How to Fix section of every doc that has an Incorrect Usage example (37 docs). Each shows the corrected version that no longer triggers the diagnostic.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,133 @@
...
+## Impact
+
+- **Area:** Client operation and parameter placement. Generation continues, but the requested `@clientLocation` move cannot be applied safely to the generated client structure.
+- **Not affected:** HTTP routes, parameter wire names, and service operation definitions are unchanged.
+
+## String target with multiple root clients
+
+### Diagnostic Message
+
+For the client location above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,133 @@
...
+
+### ✅ How to Fix
+
+Use an interface or namespace target when multiple root clients exist, or define the target sub client explicitly under the intended root client.
+
+## Operation to operation
+
+### Diagnostic Message
+
+For the client location above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,133 @@
...
+```text
+`@clientLocation` cannot be used to move an operation to another operation. Operations can only be moved to interfaces or namespaces.
+```
+
+### ✅ How to Fix
+
+Move the operation to an interface or namespace instead of another operation.
+
+## Model property conflicts with client initialization
+
```
tadelesh, line 43: This needs an example.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,133 @@
...
+For the client location above, TCGC reports:
+
+```text
+There is already a parameter called 'apiKey' in the client initialization.
+```
+
+### ✅ How to Fix
+
+Rename one of the parameters or avoid moving the model property to a client that already has a client initialization parameter with that name.
```
tadelesh: Refine with example.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Model property moved to string target
+
+### Diagnostic Message
+
+For the client location above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
```diff
@@ -0,0 +1,133 @@
...
+): void;
+```
+
+## Suppression
+
+Suppress this warning only if the conflicting `@clientLocation` is intentionally ignored or handled manually by the target emitter.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/client-location-conflict" "client location handled manually"
+```
```
tadelesh: This diagnostic should not be suppressed.

packages/typespec-client-generator-core/src/diagnostics/client-location-wrong-type.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+## Suppression
+
+Suppress this warning only if the invalid `@clientLocation` target is intentionally ignored and the operation should remain in its original client.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/client-location-wrong-type" "operation intentionally stays in original client"
+```
```
tadelesh: This diagnostic should not be suppressed.

packages/typespec-client-generator-core/src/diagnostics/client-name-ineffective.md
```diff
@@ -0,0 +1,55 @@
...
+
+For the declaration above, TCGC reports:
+
+```text
+Application of @clientName decorator to WidgetName is not effective
+```
+
+### ✅ How to Fix
+
+Move `@clientName` to the declaration that TCGC actually uses for the generated SDK name, or remove the ineffective decorator.
```
tadelesh: This sub message will never happen. Remove it from diagnostic and refine this doc.

packages/typespec-client-generator-core/src/diagnostics/client-option.md
```diff
@@ -0,0 +1,41 @@
...
+}
+```
+
+## Suppression
+
+Suppress this warning only when `@clientOption` is a deliberate temporary workaround for a specific language emitter.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/client-option" "preview feature for python"
+```
```
tadelesh: Always add this suppression since it is a temp solution for languages' emitter.

packages/typespec-client-generator-core/src/diagnostics/conflict-access-override.md
```diff
@@ -0,0 +1,43 @@
...
+op test(@body body: A): void;
+```
+
+## Suppression
+
+Suppress this warning only if the public exposure is intentional and the conflicting `@access` override is kept for another emitter or compatibility reason.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/conflict-access-override" "public exposure is intentional"
+```
```
tadelesh: This diagnostic should not be suppressed.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name-warning.md
```diff
@@ -0,0 +1,66 @@
...
+  @route("/listByParent")
+  listByParent(parent: string): void;
+}
+```
+
+## Generated name conflict
+
+### Diagnostic Message
+
+For the generated operation name above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name.md
```diff
@@ -0,0 +1,50 @@
...
+model WidgetResponse {}
+
+model Widget {}
+```
+
+## Generated name conflict
+
+### Diagnostic Message
+
+For the declaration above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/empty-client-name.md
```diff
@@ -0,0 +1,38 @@
...
+model Widget {}
+```
+
+## Suppression
+
+Suppress this warning only if the empty `@clientName` is intentionally ignored and the default generated name is acceptable.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/empty-client-name" "default client name is intentional"
+```
```
tadelesh: This should not suppress. Either rename or remove decorator.

packages/typespec-client-generator-core/src/diagnostics/empty-client-namespace.md
```diff
@@ -0,0 +1,38 @@
...
+## ✅ How to Fix
+
+Provide a non-empty namespace string or remove `@clientNamespace`.
+
+```typespec
+@clientNamespace("Contoso.Widgets")
+model Widget {}
+```
+
+## Suppression
```
tadelesh, line 32: This should not suppress. Either rename or remove decorator.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
```diff
@@ -0,0 +1,56 @@
...
+## Impact
+
+- **Area:** Example and sample loading. Generation continues without the affected example files, so generated samples or example-based tests may be incomplete.
+- **Not affected:** SDK clients, models, and service protocol metadata are still generated from TypeSpec.
+
+## Invalid example file
+
+### Diagnostic Message
+
+For the example file above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
```diff
@@ -0,0 +1,56 @@
...
+
+### ✅ How to Fix
+
+Fix the JSON syntax or contents of the example file so it can be parsed.
+
+## Examples directory cannot be read
+
+### Diagnostic Message
+
+For the example directory above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
```diff
@@ -0,0 +1,56 @@
...
+
+### ✅ How to Fix
+
+Create the configured examples directory, correct the `examples-dir` option, or omit the option when examples are not available.
+
+## Missing operationId or title
+
+### Diagnostic Message
+
+For the example file above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
```diff
@@ -0,0 +1,56 @@
...
+
+Add both `operationId` and `title` to the example JSON file.
+
+## Suppression
+
+Suppress this warning only if examples are optional for this run, such as local validation where the examples directory is intentionally absent.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/example-loading" "examples not required for this run"
+```
```
tadelesh: This can not be suppressed.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
```diff
@@ -0,0 +1,58 @@
...
+## Suppression
+
+Suppress this warning only if the mismatched example value is intentionally excluded from generated SDK examples or tests.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/example-value-no-mapping" "example value intentionally not mapped"
+```
```
tadelesh: This should not be suppressed. Fix the example.

packages/typespec-client-generator-core/src/diagnostics/external-library-version-mismatch.md
```diff
@@ -0,0 +1,72 @@
...
+model Collection {}
+```
+
+## Suppression
+
+Suppress this warning only if dependency version unification is handled outside TCGC and the emitter will use one compatible external package version.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/external-library-version-mismatch" "external package version unified by emitter"
+```
```
tadelesh: This should not be suppressed. Consolidate the version.

packages/typespec-client-generator-core/src/diagnostics/external-type-on-model-property.md
```diff
@@ -0,0 +1,54 @@
...
+}
+```
+
+## Suppression
+
+Suppress this warning only if the property-level external alternate type is intentionally ignored and the SDK type is handled manually.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/external-type-on-model-property" "property external type handled manually"
+```
```
tadelesh: This should not be suppressed. Fix the usage.

packages/typespec-client-generator-core/src/diagnostics/invalid-encode-for-collection-format.md
```diff
@@ -0,0 +1,42 @@
...
+op myOp(@header @encode(ArrayEncoding.pipeDelimited) header: string[]): void;
+```
+
+## Suppression
+
+Suppress this warning only if the fallback collection format, such as CSV, is the intended SDK behavior for this parameter.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/invalid-encode-for-collection-format" "CSV fallback is intentional"
+```
```
tadelesh: Should not suppress, fix the usage.

packages/typespec-client-generator-core/src/diagnostics/invalid-mark-as-lro-target.md
```diff
@@ -0,0 +1,44 @@
...
+op start(): StartResponse;
+```
+
+## Suppression
+
+Suppress this warning only if the operation should remain a regular non-LRO method and the legacy `@markAsLro` annotation is intentionally ignored.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/invalid-mark-as-lro-target" "regular operation despite legacy LRO marker"
+```
```
tadelesh: should not suppress. fix or remove.

packages/typespec-client-generator-core/src/diagnostics/invalid-mark-as-pageable-target.md
```diff
@@ -0,0 +1,48 @@
...
+@get
+op listWidgets(): WidgetPage;
+```
+
+## Suppression
+
+Suppress this warning only if the operation should remain a regular non-pageable method and the legacy `@markAsPageable` annotation is intentionally ignored.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/invalid-mark-as-pageable-target" "regular operation despite pageable marker"
```
tadelesh: should not suppress. fix or remove.

packages/typespec-client-generator-core/src/diagnostics/legacy-hierarchy-building-conflict.md
```diff
@@ -0,0 +1,64 @@
...
+op test(): A;
+```
+
+## Suppression
+
+Suppress this warning only if the property dropped during legacy hierarchy rebasing is intentional and generated SDK compatibility has been verified.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/legacy-hierarchy-building-conflict" "legacy rebase drops property intentionally"
+```
```
tadelesh: should not suppress. fix or remove.

packages/typespec-client-generator-core/src/diagnostics/mark-as-pageable-ineffective.md
```diff
@@ -0,0 +1,63 @@
...
+  listEmployees is ArmResourceListByParent<Employee>;
+}
+```
+
+## Suppression
+
+Suppress this warning only if the redundant `@markAsPageable` annotation is kept for source compatibility while real `@list` metadata drives paging.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/mark-as-pageable-ineffective" "redundant pageable marker kept for compatibility"
```
tadelesh: should not suppress. fix or remove.

packages/typespec-client-generator-core/src/diagnostics/multiple-param-alias.md
```diff
@@ -0,0 +1,45 @@
...
+@@paramAlias(ClientOptions.account, "accountName");
+```
+
+## Suppression
+
+Suppress this warning only if the first `@paramAlias` is the intended alias and later aliases are deliberately ignored.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/multiple-param-alias" "first parameter alias is intentional"
+```
```
tadelesh: should not suppress. fix.

packages/typespec-client-generator-core/src/diagnostics/no-corresponding-method-param.md
```diff
@@ -0,0 +1,18 @@
+This diagnostic is issued when a required HTTP operation parameter cannot be mapped to any generated method parameter. The generated client would have no reliable way to supply that required service parameter.
+
+## Impact
+
+- **Area:** Method-to-protocol parameter mapping. Blocks generation of a service method when a required HTTP parameter cannot be supplied by the generated SDK method signature.
+- **Not affected:** The HTTP operation still declares the required parameter.
+
+## Diagnostic Message
+
+For the case above, TCGC reports:
```
tadelesh:
```suggestion
TCGC reports:
```

packages/typespec-client-generator-core/src/diagnostics/no-corresponding-method-param.md
```diff
@@ -0,0 +1,18 @@
...
+
+```text
+Missing HTTP operation parameter "apiVersion" in method "getWidget". Please check the method definition.
+```
+
+## ✅ How to Fix
+
+Keep a matching method parameter or model property path for the HTTP parameter, or make the service parameter optional when appropriate.
```
tadelesh: Check any customization that will impact the parameter mapping and fix them.

packages/typespec-client-generator-core/src/diagnostics/no-emitter-name.md
```diff
@@ -0,0 +1,26 @@
...
+
+Make sure the emitter is invoked with a resolvable package name (such as `@azure-tools/typespec-csharp`) so TCGC can derive the target language.
+
+## Suppression
+
+Suppress this warning only in tooling or tests that intentionally run TCGC without a real language emitter name.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/no-emitter-name" "test run has no emitter name"
+```
```
tadelesh: This could not be suppressed. Just check the usage.

packages/typespec-client-generator-core/src/diagnostics/require-versioned-service.md
```diff
@@ -0,0 +1,56 @@
...
+}
+```
+
+## Suppression
+
+Suppress this warning only if `@clientApiVersions` is temporarily present before service versioning is added and the generated client should ignore it.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/require-versioned-service" "client API versions staged before versioning"
+```
```
tadelesh: Should not suppress. Fix or remove.

packages/typespec-client-generator-core/src/diagnostics/union-circular.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+## Suppression
+
+Suppress this warning only if the recursive union is intentionally left in the source and you accept that TCGC cannot produce a usable SDK union for it.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/union-circular" "recursive union intentionally ignored by SDK"
+```
```
tadelesh: Should not suppress. Fix the spec.

packages/typespec-client-generator-core/src/diagnostics/union-null.md
```diff
@@ -0,0 +1,26 @@
...
+
+Add at least one non-null variant to the union or remove the union from the generated client surface.
+
+## Suppression
+
+Suppress this warning only if the null-only union is intentionally excluded from usable SDK models and you accept TCGC's fallback representation.
+
+```typespec
+#suppress "@azure-tools/typespec-client-generator-core/union-null" "null-only union intentionally not generated"
+```
```
tadelesh: Should not suppress. Fix the spec.

packages/typespec-client-generator-core/src/diagnostics/no-corresponding-method-param.md
tadelesh: Done - the How to Fix now says to check any client customization (`@override`, `@paramAlias`, parameter reorder/removal) that changes the method parameters and fix it so every HTTP operation parameter maps to a method parameter.

packages/typespec-client-generator-core/src/diagnostics/client-name-ineffective.md
tadelesh: Done - removed the never-reported `default` sub-message from `client-name-ineffective` in lib.ts (the only report site uses the override message), and refined the doc to a single case. Build + tests pass.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name-warning.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/duplicate-client-name.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/no-corresponding-method-param.md
tadelesh: Applied - simplified this line to just `TCGC reports:`. Done across all diagnostic docs for consistency.

packages/typespec-client-generator-core/src/diagnostics/client-option.md
tadelesh: Done - reworded the Suppression section to say this should always be suppressed with a justification, since `@clientOption` is a temporary language-emitter mechanism. Kept the `#suppress` example.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Done - added an Incorrect Usage (with the offending line marked) and a How to Fix example for this case, grounded in the `@clientLocation` decorator logic.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Done - added an Incorrect Usage (with the offending line marked) and a How to Fix example for this case, grounded in the `@clientLocation` decorator logic.

packages/typespec-client-generator-core/src/diagnostics/client-location-conflict.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/client-location-wrong-type.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/conflict-access-override.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/empty-client-name.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/empty-client-namespace.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/example-loading.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/example-value-no-mapping.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/external-library-version-mismatch.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/external-type-on-model-property.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/invalid-encode-for-collection-format.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/invalid-mark-as-lro-target.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/invalid-mark-as-pageable-target.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/legacy-hierarchy-building-conflict.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/mark-as-pageable-ineffective.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/multiple-param-alias.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/no-emitter-name.md
tadelesh: Done - the Suppression section now states this diagnostic should not be suppressed, with the specific fix, and the `#suppress` example was removed.

packages/typespec-client-generator-core/src/diagnostics/inconsistent-multiple-service-dependency.md
tadelesh: This doc already has an Incorrect Usage example (two services merged into one client pinning a shared dependency to different versions) plus a How to Fix - added in an earlier round.

packages/typespec-client-generator-core/src/diagnostics/require-versioned-service.md
tadelesh: Done - reframed to say it should not be suppressed; add @versioned to the service or remove @clientApiVersions.

packages/typespec-client-generator-core/src/diagnostics/union-circular.md
tadelesh: Done - reframed to say it should not be suppressed; fix the spec to break the circular union reference or model the recursion via a model property.

packages/typespec-client-generator-core/src/diagnostics/union-null.md
tadelesh: Done - reframed the Suppression section to state this should not be suppressed; fix the spec so the union has a non-null variant or remove it.

packages/typespec-client-generator-core/src/diagnostics/unsupported-protocol.md
```diff
@@ -0,0 +1,18 @@
...
+
+TCGC reports:
+
+```text
+Currently we only support HTTP and HTTPS protocols
+```
+
+## ✅ How to Fix
+
+Add appropriate HTTP route and verb decorators to the operation, or remove it from the TCGC-generated client surface.
```
lirenhe, line 18: can this diagnostic be suppressed?

tadelesh: For doc without suppression part, that diagnostic is an error diagnostic that will block the code emit.

lirenhe (approved):

PR #4983, opened 2026-07-21
typespec-java, Read pinned core sources via git archive zip instead of tar

weidongxu-microsoft:
## Problem

Following #4972, `packages/typespec-java/CoreCommit.ps1` reads the pinned `core/` commit's sources by `git archive` into a temp dir and then extracts them with an external `tar`. On Windows this fails: the `tar` first on PATH is Git for Windows' GNU `tar`, which interprets the archive path `-f C:\...\core.tar` as a **remote host** (`host:path` rsh syntax) and aborts:

```
/usr/bin/tar: Cannot connect to C: resolve failed
```

Linux CI doesn't hit this (`/tmp/...` has no drive-letter colon), so it only bites Windows dev builds.

## Fix

`Get-CoreSourceRoot` now archives with `git archive --format=zip` and extracts with PowerShell's built-in `Expand-Archive` (System.IO.Compression). This keeps the whole flow inside PowerShell/.NET with **no external archiver dependency** and no drive-letter parsing quirk. `core/` is still never checked out or modified, and the temp dir is still cleaned up by `Remove-CoreSourceRoot`.

Also refreshes `CONTRIBUTING.md`, which still described the pre-#4972 in-place `core/` checkout / `git checkout .` patch flow, to match the current `git archive` read and copy-and-patch-the-generator flow.

## Verification

Pinned `core-commit.json` to a commit newer than the current `core` checkout to force the archive path, then ran `Copy-Sources.ps1`:

- Took the archive branch (*"Reading core sources from pinned commit ... without checking it out"*).
- `git archive --format=zip` + `Expand-Archive` succeeded; generator copied and `core.patch` applied; `exit=0`.
- `git -C core status` clean (submodule untouched), pin restored, no temp dir leaked.

### DIFF
--- a/packages/typespec-java/CoreCommit.ps1
+++ b/packages/typespec-java/CoreCommit.ps1
@@ -84,22 +84,23 @@ function Get-CoreSourceRoot {
     Write-Host "Reading core sources from pinned commit $targetSha (was $originSha) without checking it out"
     $tempDir = Join-Path ([System.IO.Path]::GetTempPath()) ("typespec-java-core-" + [System.Guid]::NewGuid().ToString("N"))
     New-Item -ItemType Directory -Path $tempDir | Out-Null
-    $tarPath = Join-Path $tempDir "core.tar"
+    $zipPath = Join-Path $tempDir "core.zip"
     try {
-        git -C $CoreRoot archive -o $tarPath "${targetSha}:$script:CoreJavaSubtree"
+        # Produce a zip (not tar) and expand it with the built-in Expand-Archive:
+        # this keeps the whole flow inside PowerShell/.NET and avoids depending on an
+        # external `tar`, whose behavior varies by platform (notably Git for Windows'
+        # GNU tar treats a drive-letter path like C:\...\core.tar as a remote host).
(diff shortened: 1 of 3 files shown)


packages/typespec-java/CoreCommit.ps1
```diff
@@ -84,22 +84,23 @@ function Get-CoreSourceRoot {
     Write-Host "Reading core sources from pinned commit $targetSha (was $originSha) without checking it out"
     $tempDir = Join-Path ([System.IO.Path]::GetTempPath()) ("typespec-java-core-" + [System.Guid]::NewGuid().ToString("N"))
     New-Item -ItemType Directory -Path $tempDir | Out-Null
-    $tarPath = Join-Path $tempDir "core.tar"
+    $zipPath = Join-Path $tempDir "core.zip"
```
weidongxu-microsoft, line 87: Main change is here. Windows does not have `tar` built-in.

XiaofeiCao (approved):

XiaofeiCao (approved):

PR #5015, opened 2026-07-22
Enrich TCGC linter rule docs with impact, severity, message and suppression

tadelesh:
## Summary

Enriches the reference docs for all TCGC (`@azure-tools/typespec-client-generator-core`) linter rules with structured guidance, so each rule page explains its impact, the error it produces, how to fix it, and when it may be suppressed.

The three rules covered: `require-client-suffix`, `property-name-conflict`, `csharp-no-url-suffix`.

## Doc structure

Each rule doc now follows this layout (keeping the original description):

- **Impact** — which SDK / language and plane (data-plane / management-plane) the rule affects, and what it does not affect.
- **❌ Incorrect Usage** — TypeSpec that triggers the rule.
- **Diagnostic Message** — the actual message text for that example, with a brief explanation.
- **✅ How to Fix** — the corrected TypeSpec (including `@clientName` / `@client` variants where relevant).
- **Suppression** — when it is appropriate to suppress the rule. All TCGC linter rules are `warning`, so the severity nuance lives here: `require-client-suffix` and `csharp-no-url-suffix` may be suppressed when the naming is intentional, while `property-name-conflict` should not be suppressed because the conflict breaks C# code generation and must be fixed.

## Notes

- Docs only (source markdown under `packages/typespec-client-generator-core/src/rules/*.md`). The generated website pages are produced by `tspd doc` and are not tracked.
- No changelog entry: markdown-only changes are excluded from Chronus.

### DIFF
--- a/packages/typespec-client-generator-core/src/rules/property-name-conflict.md
+++ b/packages/typespec-client-generator-core/src/rules/property-name-conflict.md
@@ -3,6 +3,27 @@ some language emitters (notably C#), where the generated property would
 collide with the type name. Rename the property, or use `@clientName` to
 rename it for the affected language.
 
+## Impact
+
+- **Area:** SDK generation, primarily **C#**. A property whose name collides with its enclosing model's name produces code that does not compile in C# (applies to both data-plane and management-plane).
+- **Not affected:** The service definition and the wire protocol are unchanged; the collision only affects the generated client code.
+
+## Severity
+
+`warning` — suppressible. Although reported as a warning, it should be treated as a must-fix for the C# SDK: leaving the conflict in place produces C# code that fails to compile.
+
+## Diagnostic message
+
+```text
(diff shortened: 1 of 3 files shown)


packages/typespec-client-generator-core/src/rules/property-name-conflict.md
```diff
@@ -3,27 +3,53 @@ some language emitters (notably C#), where the generated property would
...
+## Suppression
+
+Although reported as a `warning`, this conflict breaks C# code generation, so it should be treated as a must-fix. Suppress it only if the affected model is never generated for C#:
+
+```tsp
+#suppress "@azure-tools/typespec-client-generator-core/property-name-conflict" "model not generated for C#"
+model Widget {
+  widget: string;
+}
+```
```
tadelesh: This should not be suppressed cause it will result to C# generated code compile error.

tadelesh: Good point - removed the suppression example and the Suppression section now states this rule should not be suppressed because the conflict breaks C# code generation and must be fixed.

lirenhe (approved):

Issue #5006, opened 2026-07-22
typespec-java, eng, Remove the dependency on submodule of microsoft/typespec


weidongxu-microsoft:
### Clear and concise description of the problem

Remove the dependency on submodule of microsoft/typespec

We can let agent to brainstorm potential solutions.

One short term solution I thought of is to just move all the code here (let microsoft/typespec do the manual sync, as we seldom release the unbranded emitter).

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

weidongxu-microsoft:
## Agent suggestion

The core problem is not specifically Git submodules; it is one source tree serving two independently released packages without a two-tier release pipeline.

As a near-term bridge, consider checking an upstream source snapshot into `Azure/typespec-azure`:

- Keep `microsoft/typespec` as the authoritative source.
- Copy the relevant `packages/http-client-java` files into this repository.
- Record the upstream SHA in a metadata file.
- Use a deterministic script or workflow to create source-sync PRs.
- Keep Azure-specific changes isolated as explicit overlays or patches.
- Have CI verify that the checked-in snapshot matches the recorded SHA plus approved overlays.

This moves the cross-repository dependency from build time to review/sync time. Builds and releases become hermetic—no submodule checkout, network fetch, or hidden dependency on another repository's state—while this repository can deliberately remain on an older upstream snapshot when needed.

Possible progression:

1. Replace the submodule with a checked-in snapshot and upstream-SHA metadata.
2. Add deterministic sync and drift-validation scripts.
3. Automate sync PR creation when `packages/http-client-java` changes.
4. Require fixes to land upstream first; allow temporary Azure overlays with clear ownership and expiry.
5. Later replace the snapshot with a versioned upstream artifact when a true two-tier release becomes practical.

Alternatives such as downloading a SHA archive during build or using Git subtree remove some submodule friction but retain build-time coupling or introduce harder synchronization. Moving the canonical implementation permanently into this repository is simpler operationally, but should be an explicit product-ownership decision rather than a temporary workaround.

PR #5019, opened 2026-07-22
Unify typespec-java emitter and emitter-tests into one npm project

### LINKED ISSUE #5008: typespec-java, eng, Consider use same npm project for emitter and emitter-test
May ask agent to investigate how other emitters do this.

Also may need to follow a unification process that apply to all emitters.

XiaofeiCao:
Fixes #5008

## What

Unify the `@azure-tools/typespec-java` **emitter** and its **emitter-tests** into a single npm project, matching how `typespec-go` and `typespec-python` are structured. Previously `emitter-tests/` was a standalone npm + Maven project, deliberately excluded from the pnpm workspace, that consumed the emitter via a packed `.tgz` and pinned peer versions with npm `overrides`.

## Why

`typespec-java` was the only language emitter whose e2e tests lived in a separate npm project. This required a bespoke pack + `npm install` flow (`Build-TypeSpec.ps1`), a second `package.json`/`package-lock.json`, an `overrides` block, and a core-driven version sync. Consolidating removes that maintenance overhead and makes Java consistent with the other emitters.

## How

- **One npm project.** The e2e scripts (`regenerate`, `test:java:e2e`, `spector-*`, `sync-tests`) and the spector/http-specs devDependencies (`workspace:^`) now live in the root `packages/typespec-java/package.json`. Removed `emitter-tests/package.json`, `package-lock.json`, and `Build-TypeSpec.ps1`/`Setup.ps1`/`setup.py`.
- **Emitter resolved by self-reference.** Because `emitter-tests/` no longer has its own `package.json`, `emit: ["@azure-tools/typespec-java"]` in `emitter-tests/tspconfig.yaml` resolves to the parent package via Node package self-reference (no `.tgz`, no `--emit` argument). `tsp compile <spec>.tsp` works directly.
- **Harness scripts moved to the package root** (`Generate.ps1`, `Spector-Tests.ps1`, `SyncTests.ps1`); they `Push-Location`/`Set-Location` into `emitter-tests/` for the Maven/tsp work.
- **CI/ADO** run via `pnpm run regenerate` / `pnpm run test:java:e2e` with `working-directory: packages/typespec-java` (`ci-java.yml`, `nightly-e2e.yml`, `spector-coverage.yml`).
- Updated `.gitignore`/`.prettierignore` and developer docs; removed an orphaned `tsp-checkout.sh`.

## Behavioral note

The e2e tests now run against the **workspace/dev** peer versions (via the catalog), like typespec-go/python, instead of the previously pinned released versions. `CONTRIBUTING.md` now documents the unified workspace build using the repository’s safe Turbo wrapper.

## Validation

- `pnpm build` (Maven jar + tsc) succeeds.
- `pnpm run regenerate` generates the full local + http-specs/azure-http-specs suites through the self-reference emitter (2800+ Java files).
- `pnpm install --frozen-lockfile` is consistent.


packages/typespec-java/core.test.patch
timotheeguerin, line 1: is there no better  way to do this here?

weidongxu-microsoft: We likely able to delete this file, if we move source into typespec-azure repo (aka when no need to do Copy-Sources or SyncTests)

Short-term: Alternatively, we can have a replacement of string in SyncTests to handle the `node_modules/@typespec/http-specs/assets/` -> `../node_modules/@typespec/http-specs/assets/`
and remove this file.
The adding of `@Disable` is temporary -- for the case we mentioned yesterday: changing an existing test case still have the possibility to break downstream

timotheeguerin: ok i'm not fully sure I get what is going on but maybe will dig into more when we have the new common runner

XiaofeiCao: The `@Disabled` is for newly added property `armIdWithGroupScope` here: https://github.com/Azure/typespec-azure/pull/4891/changes#diff-e8dfd76612fab9bfdc25ee76bb899c3a060401dbbae31a3edc44ac5fae981013

It's merged to main but not released. Our current hand-written test is synced from core, which depends on released version. We need patch after sync from core, either to add test for this new property, or disable it.

After removing the dependency on core: https://github.com/Azure/typespec-azure/issues/5006, we should be able to get rid of this patch, and directly disable/update the test.

timotheeguerin (approved):

PR #5027, opened 2026-07-22
fix: detect operation name conflicts across services in same client

### LINKED ISSUE #3789: [TCGC] multiple services lint for naming collisions in operations
Consider this typespec: [link](https://azure.github.io/typespec-azure/playground/?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fdata-plane%22%5D%7D%7D&e=%40azure-tools%2Ftypespec-client-generator-core&vs=%7B%7D&c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtY2xpZW50LWdlbmVyYXRvci1jb3JlIjsKyTbIKi92ZXJzaW9uaW5n1R9odHRwIjsKCnVzaW5nIEHEbC5DxV1HyFwuQ29yZTvHIlbJT8gSSHR0cDsKCi8qCiAqIEZpcnN0IHNlcnZpY2UgZGVmaW5pdGlvbiBpbiBhIG11bHRpLcgecGFja2FnZSB3aXRoIOoApwogKi8KQMcQZWQox3JzQSkKbmFtZXNwYWNlIFPGREEgewogIGVudW3oAJlzxhMgIGF2MSzHCTLECX0KCiAgQHJvdXRlKCIvdGVzdCIpCiAgb3AgxAwoQHF1ZXJ5KCJhcGktx3wiKSBhcGnHWDrKYSk6IHZvaWQ7Cgp95AD35QD4U2Vjb25k%2FwD5%2FwD5%2BAD5QvMA%2BULyAPnGEyAgYugA%2BWL%2FAPn%2FAPnxAPlC6QD5fQoKQOYCgijEdOcA2DogW%2BgBlyzpAKhdCn3zAL8uTeQBEcgNQ29tYmluZWQge30%3D). The top level client contains both operations from the two services, and they both share the same name. For .NET, this would result in a compilation error once the code is generated.

It would be really helpful if there was at least a linter warning, so the spec author can customize the name(s) of these operations to avoid the collision.

### Checklist

iscai-msft:
## Problem

When multiple services are combined into a single client via `@client({service: [ServiceA, ServiceB]})`, operations from different services can have the same name. This causes compilation errors in generated code (e.g., duplicate method names in C#). The same issue occurs when sub-namespaces with the same name are merged into a single sub-client.

Fixes https://github.com/Azure/typespec-azure/issues/3789

## Solution

Added a duplicate operation name check in `cache.ts` after the client→operations cache is fully built. This location is ideal because:

- It naturally handles **merged sub-clients** (same-named namespaces across services like `A.Shared` and `B.Shared`)
- It respects **`@clientLocation`** moves (operations reassigned to other clients)
- It's **emitter-scope-aware** (uses `context.emitterName` for `@clientName` resolution)
- It uses the existing `duplicate-client-name` diagnostic (no new warning codes)
- It only checks clients whose operations come from multiple containers (same-namespace duplicates are already caught by `validateClientNamesPerNamespace`)

Also removed the earlier approach from `validations/types.ts` which read `@client` decorator state directly but couldn't handle merged sub-clients.

## Tests

Added tests to `test/validations/types.test.ts`:
- ✅ Error when multi-service client has operations with the same name
- ✅ No error when operations have different names
- ✅ No error when `@clientName` resolves the conflict
- ✅ No error when same-named operations are in different sub-clients
- ✅ Error when merged sub-clients have operations with the same name (A.Shared + B.Shared)
- ✅ Error only for the language where `@clientName` causes a conflict
- ✅ `@clientName` fixes conflict for one language but not others


packages/typespec-client-generator-core/src/validations/types.ts
```diff
@@ -101,6 +101,21 @@ function validateClientNames(tcgcContext: TCGCContext) {
     [...newClients.values()].map((operations) => {
       validateClientNamesCore(tcgcContext, scope, operations);
     });
+
+    // Check for operation name conflicts across service boundaries.
+    // The per-namespace check above only catches duplicates within a single namespace.
+    // When multiple services are combined into one client (via @client({service: [A, B]})),
+    // operations from different namespaces can collide at the client level.
+    for (const [, client] of listScopedDecoratorData(tcgcContext, clientKey, scope)) {
```
tadelesh: This could not handle the case that merging service with same sub clients. e.g.:
```typespec
namespace A {
  namespace B {
    op test(): void;
  }
}
namespace C {
  namespace B {
    op test(): void;
  }
}
```
I'm thinking about if we need to add such check in `cache.ts` not in validation. This is really a client level issue, and could fail only when emit clients.

iscai-msft: that's a really good point, thanks for always leaving great comments on my prs, let me switch my design over to that instead of at validation time

tadelesh (approved):
LGTM

PR #5028, opened 2026-07-22
Make TypeSpec Python setup opt-in

### LINKED ISSUE #5024: Python emitter's install and prepare scripts require a Python installation
From `typespec-python/package.json`:
```json
    "install": "tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/install.py",
    "prepare": "tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/prepare.py",
```
This reports a benign error during `pnpm install`.

However, it's a _hard_ error when building the Go emitter via `pnpm build:deps`.

Ideally, node scripts shouldn't depend on optional tooling.  At the very least, they should fail gracefully.

Another option is can we skip building `typespec-python` (and other projects) when building the Go emitter?  It seems unnecessary.

xirzec:
## Summary

- remove the Python `install` and `prepare` lifecycle hooks and add an explicit `setup:python` command
- update Python-specific CI, nightly, and release workflows to prepare the environment explicitly
- run development TypeScript scripts directly with Node, remove the `tsx` dependency, and enforce erasable syntax

## Validation

- `pnpm format`
- `pnpm lint`
- full workspace install with Python resolution forced to an invalid path
- TypeSpec Python build, explicit Python setup, lint, and formatting checks
- native execution of each converted TypeScript entry point on Node 22

Closes #5024

### DIFF
--- a/packages/typespec-python/package.json
+++ b/packages/typespec-python/package.json
@@ -31,15 +31,15 @@
   "scripts": {
     "clean": "rimraf ./dist ./temp ./venv ./node_modules",
     "build": "tsc -p .",
-    "setup:python": "tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/install.py && tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/prepare.py",
+    "setup:python": "node ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/install.py && node ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/prepare.py",
     "watch": "tsc -p . --watch",
-    "lint:extra": "tsx ./eng/scripts/ci/lint.ts",
-    "format:extra": "tsx ./eng/scripts/ci/format.ts --python",
-    "format:extra:check": "tsx ./eng/scripts/ci/format.ts --python --check",
-    "regenerate": "tsx ./eng/scripts/ci/regenerate.ts",
-    "sync": "tsx ./eng/scripts/sync.ts",
-    "sync:check": "tsx ./eng/scripts/sync.ts --check",
-    "test:python:e2e": "tsx ./eng/scripts/ci/run-tests.ts",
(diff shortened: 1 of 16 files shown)


packages/typespec-python/package.json
timotheeguerin, line 1: i don't think i like that, now you cannot run `pnpm install` then `pnpm build` and have it just work.
We setup this repo to work with mise which will deal with all the tooling necessary so you don't have to manage anything else.

xirzec: does mise get invoked by pnpm install? I don't see where mise gets used in any flow. Also are we saying to contribute to any emitter package you must install tooling for all emitters?

timotheeguerin: no, mise just needs to be setup once, as a user you need to trust the folder once(maybe again if deps change) but then it just works, it register all teh commands in your path and delegate to an install of the configured version.

This is a system I want to expand to all our -sdk repo and the service team development flow to improve the sdk generation experience.
I have never been able to build hte java emitter until it was added in this repo and then it just worked immediately with vite with nothing I had to do

timotheeguerin: happy to demo what it does in the sdk meeting tonight

xirzec: also I'm curious why this is unique to python -- Java doesn't appear to need any lifecycle hooks? https://github.com/Azure/typespec-azure/blob/a608c797f51bb98ade7f346aa86333fe7e96165b/packages/typespec-java/package.json#L30

timotheeguerin: yeah that part not sure, java must be running maven as part of the build but I think this will break at some point with some CFS requirement to split install and build

xirzec:
Going to think more about the right answer here, but for now I have a PR to improve documentation/behavior for mise: https://github.com/Azure/typespec-azure/pull/5031

PR #5055, opened 2026-07-25
feat(azure-core): add no-dollar-prefixed-query-params linter rule

xirzec:
> **Note:** This PR was authored by GitHub Copilot (an AI agent) on behalf of @xirzec.

## Summary

Adds one linter rule to `@azure-tools/typespec-azure-core`, encoding a mechanical statement from the [Azure REST API Guidelines (vNext data-plane)](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md).

**`no-dollar-prefixed-query-params`** — reports data-plane query parameters whose **wire name** is a `$`-prefixed spelling of one of the seven standard collection query options (`filter`, `orderby`, `skip`, `top`, `maxpagesize`, `select`, `expand`). Guideline anchor: [`collections-query-options-no-dollar-sign`](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections-query-options-no-dollar-sign).

Enabled in the `data-plane` ruleset, deliberately disabled in `resource-manager`.

```tsp
// ❌
op list(@query("$filter") filter?: string, @query `$top` top?: int32): Widget[];

// ✅
op list(@query filter?: string, @query top?: int32): Widget[];
```

Handles both the explicit `@query("...")` form and the backtick-identifier form. Non-standard `$`-prefixed names (e.g. `$custom`) are intentionally left alone. Each offending declaration is reported once, at the declaration, rather than once per operation that spreads it — so one suppression covers all consumers.

## Scope: data-plane only

Management-plane APIs use the OData spelling by design — `Azure.ResourceManager` ships `ArmTopParameter` as `@query("$top")`, plus `ArmFilterParameter`, `ArmSkipParameter`, and `SkipTokenParameter`. The rule is therefore off in the `resource-manager` ruleset, and the doc states this up front (thanks @markcowl for flagging that it wasn't prominent enough).

## Not a duplicate of `casing-style`

`casing-style` inspects **TypeSpec identifiers** — `model.name`, `property.name`, `operation.name`. It never reads the string argument of `@query(...)`. So this is invisible to it:

```tsp
op list(@query("$filter") filter?: string): Widget[];
```

The identifier is `filter` (clean); the **wire name** the customer types into a URL is `$filter`. This rule reads `HttpProperty.options.name`, which is a new axis of checking for the linter.

## Two rules were removed after measurement

This PR originally proposed three rules. Two were dropped after measuring them against the `azure-rest-api-specs` corpus, per @markcowl's criteria.

**`use-date-api-version` (removed).** Mark's threshold: if more than 20% of data planes use a non-standard version format, it belongs in authoring guidance rather than a lint rule. Measured: **25.8% by project (24/93)** and **26.3% by service (10/38)** — both units agree and both are over the line. Offenders include Key Vault (`7.5`, `7.6`, `7.6-preview.2`), Maps (`1.0`, `1.1`), Machine Learning (`1.0.0`, `1.5.0`, `v1.0`), Monitor (`v1`), App Insights (`v2.1`), Cognitive Services (`v1.1`), Schema Registry (`2021-10`), Translation (`3.0`), AI Face (`v1.2`), AI Agents / AI Foundry (`v1`). At that rate the signal is live disagreement about whether the guideline applies, not a fixable defect.

**`no-version-in-route` (removed).** All **254** sites it reported were unfixable by design: the OCI Distribution Spec mandates ACR's `/v2/` root, Apache Atlas mandates Purview's `/atlas/v2/`, and the Machine Learning routes are shipped. A static rule cannot distinguish an inherited version segment from a newly introduced one — that judgement needs the diff — so it moves to agentic review instead. Perfect precision with zero actionable findings is exactly the shape the criteria warn against.

## Why the surviving rule clears the bar

Its 81 sites are **legacy**, not an ongoing exception. They are OData-derived services (Search, Batch, Cosmos DB Table API) whose parameter names were fixed years ago. A **new** data-plane API using `$filter` is unambiguously wrong — there is no live disagreement about whether the guideline applies, unlike the 26% above. So a one-time suppression settles the back catalogue and the rule then earns its keep on everything new.

## Validation

- 14 tests for the rule: valid cases, invalid cases, boundary conditions, dedup regressions, and a case asserting `Azure.Core` / `Azure.ResourceManager` library types are exempt.
- Full `typespec-azure-core` suite: **55 files, 352 passed, 2 skipped**.
- `typespec-azure-rulesets` suite (incl. `validate-rules-defined`): 4/4.
- `pnpm lint`, `pnpm format`, `pnpm cspell` (0 issues), `regen-docs`, `chronus verify` all clean.
- Zero diagnostics across `packages/samples/specs/**` and `packages/azure-http-specs/specs/azure/**`.

## Follow-ups (not in this PR)

- A companion suppression PR against `azure-rest-api-specs` **`typespec-next`** for the legacy `$`-prefixed services, following #4880 / Azure/azure-rest-api-specs#44804. Not created yet — see the discussion below on `linter.disable` vs `#suppress`.
- Promoting `arm-resource-invalid-version-format` from `typespec-azure-resource-manager` into `typespec-azure-core`, which @markcowl suggested. Deliberately not bundled here since it refactors a rule this PR doesn't own; happy to do it separately.


xirzec:
> 🤖 Written by GitHub Copilot (an AI agent) on behalf of @xirzec.

**`int:azure-specs` results.** All the diagnostics are true positives, and none are realistically fixable: ACR's `/v2/` is mandated by the OCI Distribution Spec, Purview's `/atlas/v2/` is Apache Atlas compat, Batch/Search `$filter` is deliberate OData, and the Key Vault `7.5`/`7.6` versions are already shipped. Renaming any of them breaks released SDKs, so these need suppressions rather than fixes.

Investigating also turned up a real bug in two of the rules — they reported the same declaration once per consuming operation (Batch: 180 diagnostics for 36 declarations) and double-reported `is`-aliases (Purview: 178 for 89). Both fixed in this PR; **679 → ~376**, with no loss of detection.

Suppressions go to `azure-rest-api-specs` `typespec-next` (it pins the `next` dist-tag), following #4880 / https://github.com/Azure/azure-rest-api-specs/pull/44804. Not created yet — two questions first:

1. Per-project `linter.disable` (~18 tspconfigs) or per-declaration `#suppress` (~465 lines / 44 files)? I lean toward the former since these deviations are service-wide and permanent, but `#suppress` matches prior art.
2. Prepare `main` suppressions now, or wait until the release is adopted there?

Happy to expand on any of it — severity semantics, the per-cluster breakdown, or why "only lint new APIs" isn't expressible in the linter model.

packages/typespec-azure-core/src/rules/use-date-api-version.md
timotheeguerin, line 1: name wise I think `use-date-api-version` would be more inline with naming guidelines

.chronus/changes/add-api-guidelines-linter-rules-2026-07-24-18-10-00.md
timotheeguerin, line 1: are those new rules? is there any conflict with the ongoing lintdiff migration work or is this part of it? 
@catalinaperalta  @markcowl

xirzec: Yeah, this is from me asking Copilot to look at any gaps from our REST guidelines. Happy to avoid duplicating any work that Cat and Mark have in flight.

packages/typespec-azure-core/src/rules/use-date-api-version.md
xirzec: (from Copilot, acting on @xirzec's behalf) Good call, renamed the rule to `use-date-api-version` to follow the `use-<preferred-thing>` convention, updating all references, tests, rulesets, and docs. Happy to adjust further.

packages/typespec-azure-core/src/rules/no-dollar-prefixed-query-params.md
```diff
@@ -0,0 +1,57 @@
+The standard collection query options (`filter`, `orderby`, `skip`, `top`, `maxpagesize`, `select`, and `expand`) come from the OData standard, where they are spelled with a `$` prefix. Azure services use the same names, but without the `$`.
```
markcowl: This is not true for management plane - this should be narrowly focused on data plane services.

packages/typespec-azure-core/src/rules/no-version-in-route.ts
```diff
@@ -0,0 +1,65 @@
...
+  Operation,
+  paramMessage,
+} from "@typespec/compiler";
+import { getHttpOperation } from "@typespec/http";
+import { isExcludedCoreType, isTemplatedInterfaceOperation } from "./utils.js";
+
+/**
+ * Matches a path segment that looks like an API version, e.g. `v1`, `V2`, `v1.0`, `v2.1.3`.
+ */
+const versionSegmentPattern = /^v\d+(\.\d+)*$/i;
```
markcowl: I am skeptical of this rule.  We are unsure if any path like this represents an api-version and not something else.  I think we should reserve these rules for real, impactful violations that are likely

packages/typespec-azure-core/src/rules/use-date-api-version.ts
```diff
@@ -0,0 +1,60 @@
+import { createRule, fileRef, listServices, paramMessage } from "@typespec/compiler";
```
markcowl: This is duplicative of the rule that already exists in ARM.  However, unlike in ARM, this is not always true for data plane APIs, which is why this wasn't added for data planes.  I would suggest not adding this rule, and if you do, moving the ARM rule into azure-core instead.

https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-resource-invalid-version-format.ts

markcowl (changes requested):
I don't think any of these rules meet the bar for linting rules.  These some more like authoring guidance.

xirzec:
> 🤖 Written by GitHub Copilot (an AI agent) on behalf of @xirzec.

@markcowl Thanks - we measured your open questions against the spec corpus and reworked the PR down to a single rule.

**`use-date-api-version`: removed.** Against your 20% threshold it measured **25.8% by project (24/93)** and **26.3% by service (10/38)** using non-date version formats - Key Vault, Maps, Machine Learning, Monitor, App Insights, Cognitive Services, Schema Registry, Translation, AI Face/Agents/Foundry. Over the line on both units, so it belongs in authoring guidance rather than a lint rule. On promoting `arm-resource-invalid-version-format` into azure-core: agreed that is reasonable, but I have deliberately left it out since it refactors a rule this PR does not own. Happy to do it as a separate PR if you want it.

**`no-version-in-route`: removed.** Your skepticism matches the data - all 254 sites were unfixable by design (OCI Distribution Spec for ACR's `/v2/`, Apache Atlas for Purview, shipped ML routes). A static rule cannot tell an inherited version segment from a newly introduced one, so it moves to agentic review where the diff is visible.

**`no-dollar-prefixed-query-params`: kept, doc fixed.** You flagged this against the `.md`, and you were right that the doc was the problem - the rule was already data-plane-only (disabled in the resource-manager ruleset, since ARM ships `ArmTopParameter` as `@query("$top")` by design), but the doc buried that at the bottom. It is now the first thing on the page, with the ARM rationale stated explicitly.

The distinction I would offer for why this one survives when the other two do not: its 81 sites are **legacy**, not an **ongoing exception**. They are OData-derived services (Search, Batch, Cosmos DB Tables) with names fixed years ago. A new data-plane API using `$filter` is unambiguously wrong - there is no live disagreement about whether the guideline applies, which is exactly what the 26% represented for the version-format rule. One-time suppression, then the rule earns its keep on everything new.

@timotheeguerin Your question about conflicts with the ongoing lintdiff migration is still open, and we would like an answer before proposing further rules - it may affect whether some of these belong here at all.

For context, we measured the rest of the candidate list against the same criteria. Three survive: this rule, `query-param-casing` (~16% violation rate, ~3% suppression rate once `api-version` and `$`-prefixed names are exempted), and `no-secrets-in-get-response`. Dropped: `header-name-casing` (the non-kebab names are overwhelmingly canonical HTTP headers like `Content-Type` and `ETag`, and header names are case-insensitive anyway), `json-field-name-casing` (already covered by `casing-style`), `no-post-create-lro` (1 of 42 projects), and `duration-property-naming` (~60% false-positive rate; telling a duration `int32` from a page-count `int32` needs semantics, not regex). None of those are in this PR.

Happy to expand on any of the measurements.

.chronus/changes/add-api-guidelines-linter-rules-2026-07-24-18-10-00.md
catalinaperalta: Sorry about the delay here folks, I see that the api version rules were removed. And my original answer would have been that we have some api-version rules in the lintdiff migration, such as: 

| LintDiff rule | Tsp lint |
|-|-|
|ApiVersionParameterRequired|@azure-tools/typespec-azure-core/operation-missing-api-version, @azure-tools/typespec-azure-resource-manager/arm-resource-operation|
|VersionPolicy|@azure-tools/typespec-azure-core/operation-missing-api-version, tsp-lintdiff-local-linter/version-policy|

There are other rules related to Azure guidelines for api version control, like: PreviewVersionOverOneYear

As part of the lintdiff migration effort we're migrating all of the rules defined in the Azure OpenAPI Validator repo to a typespec lint. Initially our focus is solely on the ARM + common rules. Here is a link to the repo: https://github.com/Azure/azure-openapi-validator/tree/main

This branch contains from initial work on the migrated lints (IMPORTANT: the rules defined in this branch will be incorporated into existing rulesets not a new one as it's currently represented in the branch): https://github.com/Azure/typespec-azure/tree/feature/lintdiff-migration/packages/typespec-lintdiff

catalinaperalta: This might be an easier view of the rules (please note that some of the classifications may change since these were carried out with an agent and human review may change some classifications): https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration/packages/typespec-lintdiff/catalog/CATALOG.md

xirzec: > 🤖 *Written by GitHub Copilot (an AI agent) on behalf of @xirzec.*

Thanks @catalinaperalta — that catalog changes our plan, in a good way.

**`ParameterNamesConvention` is a direct hit.** It's the two rules we were about to propose next (`@query("...")` camelCase, `@header("...")` kebab-case). We're dropping them — they're yours. For whatever it's worth, our independent measurement agrees with your `Unconstrained` classification: ~16% of data-plane services have genuinely non-camel query parameter names once `api-version` and `$`-prefixed names are excluded.

**Two questions:**

1. Does `no-dollar-prefixed-query-params` (the one rule left in this PR) overlap anything in the catalog? I don't see it, but you'd know better. If it does, we'll close this rather than merge it.
2. Do `VersionConvention` / `ApiVersionEnum` cover api-version *format*? We dropped a date-format rule from this PR after measuring 26% of data-plane services using non-date versions (keyvault `7.5`, maps `1.0`, monitor `v1`, translation `3.0`, …), which is over the 20% bar in @markcowl's [when-to-write-a-linting-rule](https://azure.github.io/typespec-azure/docs/howtos/contributing/when-to-write-a-linting-rule/) guidance. If those rules do enforce a format, that measurement might be relevant to your migration too.

**One observation worth sharing.** Your tiering — 29 Infallible + 46 Template-enforced + 4 Prerequisite-blocked out of 209 — matches what we found independently from the other direction: a large share of the Azure REST API Guidelines is already true-by-construction for anyone going through the `Azure.Core` templates. Your 130 `Unconstrained` rules look like a much better scope definition for agentic review than the raw guidelines document, since they're already triaged for what a linter can and can't reach.

Separately, we're building a data-plane agentic reviewer in `azure-rest-api-specs` ([#45079](https://github.com/Azure/azure-rest-api-specs/pull/45079)) for the judgment-based residue. It keeps a hand-maintained map of which rules are linter-owned so it stays silent on them. If the catalog is going to be the authoritative version of that map, we'd rather consume yours than maintain a parallel one — happy to talk.

PR #5059, opened 2026-07-27
fix(eng): trigger all emitter CI on any pnpm-workspace.yaml change

### LINKED ISSUE #5010: Python CI not triggered when http-client-python is updated in pnpm-workspace.yaml (PR #5009)
## Problem

In PR https://github.com/Azure/typespec-azure/pull/5009, the Python CI was **not triggered**, which is unexpected.

The PR updates @typespec/http-client-python in pnpm-workspace.yaml (and pnpm-lock.yaml). Since http-client-python was changed in pnpm-workspace.yaml, the Python CI **should** have been triggered.

## Expected behavior

When http-client-python is updated in pnpm-workspace.yaml, the Python CI should be triggered.

## Actual behavior

Python CI was not triggered for PR #5009.

msyyc:
Fixes #5010.

## Problem

`@typespec/http-client-python` is not a workspace package — it is consumed from npm through the pnpm **catalog**, so bumping it only edits the root `pnpm-workspace.yaml`.

pnpm attributes root-file changes to the **root package only**, so `pnpm --filter "...[base]"` reports no affected emitter and `detect-affected` returns `python: false`. That is why #5009 ran no Python CI.

## Fix

Add `pnpm-workspace.yaml` to `sharedExtra`, so any change to it triggers every emitter target — the same treatment the `core` submodule pointer already gets.

The file changes rarely and always affects dependency resolution repo-wide, so this is deliberately blunt rather than trying to work out which catalog entry moved.

## Verification

Simulated the #5009 diff locally with `BASE_SHA=origin/main`:

```
Affected targets: {"python":true,"java":true,"typescript":true,"go":true}
```

Plus a unit test in `eng/scripts/detect-affected.test.ts`.

### DIFF
--- a/eng/scripts/detect-affected.config.json
+++ b/eng/scripts/detect-affected.config.json
@@ -1,5 +1,6 @@
 {
   "submodulePath": "core",
+  "catalogPath": "pnpm-workspace.yaml",
   "ignore": ["**/test/**", "**/tests/**", "**/*.test.ts", "**/*.md"],
   "sharedExtra": [".github/actions/setup/**"],
   "targets": {
--- a/eng/scripts/detect-affected.config.json
+++ b/eng/scripts/detect-affected.config.json
@@ -1,8 +1,7 @@
 {
   "submodulePath": "core",
-  "catalogPath": "pnpm-workspace.yaml",
   "ignore": ["**/test/**", "**/tests/**", "**/*.test.ts", "**/*.md"],
-  "sharedExtra": [".github/actions/setup/**"],
+  "sharedExtra": [".github/actions/setup/**", "pnpm-workspace.yaml"],
   "targets": {
     "python": {
       "package": "@azure-tools/typespec-python",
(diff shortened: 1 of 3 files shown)


timotheeguerin (approved):

eng/scripts/detect-affected.config.json
```diff
@@ -1,7 +1,7 @@
 {
   "submodulePath": "core",
   "ignore": ["**/test/**", "**/tests/**", "**/*.test.ts", "**/*.md"],
-  "sharedExtra": [".github/actions/setup/**"],
+  "sharedExtra": [".github/actions/setup/**", "pnpm-workspace.yaml"],
```
timotheeguerin, line 4: might want the pnpm-lock.yaml too

msyyc: I take a look for many PRs then find `pnpm-lock.yaml` is usually changed together with `pnpm-workspace.yaml` so I only put one.

PR #5065, opened 2026-07-27
feat(typespec-metadata): add apiVersion field using TCGC resolution

iscai-msft:
## Summary

Adds an `apiVersion` field to the `LanguagePackageMetadata` interface in `packages/typespec-metadata`, resolved via `createSdkContext` from `@azure-tools/typespec-client-generator-core`.

## Changes

- **`src/metadata.ts`**: Added `apiVersion?: string` to `LanguagePackageMetadata` with doc comment explaining possible values
- **`src/emitter.ts`**: Calls `createSdkContext` and resolves `sdkPackage.metadata.apiVersions` map
- **`src/collector.ts`**: Threads `resolvedApiVersion` through `collectLanguagePackages` → `buildLanguageMetadata` → `createLanguageMetadata`
- **`package.json`**: Added `@azure-tools/typespec-client-generator-core` as `"workspace:^"` dependency
- **`test/collector.test.ts`**: Added "apiVersion extraction" describe block with 5 tests

## Resolution Logic

| Map state | Resolved value |
|-----------|---------------|
| size > 1 | `"multiple-versions"` |
| size == 1 | The single value (`"all"` or actual version like `"2023-10-01"`) |
| empty/undefined | `undefined` |


packages/typespec-metadata/src/metadata.ts
```diff
@@ -20,6 +20,14 @@ export interface LanguagePackageMetadata {
...
   /** Service directory path for this language emitter. */
   serviceDir?: string;
+  /**
+   * Resolved API version for this emitter.
+   * - `"all"` when configured for all versions
+   * - `"multiple-versions"` for multi-service configs with more than one API version entry
+   * - An actual resolved version string (e.g. `"2023-10-01"`) for a single version
+   * - `undefined` when no API version information is available
+   */
+  apiVersion?: string;
```
praveenkuttappan, line 30: Should this be in language metadata or typespec metadata? Is it possible to have different API version value for each language?

iscai-msft: you're right, I guess each language could specify different values in tspconfig.yaml, so it shouldn't be the case but it could be

iscai-msft: would you prefer to ignore that possibility and move it to typespec metadata, or keep it where it is in language metadata

packages/typespec-metadata/test/tester.ts
```diff
@@ -0,0 +1,19 @@
+import { resolvePath } from "@typespec/compiler";
+import { createTester } from "@typespec/compiler/testing";
+
+export const MetadataTester = createTester(resolvePath(import.meta.dirname, ".."), {
```
timotheeguerin: I think that structure is not defiened correctly.

just need `ApiTester` and then `EmitterTester` which is the `ApiTester.emit(....)`

packages/typespec-metadata/src/metadata.ts
praveenkuttappan: Nope. Let's keep it in language metadata. It's easier to parse all get the distinct on the consumer side.

praveenkuttappan (approved):

timotheeguerin (approved):

PR #5071, opened 2026-07-27
feat(tcgc): add isPreview to SdkPackageType metadata

iscai-msft:
Adds `isPreview` field to `SdkPackageType.metadata` so language emitters can determine whether the resolved API version is a preview version.

Split out from #5065.


tadelesh:
What's the user scenario for this featue?

PR #5075, opened 2026-07-28
Add generated Python test fixtures

msyyc:
related upstream PR: https://github.com/microsoft/typespec/pull/11422

Adds generated Python test fixtures required by the test suite and syncs the marked `.gitignore` fixture rules from upstream. Missing markers now produce a warning and skip the `.gitignore` update.


l0lawrence:
I thought the point of the emitter-diff tool was to avoid requiring having to store all the generated tests in this repo? (under core submodule). Like this: https://github.com/microsoft/typespec/pull/11370/changes#issuecomment-5063100462

msyyc:
> I thought the point of the emitter-diff tool was to avoid requiring having to store all the generated tests in this repo? (under core submodule). Like this: https://github.com/microsoft/typespec/pull/11370/changes#issuecomment-5063100462

@l0lawrence I updated the regenerate logic and now we only need 30 files which contains customized code for test.

msyyc:
@copilot+claude-opus-5 resolve the merge conflicts in this pull request

packages/typespec-python/.gitignore
```diff
@@ -0,0 +1,52 @@
+# Generated SDK fixtures with hand-authored/customized code required by tests. (begin)
```
tadelesh, line 1: can we merge this into the global git ignore fire?

tadelesh (approved):

packages/typespec-python/.gitignore
msyyc: Just to keep same structure with http-client-python to make maintenance easy when sync changes between http-client-python and typespec-python.

PR #5077, opened 2026-07-28
feat(typespec-go): render doc comment lists per Go doc convention

### LINKED ISSUE #4988: Full Support of Documentation Generation
- spector test: https://github.com/microsoft/typespec/tree/main/packages/http-specs/specs/documentation

Previously, we tried to add this spector test: https://github.com/Azure/autorest.go/pull/1850, but the generated documentation has various formatting issues.

tadelesh:
Resolves #4988.

## What
Adds support for rendering **bullet and numbered lists** in generated Go doc comments following the native Go doc convention, and opts the upstream Spector `documentation` spec into generation with hand-written client tests.

## Problem
Previously the emitter word-wrapped doc text without list awareness: list items wrapped to column 0 (breaking the item), and blank lines between paragraphs/lists were collapsed. As a result `go doc`/`gofmt` did not recognize lists (see the earlier attempt Azure/autorest.go#1850, which had formatting issues).

## Fix
In `packages/typespec-go/src/codegen/core/helpers.ts`, `formatDocComment` and `formatDocCommentWithPrefix` now route the description through a new list-aware renderer that:
- detects bullet (`-` / `*` / `+`) and numbered (`1.` / `1)`) list items;
- emits items with the canonical Go indentation (`//   - item` / `//  1. item`) and aligns wrapped continuation lines under the item text;
- preserves blank lines and inserts a blank comment line around lists so `gofmt`/`go doc` recognize them.

Non-list, blank-line-free docs are byte-identical to before (all existing snapshots unchanged). Applies across models, fields, options, responses, enums/enum values, and operations. Markdown emphasis (`**bold**` / `*italic*`) is intentionally left as literal text since Go doc has no equivalent.

Example generated output:
```go
// BulletPointsModel - This tests:
//
//   - Simple bullet point. This bullet point is going to be very long to test how text wrapping is handled in bullet points within
//     documentation comments. It should properly indent the wrapped lines.
//   - Another bullet point with **bold text**.
//
// Steps to follow:
//
//  1. First step with **important** note
//  2. Second step with *emphasis*
```

## Conventions & limitations
- **Marker indentation** follows go/doc/comment: list content is indented to a 4-column tab stop (continuation lines align there too). The marker is left-padded to width 4, so a 2-char `-` bullet gets one more leading space than a 3-char `N.` item — hence `//   - ` (3 spaces) vs `//  1. ` (2 spaces). Output is gofmt-idempotent.
- **Line length**: list lines may exceed the emitter's 120-char soft wrap target by a few characters, because the text is wrapped against the base `// ` prefix and the list lead (`//   - ` etc.) is then applied. This matches the existing `formatCommentAsBulletItem` behavior; `gofmt`/`go doc` impose no line-length limit on comments.
- **Nested (multi-level) lists are not supported**: the Go doc comment format has no concept of nested lists and `gofmt` flattens indented sub-items to a single level. Source indentation therefore carries no nesting meaning and every item is rendered at one level. This is documented in `renderDocBody` / `matchDocListItem`.

## Tests
- New unit scenario `test/unittest/scenarios/doc-comment-lists.md` covering bullet + numbered lists in model, field, and enum docs.
- Opts the Spector `documentation` spec into `spector.config.http.yaml` (module `documentationgroup`) with hand-written `*_client_test.go` files (generated `zz_*` output is gitignored).
- Validated: `pnpm vitest run` (49 unit tests pass); `go vet` clean; all 6 `documentationgroup` tests pass against the Spector mock server.


tadelesh:
Superseded by #5078, which uses a branch in this repository instead of a fork.

PR #5078, opened 2026-07-28
feat(typespec-go): render doc comment lists per Go doc convention

### LINKED ISSUE #4988: Full Support of Documentation Generation

tadelesh:
Resolves #4988.

## What
Adds support for rendering **bullet and numbered lists** in generated Go doc comments following the native Go doc convention, and opts the upstream Spector `documentation` spec into generation with hand-written client tests.

## Problem
Previously the emitter word-wrapped doc text without list awareness: list items wrapped to column 0 (breaking the item), and blank lines between paragraphs/lists were collapsed. As a result `go doc`/`gofmt` did not recognize lists (see the earlier attempt Azure/autorest.go#1850, which had formatting issues).

## Fix
In `packages/typespec-go/src/codegen/core/helpers.ts`, `formatDocComment` and `formatDocCommentWithPrefix` now route the description through a new list-aware renderer that:
- detects bullet (`-` / `*` / `+`) and numbered (`1.` / `1)`) list items;
- emits items with the canonical Go indentation (`//   - item` / `//  1. item`) and aligns wrapped continuation lines under the item text;
- inserts a blank comment line immediately before and after a list so `gofmt`/`go doc` recognize it.

Non-list, blank-line-free docs are byte-identical to before (all existing snapshots unchanged). Applies across models, fields, options, responses, enums/enum values, and operations. Markdown emphasis (`**bold**` / `*italic*`) is intentionally left as literal text since Go doc has no equivalent.

Example generated output:
```go
// BulletPointsModel - This tests:
//
//   - Simple bullet point. This bullet point is going to be very long to test how text wrapping is handled in bullet points within
//     documentation comments. It should properly indent the wrapped lines.
//   - Another bullet point with **bold text**.
//
// Steps to follow:
//
//  1. First step with **important** note
//  2. Second step with *emphasis*
```

## Conventions & limitations
- **Marker indentation** follows go/doc/comment: list content is indented to a 4-column tab stop (continuation lines align there too). The marker is left-padded to width 4, so a 2-char `-` bullet gets one more leading space than a 3-char `N.` item — hence `//   - ` (3 spaces) vs `//  1. ` (2 spaces). Output is gofmt-idempotent.
- **Line length**: list lines may exceed the emitter's 120-char soft wrap target by a few characters, because the text is wrapped against the base `// ` prefix and the list lead (`//   - ` etc.) is then applied. This matches the existing `formatCommentAsBulletItem` behavior; `gofmt`/`go doc` impose no line-length limit on comments.
- **Nested (multi-level) lists are not supported**: the Go doc comment format has no concept of nested lists and `gofmt` flattens indented sub-items to a single level. Source indentation therefore carries no nesting meaning and every item is rendered at one level. This is documented in `renderDocBody` / `matchDocListItem`.

## Tests
- New unit scenario `test/unittest/scenarios/doc-comment-lists.md` covering bullet + numbered lists in model, field, and enum docs.
- Opts the Spector `documentation` spec into `spector.config.http.yaml` (module `documentationgroup`) with hand-written `*_client_test.go` files (generated `zz_*` output is gitignored).
- Validated: `pnpm vitest run` (49 unit tests pass); `go vet` clean; all 6 `documentationgroup` tests pass against the Spector mock server.

- **Blank lines**: blank source lines are still not preserved as blank comment lines (matching the pre-existing `comment()` behavior); a blank comment line is only emitted where Go requires one to delimit a list. Docs without lists are therefore byte-identical to before.


packages/typespec-go/spector.config.http.yaml
```diff
@@ -9,6 +9,7 @@ specs:
   "authentication/http/custom": { options: { module: customgroup } }
   "authentication/oauth2": { options: { module: oauth2group } }
   "authentication/union": { options: { module: unionauthgroup } }
+  documentation: { options: { module: documentationgroup } }
```
jhendrixMSFT: Can you add quotes so it's consistent?

jhendrixMSFT (commented):
Overall this is a solid, well-scoped change. The list-aware renderer (`renderDocBody` / `matchDocListItem` / `formatDocListItem`) is clean, thoroughly commented, and the limitations (no nesting, soft-wrap overflow) are honestly documented. I verified two key claims empirically with `go1.26` gofmt:

- ✅ The exact emitted format is **gofmt-idempotent**, including multi-digit numbers (`//  10. tenth` is left untouched).
- ✅ Non-list docs stay byte-identical: `comment()` already converts `\n+` → hard break, so splitting per line and re-joining produces identical output for prose.

A few findings below.

### 1. (Medium) Lists at the very start of a top-level doc comment are not recognized → non-idempotent

Go only recognizes a list when it is preceded by a paragraph/text line. `renderDocBody` emits a list with no leading blank/text when `out` is empty (`lastIsBlank()` returns `true` on an empty `out`). So for a top-level declaration whose description *begins* with a list and has no summary, gofmt de-indents the output:

```
//   - one     ->  // - one
//  1. one     ->  // 1. one
```

(Reproduced by running gofmt against `type`/`func`-level comments. Note struct **field** comments are *not* reformatted by gofmt, so field-level cases like the test's `Prop` are safe — only top-level `type`/`func`/`const` docs are exposed: operations, client types, options/response structs.)

Suggested guard: when a list would be the first emitted line, render it as gofmt would (de-indented), since Go cannot render a leading list anyway.

### 2. (Minor) `formatDocCommentWithPrefix` orphans the first item when a description is a pure list

With a prefix and no summary, `description` becomes `"Prop - - one\n- two"`. The first line `Prop - - one` doesn't match the bullet regex, so item one is absorbed into the prefix as prose and only the remainder renders as a list:

```
// Prop - - one
//
//   - two
```

### 3. (Minor) Test coverage

The scenario covers model/field/numbered lists well, but not the two edge cases above (leading list, prefix + pure-list). Worth adding a snapshot for each.

### 4. (Info) Line length > 120

Wrapped list items may exceed 120 chars. This is acknowledged in the PR description and matches the existing `formatCommentAsBulletItem` behavior; gofmt/go doc impose no limit, so this is fine.

Findings #1/#2 are narrow (they require a description that *starts* with a list on a prefix-less/summary-less top-level entity), but since gofmt-idempotency is the PR's explicit guarantee, they're worth a guard plus a snapshot test.

tadelesh:
Thanks — all four are addressed in 35df1bb9. Findings #1 and #2 were real; digging into `go/doc/comment` showed the underlying rule is a bit broader than "lists at the very start aren't recognized", so the fix is correspondingly broader.

### Root cause

`go/doc/comment` runs `unindent()` over the whole comment *before* it looks for lists ([parse.go:285](https://github.com/golang/go/blob/master/src/go/doc/comment/parse.go#L285), [506-524](https://github.com/golang/go/blob/master/src/go/doc/comment/parse.go#L506-L524)) and strips the **common** leading whitespace of every non-blank line. So the real condition is:

| comment content | result |
| --- | --- |
| every line is a list item | common prefix = 2/3 spaces, stripped -> paragraph -> gofmt rewrites it |
| at least one prose line anywhere | common prefix = 0 -> list survives |

A leading list is fine on its own — verified with gofmt that this is stable:

```go
//   - one
//
// prose
//   - two
```

What breaks is a comment with *no* prose line at all, wherever the list sits.

### #1 — fixed

Lists are rendered only when the comment has a prose line (either supplied by the caller as a lead-in, or present in the body). Otherwise the text falls back to plain wrapped prose — which is what gofmt produces *and* what the emitter produced before this PR, so there is no regression relative to `main`.

I first tried your suggestion of emitting de-indented markers (`// - item`), but that isn't stable either: when an item wraps, its continuation lines are indented relative to the marker, which trips the parser's `unindentedListOK` heuristic and gofmt re-indents the whole thing. Plain prose avoids that.

### #2 — fixed, plus two more instances

Rather than dropping the lead-in, it now moves to its own line, which fixes the orphaning *and* restores the prose line that keeps the list renderable:

```go
// Prop -
//
//   - one
//   - two
```

While fixing `formatDocCommentWithPrefix` I found the same merge-into-first-line bug in two more places, so all four now share one helper (`go.prefixDocWithName`):

- `tcgcadapter/types.ts` — the `<ModelName> - ` prefix
- `tcgcadapter/clients.ts` — the `<ClientName> - ` prefix
- `codegen/core/models.ts` — the `REQUIRED`/`READ-ONLY`/`CONSTANT` field modifiers, which were producing `// REQUIRED; - one`

### Bonus fix

A blank source line between a list and following prose swallowed the required blank comment line, because it reset the "in list" state:

```go
//   - one
//   - two
- // after      <- gofmt inserts a blank line here
```

Now tracked separately from the loose/tight-list state.

### #3 — added

`doc-comment-lists.md` now covers a pure-list model doc, a pure-list required field (modifier case), a pure-list optional field (no lead-in at all), and a pure-list enum + enum value (prefix case). Snapshots go through `gofmt -s`, so they double as idempotency assertions.

### #4 — no change

Agreed, leaving as-is.

### spector.config.http.yaml

Quoted, thanks.

### Validation

- Built a file with 17 hand-constructed edge cases (pure lists, leading lists, blank-separated items, multi-digit numbering, `*`/`+` markers, wrapped items, prefix/summary combinations) straight from the emitter's raw pre-gofmt output: `gofmt -d` reports **no diff**.
- Regenerated **every** Spector module and compared all 1314 `zz_*.go` files against the pre-change output: **0 changed**. The fixes only touch shapes that don't occur in the current corpus, so the blast radius is nil.
- 50/50 unit tests, oxlint and prettier clean.

jhendrixMSFT (commented):
## Re-review — all feedback addressed ✅

Re-reviewed at `48d69b98` (fix commit `35df1bb9`). The updates resolve every point from my earlier review, and I re-verified idempotency against `go1.26` gofmt.

### 1. (was Medium) Leading list at start of a top-level doc comment — **fixed**
The renderer now only formats lists when a prose line exists (`renderLists = hasPrecedingProse || hasProseLine(text)` in `renderDocBody`), because `go/doc/comment` strips common indentation and demotes an all-list comment to a paragraph. Two complementary paths:
- `formatDocCommentWithPrefix` routes the description through `prefixDocWithName`, which drops the `Name -` lead onto its own line (`Name -\n…`) when the text starts with a list, restoring a prose line so the list stays renderable — and passes `hasPrecedingProse = true`.
- With no prose available (e.g. an optional field that is a pure list), it falls back to plain de-indented prose (`// - …`), exactly what gofmt would produce.

Verified: the pure-list model, enum, and enum-value outputs all gofmt-idempotent, and the de-indented optional-field fallback is stable too.

### 2. (was Minor) Prefix/modifier orphaning the first list item — **fixed**
In `models.ts`, when field modifiers (`REQUIRED`, etc.) precede a description that starts with a list, the modifier is now placed on its own line (`${mods.join("; ")}\n${description}`) instead of being joined with `; `, so the first item is no longer absorbed into the modifier. Confirmed by the new `Required` field snapshot.

### 3. (was Minor) Test coverage — **fixed**
`doc-comment-lists.md` now covers the edge cases: a pure-list model (`PureListModel`), pure-list required vs. optional fields, and a pure-list enum + enum value (`PureListEnum`), across both the `models` and `constants` generated blocks.

### 4. (Info) Line length > 120 — unchanged, as expected/acceptable.

Nice refactor extracting the shared list primitives into `codemodel/docs.ts` so the TCGC adapter and generator agree on what a list item is. LGTM.

jhendrixMSFT (approved):

PR #5079, opened 2026-07-28
feat(typespec-go): support paging with a relative nextLink

### LINKED ISSUE #4994: Support paging with relative nextlink
spector test: https://github.com/Azure/typespec-azure/tree/main/packages/azure-http-specs/specs/azure/core/page

tadelesh:
Fixes #4994

### Summary

Supports paging when the service returns a `nextLink` that's relative to the endpoint, e.g. `"/azure/core/page/with-relative-next-link/page/2"`. Today this fails with `no Host in request URL`.

Per review feedback, the resolution itself lives in azcore ([Azure/azure-sdk-for-go#27291](https://github.com/Azure/azure-sdk-for-go/pull/27291)). This PR only passes the client endpoint through the new `runtime.FetcherForNextLinkOptions.Endpoint` field:

```diff
 		resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), nextLink, func(ctx context.Context) (*policy.Request, error) {
 			return client.listCreateRequest(ctx, apiVersion, options)
-		}, nil)
+		}, &runtime.FetcherForNextLinkOptions{
+			Endpoint: client.endpoint,
+		})
```

The endpoint expression is `client.internal.Endpoint()` for ARM and the client's single URI parameter otherwise. It's omitted for templated hosts (assembled per method) and when a custom next page operation builds the request itself.

Enables the previously disabled `Azure_Core_Page_withRelativeNextLink` Spector test.

### Dependency

Uses azcore main pseudo-version `v1.22.1-0.20260731225345-5811a2e9dc69`, which contains the merged [Azure/azure-sdk-for-go#27291](https://github.com/Azure/azure-sdk-for-go/pull/27291). It will be replaced with `v1.22.1` after that version is released.

### Validation

- 48/48 unit tests, oxlint and prettier clean
- all 105 generated Go modules build
- relative-nextLink Spector test passes against the azcore main pseudo-version

### DIFF
--- a/packages/typespec-go/test/unittest/scenarios/pageable-lro.md
+++ b/packages/typespec-go/test/unittest/scenarios/pageable-lro.md
@@ -102,17 +102,11 @@ func (client *PageableLROsClient) BeginListPrivateEndPoints(ctx context.Context,
 			return page.NextLink != nil && len(*page.NextLink) > 0
 		},
 		Fetcher: func(ctx context.Context, page *PageableLROsClientListPrivateEndPointsResponse) (PageableLROsClientListPrivateEndPointsResponse, error) {
-			nextLink := *page.NextLink
-			// the service can return a next link that's relative to the endpoint, however
-			// runtime.FetcherForNextLink requires an absolute URL, so resolve it here.
-			if nextLink != "" {
-				if u, err := url.Parse(nextLink); err == nil && !u.IsAbs() {
-					nextLink = runtime.JoinPaths(client.internal.Endpoint(), nextLink)
-				}
-			}
-			resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), nextLink, func(ctx context.Context) (*policy.Request, error) {
+			resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), *page.NextLink, func(ctx context.Context) (*policy.Request, error) {
 				return client.listPrivateEndPointsCreateRequest(ctx, apiVersion, resourceGroupName, resourceName, options)
-			}, nil)
+			}, &runtime.FetcherForNextLinkOptions{
+				Endpoint: client.internal.Endpoint(),
+			})
 			if err != nil {
 				return PageableLROsClientListPrivateEndPointsResponse{}, err
 			}
--- a/packages/typespec-go/test/unittest/scenarios/pageable-lro.md
+++ b/packages/typespec-go/test/unittest/scenarios/pageable-lro.md
@@ -102,7 +102,15 @@ func (client *PageableLROsClient) BeginListPrivateEndPoints(ctx context.Context,
 			return page.NextLink != nil && len(*page.NextLink) > 0
 		},
 		Fetcher: func(ctx context.Context, page *PageableLROsClientListPrivateEndPointsResponse) (PageableLROsClientListPrivateEndPointsResponse, error) {
-			resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), *page.NextLink, func(ctx context.Context) (*policy.Request, error) {
+			nextLink := *page.NextLink
+			// the service can return a next link that's relative to the endpoint, however
+			// runtime.FetcherForNextLink requires an absolute URL, so resolve it here.
+			if nextLink != "" {
+				if u, err := url.Parse(nextLink); err == nil && !u.IsAbs() {
+					nextLink = runtime.JoinPaths(client.internal.Endpoint(), nextLink)
+				}
+			}
+			resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), nextLink, func(ctx context.Context) (*policy.Request, error) {
 				return client.listPrivateEndPointsCreateRequest(ctx, apiVersion, resourceGroupName, resourceName, options)
 			}, nil)
 			if err != nil {
--- a/packages/typespec-go/test/unittest/scenarios/tenant-resource.md
+++ b/packages/typespec-go/test/unittest/scenarios/tenant-resource.md
@@ -287,16 +287,11 @@ func (client *TenantItemsClient) NewListPager(apiVersion string, options *Tenant
 			if page != nil {
 				nextLink = *page.NextLink
 			}
-			// the service can return a next link that's relative to the endpoint, however
-			// runtime.FetcherForNextLink requires an absolute URL, so resolve it here.
-			if nextLink != "" {
-				if u, err := url.Parse(nextLink); err == nil && !u.IsAbs() {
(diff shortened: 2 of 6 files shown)


tadelesh:
@jhendrixMSFT I remembered that you said we already support it in swagger cases. But I could not find the related code. Current implementation will change all the generated code, which I do not prefer. Any suggestions?

packages/typespec-go/test/unittest/scenarios/pageable-lro.md
```diff
@@ -102,7 +102,15 @@ func (client *PageableLROsClient) BeginListPrivateEndPoints(ctx context.Context,
 			return page.NextLink != nil && len(*page.NextLink) > 0
 		},
 		Fetcher: func(ctx context.Context, page *PageableLROsClientListPrivateEndPointsResponse) (PageableLROsClientListPrivateEndPointsResponse, error) {
-			resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), *page.NextLink, func(ctx context.Context) (*policy.Request, error) {
+			nextLink := *page.NextLink
+			// the service can return a next link that's relative to the endpoint, however
```
jhendrixMSFT: We shouldn't be emitting comments.

packages/typespec-go/test/unittest/scenarios/pageable-lro.md
```diff
@@ -102,7 +102,15 @@ func (client *PageableLROsClient) BeginListPrivateEndPoints(ctx context.Context,
...
+			// runtime.FetcherForNextLink requires an absolute URL, so resolve it here.
+			if nextLink != "" {
+				if u, err := url.Parse(nextLink); err == nil && !u.IsAbs() {
```
jhendrixMSFT: A few problems here.
- `err` is discarded when not `nil`
- we parse just to check `IsAbs()` then throw that away
  - this happens for all paths and IIRC relative paths are rare

Do we know at codegen time if the path is relative?

jhendrixMSFT (changes requested):

packages/typespec-go/test/unittest/scenarios/pageable-lro.md
tadelesh: No, we could not know that. It's kind of a runtime behavior. I remembered that you said we already support it in swagger cases. But I could not find the related code. Current implementation will change all the generated code, which I do not prefer. Any suggestions?

jhendrixMSFT: The swagger case was different than I remembered :(.  e.g. [here](https://github.com/Azure/autorest.go/blob/main/packages/autorest.go/test/autorest/paginggroup/zz_paging_client.go#L390-L392).  It uses a swagger-defined operation for constructing the request for the next link.

It might be best to update `azcore/runtime/FetcherForNextLink` to handle this case.  We'd still have to change all of the codegen call sites, but the logic can be centralized.

tadelesh:
Reworked per your suggestion — the resolution now lives in azcore: Azure/azure-sdk-for-go#27291.

`FetcherForNextLinkOptions` gains an `Endpoint` field, and this PR just passes the client endpoint at every pager call site. No more emitted comments, no inline `url.Parse`, and nothing is emitted for clients where the endpoint can't be determined (templated hosts, or a custom next page operation that builds the request itself).

On the azcore side:
- `url.Parse` decides whether the next link is absolute; a malformed link is passed through unchanged so the failure still surfaces from `NewRequest`.
- the join happens on the raw strings rather than the parsed components so percent-encoding (e.g. `%2F`) survives.
- resolution uses `runtime.JoinPaths` (append to the endpoint) rather than RFC 3986 reference resolution, matching how the Go SDK builds every other request URL and matching the JS/Python emitters.

This PR is blocked on the azcore v1.22.1 release; the minimum azcore version in `gomod.ts` will be bumped in a follow-up.

tadelesh:
azcore support has merged in Azure/azure-sdk-for-go#27291. Updated the emitter minimum azcore version to the current `main` pseudo-version (`v1.22.1-0.20260731225345-5811a2e9dc69`), regenerated the page client, and verified the relative-nextLink Spector test passes without a local replace. Commit: cccfaa44.

packages/typespec-go/test/unittest/scenarios/tenant-resource.md
```diff
@@ -289,7 +289,9 @@ func (client *TenantItemsClient) NewListPager(apiVersion string, options *Tenant
 			}
 			resp, err := runtime.FetcherForNextLink(ctx, client.internal.Pipeline(), nextLink, func(ctx context.Context) (*policy.Request, error) {
 				return client.listCreateRequest(ctx, apiVersion, options)
-			}, nil)
+			}, &runtime.FetcherForNextLinkOptions{
+				Endpoint: client.internal.Endpoint(),
+			})
```
tadelesh, line 294: @jhendrixMSFT Still needs to add this for all paging op. Is this acceptable?

jhendrixMSFT: I think so.

Do we have cases/support using custom next page operations in tsp?  I ask because there might be a problematic intersection in `FetcherForNextLink`.  If the next link is relative, and the custom next link operation expects it to be relative, when `FetcherForNextLink` calls `options.NextReq` we will pass in an absolute next link which would be a problem.

tadelesh: In TypeSpec, we only have one legacy case will pass `FetcherForNextLinkOptions` which is non-get next page, and it is covered by Spector test. So, I think there should be no intersection for TypeSpec cases.

jhendrixMSFT: So just to confirm, in tsp we don't support `operationName` like in https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-pageable?

tadelesh: No, we don't.

jhendrixMSFT: I have a new idea for this and will get folded together with the work mentioned in https://github.com/Azure/typespec-azure/pull/5142.

jhendrixMSFT: Does this mean we can remove `sdkMethod.pagingMetadata.nextLinkOperation` from tcgc?

tadelesh: Seems so. I think in Swagger, we used it to change next link fetch HTTP method. Do you know other usages?

tadelesh:
Replaced by https://github.com/Azure/typespec-azure/pull/5198

PR #5085, opened 2026-07-28
Fix handling of utcDateTime types

### LINKED ISSUE #4924: Example generator emits invalid Go for RFC7231 datetime values (regression from #1992)
> Migrated from https://github.com/Azure/autorest.go/issues/2026

---

## Description

The example test generator produces uncompilable Go code for `utcDateTime` properties encoded as `rfc7231`. The generated value is wrapped in `strconv.ParseInt` (the Unix-timestamp path) with an unquoted, non-numeric date string.

## Repro

Generate `armdomainservices` (or any spec with an `rfc7231`-encoded datetime, e.g. `ConfigDiagnostics.lastExecuted`).

## Actual (broken) output

```go
LastExecuted: to.Ptr(func() time.Time { t, _ := strconv.ParseInt(05 May 2021 12:00:23 GMT, 10, 64); return time.Unix(t, 0).UTC()}()),
```

- `05 May 2021 12:00:23 GMT` is unquoted -> does not compile
- Wrong parse path entirely (Unix timestamp instead of RFC1123 layout)

## Expected output

```go
LastExecuted: to.Ptr(func() time.Time { t, _ := time.Parse(time.RFC1123, "05 May 2021 12:00:23 GMT"); return t }()),
```

## Root cause

PR Azure/autorest.go#1992 ("Use RFC7231 for tsp instead of RFC1123", commit `d9189b9ac`) changed the TypeSpec adapter to emit the `RFC7231` `TimeFormat` value instead of `RFC1123`. It updated the consumers in `servers.ts`, `operations.ts`, `models.ts`, and `type.ts`, **but missed `packages/codegen.go/src/core/example.ts`**.

`getTimeValue`'s `formatMap` only handled `RFC1123` and `RFC3339`, so `RFC7231` fell through to the `else` branch, which assumes a Unix timestamp:

```ts
const formatMap: Record<string, string> = {
  PlainDate: helpers.plainDateFormat,
  PlainTime: helpers.plainTimeFormat,
  RFC1123: helpers.RFC1123Format,
  RFC3339: helpers.RFC3339Format,
  // RFC7231 missing
};
```

## Fix

Add `RFC7231: helpers.RFC1123Format` to the map (mirrors the existing mapping in `fake/servers.ts`). Both share the Go `time.RFC1123` layout.

## Regression range

Introduced in Azure/autorest.go#1992 (emitter versions after 0.14.0).

### LINKED ISSUE #4956: When can codegen assume GMT/UTC/etc time
> Migrated from https://github.com/Azure/autorest.go/issues/942

---

Investigate if there's a way to know when generated code should set the time zone.

jhendrixMSFT:
Time types defined as utcDateTime now properly convert their values to UTC before sending over the wire. This changed the semantics of the code model's Time.utc value; when true, it indicates that the value must be converted to UTC before sending (see doc comments for more info).
Updated the example codegen for time types to construct the time.Time values in place via time.Date() instead of using time.Parse() which was leaking the wire format into the examples.
Improved error message when post codegen steps fail due to a Windows MAX_PATH violation.
Added VSCode settings to avoid scanning node modules and temp files which tends to peg the CPU for long periods of time. Updated developer docs with info on how to use mise to configure the repo required tooling.

Fixes:
- https://github.com/Azure/typespec-azure/issues/4924
- https://github.com/Azure/typespec-azure/issues/4956


packages/typespec-go/src/emitter.ts
```diff
@@ -257,6 +261,38 @@ export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
...
+  if (process.platform === "win32" && error.code === "ENOENT") {
+    return (
+      `unable to run '${command}': the tool could not be launched. This is typically caused by the ` +
+      `emitter output directory path (${outputDir.length} characters) exceeding the Windows maximum ` +
+      `path length of ${WINDOWS_MAX_PATH}; Windows cannot use a working directory longer than this. ` +
+      `Emit to a shorter output directory. (output directory: ${outputDir})`
+    );
+  }
+  return error.message;
+}
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

packages/typespec-go/.vscode/settings.json
tadelesh, line 1: I'm not sure if it is valid to commit vscode config files in this repo. cc: @timotheeguerin

tadelesh (approved):

packages/typespec-go/.vscode/settings.json
jhendrixMSFT: I'm fine to remove it if needed and can add it to `.gitignore`.  But the scanning problem is a real issue (at least on Windows).

timotheeguerin: I think its fine to add that if that's the workflow you want to work with, only thing I would maybe check is that we have things in common with the top level one.

jhendrixMSFT:
The "go / emitter diff / Generate & Diff" step was skipped.  It appears to be intention since the PR comes from a fork (see [here](https://github.com/Azure/typespec-azure/blob/main/.github/workflows/ci-emitter-diff-go.yml#L38)).  Why is that?

tadelesh:
> The "go / emitter diff / Generate & Diff" step was skipped. It appears to be intention since the PR comes from a fork (see [here](https://github.com/Azure/typespec-azure/blob/main/.github/workflows/ci-emitter-diff-go.yml#L38)). Why is that?

It is the limitation for GitHub action to do the diff from fork. You could use a feature branch in current repo instead.

jhendrixMSFT:
This is a pretty significant limitation.  Are there plans to fix it, is it even possible?  I would prefer to not use a branch in upstream unless there is really no other choice.

CC @heaths

timotheeguerin:
why can't the diff run on a fork?

tadelesh:
> why can't the diff run on a fork?

From previous Libba's PR: https://github.com/microsoft/typespec/pull/11122#discussion_r3543853066

timotheeguerin:
If it's just the comment we should still run it, what breaks is commenting on pr this way, there is way around we should just copy here like how we make the chronus comment, it just need 2 workflow which is anoyoing but it works

packages/typespec-go/.vscode/settings.json
jhendrixMSFT: fixed

xirzec (approved):

jhendrixMSFT:
I opened https://github.com/Azure/typespec-azure/issues/5123 to track fixing this.

PR #5086, opened 2026-07-28
Promote 14 pagination and error-handling scenarios from Backlog to Core

iscai-msft:
## Summary

Moves 14 scenarios from Backlog to Core tier in the Can I Use dashboard.

### Pagination (12 scenarios)
These patterns are used by every Azure list operation:
- **ContinuationToken** (6): All combinations of header/query request with body/header/nested response
- **Link-based** (3): `link`, `linkString`, `nestedLink`
- **PageSize** (2): `listWithPageSize`, `listWithoutContinuation`
- **XML Pagination** (2): `listWithContinuation`, `listWithNextLink`

### Error Handling (2 scenarios)
Every service returns errors; these test distinct wire behavior:
- `Response_StatusCodeRange_errorResponseStatusCodeInRange`
- `Response_StatusCodeRange_errorResponseStatusCode404`

### Rationale

Using the rubric:
1. **Service Prevalence**: All Azure list operations use pagination; all services return error status codes
2. **Wire Correctness**: Each tests a distinct request/response pattern not covered by other Core scenarios
3. **SDK Usability**: Users would file bugs if pagination or error handling didn't work

### Current adoption (unbranded emitters)
- 4 of these are already passing in all 4 languages (Python/C#/JS/Java)
- 8 more are passing in 3/4 (only JS missing — newest emitter)
- 2 are passing in Python/C#/Java

### DIFF
--- a/website/src/pages/can-i-use/scenario-tiers.ts
+++ b/website/src/pages/can-i-use/scenario-tiers.ts
@@ -60,19 +60,19 @@ export const scenarioTiers = {
       "Type_Union_Discriminated_NoEnvelope_CustomDiscriminator_get",
       "Type_Union_Discriminated_NoEnvelope_CustomDiscriminator_put",
 
-      // Pageable: ContinuationToken — skipped by Go, Java, JS (Python skips 1 of 6);
-      // custom continuation-token pagination; no ask in Azure yet (JS);
-      // Go tracks in https://github.com/Azure/autorest.go/issues/1494
+      // Pageable: ContinuationToken with header-based patterns — legacy Azure pattern
+      // (e.g., Cosmos DB x-ms-continuation); newer TypeSpec services use query+body instead
       "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestHeaderNestedResponseBody",
       "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestHeaderResponseBody",
       "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestHeaderResponseHeader",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryNestedResponseBody",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryResponseBody",
       "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryResponseHeader",
+      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryNestedResponseBody",
+
+      // Pageable: nestedLink — no Azure service found using @nextLink in nested response
+      "Payload_Pageable_ServerDrivenPagination_nestedLink",
 
       // Pageable: AlternateInitialVerb — newly added (microsoft/typespec#9966);
-      // POST-based initial pagination request; no language has implemented this yet;
-      // fits the same pageable backlog pattern as ContinuationToken/link/PageSize
+      // POST-based initial pagination request; not yet broadly implemented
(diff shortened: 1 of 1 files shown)


website/src/pages/can-i-use/scenario-tiers.ts
```diff
@@ -60,19 +60,8 @@ export const scenarioTiers = {
...
 
-      // Pageable: ContinuationToken — skipped by Go, Java, JS (Python skips 1 of 6);
-      // custom continuation-token pagination; no ask in Azure yet (JS);
-      // Go tracks in https://github.com/Azure/autorest.go/issues/1494
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestHeaderNestedResponseBody",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestHeaderResponseBody",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestHeaderResponseHeader",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryNestedResponseBody",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryResponseBody",
-      "Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryResponseHeader",
```
weidongxu-microsoft, line 71: any azure service doing this? continuationToken on header?

tadelesh: Same concern with weidong, are these promoted tests used by read service?

iscai-msft: i think this is how storage does paging

weidongxu-microsoft: I don't know... Least is that storage won't be using all such different scenarios?

iscai-msft: i can remove the continuationToken strings from core, not something I feel very strongly about, just figured i'd add them bc of storage. But i do think most languages are doing handwritten customizations in this case right now, maybe when we want this to be auto-generated I'll add it back in

weidongxu-microsoft: The only non-XML `@continuationToken` I can find in storage is this
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/QueueStorage/routes.tsp#L173-L189

I think it is `requestQueryResponseBody` case.
https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/payload/pageable/main.tsp#L232

website/src/pages/can-i-use/scenario-tiers.ts
```diff
@@ -95,28 +92,6 @@ export const scenarioTiers = {
       "Authentication_Noauth_Union_validNoAuth",
       "Authentication_Noauth_Union_validToken",
 
-      // Pageable: Link-based pagination — skipped by Java, JS;
-      // nextLink-based pagination patterns; no ask in Azure yet (JS)
-      "Payload_Pageable_ServerDrivenPagination_link",
-      "Payload_Pageable_ServerDrivenPagination_linkString",
-      "Payload_Pageable_ServerDrivenPagination_nestedLink",
```
weidongxu-microsoft, line 102: Similarly, not sure any Azure service does this `Payload_Pageable_ServerDrivenPagination_nestedLink` . Didn't find `@nextLink` in storage.

weidongxu-microsoft:
The question I had is that this table is used for TypeSpec Azure. We should not put test case that is anti-pattern for Azure service into the "Core" section.

weidongxu-microsoft (approved):

PR #5098, opened 2026-07-29
Add missing Spector tests for ARM resource IDs

jhendrixMSFT:
Skip the test for user-defined errors as this isn't yet supported in the emitter and the test is validating the wrong thing.


jhendrixMSFT:
The Go tests didn't run which seems bad.  Is it because of [this](https://github.com/Azure/typespec-azure/blob/main/eng/scripts/detect-affected.config.json#L3)?

timotheeguerin:
That would make snese, I think the logic needs to be tweaked that this ignore should only be for upstream changes

xirzec (approved):

PR #5101, opened 2026-07-29
Add warning diagnostic for clientDefaultValue type mismatch

### LINKED ISSUE #3934: Add a lint if the clientDefaultValue type doesn't match the actual property type
```
import "@typespec/rest";
import "@typespec/http";

using TypeSpec.Http;

@service(#{ title: "Pet Store Service" })
namespace PetStoreNamespace;

interface PetStore {
  /**
   * Get a list of pets
   */
  list(
    @Azure.ClientGenerator.Core.Legacy.clientDefaultValue("10")
    @query
    pageSize?: int32,

    @Azure.ClientGenerator.Core.Legacy.clientDefaultValue(123)
    @query
    sortOrder?: string,
  ): string[];
}
```
In the example above, the type of the value inside `@clientDefaultValue` doesn't match the actual parameter type. Should we add an error from compiler or tcgc?

iscai-msft:
## Problem

When using `@clientDefaultValue`, the type of the value passed doesn't have to match the actual property type. For example, `@clientDefaultValue("10")` on an `int32` property or `@clientDefaultValue(123)` on a `string` property would silently produce incorrect defaults in generated SDKs.

## Solution

Added a `client-default-value-type-mismatch` warning diagnostic that fires when the value type doesn't match the property's scalar type:
- String values require the property to be a string scalar
- Numeric values require the property to be a numeric scalar (int32, int64, float32, etc.)
- Boolean values require the property to be a boolean scalar

## Changes

- **`src/lib.ts`** — Added diagnostic definition
- **`src/decorators.ts`** — Added type validation in `$clientDefaultValue` decorator
- **`src/diagnostics/client-default-value-type-mismatch.md`** — Diagnostic documentation
- **`test/decorators/client-default-value.test.ts`** — Added test cases for type mismatch scenarios

## Example

```typespec
// ⚠️ Warning: Client default value type "string" does not match property type "int32"
@Azure.ClientGenerator.Core.Legacy.clientDefaultValue("10")
@query pageSize?: int32;

// ✅ Correct: types match
@Azure.ClientGenerator.Core.Legacy.clientDefaultValue(10)
@query pageSize?: int32;
```

Fixes #3934


packages/typespec-client-generator-core/src/lib.ts
```diff
@@ -596,6 +596,13 @@ export const $lib = createTypeSpecLibrary({
         default: "Auto-merging service client must be empty.",
       },
     },
+    "client-default-value-type-mismatch": {
+      ...doc("client-default-value-type-mismatch"),
+      severity: "warning",
```
tadelesh, line 601: Should this be an error?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1784,13 +1786,43 @@ export function getNextLinkVerb(context: TCGCContext, entity: Operation): "GET"
...
 export const $clientDefaultValue: ClientDefaultValueDecorator = (
   context: DecoratorContext,
   target: ModelProperty,
   value: string | boolean | Numeric,
   scope?: LanguageScopes,
 ) => {
   const actualValue = isNumeric(value) ? value.asNumber() : value;
+
+  // Validate that the value type matches the property type
+  const propertyType = target.type;
```
timotheeguerin: can't you just use the typekit to check if the value is of type?

timotheeguerin: (same logic we use in checker for default)

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1804,19 +1804,25 @@ export const $clientDefaultValue: ClientDefaultValueDecorator = (
 ) => {
   const actualValue = isNumeric(value) ? value.asNumber() : value;
 
-  // Validate that the value type matches the property type
-  const propertyType = target.type;
-  if (propertyType.kind === "Scalar") {
+  // Validate that the value type matches the property type (or the alternate type if present)
+  const alternateTypeDecorator = target.decorators.find(
```
timotheeguerin: do not try to look for other decorators like that, I think this might be a good use of post validation where you can then safely call getAlernateType

packages/typespec-client-generator-core/src/validations/decorators.ts
```diff
@@ -0,0 +1,47 @@
+import { ModelProperty, Type } from "@typespec/compiler";
+import { $ } from "@typespec/compiler/typekit";
+import { clientDefaultValueKey, getAlternateType } from "../decorators.js";
+import { TCGCContext } from "../interfaces.js";
+import { AllScopes } from "../internal-utils.js";
+import { reportDiagnostic } from "../lib.js";
+
+export function validateDecorators(context: TCGCContext) {
```
timotheeguerin: I meant we now have post validator for decorators so we don't need to use #onValidate

packages/typespec-client-generator-core/src/validations/decorators.ts
```diff
@@ -0,0 +1,47 @@
...
+    const typeToCheck: Type | undefined =
+      alternateType !== undefined && alternateType.kind !== "externalTypeInfo"
+        ? alternateType
+        : undefined;
+    const effectiveType = typeToCheck ?? property.type;
+
+    if (!tk.scalar.is(effectiveType)) continue;
+
+    const valueType = typeof value;
+    const isMatch =
```
timotheeguerin: can you not do `tk.value.isAssignableTo(value, effectiveType)`?

timotheeguerin (approved):

Issue #3066, opened 2025-07-28
Remove ad-hoc usage of OpenAPI decorators in ARM Specs


markcowl:
Older ARM specs use OpenApi decorators in lieu of using higher-level concepts (for example, using `@extension` decorators rather than using core types that automatically add these decorators, especially for custom LROs)

- Refactor specs to remove usage of OpenAPI decorators
- Add a linting rule removing OpenAPI decorator usage from ARM

PR #5109, opened 2026-07-30
Add no-openapi-client-extensions linter rule

markcowl:
## Summary

Adds a new `typespec-azure-core` linter rule, `no-openapi-client-extensions`, that flags use of the `@typespec/openapi` `@extension` decorator to emit **client-altering** `x-ms-*` / `x-nullable` OpenAPI extensions.

These extensions change how clients, SDKs, and the ARM platform interpret an API (long-running, pageable, secret, ARM resource, client name, etc.). When hand-written with the raw `@extension` decorator they only appear in the OpenAPI output, so every other emitter works from the semantic TypeSpec model and never sees them — producing an incorrect representation of the API. Each has a first-class TypeSpec construct that should be used instead, which the docs enumerate.

Extensions flagged: `x-ms-skip-url-encoding`, `x-ms-enum`, `x-ms-parameter-grouping`, `x-ms-parameter-location`, `x-ms-client-name`, `x-ms-discriminator-value`, `x-ms-client-flatten`, `x-ms-parameterized-host`, `x-ms-pageable`, `x-ms-long-running-operation`, `x-ms-long-running-operation-options`, `x-nullable`, `x-ms-internal`, `x-ms-azure-resource`, `x-ms-arm-id-details`, `x-ms-secret`.

Motivated by the discussion in Azure/typespec-azure#3066 (bar use of client-altering extensions).

## Changes

- Rule implementation, tests, and docs in `packages/typespec-azure-core`.
- Registered in `linter.ts` and in **both** the `data-plane` and `resource-manager` rulesets.
- Docs include **Impact**, **LintDiff Equivalent** (correlates to the LintDiff rules that require/validate these extensions, e.g. R2007/R2005/R2064 for LRO, R2029/R2060/R4012 for pageable, R2018/R4005 for enum, R2012/R2013 for client-name, R2001 for client-flatten, R2019/R2062 for azure-resource, XMSSecretInResponse for secret), examples, and **Suppression** sections.
- Changeset (feature) for `@azure-tools/typespec-azure-core` and `@azure-tools/typespec-azure-rulesets`.

## Validation

- `pnpm build`, targeted rule tests (12, all passing), and rulesets `validate-rules-defined` tests pass.
- `pnpm format`, `lint`, and `cspell` clean.
- `int:azure-specs` label to be applied to run the External Integration check against all Azure specs.


packages/typespec-azure-core/src/rules/no-openapi-client-extensions.md
timotheeguerin, line 1: how is that diffferent from the existing `no-openapi` rule?

markcowl: This is specifically about extension decorators that impact the API, and ignores other decorators.  I want to specifically find out which specs have these violations, regardless of suppressions.  Before merging, I think we should either reconcile these with special emphasis on extension decorators, or we should exclude extension decorators from the other check.

markcowl: Updated to fix this

markcowl:
PR for spec suppressions: https://github.com/Azure/azure-rest-api-specs/pull/45323

packages/typespec-azure-core/src/rules/no-openapi-client-extensions.md
```diff
@@ -0,0 +1,129 @@
+Azure specs should not use the `@typespec/openapi` `@extension` decorator to emit client-altering `x-ms-*` (and `x-nullable`) OpenAPI extensions.
```
catalinaperalta, line 5: I think this is missing the title section: 

```suggestion
---
title: "no-openapi-client-extensions"
---

Azure specs should not use the `@typespec/openapi` `@extension` decorator to emit client-altering `x-ms-*` (and `x-nullable`) OpenAPI extensions.
```

catalinaperalta (approved):
just a small nit in the docs

PR #5112, opened 2026-07-30
feat(typespec-go): emit apiview-properties.json with cross-language definition IDs

### LINKED ISSUE #4931: Output cross-language definition IDs
> Migrated from https://github.com/Azure/autorest.go/issues/1548

---

See [this gist](https://gist.github.com/iscai-msft/600f6b6042753c3f23035a37ebf2a5ae) for details. Related: Azure/azure-sdk-tools#4282

Place in testdata folder to avoid this cluttering the top level folder and to avoid it ending up in teh package

tadelesh:
Fixes #4931

## What

The Go emitter now writes `testdata/apiview-properties.json` alongside `testdata/_metadata.json`. It maps the Go APIView parser's line IDs for generated models, enums, enum values, clients, client factories, constructors, and methods to their TCGC `crossLanguageDefinitionId` values.

## Line IDs

Keys match [`src/go/cmd/token_makers.go`](https://github.com/Azure/azure-sdk-tools/blob/main/src/go/cmd/token_makers.go):

| Symbol | Line ID |
| --- | --- |
| Struct / enum / enum value / client | `<rel>.<Name>` |
| Package-level function | `<rel>-<FunctionName>` |
| Method | `<rel>-(<receiver> <receiverType>) <MethodName>` |

`<rel>` is the package path relative to the module root, with a `/vN` module suffix removed. For `containing-module`, the emitter uses the nearest `go.mod` to compute the complete relative path. If the module root is unavailable or does not contain the output path, it falls back to the output directory name.

Method names use `fixUpMethodName`, so LRO and pageable methods match their generated `Begin<Name>` and `New<Name>Pager` names. Only exported symbols are included.

## Implementation

- Threads `crossLanguageDefinitionId` from TCGC through the Go code model and carries `crossLanguageVersion` on `Info`.
- Walks the generated package tree and emits a deterministic, sorted mapping, including ARM `ClientFactory`, `NewClientFactory`, and every `New<Client>` accessor.
- Falls back to the output directory name when a module-relative path cannot be computed, preserving the previous behavior without blocking generation.

## Notes

The APIView side is tracked by [Azure/azure-sdk-tools#4282](https://github.com/Azure/azure-sdk-tools/issues/4282). This PR supplies the emitter metadata it will consume.

## Validation

- 52/52 unit tests pass, including module-root and multi-level `containing-module` paths
- TypeScript build, oxlint, Prettier, and cspell clean
- Real `containingmod/v2` fixture generates APIView IDs prefixed with `containingmod/subpkg`


packages/typespec-go/test/unittest/scenarios/apiview-properties.md
```diff
@@ -0,0 +1,124 @@
...
+}
+```
+
+```yaml
+containing-module: github.com/contoso/module
+emitter-output-dir: "{output-dir}/widgets"
+```
+
+```json apiview-properties
+// (file was not generated)
```
tadelesh: Why this file is not generated?

packages/typespec-go/src/codegen/core/apiview-properties.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  return JSON.stringify(properties, null, 2) + "\n";
+}
+
+/**
+ * recursively collects the line ID to cross-language definition ID mappings for a package.
+ *
+ * @param pkg the package to collect
+ * @param relName the package name relative to the module root (e.g. azblob/blob)
```
tadelesh, line 86: `azblobl` needs add to the cspell ignore list.

packages/typespec-go/src/codegen/core/apiview-properties.ts
```diff
@@ -0,0 +1,123 @@
...
+ * so that APIView can link the Go API surface to the other languages.
+ *
+ * @param codeModel the code model for which to generate the mapping
+ * @returns the JSON content or the empty string when there's nothing to emit
+ */
+export function generateApiViewProperties(codeModel: go.CodeModel): string {
+  if (codeModel.root.kind !== "module") {
+    // the module-relative package path is only knowable when emitting a module
+    return "";
+  }
```
tadelesh: `ContainingModule` should also export this apiview metadata.

packages/typespec-go/test/unittest/scenarios/apiview-properties.md
tadelesh: Fixed in 4ec832ad - it is now generated for containing modules too. The scenario has been updated to assert the emitted file.

packages/typespec-go/src/codegen/core/apiview-properties.ts
tadelesh: Added `azblob` to the `words` list in `cspell.yaml` (under the Go emitter section).

packages/typespec-go/src/codegen/core/apiview-properties.ts
tadelesh: Done. Removed the module-only guard and added a `relPackageName()` helper that resolves the module-relative package path for `ContainingModule` roots (stripping the `/vN` major-version suffix), matching the Go APIView parser's `relName`. e.g. `github.com/contoso/widget/v2` + package `subpkg` -> `widget/subpkg`.

tadelesh:
Fixed containing-module LineID qualification in c680046e. The emitter now derives the complete package path from `emitter-output-dir` relative to the nearest `go.mod`, matching APIView `Pkg.Name()` for module-root, one-level, and multi-level packages. Added root and nested scenario coverage.

tadelesh:
Adjusted in d4c3bc52: removed the new containing-module validation. The full path is used when the nearest `go.mod` establishes containment; otherwise generation continues and falls back to the output directory name. Added fallback scenario coverage.

tadelesh:
Refactored in c2b648aa. `Adapter` is restored to its original API and construction logic. `$onEmit` now applies the module-relative path after code model creation; the scenario runner mirrors only that orchestration step. The fallback remains in `apiview-properties.ts`.

tadelesh:
Added the missing ARM factory symbols in 5dea2139: `ClientFactory` and `NewClientFactory` map to the package cross-language ID, while every `(c *ClientFactory) New<Client>` method maps to that client's cross-language ID. The ARM scenario now has two clients to verify all accessors are included.

packages/typespec-go/src/codegen/core/apiview-properties.ts
```diff
@@ -0,0 +1,157 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from "path";
+import * as go from "../../codemodel/index.js";
+import { fixUpMethodName } from "./operations.js";
+
+/** the shape of the apiview-properties.json file */
+interface ApiViewProperties {
+  CrossLanguagePackageId?: string;
```
jhendrixMSFT: It's not clear why the `?` is necessary.  Can this really be empty?  Question also applies to `CrossLanguageVersion`.

packages/typespec-go/src/codemodel/module.ts
```diff
@@ -24,6 +24,12 @@ export interface ContainingModule {
    */
   identity: string;
 
+  /**
+   * the directory containing the emitted package, relative to the module root.
+   * empty when emitting into the module root.
+   */
+  relativePackagePath?: string;
```
jhendrixMSFT: It's not clear why this is required.

tadelesh: Agreed that this filesystem-derived APIView detail does not belong on the generic `ContainingModule` code model. Removed the property in 78b503d3; emitter orchestration now computes the path and passes it directly to APIView generation, preserving the existing fallback when it cannot be calculated.

packages/typespec-go/src/codegen/core/apiview-properties.ts
tadelesh: Good catch. TCGC declares and populates both values as required strings. Updated `ApiViewProperties` and the Go code-model `Info` accordingly, and now pass both through the `Info` constructor in 78b503d3.

PR #5119, opened 2026-07-30
Add `decorator-requires-scope` validation for @convenientAPI and @clientOption

### LINKED ISSUE #3751: [tcgc] add linter warning if `@convenientApi` decorator isn't scoped to java or sharp

iscai-msft:
## Summary

Adds a new `onValidate`-phase validation (`decorator-requires-scope`) that warns when TCGC decorators that require a language scope are used without one or with an invalid scope.

### Decorators checked:
- **`@convenientAPI`** — must be scoped to `"java"` and/or `"csharp"` (the only emitters that use this information)
- **`@clientOption`** — must be scoped to any specific language

This subsumes the previous inline `client-option-requires-scope` diagnostic with a centralized validation approach that runs during the `$onValidate` phase.

### Changes:
- New `src/validations/decorators.ts` with `validateDecoratorScopes()` function
- New `decorator-requires-scope` paramMessage diagnostic in `lib.ts`
- Removed old inline scope check from `$clientOption` decorator
- Updated all tests to use proper `"java"` scope for `@convenientAPI`
- Updated all references from `client-option-requires-scope` to `decorator-requires-scope`

Fixes https://github.com/Azure/typespec-azure/issues/3751

Related specs prs:
https://github.com/Azure/azure-rest-api-specs-pr/pull/29828
https://github.com/Azure/azure-rest-api-specs/pull/45196


packages/typespec-client-generator-core/src/diagnostics/decorator-requires-scope.md
```diff
@@ -0,0 +1,63 @@
...
+op yetAnotherOperation(): void;
+
+#suppress "@azure-tools/typespec-client-generator-core/client-option" "temporary workaround"
+@clientOption("enableFeatureFoo", true, "python")
+model Widget {}
+```
+
+## Suppression
+
+Suppress this warning only if the decorator is intentionally applied without a language scope:
```
tadelesh:
```suggestion
Suppress this warning only if the decorator is intentionally shared by all emitters rather than scoped to one language.
```

tadelesh (commented):
Seems a big breaking. Could you please check if any spec or spector test need to apply this breaking?

packages/typespec-client-generator-core/src/diagnostics/decorator-requires-scope.md
iscai-msft: i don't understand this suggested change, i'm trying to move away from "client option" and just make this warning against any decorator that we want scoped being applied without a language scope

iscai-msft:
> Seems a big breaking. Could you please check if any spec or spector test need to apply this breaking?

It's not very breaking, went through the specs and these are the only changes
https://github.com/Azure/azure-rest-api-specs/pull/45196
https://github.com/Azure/azure-rest-api-specs-pr/pull/29828

packages/typespec-client-generator-core/src/diagnostics/decorator-requires-scope.md
tadelesh: Oh, sorry, edited the comment.

tadelesh (approved):
LGTM. Please help to fix the ci before merge.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -404,13 +406,60 @@ export function listOperationsInClient(
...
+            },
+            target: entity,
+          }),
+        );
+      } else {
+        const parsedScopes = scope
+          .split(",")
+          .map((s) => s.trim().toLowerCase());
+        const hasValidScope = parsedScopes.some((s) =>
+          VALID_SCOPES.some((allowed) => s.includes(allowed)),
```
weidongxu-microsoft: **Scope matching is too permissive.** `includes` accepts scopes such as `"javascript"`, `"notjava"`, and `"java, python"`. The last case applies the decorator to Python too, but no diagnostic is emitted because one token contains `java`. Negation patterns such as `"!(python, csharp)"` are also misclassified. Please parse the scope with the existing scope parser and require every effective positive scope to be exactly `java` or `csharp`.

packages/typespec-client-generator-core/src/context.ts
```diff
@@ -190,6 +191,37 @@ export async function createSdkContext<
     context.options["generate-protocol-methods"] ?? tcgcContext.generateProtocolMethods;
   const generateConvenienceMethods =
     context.options["generate-convenience-methods"] ?? tcgcContext.generateConvenienceMethods;
+
+  // Warn if non-java/csharp emitter sets convenience/protocol options
+  const resolvedEmitterName = (emitterName ?? context.options["emitter-name"] ?? "").toLowerCase();
+  const isJavaOrCsharp = ["java", "csharp"].some((lang) => resolvedEmitterName.includes(lang));
```
weidongxu-microsoft: **JavaScript is classified as Java here.** The real emitter name `@azure-tools/typespec-javascript` contains the substring `java`, so `isJavaOrCsharp` becomes true and the new `unnecessary-emitter-option` warning is skipped for JavaScript. Please compare the parsed emitter language exactly rather than searching the package name by substring.

packages/typespec-client-generator-core/src/diagnostics/decorator-requires-scope.md
```diff
@@ -0,0 +1,63 @@
+This diagnostic is issued when a TCGC decorator that requires a language scope is used without one, or with an invalid scope.
+
+## Impact
+
+- **Area:** Language-scoped decorator behavior. The decorator may apply globally or to unintended emitters without a proper scope.
+- **Decorators checked:**
+  - `@convenientAPI(true)` — must be scoped to `"java"` and/or `"csharp"`. `@convenientAPI(false)` is always allowed without scope since opting out of convenience methods is safe for any language.
```
weidongxu-microsoft: **The documented exception does not exist in the implementation.** `validateJavaCsharpScope` warns whenever `scope` is undefined, regardless of the decorator value, and the added test explicitly expects a warning for `@convenientAPI(false)`. This page also repeats the invalid unscoped example below. Please either implement the exception or update the documentation to match the chosen behavior.

weidongxu-microsoft (changes requested):
The new validation has two substring-matching gaps that let JavaScript bypass the intended checks, and the diagnostic documentation contradicts the implemented behavior.

packages/typespec-client-generator-core/src/diagnostics/decorator-requires-scope.md
iscai-msft: oh thanks! that was from a previous implementation, i didn't clean it up well enough

packages/typespec-client-generator-core/src/context.ts
iscai-msft: great catch, thanks so much! i've added tests for this

weidongxu-microsoft (approved):

PR #5120, opened 2026-07-30
feat(examples): add examples-migrate tool

timotheeguerin:
Part of the Unified Examples Format epic (Azure/typespec-azure#4831), tracking Azure/typespec-azure#4833.

## Stack
- Azure/typespec-azure#4908 — `examples.yaml` schema + `examples-validate` (base)
- **This PR** → `examples-migrate` (base: `examples-file-format`)
- `examples-resolve` (stacked on top of this)

> Stacked PR — review/merge after the PR below it. Targets `examples-file-format`, not `main`.

## What
Adds the `examples-migrate` tool + `examples-migrate` CLI to `@azure-tools/typespec-azure-examples`, which converts versioned Swagger `x-ms-examples` into the unified `examples.yaml` format.

- Operation key derived by splitting `operationId` on the first `_` and lowercasing the method segment (e.g. `CaCertificates_Get` → `CaCertificates.get`).
- De-duplicates example variants across API versions into `since`-anchored lineages.
- Emits `examples.yaml` (bare integer status keys, quoted `since`, single-file or per-interface split).

Tests added (vitest); `tsc`, `oxlint`, `prettier` clean.

## Update — preserve legacy file names and keys

Migration now records the original `x-ms-examples` file name and key so the round-trip can reproduce the exact legacy files. It emits the **minimal** deviation from convention: nothing when the key equals the `operationId` and the file is `<OperationId>.json`, just `legacyFilename` when the key is recoverable from the file name, and an explicit `title` (plus `legacyFilename` when needed) otherwise. The convention itself comes from the shared naming helper in Azure/typespec-azure#4908.

### DIFF
--- a/.chronus/changes/add-examples-migrate-tool-2026-07-30-12-30-00.md
+++ b/.chronus/changes/add-examples-migrate-tool-2026-07-30-12-30-00.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-examples"
+---
+
+Add the `examples-migrate` tool that converts existing `x-ms-examples` JSON into the unified `examples.yaml` format (crawls versioned Swagger, normalizes `{api-version}`, and dedups across versions into `since` lineages).
--- a/packages/typespec-azure-examples/README.md
+++ b/packages/typespec-azure-examples/README.md
@@ -1,7 +1,7 @@
 # @azure-tools/typespec-azure-examples
 
-Tooling for the Azure **unified examples format** (`examples.yaml`): the published JSON Schema
-and the `examples-validate` CLI.
+Tooling for the Azure **unified examples format** (`examples.yaml`): the published JSON Schema,
+the `examples-validate` CLI, and the `examples-migrate` CLI.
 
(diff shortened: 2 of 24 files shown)


timotheeguerin:
Addressed review feedback:
- **api-version normalization** no longer substring-matches. It was rewriting any string that merely contained a service-version date — e.g. an ISO timestamp `2023-01-01T09:00:00Z` became `{api-version}T09:00:00Z` and was then re-materialized to the wrong date. Added a trailing token boundary `(?![T\\d])` so only whole api-version tokens match, and each example is now normalized against its own version rather than the full version set. Added a regression test.

timotheeguerin:
Added handling for the "example removed in a later version" case (raised in the stack review). The format intentionally has no `until`/removal marker for now, so migration instead emits an `example-removed-before-latest` **warning** whenever an example lineage stops appearing before the service's latest version — since such an example would otherwise be re-materialized for versions it never shipped in. Documented as a known limitation in the package README. Covered by new e2e tests (warns on drop; no false positives when every example carries forward, including the real Microsoft.Contoso round-trip).

Issue #4834, opened 2026-07-06
examples-resolve tool


timotheeguerin:
Part of #4831.

RFC: https://github.com/Azure/azure-rest-api-specs/blob/rfc/unified-examples-format/documentation/rfc/unified-examples-format.md

Build **`examples-resolve`**: given a target API version, resolve the applicable example for each operation.

### Scope
- Apply `since`: within a lineage pick the entry with the greatest `since` that is `<=` the target version.
- Use the **single linear version order from `service.yaml`** (see service.yaml epic #4825); no preview/stable special-casing.
- Substitute the `{api-version}` placeholder with the target version on materialization.

Depends on #4832 and on `service.yaml` version ordering (#4826).

PR #5121, opened 2026-07-30
feat(examples): add examples-resolve tool

timotheeguerin:
Part of the Unified Examples Format epic (Azure/typespec-azure#4831), tracking Azure/typespec-azure#4834.

## Stack
- Azure/typespec-azure#4908 — `examples.yaml` schema + `examples-validate` (base)
- Azure/typespec-azure#5120 — `examples-migrate`
- **This PR** → `examples-resolve` (base: `examples-migrate`)

- Azure/typespec-azure#5465 — `typespec-autorest` emitter integration (stacked on top of this)

> Review/merge after the two PRs below it. Targets `examples-migrate`, not `main`.

## What
Adds the `examples-resolve` tool + `examples-resolve` CLI to `@azure-tools/typespec-azure-examples`, which resolves the applicable example for each operation at a target API version.

- Per lineage, selects the entry with the greatest `since <= target` (base entry applies from the earliest version), using the linear `service.yaml` version order.
- Materializes the `{api-version}` placeholder to the target version.
- Errors on an unknown target version; omits lineages with no applicable entry.

Tests added (vitest); `tsc`, `oxlint`, `prettier` clean.

## Update — `legacyFilename`

Resolution now threads the `legacyFilename` recorded in `examples.yaml` through to each resolved example, so the `typespec-autorest` emitter (Azure/typespec-azure#5465) can reproduce the original `x-ms-examples` file names during rollout.

### DIFF
--- a/.chronus/changes/add-examples-resolve-tool-2026-07-30-14-05-00.md
+++ b/.chronus/changes/add-examples-resolve-tool-2026-07-30-14-05-00.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-examples"
+---
+
+Add the `examples-resolve` tool that resolves the applicable example for each operation at a target API version (greatest `since <= target` per lineage, using the `service.yaml` order) and materializes the `{api-version}` placeholder.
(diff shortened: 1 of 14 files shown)


timotheeguerin:
Addressed review feedback:
- The `legacy-expand` CLI rebuilt each example with an explicit field whitelist that dropped `legacyFilename`, so the resolved field never reached the output — defeating its purpose of reconstructing the original per-version files. It's now included in the emitted object. Added an e2e assertion.

PR #5122, opened 2026-07-30
Update @typespec/http-client-python to 0.35.1

l0lawrence:
Bump the `@typespec/http-client-python` emitter dependency to **0.35.1** (released from microsoft/typespec).

## Changes
- `pnpm-workspace.yaml`: catalog range `>=0.35.0 <1.0.0` → `>=0.35.1 <1.0.0`
- `pnpm-lock.yaml`: resolved version `0.35.0` → `0.35.1` (with updated sha512 integrity)

## Released changes in 0.35.1 (from microsoft/typespec)
- Use wire names in TypedDict docstrings
- Fix `@api_version_validation` decorator to read the correct client config attribute for the API version
- Fix generated request builders serializing a `None` `content-type` header for an optional body whose content-type is required/constant

Follows the same shape as prior bumps (e.g. #5003).

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -223,8 +223,8 @@ catalogs:
       specifier: ^17.0.35
       version: 17.0.35
     '@typespec/http-client-python':
-      specifier: '>=0.35.0 <1.0.0'
-      version: 0.35.0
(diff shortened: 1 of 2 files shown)


l0lawrence:
ah I guess this will include the .1 anyway -- closing

PR #5125, opened 2026-07-31
Set typespec-metadata test timeout to 30s

jhendrixMSFT:
The inherited 5s seems to be too short in some cases, leading to spurious timeouts in CI.

### DIFF
--- a/packages/typespec-metadata/vitest.config.ts
+++ b/packages/typespec-metadata/vitest.config.ts
@@ -1,4 +1,11 @@
 import { defineConfig, mergeConfig } from "vitest/config";
 import { defaultTypeSpecVitestConfig } from "../../core/vitest.config";
 
-export default mergeConfig(defaultTypeSpecVitestConfig, defineConfig({}));
+export default mergeConfig(
+  defaultTypeSpecVitestConfig,
+  defineConfig({
+    test: {
+      testTimeout: 30000,
+    },
+  }),
+);
--- a/.chronus/changes/metadata-test-timeout-2026-6-31-10-49-19.md
+++ b/.chronus/changes/metadata-test-timeout-2026-6-31-10-49-19.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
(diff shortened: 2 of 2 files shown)


packages/typespec-metadata/vitest.config.ts
timotheeguerin, line 1: i have never seen that, weird that typespec-metadata has anything that is not instant, do you have a ci or which test timesout locally?

jhendrixMSFT: I saw this today in https://github.com/Azure/typespec-azure/pull/5124 and reran the job.  Also saw in at least one of my Go emitter PRs (I think https://github.com/Azure/typespec-azure/pull/5118 but not sure now).

jhendrixMSFT: It was always the first test `"single versioned service emits latest version as apiVersion"`.

jhendrixMSFT: This is what Copilot told me.

Why it's transient and hits line 26 specifically: That `it` on `api-version.test.ts:26` is the first test to actually invoke the tester. `createTester(...)` in `tester.ts:4` is lazy — the whole library graph (@typespec/http, rest, versioning, azure-core, typespec-client-generator-core, and typespec-metadata itself) isn't loaded and compiled until the first `compileAndDiagnose` call. So the very first test in the run pays the entire cold-start cost: module resolution, JITing, and the initial TypeSpec program compile.

That one-time cost lands inside the default vitest testTimeout of 5000ms (you're inheriting it from core/vitest.config.ts — nothing overrides it). On a warm dev machine it's well under 5s, but on a loaded/cold CI runner (especially with `isolate: false`, where the first worker also does all the initial ESM loading) it occasionally tips past 5s. Whichever test happens to run first eats this — right now that's line 26.

Things that make it worse on CI: shared/noisy runners, cold filesystem cache, node 22's module loading, and the fact that all six libraries resolve on that first call.

timotheeguerin (approved):

PR #5142, opened 2026-08-03
[typespec-go] Support paging with parameterized nextLink

### LINKED ISSUE #4993: Support paging with parameterized nextlink
spector test: https://github.com/Azure/typespec-azure/tree/main/packages/azure-http-specs/specs/azure/core/page

tadelesh:
Fixes #4993

## Summary

Adds support for `Azure.Core.Legacy.parameterizedNextLink` in the Go emitter. Parameterized next links identify request query parameters that are not guaranteed to be included in a service-provided `nextLink` and therefore must be copied from the original method arguments into each subsequent page request.

## Changes

- Adds `reinjectedParams` to the Go `PageableStrategyNextLink` code model.
- Reads TCGC's `nextLinkReInjectedParametersSegments` metadata when adapting pageable methods.
- Resolves each reinjection segment to its generated Go method parameter.
  - Query scalar and collection parameters are retained for reinjection.
  - Header parameters are skipped because they are handled by the normal request pipeline.
  - Missing or unexpected parameter metadata is reported as an internal adapter error instead of producing incomplete paging code.
- Reuses the normal Go query-parameter serialization and optionality logic when generating reinjection code, including scalar formatting, collection formats, client-side defaults, and options-bag nil checks.
- Enables the `azure/core/page` parameterized nextLink scenario for Go and restores its two-page Spector test.
- Adds a generated-code snapshot test for the emitted pager implementation.

## Reinjection algorithm

For the first page, the pager uses the existing request builder, so all original query parameters are serialized normally.

For every non-empty service-provided `nextLink`, generated code:

1. Parses the nextLink with `net/url`.
2. Parses its existing raw query after escaping raw semicolons as `%3B`, matching azcore's handling and avoiding Go's `url.ParseQuery` rejection of unescaped semicolons.
3. Keeps all query parameters already supplied by the service, for example `select=name`.
4. For each parameter listed by TCGC as requiring reinjection:
   - checks whether an optional value is present;
   - formats the value using the same helper used by initial request generation;
   - uses `Set` for scalar/single-value formats, replacing an existing value with the original method argument;
   - uses `Add` for multi-value collection formats.
5. Re-encodes the complete query and replaces `+` with `%20`, consistent with existing Go emitter URI encoding.
6. Passes the rebuilt nextLink to `runtime.FetcherForNextLink`.

Example:

```text
Original request:  /widgets?includePending=true&select=name
Returned nextLink: /widgets/next?select=name
Next page request: /widgets/next?includePending=true&select=name
```

## Validation

- `pnpm build`
- `pnpm test`
- `pnpm lint`
- `pnpm cspell`
- Generated-code snapshot for parameterized nextLink paging
- Spector `TestPageClient_NewWithParameterizedNextLinkPager`


packages/typespec-go/test/unittest/scenarios/parameterized-next-link.md
```diff
@@ -0,0 +1,143 @@
...
+	return runtime.NewPager(runtime.PagingHandler[WidgetsClientListResponse]{
+		More: func(page WidgetsClientListResponse) bool {
+			return page.NextLink != nil && len(*page.NextLink) > 0
+		},
+		Fetcher: func(ctx context.Context, page *WidgetsClientListResponse) (WidgetsClientListResponse, error) {
+			nextLink := ""
+			if page != nil {
+				nextLink = *page.NextLink
+			}
+			if len(nextLink) > 0 {
```
jhendrixMSFT, line 92: I want to move the param reinjection into `azcore` and am working on it.

tadelesh:
Replaced by https://github.com/Azure/typespec-azure/pull/5198.

PR #5151, opened 2026-08-04
Fix Go path API version overrides

### LINKED ISSUE #4940: API version in path can't be changed by ClientOptions.APIVersion in some cases
> Migrated from https://github.com/Azure/autorest.go/issues/1743

---

For cases where the API version is a path parameter (_not_ a URI parameter), setting `ClientOptions.APIVersion` will have no effect. This is because the path segement in the `*CreateRequest` method is replaced with the literal value.

tadelesh:
## Summary

- make `ClientOptions.APIVersion` override a defaulted API version embedded in an operation path
- resolve the generated API-version literal and optional override once in the client constructor
- reject an empty resolved path API version before constructing the generated client
- derive path-override support from the client parameters already present in the code model, without adding adapter or code-model state
- keep existing client-parameter collection, subclient behavior, and query/header API-version policy behavior unchanged
- support ARM clients by normalizing nil options before reading `options.APIVersion`
- restore the Azure API-version path Spector test that was previously skipped

## Override algorithm

1. For a directly constructable client with a defaulted path API-version parameter, the constructor starts with the generated API-version literal.
2. A non-empty `ClientOptions.APIVersion` replaces that default in the constructor.
3. The constructor rejects an empty resolved value, stores the result on the client, and each request builder uses the stored value directly when substituting the path segment.
4. Query and header API versions continue to use azcore's existing API-version policy behavior.

This keeps generated literals as the source of default versions and avoids relying on azcore's path API-version policy, which intentionally does not rewrite URL paths.

## Tests

- generated-code scenarios cover regular and ARM clients, including nil ARM options and constructor-time resolution
- the default path test runs against the Spector mock server
- the override test uses a capture transport that returns HTTP 200 and asserts the outgoing request path, so it can verify an arbitrary API version without requiring mock-server support for that version
- regenerated `versionedgroup` retains its existing query API-version policy and its handwritten test remains unchanged
- typespec-go unit tests, repository formatting, and lint pass

Closes #4940


packages/typespec-go/test/unittest/scenarios/arm-path-api-version-override.md
```diff
@@ -0,0 +1,99 @@
...
+	if !runtime.HasStatusCode(httpResp, http.StatusNoContent) {
+		return TestClientGetResponse{}, runtime.NewResponseError(httpResp)
+	}
+	return TestClientGetResponse{}, nil
+}
+
+// getCreateRequest creates the Get request.
+func (client *TestClient) getCreateRequest(ctx context.Context, _ *TestClientGetOptions) (*policy.Request, error) {
+	urlPath := "/api-version/{apiVersion}"
+	apiVersion := version20220101
```
jhendrixMSFT: Since this is a client parameter, we should resolve this in the client constructor instead of per method.  Also, we will need to check for the empty string and return an error in the case.

tadelesh: Updated in c409faf5e. The constructor now resolves the generated literal with a non-empty ClientOptions.APIVersion override, rejects an empty resolved value, and stores it on the client. Request builders use client.apiVersion directly without per-method fallback logic.

jhendrixMSFT (approved):

jhendrixMSFT:
Test failure uncovered that we've been generating incorrect code for `server/versions/versioned`.  I'm working on a fix.

Also note that the Go mock API tests aren't required to be passing in order to merge.  We need to fix that.

jhendrixMSFT:
Fixing the test is much more involved.  I also don't think it's a real scenario (we should remove this from core).  I opened https://github.com/Azure/typespec-azure/issues/5226 to track it.

Also note that path param API version for ARM is broken.  I opened https://github.com/Azure/azure-sdk-for-go/issues/27385 to track it.

tadelesh:
> Test failure uncovered that we've been generating incorrect code for `server/versions/versioned`. I'm working on a fix.
> 
> Also note that the Go mock API tests aren't required to be passing in order to merge. We need to fix that.

Tim disagree to set it to be required. Currently, any up stream changes will run all emitter's check. Could you talk with him about your requirement?

packages/typespec-go/test/unittest/scenarios/arm-path-api-version-override.md
```diff
@@ -0,0 +1,96 @@
...
+	if !runtime.HasStatusCode(httpResp, http.StatusNoContent) {
+		return TestClientGetResponse{}, runtime.NewResponseError(httpResp)
+	}
+	return TestClientGetResponse{}, nil
+}
+
+// getCreateRequest creates the Get request.
+func (client *TestClient) getCreateRequest(ctx context.Context, _ *TestClientGetOptions) (*policy.Request, error) {
+	urlPath := "/api-version/{apiVersion}"
+	urlPath = strings.ReplaceAll(urlPath, "{apiVersion}", url.PathEscape(client.apiVersion))
```
tadelesh, line 89: I'm confused about this: https://github.com/Azure/azure-sdk-for-go/issues/27385. The path API version is set in the generated code. Do you mean the policy's path location set is useless? From the comment, it should be intentional to ignore the path location and do the set in the generated code.

jhendrixMSFT: The problem is that the [policy](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_api_version.go#L57-L78) will add a query param for the API version in this case.  This is because we can't set [`APIVersionOptions.Location`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_api_version.go#L17) to `APIVersionLocationPath` which tells the policy to skip setting anything.

PR #5152, opened 2026-08-04
[typespec-go] Support array encoding

### LINKED ISSUE #4927: [http-client-go] CSV encoding for model properties Implementation
> Migrated from https://github.com/Azure/autorest.go/issues/1771

---

_No description was provided in the original issue._

JiaqiZhang-Dev:
## Summary

- preserve TypeSpec array encoding metadata in the Go code model
- generate delimiter-aware JSON serialization and deserialization for string and string-backed enum arrays
- support comma, space, pipe, and newline delimiters
- add generated-output and Spector coverage for the supported encodings

Fixes #4927

## Testing

- generated-output scenario for pointer-element slices
- 12 Spector round-trip cases for string, fixed enum, and extensible enum arrays


packages/typespec-go/src/codemodel/type.ts
```diff
@@ -237,8 +237,15 @@ export interface ModelFieldAnnotations {
 
   /** unmarshal an empty string as nil. the default is false */
   unmarshalEmptyStringAsNil: boolean;
+
+  /** delimiter encoding used to serialize a string-like array as one string */
+  arrayEncoding?: ArrayEncoding;
```
jhendrixMSFT: We should model this as an explicit type given it's non-standard and has special behavior.  We'd create one of these during adaptation.
```ts
/** specialized slice type for arrays represented as delimited strings */
export interface SliceArray {
  kind: "sliceArray";

  /** the element type for this slice */
  elementType: String;

  /** the delimiter used to separate elements */
  delimiter: SliceArrayDelimiter;
}

/** the set of slice array delimiters */
export type SliceArrayDelimiter = "comma" | "space" | "pipe" | "newline";
```

jhendrixMSFT (changes requested):

packages/typespec-go/src/codemodel/type.ts
JiaqiZhang-Dev: Updated

packages/typespec-go/src/codemodel/type.ts
```diff
@@ -320,6 +321,23 @@ export interface Slice {
   elementTypeByValue: boolean;
 }
 
+/** specialized slice type for arrays represented as delimited strings */
+export interface SliceArray {
+  kind: "sliceArray";
+
+  /** the element type for this slice */
+  elementType: SliceElementType;
```
jhendrixMSFT: Is this accurate?  e.g. can we have an array of pipe-delimited models?

JiaqiZhang-Dev: introduce a new type `SliceArrayElementType`, which only support string or constant

jhendrixMSFT (approved):

PR #5154, opened 2026-08-04
[typespec-ts] Rename reserved-word operations using singularized group name

maorleger:
## Summary

Reserved-word operations (e.g. `delete`) that belong to an operation group are now renamed by suffixing the **singularized** operation group name, instead of emitting a `@fixme` doc comment on a `$`-guarded name.

For example, given an operation `delete` in the `Conversations` operation group:

```ts
// Before
export interface ConversationsOperations {
  /**
   *  @fixme delete is a reserved word that cannot be used as an operation name.
   *         Please add @clientName(...) to the operation to override the generated name.
   */
  delete: (conversationName: string, options?: ConversationsDeleteOptionalParams) => Promise<void>;
}

// After
export interface ConversationsOperations {
  deleteConversation: (conversationName: string, options?: ConversationsDeleteOptionalParams) => Promise<void>;
}
```

This mirrors the manual in-SDK customization done in [Azure/azure-sdk-for-js#39499](https://github.com/Azure/azure-sdk-for-js/pull/39499), applying it automatically in the emitter.

## Data plane only (mgmt is intentionally untouched)

For mgmt plane we already decided a bare `delete` method is acceptable and suppressed the `@fixme` a while back (see [autorest.typescript#3749](https://github.com/Azure/autorest.typescript/issues/3749)), and `delete` has been generated that way going back to HLC. Renaming it now would be a breaking change for shipped mgmt SDKs, so ARM keeps emitting plain `delete` with no rename and no `@fixme`.

## Backwards-compat escape hatch via `@clientName`

The rename itself is a breaking change for any data-plane library that already shipped with the guarded `delete` name. To give those libraries an out, an explicit `@clientName` on the operation is now treated as an intentional choice: we honor the requested name verbatim and skip both the rename and the `@fixme`.

```tsp
@route("/conversations")
interface Conversations {
  // Keep `delete` as the generated method name instead of `deleteConversation`.
  @delete
  @clientName("delete", "javascript")
  delete(@path conversationName: string): void;
}
```

In that case the public method keeps the reserved word (`delete`), while the generated API-layer function stays guarded (`$delete`) because a reserved word isn't a valid function binding in JS.

## Details

- The rename is applied consistently to the API-layer function (`deleteConversation`) and its `_deleteConversationSend` / `_deleteConversationDeserialize` helpers, as well as the public method exposed on clients and operation groups.
- The options parameter type (e.g. `ConversationsDeleteOptionalParams`) is intentionally left unchanged.
- Singularization uses `pluralize.singular`, which correctly handles cases like `KnowledgeBases` → `KnowledgeBase`, `Conversations` → `Conversation`, `Categories` → `Category`.
- Operations **without** an operation group (e.g. top-level operations) continue to fall back to the previous guarded name (`$delete`) and `@fixme` guidance, since there's no group name to disambiguate with — unless a `@clientName` override is present, in which case the escape hatch above applies. This is a very rare case.

## Validation

- Build, lint, and format: clean.
- `modular-unit`: 661/661.
- `test-next`: 254/254.


packages/typespec-ts/src/modular/helpers/naming-helpers.ts
```diff
@@ -36,10 +66,25 @@ export function getOperationName(
...
+ * operation group it belongs to. The innermost operation group name is singularized
+ * so that, for example, the `Conversations` group yields `Conversation`.
+ */
+function getReservedNameGroupSuffix(prefixes: string[]): string | undefined {
+  const groups = prefixes.filter((prefix) => prefix && prefix.length > 0);
+  const innermostGroup = groups[groups.length - 1];
+  if (!innermostGroup) {
+    return undefined;
+  }
+  return pluralize.singular(normalizeName(innermostGroup, NameType.Interface));
```
xirzec, line 85: ah, a handy dependency that everyone uses but hasn't been updated in 8 years. Probably this is fine.

xirzec (approved):
Seems reasonable to me, though please get @JialinHuang803 to take a look as well

JialinHuang803 (approved):
LGTM! Just curious, though it's a rare case, what will happen if both `delete` and `deleteConversation` exist in an operation group?

maorleger:
> LGTM! Just curious, though it's a rare case, what will happen if both `delete` and `deleteConversation` exist in an operation group?

Good Q! In that case it'll show up as an overloaded method. Not ideal, but still usable and like you said it's quite the edge case

PR #5166, opened 2026-08-05
Disable C# standard acronym rule

haiyuazhang:
Disable `csharp-use-standard-acronyms` in the client SDK ruleset until the C# emitter handles standard acronym casing.

This avoids requiring widespread suppressions and C# client-name overrides for common IP/DB-style names.

### DIFF
--- a/.chronus/changes/disable-csharp-standard-acronyms-2026-08-06.md
+++ b/.chronus/changes/disable-csharp-standard-acronyms-2026-08-06.md
@@ -1,5 +1,5 @@
 ---
-changeKind: fix
+changeKind: internal
 packages:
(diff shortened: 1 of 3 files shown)


timotheeguerin (approved):

.chronus/changes/disable-csharp-standard-acronyms-2026-08-06.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
```
timotheeguerin:
```suggestion
changeKind: internal
```

The change isn't released so we don't need to document

PR #5171, opened 2026-08-06
Revert "Split `arm-resource-operation` lint rule into three focused rules"

catalinaperalta:
Reverts #4808 and commit `a55c28f74db70c86f6477cb0204ef7b6de5b0240`.

This is a direct revert of the original squash commit.


timotheeguerin:
as talked offline I think the new suppressions are good so we shouldn't need to revert.

PR #5173, opened 2026-08-06
feat(tcgc): support overriding method responses

### LINKED ISSUE #3542: [tcgc] do we want to allow users to `@override` to a different response?
@dargilco got feedback from arch board that they have several different "delete" operations defined in TypeSpec, each returning a unique object when succeeded, they should return `None` when succeeded instead. To change the outputted sdk, we would have to use `@override` to modify the return type. We started `@override` with a strict ethos of one-to-one mapping from original to what we want to override (today we don't fail, we just don't update the response method in the outputted method.

What should we do here? I think since we've been loosening requirements for override methods, and this is the way to control the client signature, we should go ahead and permit this usage. Maybe add a linter.

iscai-msft:
fixes #3542


.chronus/changes/replace-client-response-2026-08-06-10-49-09.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
```
heaths: Should this be "feat" or something like your commit title?

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
heaths, line 1: Should add a test that returns:

```typespec
op do(): {
    @body body: bytes;
}
```

packages/typespec-client-generator-core/src/methods.ts
```diff
@@ -167,7 +167,7 @@ function getSdkPagingServiceMethod<TServiceOperation extends SdkServiceOperation
   const diagnostics = createDiagnosticCollector();
 
   const baseServiceMethod = diagnostics.pipe(
-    getSdkBasicServiceMethod<TServiceOperation>(context, operation, client),
+    getSdkBasicServiceMethod<TServiceOperation>(context, operation, client, false),
```
tadelesh: Why here we pass `false`? This flag looks quick hacky to me.

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      name: string;
+    }
+
+    model DeleteResult {
+      deleted: boolean;
+    }
+
+    @post op create(): Widget;
+
+    op customizedCreate(): DeleteResult;
```
tadelesh: Please pay attention to the usage and serialization info of new response model.

tadelesh: as well as the usage for original model.

packages/typespec-client-generator-core/lib/functions.tsp
```diff
@@ -130,6 +130,24 @@ extern fn reorderParameters(
...
+ * @returns A new operation with the response type replaced.
+ *
+ * @example Replace a response with void
+ * ```typespec
+ * alias DeleteResponse = replaceResponse(MyService.delete, void);
+ * @@override(MyService.delete, DeleteResponse);
+ * ```
+ */
+#suppress "experimental-feature" "replaceResponse uses extern fn which is experimental but provides essential response transformation functionality"
+extern fn replaceResponse(operation: Reflection.Operation, response: unknown): Reflection.Operation;
```
tadelesh, line 152: When we replace the response, how could we know how to map the method level response to the HTTP level response?

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    @error
+    model Error {
+      code: string;
+    }
+
+    model Widget {
+      name: string;
+    }
+
+    @post op create(): Widget | Error;
```
weidongxu-microsoft, line 23: Java does not support union of 2 unrelated models...

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    model DeleteResult {
+      deleted: boolean;
+    }
+
+    @post op create(): Widget;
+
+    op customizedCreate(): DeleteResult;
+    @@override(TestService.create, TestService.customizedCreate);
```
weidongxu-microsoft: Why is this allowed?

packages/typespec-java/emitter-tests/tsp/method-override.tsp
```diff
@@ -124,7 +124,7 @@ namespace Customization {
   ): void;
   op groupExcludeBodyCustomization(
     ...TspTest.MethodOverride.GroupExcludeBody,
-    options?: TspTest.MethodOverride.GroupQueryOptions,
+    @query options?: TspTest.MethodOverride.GroupQueryOptions,
```
weidongxu-microsoft: Emm, I am OK if this is only to unblock. But unsure why? What if we group query and header parameter in the model?

packages/typespec-client-generator-core/lib/functions.tsp
iscai-msft: we can't, except in instances where the user now just returns the raw bytes or void. since heath's use case is just returning bytes, should i make it more narrow and just add a function like `returnRawResponse`? I made it more wide, but in the cases besides raw bytes or void, sdk authors would probably have to do some handwritten conversion

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
iscai-msft: this is a union of a model with an error though, i think we have those elsewhere as well. i can change the test here if need be

tadelesh: This is a common case for separating normal response and error response. But it seems not related with current test.

packages/typespec-client-generator-core/lib/functions.tsp
iscai-msft: gentle bump @tadelesh what are your thoughts?

tadelesh: I prefer to make it a little bit narrow.

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
weidongxu-microsoft: Same in HEAD.

I don't think we should allow replace response to an unrelated model.

If service wrote a wrong REST API, ask them to fix.

```
    model Widget {
      name: string;
    }

    model DeleteResult {
      deleted: boolean;
    }

    @post op create(): Widget;

    op customizedCreate(): DeleteResult;
```

iscai-msft: Good point, and I agree. I've narrowed the API so this is no longer possible: instead of a single `replaceResponse(operation, response: unknown)` that accepted any type, there are now two purpose-built functions, `replaceResponseWithVoid` and `replaceResponseWithBytes`, which only allow replacing the response with `void` or `bytes` respectively. Replacing with an arbitrary unrelated model (like the `DeleteResult` example) is no longer expressible, so this case can't happen anymore. Updated the tests accordingly.

weidongxu-microsoft (commented):
Does override still able to do any replacement on return type? Do we error or warning on it?

There is many real case in client.tsp do this wrong.
E.g.
"return type is void" (wrongly and not possible to capture -- it is just wrong, dev is lazy, that's all)
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/databoxedge/resource-manager/Microsoft.DataBoxEdge/DataBoxEdge/client.tsp#L128-L151

"return type mismatch"
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/databoxedge/resource-manager/Microsoft.DataBoxEdge/DataBoxEdge/DataBoxEdgeDevice.tsp#L68-L76
->
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/databoxedge/resource-manager/Microsoft.DataBoxEdge/DataBoxEdge/client.tsp#L1638-L1656

There appear to be more. I didn't really verify this one by one. Just let agent find some case that potentially having problems if we enable return type replacement.

iscai-msft:
@weidongxu-microsoft i've added this verification https://github.com/Azure/typespec-azure/pull/5173/changes#diff-d276532a6e7d2da177119679a7dfdc92a59980b7bbb70000f99ad0e4d50f3183R270

basically we error if you're trying to switch it to a completely different response type, but we just give a warning if you switch it to `void` or `bytes`

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    model Widget {
+      name: string;
+    }
+
+    op create(): Widget;
+  `;
+
+  const customizationCode = `
+    #suppress "experimental-feature" "testing replaceResponseWithVoid"
```
timotheeguerin, line 141: this shouldn't report here no?

packages/typespec-client-generator-core/test/decorators/override.test.ts
```diff
@@ -55,6 +55,72 @@ it("basic", async () => {
   strictEqual(method.operation.bodyParam.correspondingMethodParams[0], paramsParam);
 });
 
+it("reports an error for an unrelated override response type", async () => {
+  const [, diagnostics] = await SimpleBaseTester.compileAndDiagnose(
```
timotheeguerin: use `.diagnose` if you only want diagnostics

iscai-msft: Updated this diagnostic-only test to use `SimpleBaseTester.diagnose(...)` instead of `compileAndDiagnose(...)`.

packages/typespec-client-generator-core/test/functions/replace-response.test.ts
iscai-msft: Good catch. That `experimental-feature` suppression was unnecessary for this test and has been removed.

timotheeguerin (approved):

packages/typespec-client-generator-core/src/diagnostics/override-response-mismatch.md
```diff
@@ -0,0 +1 @@
+This diagnostic is issued when an override operation declares a response type that is not compatible with the original operation. Use `replaceResponseWithVoid` or `replaceResponseWithBytes` for an intentional response replacement.
```
tadelesh, line 1: Could you make sure these docs follow the same pattern with previous diagnostic doc?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -853,6 +853,31 @@ export const $override = (
...
+  const returnTypesMatch =
+    original.returnType === override.returnType ||
+    $(context.program).type.isAssignableTo(override.returnType, original.returnType) ||
+    (original.returnType.kind === "Model" &&
+      override.returnType.kind === "Model" &&
+      original.returnType.name === override.returnType.name &&
+      original.returnType.namespace !== undefined &&
+      override.returnType.namespace !== undefined &&
+      getNamespaceFullName(original.returnType.namespace) ===
+        getNamespaceFullName(override.returnType.namespace));
```
tadelesh, line 866: Is there any blocker that we just check the type directly for model?

weidongxu-microsoft:
> Weidong Xu (@weidongxu-microsoft) i've added this verification https://github.com/Azure/typespec-azure/pull/5173/changes#diff-d276532a6e7d2da177119679a7dfdc92a59980b7bbb70000f99ad0e4d50f3183R270
> 
> basically we error if you're trying to switch it to a completely different response type, but we just give a warning if you switch it to `void` or `bytes`

Java emitter will not support `@override` that changes response, unless all incorrect TypeSpec is solved.

Here these errors are not resolved
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/databoxedge/resource-manager/Microsoft.DataBoxEdge/DataBoxEdge/client.tsp#L128-L151
It seems not even trigger the error/warning? Or worse, it only trigger error when SDK emitter runs, but not when TCGC validate?

<-- OK, I see. Currently these "wrong" override will not have effect and will not report error/warning. Then it is OK.

PR #5176, opened 2026-08-07
Move legacy API version override to Azure Core

markcowl:
## Summary
- add \Azure.Core.Legacy.overrideApiVersion\ for namespace and interface API-version wire overrides
- expose direct and inherited Azure Core accessors for namespaces, interfaces, and operations
- update TCGC, AutoRest, TypeScript, and Go to consume the Azure Core override while preserving ARM feature-file version precedence

## Validation
- affected package builds
- Azure Core, TCGC, ARM, AutoRest, TypeScript, and Go unit tests
- TypeScript Spector integration tests
- repository format and lint


packages/typespec-azure-core/src/decorators/override-api-version.ts
```diff
@@ -0,0 +1,183 @@
...
+ * Returns the API-version override configured directly on a namespace or interface.
+ *
+ * @param program The TypeSpec program.
+ * @param target The directly decorated namespace or interface.
+ * @param emitterName The emitter language scope, such as `python`.
+ */
+export function getApiVersionOverride(
+  program: Program,
+  target: Namespace | Interface,
+  emitterName?: string,
```
tadelesh, line 49: No Azure Core decorators have scope. Shall we remove the `emitterName`?

packages/typespec-go/src/tcgcadapter/clients.ts
tadelesh, line 1: You do not need to add new param in TCGC. Just change the logic inner `updateWithApiVersionInformation` and all emitters could get transparent value for this.

packages/typespec-azure-core/src/decorators/override-api-version.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  const values = { ...current };
+  for (const language of scopes) {
+    values[language] = version;
+  }
+  setApiVersionOverrideState(program, target, values);
+}
+
+function parseScopes(scope: string): [negationScopes: string[], scopes: string[]] {
```
markcowl: because we do not need a scope, none of these scope functions are necessary.

packages/typespec-azure-core/src/decorators/override-api-version.ts
```diff
@@ -0,0 +1,183 @@
...
+const [getApiVersionOverrideState, setApiVersionOverrideState] = useStateMap<
+  Namespace | Interface,
+  ScopedApiVersionOverride
+>(AzureCoreStateKeys.apiVersionOverride);
+
+export const $overrideApiVersion: OverrideApiVersionDecorator = (
+  context: DecoratorContext,
+  target: Namespace | Interface,
+  version: string,
+  scope?: string,
```
markcowl: We do not need any scope in this case.  The override api version always applies for all emitters.

packages/typespec-client-generator-core/src/clients.ts
```diff
@@ -279,10 +290,6 @@ function addDefaultClientParameters<
       multipleServiceApiVersionParam.optional = true;
       defaultClientParamters.push(multipleServiceApiVersionParam);
     } else {
-      // For single-service clients, API version parameters are optional only when they have a client default value
```
markcowl: why remove this code?

packages/typespec-client-generator-core/src/clients.ts
```diff
@@ -262,6 +264,15 @@ function addDefaultClientParameters<
...
       apiVersionParam = context.__clientParametersCache.get(sc)?.find((x) => x.isApiVersionParam);
       if (apiVersionParam) {
+        const explicitDefault = apiVersionParam.__raw
+          ? getClientDefaultValue(context, apiVersionParam.__raw)
+          : undefined;
+        apiVersionParam = createClientScopedMethodParameter(
+          context,
+          apiVersionParam,
+          client.__raw,
+          typeof explicitDefault === "string" ? explicitDefault : undefined,
```
markcowl: I don't think we want to set and use this value if explicitDefault is not a string, so I think you need to condition the careation of this parameter on explicitDeaftul being string, and only use the value it it is set.  I think this also means you need to retain the deleted section below that ensures that the api version parameter is optional only if there is a clientDefaultValue

packages/typespec-client-generator-core/test/decorators/override-api-version.test.ts
```diff
@@ -0,0 +1,446 @@
...
+    deepStrictEqual(
+      context.sdkPackage.clients.map((client) => client.name),
+      ["WidgetsClient"],
+    );
+    const client = requireClient(context.sdkPackage.clients, "WidgetsClient");
+
+    strictEqual(getApiVersionDefault(client), "2099-01-01");
+  });
+
+  it("selects the override by emitter scope", async () => {
```
markcowl: we can remove all of the tests over scope - this should always apply

packages/typespec-go/src/tcgcadapter/clients.ts
tadelesh: I refined the TCGC logic.

markcowl: Yep, thanks! Incorporated it into the pr cleanup

markcowl:
@tadelesh Please take a look

JoshLove-msft:
What is the expected behavior when a user sets a runtime override for one these operations? Is this decorator intended to specify an operation-specific default, or the only supported api-version?

markcowl:
Per discussion, the updated network spec using this feature is here: https://github.com/markcowl/azure-rest-api-specs/tree/copilot/test-net-core-legacy-version/specification/network/resource-manager/Microsoft.Network/Network  (no PR because the package.json changes would be confusing)

markcowl:
> What is the expected behavior when a user sets a runtime override for one these operations? Is this decorator intended to specify an operation-specific default, or the only supported api-version?

It isn't intended to overcome the user trying to manually override the api-version for an operation.   If the user selects one of the valid api-versions for the client, though, correct functionality would be for this value to override it.  Here I'm assuming you have in mind the c# management sdk feature of being able to specify a previous api-version for the network client (for example).  Ideally, this would not impact the version sent for these APIs.

JoshLove-msft:
> If the user selects one of the valid api-versions for the client, though, correct functionality would be for this value to override it.

Sorry, do you mean the version that the user selects should win, or the overrideApiVersion should win? Yes, this is for the scenario where the user sets the api-version to use in the mgmt libraries.

weidongxu-microsoft:
> Per discussion, the updated network spec using this feature is here: https://github.com/markcowl/azure-rest-api-specs/tree/copilot/test-net-core-legacy-version/specification/network/resource-manager/Microsoft.Network/Network (no PR because the package.json changes would be confusing)

It may need some fix to your branch https://github.com/markcowl/azure-rest-api-specs/pull/12/changes

Still WIP on testing now based on HEAD of above PR

weidongxu-microsoft (approved):
Java is good on https://github.com/Azure/typespec-azure/pull/5176#issuecomment-5536122874 (there is 2 extra `@@clientLocation` for java, in addition to the fix on client.tsp)

Image

Image

timotheeguerin (approved):

tadelesh (approved):

msyyc (approved):

JiaqiZhang-Dev (approved):

markcowl:
@JoshLove-msft 
> > If the user selects one of the valid api-versions for the client, though, correct functionality would be for this value to override it.
> 
> Sorry, do you mean the version that the user selects should win, or the overrideApiVersion should win? Yes, this is for the scenario where the user sets the api-version to use in the mgmt libraries.

To be very specific.  If the latest api-version for the Foo client is 2026-09-01 and operation 'bar' is set to use api-version override 2023-10-01, there should be the following behaviors:

If the user does not specify a client api-version, then bar should send api-version 2023-10-01on the wire
If the user specifies an api version for client Foo like 2024-01-01, then bar should send api-version 2023-10-01 on the wire
If the user specifies a specific api-version (2026-03-01) for operation bar, then bar shoudl send api-version 2026-03-01 on the wire

Does that make sense?

JoshLove-msft (approved):

JialinHuang803 (approved):

JialinHuang803:
Looks good for JS.

necusjz (approved):
looks good for py.

Pan-Qi (approved):

PR #5183, opened 2026-08-07
Document public API surface ahead of the library-linter documentation rules

timotheeguerin:
[microsoft/typespec#11543](https://github.com/microsoft/typespec/pull/11543) adds `missing-documentation` and `extraneous-documentation` rules to `@typespec/library-linter`. Every Azure library already runs the linter with `--warn-as-error` as part of its `build` script, so the moment that PR lands our builds break — 94 warnings across azure-core, ARM, TCGC and portal-core.

This documents the gaps so the core PR can merge cleanly.

Two categories of problems turned up, and both were hiding real content from the published reference docs:

**Doc comments starting with a tag swallowed the whole description.** `@deprecated` isn't a tag the compiler knows, so everything after it became tag content and the description vanished:

```diff
 /**
- * @deprecated Use `@client` instead. The `@operationGroup` decorator is deprecated. Sub clients should be represented using `@client`.
  * Define the sub client generated in the client SDK.
  * ...
+ *
+ * Deprecated: use `@client` instead. Sub clients should be represented using `@client`.
  */
 extern dec operationGroup(...);
```

`@operationGroup` and `ResourceOperations` now actually show their description on the website. Same story for `Azure.Core.ResourceOperationStatus`, whose only doc was behind a `@dev` tag.

**`@template` tags pointing at parameters that don't exist.** Operations inside `ResourceOperations`, `PrivateLinks` and friends documented their *interface's* template parameters, and a few models still referenced pre-rename names (`@template T` on `LocationOfCreatedResourceResponse<Resource>`). Those are now either corrected or removed.

The rest is plain missing docs on properties, enum members, template parameters and decorator parameters.

### Emitted output

Descriptions on ARM common types flow into the generated swagger, so `packages/samples/common-types` picks up a few additive `description` fields on `NetworkSecurityPerimeterConfigurationProvisioningState` and `Severity` enum values — consistent with `IssueType` and `ResourceAssociationAccessMode`, which already carry them.

Verified with `tsp compile . --warn-as-error --import @typespec/library-linter` against the linter build from the core branch: all six libraries are clean, and the full test suites for the touched packages pass unchanged.

### DIFF
--- a/packages/typespec-azure-resource-manager/lib/interfaces.tsp
+++ b/packages/typespec-azure-resource-manager/lib/interfaces.tsp
@@ -33,11 +33,12 @@ interface Operations {
 }
 
 /**
- * @deprecated Use Azure.ResourceManager.TrackedResourceOperations instead
  * A composite interface for resources that include `ResourceInstanceOperations<Resource, Properties>`
  * and `ResourceCollectionOperations<Resource>`. It includes: `GET`, `PUT`, `PATCH`, `DELETE`, ListByParent,
  * ListBySubscription operations. The actual route depends on the resource model.
  * This is the most common API pattern for Tracked Resources to use.
+ *
+ * Deprecated: use `Azure.ResourceManager.TrackedResourceOperations` instead.
  * @template Resource the ArmResource that provides these operations
  * @template Properties RP-specific property bag for the resource
  * @template BaseParameters The http parameters that are part of the request
@@ -147,7 +148,6 @@ interface ResourceCollectionOperations<
 interface ResourceListBySubscription<Resource extends Foundations.SimpleResource> {
   /**
    * List resources by subscription.
-   * @template Resource The ArmResource to list.
    */
   @doc("List {name} resources by subscription ID", Resource)
   listBySubscription is ArmListBySubscription<Resource>;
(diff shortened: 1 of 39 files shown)


packages/typespec-azure-resource-manager/lib/interfaces.tsp
```diff
@@ -33,11 +33,12 @@ interface Operations {
 }
 
 /**
- * @deprecated Use Azure.ResourceManager.TrackedResourceOperations instead
  * A composite interface for resources that include `ResourceInstanceOperations<Resource, Properties>`
  * and `ResourceCollectionOperations<Resource>`. It includes: `GET`, `PUT`, `PATCH`, `DELETE`, ListByParent,
  * ListBySubscription operations. The actual route depends on the resource model.
  * This is the most common API pattern for Tracked Resources to use.
+ *
+ * Deprecated: use `Azure.ResourceManager.TrackedResourceOperations` instead.
```
markcowl, line 41: This will show up in specs that use it though (unlike the non-existent tag above, I think)
We should probably add an @doc("") here - but that would likely make some spec changes as well.

markcowl (commented):
One nit.  I think we should see the impact on the specs repo (if any) and correct in typespec-next before merging.

markcowl (approved):
Need to have a pr fixing typespec-next (and hopefully in the private repo as well)

PR #5189, opened 2026-08-10
add spector test cases except union cases

kazrael2119:
add a new function to filter out go scope for union cases, and add other spector test cases


jhendrixMSFT:
FYI the tsp unions support work will land some time next week.

tadelesh (approved):

tadelesh:
Let's temporally add this filter and remove it after union support.

packages/typespec-go/test/azure-http-specs/azure/resource-manager/commonpropsgroup/error_client_test.go
```diff
@@ -14,7 +14,6 @@ import (
 )
 
 func TestErrorClient_CreateForUserDefinedError(t *testing.T) {
-	t.Skip("https://github.com/Azure/typespec-azure/issues/4946")
```
jhendrixMSFT, line 17: This is incorrect and will be reverted.

packages/typespec-go/test/http-specs/special-headers/repeatabilitygroup/repeatability_client_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"repeatabilitygroup"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRepeatabilityClientImmediateSuccess(t *testing.T) {
+	client, err := repeatabilitygroup.NewRepeatabilityClientWithNoCredential("http://localhost:3000", nil)
+	require.NoError(t, err)
+
+	resp, err := client.ImmediateSuccess(
```
jhendrixMSFT, line 20: These are not to be treated as regular header parameters (see https://github.com/Azure/typespec-azure/issues/4957).  This test will be removed.

PR #5196, opened 2026-08-10
feat(typespec-ts): streaming

l0lawrence:
## Summary

Adds structured JSONL + SSE streaming code generation to the `@azure-tools/typespec-ts` modular emitter. A response whose TCGC metadata indicates a structured stream now generates a `Promise<AsyncIterable<T>>` of deserialized items/events instead of returning the raw binary `Uint8Array` body.

## What is generated

- **JSONL** (`JsonlStream<T>`) -> `Promise<AsyncIterable<T>>` that lazily decodes JSON Lines.
- **SSE** (`SSEStream<T>`) -> `Promise<AsyncIterable<...>>` that dispatches each Server-Sent Event by its `event:` name (via `@azure/core-sse`), deserializes the payload, and stops at the terminal event. Full heterogeneous multi-event dispatch and `@terminalEvent` handling are supported (named terminals keyed on the terminal flag; unnamed `data: [DONE]` sentinel terminals). Generation is driven by TCGC `streamMetadata` / `sseMetadata`.

Operations connect eagerly (HTTP status/errors surface at call time) and decode the body lazily.

## Implementation

- New static helpers `static/static-helpers/streamingHelpers.ts` (`getStreamResponse`, `readJsonlStream`, `ensureStreamStatus`) and `sseStreamingHelpers.ts` (`readSseStream`, `SseEventDescriptor`), registered in `static-helpers-metadata.ts` and loaded only when the package has a structured-stream operation (`packageHasStructuredStreaming` / `packageHasSseStreaming`, side-effect-free metadata checks). `sseStreamingHelpers.ts` is the only file importing `@azure/core-sse`.
- `operation-helpers.ts`: `getStructuredStreamInfo` resolves stream metadata (gated purely on `response.streamMetadata` / `sseMetadata`); branches in the operation and deserialize builders emit the AsyncIterable shape.
- `@azure/core-sse` added to `AzureCoreDependencies` and injected into the generated package's dependencies only when an SSE op is present.


packages/typespec-ts/spector.config.yaml
```diff
@@ -115,6 +115,7 @@ specs:
   azure/client-generator-core/client-doc: true
   azure/client-generator-core/response-as-bool: true
   documentation: true
+  streaming/sse: true
```
l0lawrence, line 118: is there anything else that should be added?

packages/typespec-ts/static/static-helpers/streamingHelpers.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Connects to a streaming operation and returns the raw response with its body exposed as a
+ * readable stream. This bypasses Core's default response handling so the streamed body is not
+ * buffered or coerced into UTF-8 before it can be decoded.
+ */
+export async function getStreamResponse(
+  streamableMethod: StreamableMethod,
+): Promise<StreamResponse> {
+  const response = await streamableMethod.asNodeStream();
```
maorleger, line 20: So in JS we have multiple environments to think about:

- node / nodeLike
- browser
- react-native

We handle this by generating polyfills for each environment with a convention for file-names. Then, the build system automatically placed the proper file in the right exported bundle.

The best comparison is the handling for binary response:

- https://github.com/Azure/typespec-azure/blob/main/packages/typespec-ts/static/static-helpers/serialization/get-binary-response.ts
- https://github.com/Azure/typespec-azure/blob/main/packages/typespec-ts/static/static-helpers/serialization/get-binary-response-browser.mts
- https://github.com/Azure/typespec-azure/blob/main/packages/typespec-ts/static/static-helpers/serialization/get-binary-response-react-native.mts

I wonder if this info is enough to have copilot provide similar polyfills for streaming but that's the recipe to follow.

I _think_ core-client-rest has asNodeStream and asBrowserStream so it might just be a matter of calling the right one.

packages/typespec-ts/src/modular/helpers/operation-helpers.ts
```diff
@@ -1104,6 +1121,287 @@ export function getOperationFunction(
...
+    }
+    const itemType =
+      payloadTypeExpressions.length > 0
+        ? Array.from(new Set(payloadTypeExpressions)).join(" | ")
+        : getTypeExpression(context, streamMetadata.streamType);
+    return { kind: "sse", itemType, events };
+  }
+
+  const streamType = streamMetadata.streamType;
+  if (streamType.kind !== "model" && streamType.kind !== "union") {
```
maorleger, line 1253: I'm not sure, but would it be better to use the contentTypes in SdkStreamMetadata to determine whether this is an sse or jsonl stream? 

- application/jsonl
- text/event-stream

I just worry that something like a custom stream template might be incorrectly interpreted as jsonl but we should verify, I might just be thinking about it wrong and it may be a non-issue

packages/typespec-ts/src/modular/helpers/operation-helpers.ts
```diff
@@ -1104,6 +1121,287 @@ export function getOperationFunction(
...
+        } else if (!sseEvent.isTerminalEvent) {
+          // A non-terminal payload whose type needs no model deserializer (primitive/scalar/
+          // enum). Yield the raw payload via an identity deserializer instead of dropping it.
+          event.identityDeserialize = true;
+        }
+        if (sseEvent.payloadContentType !== undefined) {
+          event.contentType = sseEvent.payloadContentType;
+        }
+        if (!sseEvent.isTerminalEvent) {
+          payloadTypeExpressions.push(getTypeExpression(context, sseEvent.payloadType));
```
maorleger: Can you have a terminal event with data / model? If so, would this lose type information?

To help me understand a bit, I asked :copilot: and decided to shamelessly paste its output here verbatim 😄 

```md
 terminal event *with a model payload* still gets a deserializer. The helper yields it before the
generator returns (`sseStreamingHelpers.ts:78-85`). Executed against the PR's own helper bodies:

| Case | Observed | Problem |
| --- | --- | --- |
| Terminal event, model payload | `[{desc:"a"}, {desc:"FINAL"}]` | Yielded, but its type is absent from the declared union — unsound |
| Terminal event, primitive payload | `[{desc:"a"}]` | Silently dropped — inconsistent with the model case |
| Unrelated `event: delta` carrying `data: [DONE]` | `[]` | Sentinel matched on data only; the wrong event ends the stream |

The third case happens because the terminal check runs before descriptor matching. It ignores the
event name.

**Impact**: a typed terminal event often carries the final logical result. Runtime delivers it, but
the type system cannot see it. Consumers need a cast to narrow to it.

**Fix**: `@terminalEvent` says only that the client must disconnect. It does not dictate whether the
triggering event is yielded. Pick one policy: yield typed terminal events, or consume all terminal
events internally. Then make the runtime value, declared item union, and primitive/model paths
match. Name-sensitive sentinel matching is secondary hardening, not part of the blocker.

> **Cross-language note**: matching the terminal by data value mirrors Python's `terminal_event`
> with `event.data == self._terminal_event`. That half is a deliberate shared choice. Raise
> name-matching with both languages.
```

packages/typespec-ts/test/azure-modular-integration/streaming-sse.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      events.map((e) => e.desc),
+      ["one", "two", "three"],
+    );
+  });
+
+  it("should stream heterogeneous named events and stop at terminal [DONE]", async () => {
+    const events = await collect<ResponseCreated | ResponseDelta>(await client.named.receive());
+    // Terminal `data: [DONE]` must not be yielded.
+    assert.strictEqual(events.length, 3);
+    assert.strictEqual((events[0] as ResponseCreated).id, "resp_1");
```
maorleger: I think we probably want to keep the event name for better type information?

Something like: 

```ts
export type NamedReceiveEvent =
  | { event: "responseCreated"; data: ResponseCreated }
  | { event: "responseDelta"; data: ResponseDelta };
```

Will allow the user to narrow down the event based on the kind like in [this playground](https://www.typescriptlang.org/play/?#code/PQKhAIEkHIFtwFYFcDOAXcsCGBrAlgHYDm4aAFniuEgA7gBGSGe2RhApuDVgJ4A2AeywATKuSwZsPcAIBOmXuAICMwvADN17eSGABYAFCHCabeqwBjTgCV2KGgIIp2AYVnsJ7YeADeh8DL0CABc4OiyhEQA3P5haBKooeGRMQYBKAKw7ADy5NoACrICNNpoPKH0AgJ8HgSGAL6GxgSmsuZW4Lb2js4AIux88b6xAkFJaBHEqenxaIlxk9ENTUYGoBAA0oTCWOB8eDiceeAWSLLuLeAsNDWwLRJ4juAXwtpUuoZlJZ3sVngAbuwAKKAy4AXh+3Scrncnm8AB9IQ5of1BlhDBYehgkAQCFgssIQewWqFbH9AUTwb5RiFwAAidRVOkAGjiCRQoTp9CwshZYUyOTyskKxVK5VIsiQ7Eaq0M63AAHU8ORwLs1CgLBFYIQJHJwB8DF9OAA5fFeMnsAHA0EYMGxRE+cDsG2c9xQ5xuDymYR0qLgHbxUl2ZEe2He8AygIOp0u+lukPsVHxX3+iRYIPuxMDIb1VKYpwYPEEyloUKmgkWq0l8AQnzO4mluPBnowr1ePkB9PUXFmwk2mUrFAAd2VFjI4AAFEWvCWAHT1loASmGaROWGcTcznrhdOCsQC+YyNVngiIU97c87s4yWVyZAKRRKsjKy+AwHAZHXpB4JW8KFoDjPvuDCwjg0xrhudLxi2SZYLuwGHtU7AngIZ7Tn2DazleN6CvewqPmKr7vkaTrnHqX7sAAHiUFjhsB9CgakA6rG+4CMvIOLoTGDbgEOnCfoCSgCN+v5hABcgYFgBDeAJRzCZqbYyGcMhDgQImcHiZEjsQe6rIhx6nhOE6cReNqqlQXQJtu3qLteAp3g+orPjwRHdigWBaOAxjqJOdI4Q5+FOWUdJXGpJnFjay6OqxygYLCfB8NIeBqJkDwWOAAAqADKsT6chhnhTONp2beQoik+L5MSsQA)

In general it's probably better to keep the envelope shape if one is defined. I noticed TCGC has that information. I think JSONL needs to stay payload only though

packages/typespec-ts/static/static-helpers/sseStreamingHelpers.ts
```diff
@@ -0,0 +1,87 @@
...
+      return;
+    }
+    const descriptor = named.get(event.event) ?? unnamed;
+    if (!descriptor) {
+      continue;
+    }
+    if (descriptor.deserialize) {
+      const payload = event.data
+        ? isJsonContentType(descriptor.contentType)
+          ? JSON.parse(event.data)
```
maorleger: hmm ok so if there's a JSON parse error, the customer loses all information about the event like event name, status, etc.

This is minor, since it involves an edge case (invalid JSON payload) but still, would it make sense to throw a more useful error here?

```ts
try {
      const payload = event.data
        ? isJsonContentType(descriptor.contentType)
          ? JSON.parse(event.data)
catch (e) {
    throw new Error(`Unable to deserialize event ${event.event} as JSON", { cause: e });
}
   ...

Or something along those lines?

maorleger (commented):
This is coming along quite nicely!! Thanks so much for all the effort in getting this together.

I still have lots to review, but the overall direction seems ok. I noted a few findings that we should fix though

packages/typespec-ts/test/azure-modular-integration/streaming-sse.test.ts
l0lawrence: you can have named and unamed events here based on the spector test like `Info vs responseCreated: Info `- so if there is an unamed event it will just be the payload as well

packages/typespec-ts/src/modular/helpers/operation-helpers.ts
l0lawrence: So from my understanding the case we are supporting now is something like the following:

data: hello this is blah

data: [DONE]


where `@terminalEvent` in typespec is defined as "[DONE]"

maorleger: I see, ok so as long as we're guaranteeing that the terminal event is just a sentinel value and doesnt carry data this should be fine. Thanks for confirming! 👍

packages/typespec-ts/test/azure-modular-integration/streaming-sse.test.ts
maorleger: I see, but for the case of an envelope [see here, just scrolling down a bit](https://gist.github.com/chrisradek/b2656d7ee5db1b29768d50db0628f033#typespecevents) how do we represent it in other languages? Do we drop the envelope there as well and just return the payload?

l0lawrence: It looks like we dont have spector tests for envelope scenarios yet (`@data`), so at least in python I did not have support for them yet

packages/typespec-ts/src/modular/helpers/operation-helpers.ts
l0lawrence: I went looking at the search typespec instead of just the spector tests and my understanding is actually wrong they have a terminal event as a model not a literal in their typespec that this would need to support:

`@terminalEvent`
`response.completed`: KnowledgeBaseResponseCompletedEvent

so I believe we want to yield it to look like:

{event: `response.completed`; data: KnowledgeBaseResponseCompletedEvent} ?

maorleger: Oh nice find!! Thanks for digging into it. And yea, what you're proposing makes sense at first glance but I trust your judgement

PR #5197, opened 2026-08-10
fix(arm): Remove modelDeploymentRef from Agent Appliance deployment model definition.

vayada:
The appliance owns and reports the entire agent definition, so `modelDeploymentRef` does not belong in `AgentDefinitionAppliance`. It is now Platform-only and remains available on `AgentDefinitionPlatform`.

- `AgentDefinitionAppliance` drops the property and its now-unused `HasModelDeploymentRef` template parameter, so the signature becomes `AgentDefinitionAppliance<HasInstructions>`.
- `AgentPropertiesAppliance.definition` is `@visibility(Lifecycle.Read)` again. It was only left writable in #5178 so clients could send `modelDeploymentRef`; with the property gone every appliance definition field is service-owned.
- Aligns the ARM agent sample api-version with the Agent base type contract version (2024-06-01 -> 2026-04-01).

Regenerated reference docs, extern signatures and the sample swagger.


.chronus/changes/agent-base-type-remove-appliance-model-deployment-ref-2026-08-10-11-55-00.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: breaking
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Remove `modelDeploymentRef` from the experimental Agent base type `AgentDefinitionAppliance` model. The appliance owns the entire agent definition, so the property is now Platform-only and remains available on `AgentDefinitionPlatform`. As a result `AgentDefinitionAppliance` takes a single template parameter (`AgentDefinitionAppliance<HasInstructions>` instead of `AgentDefinitionAppliance<HasModelDeploymentRef, HasInstructions>`), and `AgentPropertiesAppliance.definition` is read-only again since it no longer needs to carry a client-writable field.
```
markcowl, line 7: Shorten this to just simply describe the api change (e.g. remove 'modelDeploymentRef' from AgentDefinitionAppliance and make the definition property readOnly)

packages/samples/specs/resource-manager/resource-types/agent/main.tsp
```diff
@@ -19,9 +19,9 @@ namespace Microsoft.ContosoAgent;
 
 /** Contoso Agent API versions */
 enum Versions {
-  /** 2024-06-01 version */
+  /** 2026-04-01 version */
```
markcowl, line 22: these changes are unnecessary to this fix

packages/samples/specs/resource-manager/resource-types/agent/main.tsp
```diff
@@ -33,11 +33,11 @@ enum Versions {
 // AgentDefinitionPlatform: Use for the Platform deployment model where the client
 // owns and manages agent configuration. Definition fields are writable.
 // Template params control optional definition fields:
-//   HasModelDeploymentRef: include a modelDeploymentRef property
 //   HasInstructions: include an instructions property
+//   HasModelDeploymentRef (Platform only): include a modelDeploymentRef property
```
markcowl, line 37: Why is this added back?

markcowl (changes requested):

packages/samples/specs/resource-manager/resource-types/agent/main.tsp
vayada: This is because Platform definition will have modelDeploymentRef but Appliance won't. (Being left for RPs to choose and add it by themself if needed)

markcowl (approved):

PR #5204, opened 2026-08-11
Fix nested XML model element name precedence

### LINKED ISSUE #5179: [typespec-go] Nested XML model serialization ignores the property element name
### Description

When a model property references another model with a custom XML name, the generated Go XML serializer uses the referenced model's XML name instead of the property element name.

According to the TypeSpec scenario, the property name should take precedence.

Relevant scenario:

- https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/payload/xml/main.tsp
- https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/payload/xml/mockapi.ts

The generated containing model correctly has:

```go
Author *Author `xml:"author"`
```

However, the generated `Author.MarshalXML()` unconditionally overrides the incoming element name:

```go
func (a Author) MarshalXML(enc *xml.Encoder, start xml.StartElement) error {
    start.Name.Local = "XmlAuthor"
    // ...
}
```

### Reproduction

1. Generate the `payload/xml` Spector client.
2. Start Spector with `pnpm spector --start`.
3. Run `TestXMLModelWithRenamedNestedModelValueClient_Put` after removing its temporary skip.

The request fails with HTTP 400.

Expected body:

```xml
<?xml version='1.0' encoding='UTF-8'?>
<ModelWithRenamedNestedModel>
  <author>
    <name>foo</name>
  </author>
</ModelWithRenamedNestedModel>
```

Actual body:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ModelWithRenamedNestedModel><XmlAuthor><name>foo</name></XmlAuthor></ModelWithRenamedNestedModel>
```

### Expected behavior

When `Author` is serialized as the `author` property of `ModelWithRenamedNestedModel`, the generated serializer should preserve the property element name passed through `xml.StartElement`.

The type-level XML name `XmlAuthor` should only be used when `Author` is serialized as a root element or when no property-level element name applies.

### Environment

- `@azure-tools/typespec-go`: `0.15.0`
- TypeSpec compiler: `1.14.0`

tadelesh:
## Summary

- make generated model `MarshalXML` methods apply custom model XML names for direct/root serialization
- route nested model properties and array items through an internal context-aware marshaller so their generated XML tags take precedence
- handle property names that collide with the Go model name without relying on string comparisons
- let client requests and fake-server responses use the same model marshalling behavior instead of separate root wrappers
- enable the existing nested renamed-model Spector PUT regression and add fake-server and emitter snapshot coverage

## Testing

- built and type-checked `@azure-tools/typespec-go`
- regenerated `azblob` and `payload/xml`
- ran the direct-root and nested-property azblob tests
- ran the XML name-precedence unit snapshots and scenario-suite guard
- ran the targeted fake-server XML tests
- ran the complete `payload/xml` Spector Go module uncached, including wrapped and unwrapped model arrays
- ran repository formatting and linting

Fixes #5179


tadelesh:
@jhendrixMSFT I've tried several solutions to handle different naming of root usage vs nested usage for a model. The current one is more general. What do you think?

jhendrixMSFT:
IMO we need to retool how we model XML info.  There were some cases between Autorest and tsp that were different leading to some awkward code.  Now that we don't have to share anything with Autorest we can retool it.

jhendrixMSFT (changes requested):
Will retool XML modeling now that autorest is out of the picture.

Also, will wait until after storage tsp migration has landed.

jhendrixMSFT:
Superseded by https://github.com/Azure/typespec-azure/pull/5466

PR #5212, opened 2026-08-12
Upgrade deps [skip chg]

iscai-msft:
Upgrade general deps following `pnpm upgrade --latest -r -i`


packages/typespec-ts/test/commands/run.ts
```diff
@@ -264,10 +264,20 @@ async function runTypespecHelper(env: GenEnv): Promise<void> {
     // Defaults are merged in api-extractor when the config file is read from disk with
     // `ExtractorConfig.loadFile`. This is derived from that method.
     // https://github.com/microsoft/rushstack/blob/1a92f17fa537b55529adbec80203bd99afd8cd24/apps/api-extractor/src/api/ExtractorConfig.ts#L624-L627
-    const configObject = deepMerge(
-      structuredClone((ExtractorConfig as any)._defaultConfig),
-      baseConfigObject,
+    // The defaults file uses JSONC (with comments), so it must be parsed with the same
+    // JsonFile loader api-extractor itself uses rather than plain JSON.parse.
+    const apiExtractorRequire = createRequire(
```
iscai-msft: Spector test harness reached into `ExtractorConfig._defaultConfig`, an internal api-extractor API that newer versions no longer expose as a static class property (it's now a private module-level constant), so the config merge blew up with `Cannot read properties of undefined (reading 'apiReport')`. Updated `run.ts` to load `schemas/api-extractor-defaults.json` directly via api-extractor's own `JsonFile` helper, which handles the JSONC comments that plain `JSON.parse` chokes on.

pnpm-workspace.yaml
```diff
@@ -19,6 +19,14 @@ overrides:
   "diff@>=6.0.0 <8.0.3": ">=8.0.3"
   dompurify: ^3.3.3
   "yaml@>=2.0.0 <2.8.3": ">=2.8.3"
+  # Pinned exactly (rather than left to range resolution) to match the versions baked into
+  # core/packages/typespec-vscode/ThirdPartyNotices.txt (which we cannot regenerate here).
```
timotheeguerin: hhm this just sounds like you should regenerate both lock file(delete it, delete your node_modules and rerun pnpm install)
it should pick all the same versions
we shouldn't need to pin things like that.

packages/typespec-ts/test/commands/run.ts
timotheeguerin: this is quite nasty, can we omit this library from the uprade and file an issue for typespec-ts to resolve

pnpm-workspace.yaml
iscai-msft: Done — removed all three pinning overrides and the explanatory comment block, deleted the lockfile, and regenerated from scratch with `pnpm install`.

packages/typespec-ts/test/commands/run.ts
iscai-msft: Done — reverted `run.ts` back to the original `_defaultConfig` approach and pinned `@microsoft/api-extractor` to `~7.58.1` via override to avoid 7.59.0. Filed #5370 to track the proper fix.

pnpm-workspace.yaml
```diff
@@ -19,6 +19,10 @@ overrides:
   "diff@>=6.0.0 <8.0.3": ">=8.0.3"
   dompurify: ^3.3.3
   "yaml@>=2.0.0 <2.8.3": ">=2.8.3"
+  # Pin api-extractor to 7.58.x — 7.59.0 moved _defaultConfig to a private module-level
+  # constant, breaking the in-memory config construction in typespec-ts spector tests.
+  # See: https://github.com/Azure/typespec-azure/issues/XXXX
```
iscai-msft: Tracking issue: https://github.com/Azure/typespec-azure/issues/5370

timotheeguerin (approved):

PR #5215, opened 2026-08-13
Allow C# operation overloads in duplicate-name validation

JoshLove-msft:
## Summary

- keep duplicate operation-name validation enabled for C# and report overload collisions as `duplicate-client-name-warning`
- share the duplicate client-name diagnostic creation logic between the scoped and post-generation validators
- add regression coverage that verifies both C# overload warnings
- add the required Chronus change entry for `@azure-tools/typespec-client-generator-core`

[#5027](https://github.com/Azure/typespec-azure/pull/5027) added a second validation pass over generated client methods. That pass grouped methods by name only, so it reported `duplicate-client-name` errors for valid C# overloads with different parameter lists. The scoped validator already reports C# operation name collisions as `duplicate-client-name-warning`; this change applies the same behavior to the post-generation validator while preserving errors for other emitters.

### DIFF
--- a/.chronus/changes/josh-fix-overload-diagnostic-suppression-2026-7-12-21-4-45.md
+++ b/.chronus/changes/josh-fix-overload-diagnostic-suppression-2026-7-12-21-4-45.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Allow C# emitters to generate operation overloads with the same client name.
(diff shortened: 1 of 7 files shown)


tadelesh (dismissed):

weidongxu-microsoft (approved):

weidongxu-microsoft:
Python failure likely caused by TCGC commit a few days ago. Unlikely caused by this fix.

JoshLove-msft:
CI investigation: the `Python / Mock API Tests` failure is not caused by this PR.

- This PR changes only TCGC context validation, its TypeScript test, and a Chronus entry.
- TCGC TypeScript build/unit/Spector, lint, and format checks pass.
- The same 24 boolean/numeric/bytes encoding tests fail on unrelated PR #5214 and on its merge-queue run: https://github.com/Azure/typespec-azure/actions/runs/31668672805/job/94350403270
- The failures are serialization expectation mismatches such as sending `b'test'` where the mock API expects base64 `dGVzdA==`.

This is a current repository baseline/release-wave failure rather than a regression from the C# overload validation change.

--generated by Copilot

PR #5230, opened 2026-08-14
[Swagger Linter Migration] PathParameterNames (origin)

msyyc:
## Original Swagger linter

**Rule:** `PathParameterNames`

**Docs:** https://github.com/Azure/azure-openapi-validator/blob/main/docs/path-parameter-names.md

**Source:** https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/path-param-names.ts

The original Swagger linter checks:

- [x] Traverse the emitted OpenAPI `paths` object.
- [x] Split each path key into `/`-separated segments.
- [x] For every path parameter segment after the first segment, remember the parameter name associated with the immediately preceding literal/segment value.
- [x] Report when a later path uses the same preceding segment with a different parameter name.
- [x] Report the diagnostic at the offending Swagger path key: `['paths', pathKey]`.
- [x] Emit one diagnostic per offending emitted path key, not per HTTP method.

The Swagger implementation receives the OpenAPI `paths` object and iterates `Object.keys(paths)`. It keeps a document-local `paramNameForSegment` map keyed by the preceding path segment. When a path parameter follows a segment that already has a remembered parameter name, the rule reports if the current parameter name differs.

This means the rule is tied to emitted OpenAPI document boundaries and emitted path keys. It does not inspect TypeSpec source operations directly, does not evaluate `x-ms-paths`, and naturally sees only the service/version document selected for validation. It can also observe one path once even when multiple HTTP methods share that path.

The TypeSpec rule now mirrors those emitted-document boundaries by traversing HTTP services at `root`, skipping ARM provider namespaces because this Swagger rule is data-plane only, and evaluating each original service independently. For versioned services it uses `getVersioningMutators` with `unsafe_mutateSubgraphWithNamespace` so routes are compared only within each emitted API-version document.

Within a service document, the rule deduplicates by emitted path, excludes literal-query routes that AutoRest emits under `x-ms-paths`, and orders paths with AutoRest-style URL ordering before establishing the expected parameter name. Diagnostics target the corresponding TypeSpec path parameter when available, falling back to the operation. Repeated diagnostics across version projections are deduplicated by source target plus emitted path plus segment position, scoped to the original service, so distinct paths or distinct services that reuse the same parameter property remain independently reported.

## Migration evidence

Detailed migration evidence, including focused fixtures/tests, report reconciliation, full-corpus counts, one-sided project explanation, compile failures, and remaining historical uncertainty is documented in:

https://github.com/Azure/typespec-azure/blob/feature/lintdiff-path-parameter-names/packages/typespec-lintdiff/test/fixtures/PathParameterNames/migration.md

Summary: the final full rerun processed 468 projects with six ARM compile failures, and the corrected `PathParameterNames` row had 0 validator projects, 0 TypeSpec projects, 0 overlap, no one-sided project sets, and 0 diagnostics. Generated corpus artifacts under `packages/typespec-lintdiff/specs` are validation evidence only and are not included in this PR.


msyyc:
This is not ARM linter rule

PR #5265, opened 2026-08-20
[Swagger Linter Migration] EnumInsteadOfBoolean

msyyc:
## Original Swagger linter

- linter code: [EnumInsteadOfBoolean](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/enum-insteadof-boolean.ts)
- linter doc: [enum-instead-of-boolean.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/enum-instead-of-boolean.md)
- Validator ruleset registration: https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/az-common.ts

Specific checks promoted from the done lintdiff rule:

- [x] Boolean model properties should be reported.
- [x] Boolean operation parameters should be reported, including path parameters.
- [x] Boolean request bodies should be reported.
- [x] Boolean response bodies should be reported.
- [x] Comparable non-boolean shapes should not be reported.

The Swagger rule is a Spectral rule from the common ruleset. It inspects emitted OpenAPI schema objects and reports schemas whose type is `boolean`, with diagnostics located on the emitted schema path. The lintdiff migration evidence accepts that the Swagger and TypeSpec implementations run at different representation layers: Swagger can report repeated emitted schema copies, while the TypeSpec rule reports the authorable source target.

Source branch: `feature/lintdiff-migration-new`

Source rule: https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/src/rules/enum-instead-of-boolean.ts

The source worktree had no uncommitted changes for the lintdiff source rule or `EnumInsteadOfBoolean` fixture directory when this promotion was created. The user-marked done lintdiff source rule was not modified during promotion.

## Destination analysis

The rule belongs in `@azure-tools/typespec-azure-core` because the lintdiff metadata marks `EnumInsteadOfBoolean` as `applicability: Both` with `sources: ["common"]`, and the local TypeSpec rule only depends on compiler/http APIs. It does not inspect ARM resources, provider namespaces, ARM resource paths, ARM lifecycle operations, or ARM envelopes.

Because the rule applies to both ARM and data-plane specs, this PR enables `@azure-tools/typespec-azure-core/enum-instead-of-boolean` in both `typespec-azure-rulesets` data-plane and resource-manager rulesets.

The promoted Azure Core rule preserves the done lintdiff rule behavior:

- Visits model properties and reports a diagnostic when the property type is the intrinsic `boolean` scalar.
- Visits HTTP operation responses via `getHttpOperation` and reports boolean response bodies on the operation or authored body property target.
- Keeps the diagnostic as a warning and adapts the message/docs to Azure Core convention by recommending descriptive extensible enums.
- Registers the rule in the Azure Core linter and rulesets without adding any dependency on Azure Resource Manager libraries.

- [boolean-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/EnumInsteadOfBoolean/boolean-property/main.tsp) -> `emits warning for boolean model properties`
- [boolean-path-param](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/EnumInsteadOfBoolean/boolean-path-param/main.tsp) -> `emits warning for boolean path parameters`
- [boolean-body](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/EnumInsteadOfBoolean/boolean-body/main.tsp) -> `emits warning for boolean request bodies`
- [boolean-response](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/EnumInsteadOfBoolean/boolean-response/main.tsp) -> `emits warning for boolean response bodies`
- [non-boolean-shapes](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/EnumInsteadOfBoolean/non-boolean-shapes/main.tsp) -> `allows comparable non-boolean shapes`

The official package tests use direct TypeSpec snippets and expected diagnostics. Lintdiff harness snapshots were not copied.

## Migration evidence

Migration evidence: https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/EnumInsteadOfBoolean/migration.md

The detailed focused tests, real-service project comparison, corpus counts, one-sided project explanations, compile-failure handling, and remaining uncertainty are recorded there rather than duplicated in this PR description.

## Promotion sync policy

Semantic gaps found after promotion should block the promotion PR until the user explicitly reopens lintdiff repair. This PR does not include unapproved source-rule edits in `packages/typespec-lintdiff`.


.chronus/changes/add-enum-instead-of-boolean-rule-2026-08-20-13-13-00.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-core"
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Add the `enum-instead-of-boolean` lint rule that recommends descriptive extensible enums instead of boolean API shapes when semantic values matter.
```
catalinaperalta: nit: 
```suggestion
Add the `use-enum-instead-of-boolean` lint rule that recommends descriptive extensible enums instead of boolean API shapes when semantic values matter.
```

packages/typespec-azure-rulesets/src/rulesets/resource-manager.ts
```diff
@@ -9,6 +9,7 @@ export default {
     "@azure-tools/typespec-azure-core/byos": true,
     "@azure-tools/typespec-azure-core/casing-style": true,
     "@azure-tools/typespec-azure-core/composition-over-inheritance": true,
+    "@azure-tools/typespec-azure-core/use-enum-instead-of-boolean": true,
```
catalinaperalta: FYI you left this as true if we want this PR merged in soon

catalinaperalta (approved):

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.md
```diff
@@ -0,0 +1,60 @@
+---
+title: "use-enum-instead-of-boolean"
+---
+
+```text title="Full name"
+@azure-tools/typespec-azure-core/use-enum-instead-of-boolean
+```
```
timotheeguerin, line 7: I don't think you should have the heading here anymore those get added by tspd

timotheeguerin: Image
https://tspwebsitepr.z22.web.core.windows.net/typespec-azure/prs/5265/docs/libraries/azure-core/rules/use-enum-instead-of-boolean/

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.ts
```diff
@@ -0,0 +1,57 @@
+import type { Type } from "@typespec/compiler";
+import { createRule, fileRef } from "@typespec/compiler";
+import { getHttpOperation } from "@typespec/http";
+
+export const useEnumInsteadOfBooleanRule = createRule({
+  name: "use-enum-instead-of-boolean",
```
timotheeguerin, line 6: I think `no-boolean` would be a better name, this is quite long

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.ts
```diff
@@ -0,0 +1,57 @@
...
+        if (!isBooleanScalar(property)) {
+          return;
+        }
+
+        context.reportDiagnostic({
+          target: property,
+        });
+      },
+      operation: (operation) => {
+        const [httpOperation] = getHttpOperation(context.program, operation);
```
timotheeguerin, line 28: hhm do we really need to use getHttpOperation here? what are you trying to colect? This feels like a simple rule that should be http agnostic

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.ts
```diff
@@ -0,0 +1,57 @@
...
+              target: content.body.property ?? operation,
+            });
+          }
+        }
+      },
+    };
+  },
+});
+
+function isBooleanScalar(type: Type): boolean {
```
timotheeguerin, line 53: use typekit instead

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.md
```diff
@@ -0,0 +1,60 @@
...
+values can be added without a breaking change.
+
+## Impact
+
+- **Area:** SDK, API
+
+Boolean shapes can make generated clients less readable and can force future breaking changes if the
+API later needs more than two values.
+
+## LintDiff Equivalent
```
timotheeguerin, line 20: do we really need to mention this, this is all going away when we are done

packages/typespec-azure-core/src/rules/no-boolean.md
```diff
@@ -0,0 +1,60 @@
...
+values can be added without a breaking change.
+
+## Impact
+
+- **Area:** SDK, API
+
+Boolean shapes can make generated clients less readable and can force future breaking changes if the
+API later needs more than two values.
+
+## LintDiff Equivalent
```
catalinaperalta: Already discussed offline, but we decided to include the LintDiff equivalent to help facilitate ARM reviewer workflow

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.ts
msyyc: Renamed the rule to `no-boolean` across its files, registration, rulesets, tests, docs, and changeset in 86d254855.

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.md
msyyc: Removed the manually authored title and `Full name` block in 9faac5a6a so `tspd` owns that generated metadata.

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.ts
msyyc: Removed the HTTP projection in 98bfbf476. Model properties cover parameters and body properties, while the operation visitor now checks `operation.returnType` directly.

packages/typespec-azure-core/src/rules/use-enum-instead-of-boolean.ts
msyyc: Replaced the custom scalar-name helper with TypeKit's `tk.scalar.isBoolean(...)` in fda44d3d4.

msyyc:
@timotheeguerin Thanks for your review and I updated the PR with your suggestions. Pls take a review again when spare.

timotheeguerin (approved):

PR #5271, opened 2026-08-20
[Swagger Linter Migration] LatestVersionOfCommonTypesMustBeUsed

msyyc:
## Original Swagger linter

- Linter code: [LatestVersionOfCommonTypesMustBeUsed](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/latest-version-of-common-types-must-be-used.ts)
- Linter documentation: [latest-version-of-common-types-must-be-used.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/latest-version-of-common-types-must-be-used.md)
- ARM ruleset registration: https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/az-arm.ts

**Original checks:**
- Runs on every Swagger `$ref` value via `given: "$..['$ref']"` in the ARM ruleset.
- Ignores `null` refs and refs that do not match `/common-types/resource-management/v\d+/\w+.json#`.
- Extracts the referenced common-types version and file name from the `$ref` path.
- Uses `isLatestCommonTypesVersionForFile` and `LATEST_VERSION_BY_COMMON_TYPES_FILENAME` to decide whether that file is referenced through the latest known common-types version.
- Emits one warning at the `$ref` path when the referenced version differs from the latest map entry.

The Swagger validator is a Spectral function over emitted OpenAPI 2.0 documents. It does not understand TypeSpec symbols, ARM service namespaces, `@armCommonTypesVersion`, or version projections. Instead, it scans each emitted `$ref` string, filters to ARM common-types references under `/common-types/resource-management/v*/`, extracts `{version, file}`, and compares the extracted version with a hard-coded filename-to-latest-version map.

That string-level traversal means Swagger diagnostic locations are emitted `$ref` paths and raw diagnostic cardinality is tied to every repeated emitted occurrence. It also means the validator can be stale when the hard-coded map is stale. The lintdiff evidence records known stale-map false positives for valid v6 `networksecurityperimeter.json` and `managedidentitywithdelegation.json` references that the validator still maps to v5; the promoted rule intentionally does not copy those stale data defects.

**Source lintdiff rule id:** `LatestVersionOfCommonTypesMustBeUsed`

**Local rule name / canonical validator rule slug:** `latest-version-of-common-types-must-be-used`

**Official TypeSpec rule name:** `use-latest-version-of-common-types` (renamed during promotion to follow TypeSpec linter naming conventions)

**Source branch:** `feature/lintdiff-latest-version-common-types`

**Immutable source commit inspected:** `be310817b95a8c2fba64b7906e23624ec38efab0`

**Original source worktree:** `C:\dev\worktrees\lintdiff-latest-version-common-types-source`

**Source rule:** https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/src/rules/latest-version-of-common-types-must-be-used.ts

The original promotion recorded a clean source worktree with no uncommitted rule changes. The source rule was treated as done for the promotion and was not modified during promotion.

**Source-ref restoration (2026-09-15):** The branch above was missing when the handoff links were audited. It has been restored in `Azure/typespec-azure` at exactly `be310817b95a8c2fba64b7906e23624ec38efab0`, repairing the branch-based source, fixture, and migration links. The immutable source SHA and source contents are unchanged; this is not a source refresh or a switch to a newer implementation.

## Destination analysis

Selected package: `@azure-tools/typespec-azure-resource-manager`.

`@azure-tools/typespec-azure-core` was considered but rejected because this rule is ARM-specific and depends on ARM-only concepts and helpers. The validator metadata marks applicability as ARM with source `arm`; the fixture rule docs say it applies to Resource Manager; the implementation checks ARM provider namespaces, reads effective `@armCommonTypesVersion`, resolves ARM common-type records, and uses ARM helpers such as `getArmCommonTypesVersion`, `getArmCommonTypesVersions`, `isArmCommonType`, and `findArmCommonTypeRecord`. Moving this to core would require rewriting away ARM package dependencies and would violate the dependency direction that core must not depend on resource-manager.

The existing official ARM `arm-common-types-version` rule is related but not equivalent: it requires specifying a common-types version, while this promoted rule requires selecting the latest common-types version and catches latest-version APIs that still use legacy common-type symbols.

The rule is explicitly listed in `@azure-tools/typespec-azure-rulesets/resource-manager` but disabled by default for staged rollout. Existing Azure service specs and repo samples still select older ARM common-types versions, so enabling it immediately would convert those existing warnings into CI failures. Teams can opt in directly, and the default ruleset can be flipped after specs are updated or intentionally suppressed.

This describes the implementation at PR head `7c1764c3ce19e088b3c4f1f5f05eb79e32696054`, not the historical lintdiff implementation.

The promoted rule runs once at program root and visits TypeSpec services registered as ARM provider namespaces. It discovers the latest available ARM common-types version from `Azure.ResourceManager.CommonTypes.Versions`, compares the effective `@armCommonTypesVersion` on the service namespace or each version enum member, and reports older selections on the namespace or enum-member target. An undefined selection does not trigger the latest-selected usage check.

Selection alone is insufficient: a service selecting v6 can still use the deprecated `Azure.ResourceManager.LocationParameter`, whose records stop at v5, or `Legacy.ManagedServiceIdentityV4`, which resolves to v4 `managedidentity.json`. The resolver's closest-version/default fallback allows these older records, so the usage check catches cases a selection-only rule would miss.

When an API version selects the latest common-types version, the rule walks HTTP operation parameters, request bodies, and response bodies. It uses `resolveVersions` and the resolved service/dependency version map, checks availability through `getAddedOnVersions` and `getRemovedOnVersions`, and follows historical property and return types through `getTypeChangedFrom` and `getReturnTypeChangedFrom`. This direct version-history traversal replaced the immutable source's whole-service projection with versioning mutators; the promoted copy no longer calls `unsafe_mutateSubgraphWithNamespace`. Native regressions cover dependency versions, child namespaces, additions/removals, and historical property/return types.

HTTP metadata visibility (`resolveRequestVisibility`, `Visibility.Read`, and `createMetadataInfo().isPayloadProperty`) filters payload properties. The traversal preserves the distinction between array-item and record-value visibility and follows property source chains. The HTTP parameter metadata traversal also includes explicit body properties.

Only usages recognized by native `isArmCommonType` metadata enter common-type resolution. For each usage, `findArmCommonTypeRecord(program, usage.type, { service, version: apiVersion })` supplies the record's version, reference file, kind, and name. This uses the existing selected-version, closest-supported-version, and default-record fallback policy. The linter no longer constructs and parses OpenAPI reference strings. It reports an older resolved record when the API already selects the latest version, preserving the record's native name and file spelling in diagnostics.

Resolution failures are explicitly handled as rule-owned `resolution` messages through `context.reportDiagnostic`, retaining the resolver's explanation but using this rule's diagnostic code and the user's usage target. This promotion reporting adaptation preserves linter project-location filtering and rule suppression; the rule does not forward diagnostics directly to `program.reportDiagnostics` or change the emitter's independent validation behavior.

Outdated-record diagnostics are deduplicated per source target plus resolved record identity (file, version, kind, and name). Repeated traversal of the same target does not spam, while distinct operations using the same legacy common type each receive actionable diagnostics. This intentionally differs from the Swagger validator's emitted-`$ref` occurrence count.

The branch-based fixture links below refer to the restored immutable source ref. Native cases are in `packages/typespec-azure-resource-manager/test/rules/use-latest-version-of-common-types.test.ts`. Titles are copied exactly from that file. Tests use direct TypeSpec snippets and diagnostic assertions; lintdiff snapshots were not copied.

| --- | --- | --- |
| [older-common-types](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/older-common-types/main.tsp) | `it("emits diagnostic when a versioned ARM service selects an older namespace common-types version")` | Reports inherited v3 selection; explicitly asserts the API version enum-member target. |
| [latest-common-types](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/latest-common-types/main.tsp) | `it("does not emit when a versioned ARM service selects the latest namespace common-types version")` | Compliant latest-version namespace selection. |
| [version-override-older-common-types](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/version-override-older-common-types/main.tsp) | `it("emits diagnostic when a version enum member overrides the namespace back to an older common-types version")` | Reports a version-member override from latest to v3. |
| [version-override-latest-common-types](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/version-override-latest-common-types/main.tsp) | `it("does not emit when a version enum member overrides an older namespace to the latest common-types version")` | Compliant version-member override from v3 to latest. |
| [legacy-location-parameter](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/legacy-location-parameter/main.tsp) | `it("emits diagnostic when a latest-version service uses a legacy common parameter")` | Reports `LocationParameter` resolving to v5 `types.json` under a latest-version API. |
| [legacy-managed-identity](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/legacy-managed-identity/main.tsp) | `it("emits diagnostic when a latest-version service uses a legacy common model")` | Reports v4 `managedidentity.json`; explicitly asserts separate read and create/update operation targets. |
| [versioned-legacy-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/versioned-legacy-property/main.tsp) | `it("reports a versioned legacy property only for the projected API version that contains it")` | An added legacy property is reported only for the version that contains it. The historical test title is retained; implementation now uses direct version metadata rather than mutation. |
| [repeated-legacy-reference](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/repeated-legacy-reference/main.tsp) | `it("reports each operation that emits the same legacy common type reference")` | Asserts two diagnostics with distinct operation targets for the same legacy type. |
| [payload-visibility-excludes-legacy](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/payload-visibility-excludes-legacy/main.tsp) | `it("does not report legacy common types excluded by request or response payload visibility")` | Compliant request/response visibility exclusions and metadata inside record values. |
| N/A (review regression) | `it("evaluates dependency type changes using the selected common-types version")` | Uses the selected dependency version when checking a common model's historical types. |
| N/A (review regression) | `it("applies the service version to types in child namespaces")` | Applies service-version availability to an added legacy property in a child namespace. |
| N/A (review regression) | `it("reports a removed legacy property only for the API version that contains it")` | Honors `@removed` while checking each API version. |
| N/A (review regression) | `it("reports a legacy property type only before it changes")` | Honors `@typeChangedFrom` for a legacy property type. |
| N/A (review regression) | `it("reports a legacy return type only before it changes")` | Honors `@returnTypeChangedFrom` for an operation return type. |
| N/A (review regression) | `it("follows an explicit response body in a historical return type")` | Follows a legacy `@body` payload inside a historical response wrapper. |
| N/A (review regression) | `it.each(["Legacy.json", "nested/Legacy.json"])("checks resolved metadata independently of reference layout for %s")` | Input `Legacy.json`: preserves native filename casing and the literal `Legacy%Identity` name without URI decoding; asserts the diagnostic location. |
| N/A (review regression) | `it.each(["Legacy.json", "nested/Legacy.json"])("checks resolved metadata independently of reference layout for %s")` | Input `nested/Legacy.json`: reports the legacy record even though the old reference regex would reject the nested path; asserts the diagnostic location. |
| N/A (review regression) | `it("reports common-type resolution failures instead of silently skipping them")` | Uses the standard linter tester to assert a rule-owned resolution warning at `getIdentity`, with no direct program diagnostics. |
| N/A (review regression) | `it.each([false, true])("respects rule suppression for resolution failures: %s")` | Input `false`: full compiler execution reports the resolution failure under this rule's code at the operation usage. |
| N/A (review regression) | `it.each([false, true])("respects rule suppression for resolution failures: %s")` | Input `true`: suppressing this rule at the operation suppresses the resolution warning. |
| N/A (review regression) | `it("does not report resolution failures for operations declared in libraries")` | Imported-library operations are excluded by the linter's project-location filtering. |

## Migration evidence

The immutable lintdiff source's migration report is available at: https://github.com/Azure/typespec-azure/blob/feature/lintdiff-latest-version-common-types/packages/typespec-lintdiff/test/fixtures/LatestVersionOfCommonTypesMustBeUsed/migration.md

That report contains the historical focused fixture tests, pinned real-service corpus comparison, population adjustments, known stale validator-map behavior, diagnostic-count differences, compile failures, and remaining uncertainty. Its projection-based implementation description and corpus results describe the pinned lintdiff source, not a new corpus run of the later native promotion adaptations. The source and report were not modified by the source-ref restoration or these promotion follow-ups.

### Historical promotion record

The original PR recorded these commands and checks. They are retained as historical evidence, not presented as freshly rerun checks of the latest follow-up:

- `pnpm --filter @azure-tools/typespec-azure-resource-manager exec vitest run test/rules/use-latest-version-of-common-types.test.ts`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager build`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager lint`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager regen-docs`
- `pnpm --filter @azure-tools/typespec-azure-rulesets build`
- `pnpm --filter @azure-tools/typespec-azure-rulesets test`
- `pnpm --filter @azure-tools/typespec-azure-playground-website exec vitest run test/validate-samples.test.ts`
- `pnpm --filter @azure-tools/typespec-samples exec vitest run test/samples.test.ts -t "legacy/legacy-operations"`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager test`
- Focused code review of promotion diff: no blocking findings reported at that stage.
- `git diff --check`

### Later native-metadata and review follow-up (2026-09-14)

- Native metadata change `3e13a32fa`: focused latest-version rule tests plus ARM common-type registration tests passed, 37 tests total, with `--testTimeout 30000`. An earlier attempt hit the default 10-second timeout in a registration test; the timeout-adjusted rerun passed. ARM package build and repository formatting/lint also passed.
- Two Balanced Copilot review rounds completed on the subsequent PR heads. The canonical ARM inventory omission was fixed in `da46ae1fb`; documentation formatting, repository formatting/lint, and diff checks passed. The second-round explicit-body traversal suggestion was rejected because the body property is already included in `parameters.properties`. Both processed threads were answered and resolved.
- No lintdiff corpus rerun was performed during these promotion follow-ups. Historical corpus evidence remains in the pinned source migration report.

### Handoff refresh (2026-09-15)

Restored the missing source branch at the unchanged immutable SHA, verified the source/migration links and original fixture paths, and matched the table against native test titles. That update changed only the PR description and restored the source ref; no production code or tests changed and no build or corpus rerun was claimed.

### Linter diagnostic reporting follow-up (2026-09-18)

Commit `7c1764c3ce19e088b3c4f1f5f05eb79e32696054` changes resolution-error reporting to the linter context, with user-usage targets and rule-owned diagnostics. Existing version selection, fallback, and legacy-usage traversal are unchanged.

- ARM package build passed.
- Focused native rule suite passed: 21 tests, including suppression on/off and imported-library filtering, using `--testTimeout 30000`.
- ARM documentation regeneration completed with missing-documentation warnings for existing unrelated declarations/rules; no generated tracked changes resulted.
- Repository `pnpm format`, `pnpm lint`, and `git diff --check` passed.
- Initial validation encountered stale built dependency output and then missing compiler dependencies after updating to the PR's current base. Dependency restoration's unrelated Python prepare hook failed with a TLS handshake error; `pnpm install --frozen-lockfile --ignore-scripts` restored the required JS dependencies, and dependency builds then completed. A diagnostic message-format type error and an import-order error in the new library test were corrected before the successful final build/test runs.
- No lintdiff corpus was run; no source repair, ruleset change, or reduction to selection-only checking was made.

## Validation blocker

The original promotion recorded that `pnpm validate:pr` hung after printing only `Branch is up to date` for more than 13 minutes and was stopped. The narrower promotion validations above completed successfully. This is a retained historical limitation, not a new failure or a claim that the command was rerun during the follow-ups.

## Promotion sync policy

If review finds a semantic gap after promotion, this PR should be blocked until the user explicitly reopens lintdiff rule repair. The done lintdiff source rule was treated as immutable during promotion and was not edited here. Restoring its branch at the existing pinned SHA does not reopen repair or change source provenance.


packages/typespec-azure-resource-manager/src/rules/latest-version-of-common-types-must-be-used.ts
```diff
@@ -0,0 +1,425 @@
...
+import {
+  getArmCommonTypeOpenAPIRef,
+  getArmCommonTypesVersion,
+  getArmCommonTypesVersions,
+  isArmCommonType,
+} from "../common-types.js";
+import { getArmProviderNamespace } from "../namespace.js";
+
+export const latestVersionOfCommonTypesMustBeUsedRule = createRule({
+  name: "latest-version-of-common-types-must-be-used",
```
catalinaperalta: For all the new rules we're adding we should follow the naming convention guide added by the typespec team. See here: https://typespec.io/docs/extending-typespec/linters/#naming-convention

catalinaperalta: You could update the skill to take the conventions into account when migrating the rules

catalinaperalta: In this case to make it shorter and follow convention maybe it would be something like: 
```suggestion
  name: "use-latest-version-of-common-types",
```

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,108 @@
+ARM services should use the latest ARM common-types version available in
```
catalinaperalta, line 9: Nit we need a title section, you can refer to this doc as an example: https://github.com/msyyc/typespec-azure/blob/440a51db101acdb7099a9e443c2dd02d6c5925b8/packages/typespec-azure-resource-manager/src/rules/use-api-version.md

packages/typespec-azure-resource-manager/src/rules/latest-version-of-common-types-must-be-used.md
```diff
@@ -0,0 +1,108 @@
+ARM services should use the latest ARM common-types version available in
+`Azure.ResourceManager.CommonTypes.Versions`. This keeps generated Swagger and
```
catalinaperalta: Is the common types update truly in the interest of keeping the swagger up to date? Or is it just generally to keep the project up to date? Which would imply keeping the typespec up to date as well.

packages/typespec-azure-resource-manager/src/rules/latest-version-of-common-types-must-be-used.md
```diff
@@ -0,0 +1,108 @@
...
+service version. When the selected version is current, it also checks common
+types reachable from HTTP operation parameters and payloads so older legacy
+symbols are not emitted through an otherwise current API version.
+
+## Impact
+
+- **Area:** API, SDK
+
+Older ARM common-types versions can emit stale shared schemas or parameters into
+Swagger even when newer definitions are available.
```
catalinaperalta: I dont think we should focus the docs on swagger, instead we need to focus on TypeSpec and SDKs

packages/typespec-azure-resource-manager/src/rules/latest-version-of-common-types-must-be-used.md
msyyc: Let me update the docs to focus on Typespec instead of Swagger.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+## Correct
+
+```tsp
+@armProviderNamespace
+@service(#{ title: "Contoso" })
+@versioned(Versions)
+@armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
```
catalinaperalta, line 46: This line might become outdated if a new version of common types is added. So maybe we should add a note in the docs to say something like: `As of August 2026 'v6' was the latest version of common types, but newer versions may exist so check for the latest version in the Azure.ResourceManager.CommonTypes.Versions enum.`

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,121 @@
...
+  name: string;
+}
+
+model WidgetProperties {
+  description?: string;
+}
+
+@route("/identity")
+@get
+op getIdentity(): Widget;
```
catalinaperalta: This example seems much longer than the `incorrect` example, isnt there an equivalent ManagedServiceIdentity you can return from the latest common types?

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,425 @@
...
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/use-latest-version-of-common-types",
+  messages: {
+    default: paramMessage`Use the latest ARM common-types version '${"latestVersion"}' instead of '${"currentVersion"}'.`,
+    reference: paramMessage`This API version already selects the latest ARM common-types version '${"latestVersion"}', but the common-type ${"referenceKind"} '${"referenceName"}' resolves to '${"fileName"}' version '${"currentVersion"}'. Replace the TypeSpec usage that produces this legacy reference with a common type supported in '${"latestVersion"}'.`,
+  },
+  create(context) {
+    return {
+      root: (program) => {
+        const latestVersion = getLatestArmCommonTypesVersion(program);
+        if (latestVersion === undefined) {
```
catalinaperalta, line 54: could this happen?

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,425 @@
...
+          }
+
+          const versioning = getVersioningMutators(program, service.type);
+          if (versioning?.kind === "versioned") {
+            for (const snapshot of versioning.snapshots) {
+              const currentVersion =
+                getArmCommonTypesVersion(program, snapshot.version.enumMember) ??
+                getArmCommonTypesVersion(program, service.type);
+              if (
+                reportIfOutdated(
```
catalinaperalta: nit: can we simplify this method name to be `isOutdated`

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,425 @@
...
+                  context,
+                  snapshot.version.enumMember,
+                  currentVersion,
+                  latestVersion,
+                ) ||
+                currentVersion !== latestVersion
```
catalinaperalta: I believe this `|| currentVersion !== latestVersion` is already considered in the `reportIfOutdated method since any case where currentVersion is not undefined or equal to the latest version produces a diagnostic

catalinaperalta (commented):
looks good, just some comments to improve clarity

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,425 @@
...
+              ) {
+                continue;
+              }
+
+              const projected = unsafe_mutateSubgraphWithNamespace(
```
catalinaperalta, line 87: Just wondering if it's ok to use this function @markcowl or there's a better function to use?

msyyc: Context: we need this projection step because the rule walks HTTP operations and payload types for each effective API version. Versioning decorators like `@added`, `@removed`, and `@typeChangedFrom` don't rewrite the original service graph, so walking the unprojected namespace could report legacy common types for the wrong API version.

`getVersioningMutators(...)` gives us the per-version mutator, and `unsafe_mutateSubgraphWithNamespace(...)` applies it to produce the same kind of version-specific namespace used by the OpenAPI3 and Autorest emitters before generating per-version output. I don't see a stable replacement API in this repo right now, so keeping this matches the existing emitter pattern.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
msyyc: Good point. Added a note that `v6` is latest as of August 2026 and to check `Azure.ResourceManager.CommonTypes.Versions` for newer versions.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
msyyc: Yes, there is an equivalent latest common type. Shortened the example to return `Azure.ResourceManager.CommonTypes.ManagedServiceIdentity` directly.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
msyyc: Only if common-types metadata is not initialized or is empty. I renamed the helper to `tryGetLatestArmCommonTypesVersion` to make that defensive path explicit.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
msyyc: Refactored this into a pure `isOutdated(...)` helper and kept reporting in a separate `reportOutdatedSelection(...)` function.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
msyyc: Good catch. Removed the redundant check and now handle the `undefined` case explicitly after reporting outdated versions.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
timotheeguerin: yeah we definitely do not want to start mutating everything here this is extremely costly

msyyc: @catalinaperalta / @timotheeguerin  Addressed in f15a77bfa. I removed `unsafe_mutateSubgraphWithNamespace` and replaced the per-version whole-program projection with direct traversal using resolved version metadata, including dependency-specific versions, availability, and historical property/return types. The existing cases remain covered, and I added regressions for added/removed properties, child namespaces, dependency type changes, and `@typeChangedFrom`/`@returnTypeChangedFrom`.

msyyc:
Follow-up from the review of #5451: https://github.com/Azure/typespec-azure/pull/5451#discussion_r3989121881

The review principle is that a native TypeSpec linter should use TypeSpec types and metadata rather than infer their meaning from emitted OpenAPI references. In #5451, we removed reference-string matching that allowed a custom TypeSpec model to pass merely because its external reference pointed to the standard ErrorResponse.

I checked this PR at `aeb0eabae3876d757859e071e5e79ddd2f6d88b0`. It does **not** have that same custom-model acceptance issue: usages are gated through native `isArmCommonType` metadata. However, there is a related, unnecessary dependency on OpenAPI reference formatting:

- [The version check](https://github.com/msyyc/typespec-azure/blob/aeb0eabae3876d757859e071e5e79ddd2f6d88b0/packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts#L498-L502) calls `getArmCommonTypeOpenAPIRef(...)`, then passes the result to `parseCommonTypesReference(...)`.
- [The parser](https://github.com/msyyc/typespec-azure/blob/aeb0eabae3876d757859e071e5e79ddd2f6d88b0/packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts#L532-L541) extracts the version, reference file, kind, and name using a regex tied to a particular Swagger reference layout.
- If parsing returns `undefined`, the version check continues without reporting the outdated usage. This makes the check depend on reference formatting in addition to native version metadata. I have not established an affected current service; this is an architectural concern, not a demonstrated production failure.

**Suggested change:** use `findArmCommonTypeRecord(program, usage.type, { service, version: apiVersion })` directly and consume the record's `version`, `referenceFile`, `kind`, and `name`. Those are the same fields the OpenAPI helper uses to construct its reference, so the string construction/parsing round trip should not be needed. Preserve the existing version-selection/fallback policy and handle returned diagnostics explicitly.

Regression coverage should keep the current version-selection and diagnostic-target behavior, and verify that checking a resolved common-type record does not depend on the generated reference path matching the old regex. This suggestion is specific to the linter; the OpenAPI reference helper remains appropriate for emitters.

msyyc:
Follow-up audit against the updated `develop-lintdiff-rule` and `lintdiff-rule-promote` skills, including the [metadata-over-reference boundary](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/develop-lintdiff-rule/SKILL.md#native-typespec-implementation-boundary).

**The reference-parsing concern in [the earlier comment](https://github.com/Azure/typespec-azure/pull/5271#issuecomment-5661318416) is fixed at `473a14a41ccc96d25994a646d7f20ba395a3f2cd`.** The [current implementation](https://github.com/Azure/typespec-azure/blob/473a14a41ccc96d25994a646d7f20ba395a3f2cd/packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts#L498-L503) consumes `findArmCommonTypeRecord(...)` and reports returned diagnostics. Native tests also cover reference-layout independence and incompatible-version diagnostics. I am not asking to repeat that fix.

The remaining findings concern promotion evidence:

- Replace the PR body's fixture-mapping bullets with the required `Original lintdiff fixture | Native vitest case | Coverage note` table, using exact `it("...")` titles, including the new metadata regressions.
- Refresh the source-rule, fixture, and migration links. The body points to `feature/lintdiff-latest-version-common-types`, which was not resolvable during this audit. The recorded source commit remains available, but the branch-based reader links are stale. Restore an appropriate source ref or link a resolvable source branch while retaining the immutable source SHA and explaining any provenance change.

These are requirements of the [promotion PR handoff](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#10-review-commit-push-and-create-a-draft-pr), not newly demonstrated production failures. This was a static review; it does not assert that historical validation was skipped.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,106 @@
...
+
+@route("/identity")
+@get
+op getIdentity(): Azure.ResourceManager.CommonTypes.ManagedServiceIdentity;
+```
+
+## LintDiff Equivalent
+
+This rule corresponds to the LintDiff rule
+[LatestVersionOfCommonTypesMustBeUsed](https://github.com/Azure/azure-openapi-validator/blob/main/docs/latest-version-of-common-types-must-be-used.md).
```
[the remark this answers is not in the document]
msyyc: Fixed in da46ae1fb423730d09dbf7b278992ecce0d8ba06. Added `use-latest-version-of-common-types` to the canonical ARM rules inventory with its `LatestVersionOfCommonTypesMustBeUsed` mapping and API/SDK impact. This is documentation-only; the rule behavior is unchanged.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,532 @@
...
+      addPropertyUsages(parameter.property, httpOperation.operation, requestContext);
+    }
+    if (httpOperation.parameters.body) {
+      const body = httpOperation.parameters.body;
+      visitType(body.type, httpOperation.operation, {
+        visibility: requestContext.visibility,
+        inExplicitBody:
+          body.bodyKind === "single" && body.isExplicit && body.containsMetadataAnnotations,
+      });
+    }
```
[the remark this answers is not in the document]
msyyc: The explicit body property is already covered by the preceding `httpOperation.parameters.properties` loop. TypeSpec HTTP returns its payload metadata as `parameters.properties`, including the `body`/`bodyRoot` entry whose `property` is also assigned to `body.property`. Both this rule and the pinned lintdiff source call `addPropertyUsages(parameter.property, ...)` on that entry before visiting `body.type`, so its native ARM common-parameter metadata and `sourceProperty` chain are checked. Calling it again on `body.property` would duplicate traversal. AutoRest also supports arbitrary external references, but those intentionally do not qualify as ARM common types without native metadata. No code change is needed for this finding.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,532 @@
...
+    const requestContext = {
+      visibility: resolveRequestVisibility(program, httpOperation.operation, httpOperation.verb),
+      inExplicitBody: false,
+    };
+    for (const parameter of httpOperation.parameters.properties) {
+      addPropertyUsages(parameter.property, httpOperation.operation, requestContext);
+    }
+    if (httpOperation.parameters.body) {
+      const body = httpOperation.parameters.body;
+      visitType(body.type, httpOperation.operation, {
```
timotheeguerin: **[P2] Historical body types are evaluated using unprojected HTTP metadata.**

`getHttpService` is resolved once on the original service, so `body.type` is the current type, not necessarily the type for the API version being checked. The same issue affects the response-body traversal below.

With two API versions both selecting common-types v6:

- A request parameter `@body @typeChangedFrom(Versions.v2, string) body: Azure.ResourceManager.Legacy.ManagedServiceIdentityV4` produces two warnings instead of one. The property traversal correctly sees `string` in v1, but this second traversal visits today's legacy body type anyway.
- A response wrapper property `@body @typeChangedFrom(Versions.v2, Azure.ResourceManager.Legacy.ManagedServiceIdentityV4) body: string` produces no warning for v1. The operation's return model is unchanged, so the response path visits the current `string` body and never reads the property's history.

These are concrete false positives/negatives from mixing historical type metadata with current HTTP metadata. If the usage check remains, both body paths need to use the effective versioned property type consistently; preferably remove this reconstruction from the version-selection rule as described in the review summary.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,532 @@
...
+        addUsage(type, target);
+        if (type.indexer) {
+          visitType(type.indexer.value, target, {
+            ...payloadContext,
+            visibility: isArrayModelType(type)
+              ? payloadContext.visibility | Visibility.Item
+              : payloadContext.visibility,
+          });
+        }
+        for (const property of walkPropertiesInherited(type)) {
```
timotheeguerin: **[P2] Walking inherited properties does not check the common-type base model.**

For a service selecting v6, this produces no warning:

```tsp
model Identity extends Azure.ResourceManager.Legacy.ManagedServiceIdentityV4 {
  description?: string;
}

@route("/identity")
@get
op read(): Identity;
```

`addUsage` checks `Identity`, which is not itself registered as an ARM common type. `walkPropertiesInherited` then visits the base's properties but never the base model itself, so the v4 common-type metadata is missed even though AutoRest retains the base reference in `allOf`.

If this traversal remains, it needs to account for the nominal base-model reference, not just its inherited properties. A source-level legacy-reference check already has a natural place to handle this through `model.baseModel`.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,532 @@
...
+  latestVersion: string,
+): void {
+  const reported = new Map<ModelProperty | Operation, Set<string>>();
+
+  for (const usage of usages) {
+    const [record, diagnostics] = findArmCommonTypeRecord(context.program, usage.type, {
+      service,
+      version: apiVersion,
+    });
+    context.program.reportDiagnostics(diagnostics);
```
timotheeguerin: Please use `context.reportDiagnostic` rather than reporting directly to the program. This bypasses the linter's project-location filtering and reports under a different code, so suppressing this rule does not suppress the warning. Resolution validation should live outside the rule, or become a rule-owned diagnostic.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
```diff
@@ -0,0 +1,532 @@
...
+  target: ModelProperty | Operation;
+  type: Model | ModelProperty | Enum | Union;
+}
+
+interface PayloadContext {
+  visibility: Visibility;
+  inExplicitBody: boolean;
+}
+
+function collectCommonTypeUsages(
```
timotheeguerin, line 179: This seems much more complex than checking the selected common-types version. Why do we need to go through all this ?

msyyc: The extra check distinguishes the selected version from the version a used common-type symbol actually resolves to. For example, an API selecting v6 can still use the deprecated `Azure.ResourceManager.LocationParameter`, whose records stop at v5, or `Legacy.ManagedServiceIdentityV4`, which resolves to v4 `managedidentity.json`. The common-type resolver intentionally falls back to an older supported/default record, so checking only the namespace/version selection misses both cases. These are covered by the native legacy-parameter and legacy-model cases.

The traversal finds those reachable usages; visibility checks exclude properties absent from payloads, and version-history checks avoid reporting added/removed or changed types in the wrong API version. Whole-service mutation was already removed in favor of direct version metadata.

I clarified this distinction in the rule docs in 7c1764c3ce19e088b3c4f1f5f05eb79e32696054 and kept the existing coverage. A selection-only rule or a separate legacy-usage rule is a possible scope change, but it would not be behaviorally equivalent to the current migration, so I have not silently removed that coverage.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
msyyc: Fixed in 7c1764c3ce19e088b3c4f1f5f05eb79e32696054. Resolution failures now use a rule-owned `resolution` message through `context.reportDiagnostic`, targeted at the user's usage rather than the common-type declaration. They retain the resolver's explanation but use this rule's code and normal project filtering. Replaced the direct-program assertion with the standard linter tester and added full-compiler suppression coverage (both suppressed and unsuppressed) plus an imported-library service case.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
timotheeguerin: Yeah but we already have different warnings for those deprecated legacy types this just make this rule way more complex for a duplicate warning no?

markcowl (changes requested):
Combine with this rule: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-common-types-version.ts

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,113 @@
+ARM services should use the latest ARM common-types version available in
+`Azure.ResourceManager.CommonTypes.Versions`. This keeps TypeSpec services,
+generated SDKs, and Azure tooling aligned with the current ARM common schemas.
+
+The rule checks the effective `@armCommonTypesVersion` on each ARM service or
```
markcowl, line 5: Paragraphs 2 and 3 are implementationd etails that should not be in the documentation

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,113 @@
...
+namespace Microsoft.Contoso;
+
+enum Versions {
+  @useDependency(Azure.ResourceManager.CommonTypes.Versions.v6)
+  v2024_01_01: "2024-01-01",
+}
+
+@route("/identity")
+@get
+op getIdentity(): Azure.ResourceManager.CommonTypes.ManagedServiceIdentity;
```
markcowl, line 101: Always use standard arm templates in correct examples

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
```diff
@@ -0,0 +1,113 @@
...
+namespace Microsoft.Contoso;
+
+enum Versions {
+  @useDependency(Azure.ResourceManager.CommonTypes.Versions.v6)
+  v2024_01_01: "2024-01-01",
+}
+
+@route("/identity")
+@get
+op getIdentity(): Azure.ResourceManager.Legacy.ManagedServiceIdentityV4;
```
markcowl, line 79: Use a standard arm operation template inside an interface

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
markcowl: Agreed, this rule is doing too much.  It should just look for the effective common-types version

markcowl (changes requested):

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.ts
msyyc: Consolidated in 1722f99264f78dea42a5f084ca936a7b77161b47. The existing `arm-common-types-version` rule now preserves its missing-configuration warning and checks each API version's effective selection, honoring enum-member overrides and namespace defaults. Removed the separate rule and all HTTP/payload, version-history, legacy-reference, and resolution-error traversal.

This is an intentional scope reduction rather than full equivalence with the old reference checker; the PR mapping marks removed legacy-usage coverage explicitly. The existing rule stays enabled in the resource-manager ruleset, so latest-version warnings are now active there. That behavior change and the existing rule's suppression path are documented in the changelog and PR description.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
msyyc: Addressed in 1722f99264f78dea42a5f084ca936a7b77161b47. Removed the standalone `@get` legacy-identity examples along with the out-of-scope usage check. The consolidated rule's correct example now uses `get is ArmResourceRead<Widget>` inside an `@armResourceOperations` interface.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
msyyc: Addressed in 1722f99264f78dea42a5f084ca936a7b77161b47. Removed the separate rule's documentation and consolidated user-facing guidance into `arm-common-types-version.md`. The docs now explain explicit/latest selections, namespace defaults, API-version overrides, impact, and suppression without the metadata-resolution or diagnostic-routing implementation details.

packages/typespec-azure-resource-manager/src/rules/use-latest-version-of-common-types.md
msyyc: Addressed in 1722f99264f78dea42a5f084ca936a7b77161b47. The consolidated rule documentation uses a `TrackedResource<WidgetProperties>` with `ResourceNameParameter<Widget>` and `ArmResourceRead<Widget>` inside an `@armResourceOperations` interface. The example compiles with the consolidated rule enabled and no diagnostics.

msyyc:
CI follow-up in db29d80c1cb576b30c00ae7da944361d1c9a8272:

- Updated 21 current ARM samples to select common-types v6 and regenerated their OpenAPI snapshots. Snapshot changes are exclusively v5-to-v6 common-type references; legacy references without v6 support remain unchanged.
- Added scoped, justified suppressions to the 10 grandfathered samples and the older-version private-links compatibility example. Their generated schemas remain unchanged.
- All 40 sample tests pass without snapshot recording, and all 21 playground sample checks pass. Repository formatting, lint, and diff checks pass.
- The consolidated `arm-common-types-version` rule remains enabled. No weakening of its checks or test assertions was introduced.

External Integration is a separate, still-unresolved rollout dependency: existing azure-rest-api-specs services select v3-v5 and now fail under warnings-as-errors (see https://github.com/Azure/typespec-azure/actions/runs/35491909140). The user explicitly chose to fix repository samples only and defer external-spec compatibility changes. This commit does not claim to resolve that check; downstream compatibility updates or scoped suppressions must land before it can pass.

.chronus/changes/promote-latest-version-common-types-2026-08-20-14-00-00.md
```diff
@@ -0,0 +1,15 @@
...
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Extend `arm-common-types-version` to warn when an ARM service or API version selects an older common-types version, while preserving the existing warning for missing explicit version configuration.
+
+Latest-version warnings are now active wherever this existing rule is enabled, including the resource-manager ruleset. Update `@armCommonTypesVersion` to the latest version in `Azure.ResourceManager.CommonTypes.Versions`, or suppress `@azure-tools/typespec-azure-resource-manager/arm-common-types-version` when an older version is required for compatibility.
+
+The rule checks effective version selections only; it does not inspect legacy type usages or emitted OpenAPI references.
```
msyyc: The rule documentation already keeps the emitted-reference comparison in `LintDiff Equivalent`; please keep this user-facing release note focused on native behavior too. Retain "The rule checks effective version selections only" and remove the comparison with legacy type usages and emitted OpenAPI references. The migration distinction can stay in the dedicated migration section and PR description.

PR #5273, opened 2026-08-20
[Swagger Linter Migration] ValidQueryParametersForPointOperations

msyyc:
## Original Swagger linters

This promoted TypeSpec rule intentionally covers **two** Swagger validator rules:

- linter code: [ValidQueryParametersForPointOperations](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/valid-query-parameters-for-point-operations.ts)
- linter doc: [valid-query-parameters-for-point-operations.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/valid-query-parameters-for-point-operations.md)
- linter code: [ParametersInPointGet](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/parameters-in-point-get.ts)
- linter doc: [parameters-in-point-get.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/parameters-in-point-get.md)

`ValidQueryParametersForPointOperations` implements ARM RPC guideline `RPC-Uri-V1-13`.
`ParametersInPointGet` implements ARM RPC guideline `RPC-Get-V1-08`.

**Official TypeSpec rule name:** `point-operation-invalid-query-parameter` (renamed during promotion to follow TypeSpec linter naming conventions).

`ValidQueryParametersForPointOperations` performs these checks:

- [x] Traverses resolved Swagger `paths` and `x-ms-paths` entries.
- [x] Classifies a URI as a point path only when its final provider-qualified portion contains `/providers/{namespace}` followed by one or more resource-type/resource-name pairs; resource names may be `{parameters}` or `default`.
- [x] Checks GET, PUT, PATCH, and DELETE operations.
- [x] Reads each operation's `parameters` array and ignores non-query parameters.
- [x] Allows `api-version` and reports every other query parameter independently.
- [x] Excludes collection paths, unmatched trailing path segments, and providerless paths.

`ParametersInPointGet` is the older production GET-only form of the same policy. It performs these checks:

- [x] Traverses Swagger `paths` entries.
- [x] Classifies a URI as a point resource path with `getResourcesPathHierarchyBasedOnResourceType`.
- [x] Checks only GET operations.
- [x] Allows GET query parameters named exactly `api-version`.
- [x] Reports every other GET query parameter at the operation `parameters` array.

## How the Swagger linters work

`ValidQueryParametersForPointOperations` is invoked over the resolved maps selected by `$[paths,'x-ms-paths']`. The function loops each URI, applies the shared `isPointOperation` regex to the portion beginning at the last `/providers/`, then loops GET/PUT/PATCH/DELETE and filters each resolved parameter array for query parameters other than `api-version`. Each rejected emitted parameter produces one error located at that operation's `parameters` array.

`ParametersInPointGet` uses a narrower production path: it loops the Swagger `paths` object, classifies each URI with the ARM resource hierarchy helper, checks only the `get` operation, and reports query parameters other than `api-version`. This is a subset of the broader point-operation rule, so it should map to the same TypeSpec implementation rather than to a second production TypeSpec rule.

The broader Swagger rule is `stagingOnly`, so normal production AutoRest validation disables it and reports 0 projects. The checked-in migration investigation instead ran the actual Spectral staging rule. Its initial 64-project same-corpus population became 62 projects after restricting both sides to the 462 projects whose TypeSpec compiled, with all 62 projects overlapping. The production `ParametersInPointGet` row separately shows 40 Swagger projects, all overlapping the same TypeSpec rule.

Raw diagnostic counts are intentionally not one-to-one: Swagger reports emitted operation-parameter occurrences, while a shared TypeSpec parameter declaration can instantiate into several operations and projected versions. The final staging comparison is 321 Swagger diagnostics versus 724 TypeSpec diagnostics, but 62 versus 62 projects with no one-sided projects. The GET-only production comparison is 189 Swagger diagnostics versus the same raw 724 TypeSpec diagnostics, with 40 overlapping Swagger projects and 22 TypeSpec-only projects explained by PUT/PATCH/DELETE coverage from the broader staging rule. The validator also accepts `x-ms-paths`, while a clean TypeSpec `@sharedRoute` equivalent was not included because its emitted `?_overload=...` disambiguator does not reproduce the upstream rule.

The user-marked done source is [`tsp-lintdiff-local-linter/valid-query-parameters-for-point-operations`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/src/rules/valid-query-parameters-for-point-operations.ts) on branch `feature/lintdiff-migration-new`, captured at intake commit `e9d5541c2ef3e354ce1bf6ab3332dfb87620a8d2`. The source worktree was clean at intake and had no uncommitted rule changes. This promotion did not modify the lintdiff source, fixtures, snapshots, manifests, or documentation.

## Destination analysis

This rule belongs in `@azure-tools/typespec-azure-resource-manager` because validator metadata declares `applicability: ARM`, `sources: ["arm"]`, and the fixture explicitly selects the resource-manager ruleset. Its semantics depend on ARM provider-qualified resource paths and ARM RPC guidance. Azure Core was technically possible because the implementation needs only compiler and HTTP APIs, but placing ARM-only URI policy in the shared data-plane package would expose the wrong dependency and ruleset surface. No equivalent official ARM or Core rule exists.

The promoted rule visits TypeSpec operations and resolves each with `getHttpOperation`. It filters to GET, PUT, PATCH, and DELETE, then applies the same provider/resource-pair path regex as the Swagger helper to the portion after the last `/providers/`. For eligible point paths, it examines resolved HTTP parameters, ignores non-query parameters and case-insensitive `api-version`, and reports each additional query parameter on its authorable TypeSpec parameter declaration.

This one TypeSpec rule therefore covers both Swagger inputs: it covers `ParametersInPointGet` when the verb is GET, and it covers `ValidQueryParametersForPointOperations` across GET/PUT/PATCH/DELETE. The implementation intentionally does not use ARM operation-kind metadata because that would misclassify list-shaped reads and providerless resource-group DELETE operations that the Swagger path matcher excludes. It also performs no promotion-only deduplication, preserving the done lintdiff rule's operation traversal and source-target behavior. No version projection or semantic behavior was added during promotion.

| --- | --- | --- |
| [`extra-query-param`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/extra-query-param/main.tsp) | `it("emits for extra query parameters on top-level GET, PUT, PATCH, and DELETE operations")` | Covers top-level point-resource GET, PUT, PATCH, and DELETE operations, each with one disallowed query parameter. |
| [`nested-extra-query-param`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/nested-extra-query-param/main.tsp) | `it("emits for extra query parameters on nested point GET and PUT operations")` | Covers nested point-resource GET and PUT operations with disallowed query parameters. |
| [`multiple-query-params`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/multiple-query-params/main.tsp) | `it("emits one diagnostic for each extra query parameter")` | Preserves one diagnostic per extra query parameter on the same point operation. |
| [`legacy-action-point-get`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/legacy-action-point-get/main.tsp) | `it("classifies a GET operation by point-path shape regardless of its authoring template")` | Reduces the legacy template to its semantic requirement and verifies that point-path shape, not authoring template, controls eligibility. |
| [`api-version-only`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/api-version-only/main.tsp) | `it("allows point operations whose only query parameter is api-version")` | Covers compliant point operations where the only query parameter is `api-version`. |
| [`list-operation`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/list-operation/main.tsp) | `it("allows query parameters on collection operations")` | Covers compliant collection paths, where non-`api-version` query parameters are allowed. |
| [`list-shaped-read`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/list-shaped-read/main.tsp) | `it("allows query parameters on list-shaped read paths")` | Covers the compliant list-shaped read regression for a path with an unmatched trailing segment. |
| [`providerless-delete`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/providerless-delete/main.tsp) | `it("allows query parameters on providerless DELETE operations")` | Covers the compliant providerless resource-group-style DELETE regression. |

No validator snapshots or lintdiff corpus artifacts were copied.

## Migration evidence

The focused fixtures, real-service comparison, full-corpus counts, former TypeSpec-only projects, projection fix, diagnostic-cardinality analysis, compile-success population, and remaining uncertainty for `ValidQueryParametersForPointOperations` are documented in [`migration.md`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ValidQueryParametersForPointOperations/migration.md).

The companion `ParametersInPointGet` investigation confirms the same TypeSpec rule covers every assessable production GET-only Swagger project and explains the 22 TypeSpec-only projects as broader PUT/PATCH/DELETE coverage, not as false positives or a need for another TypeSpec rule: https://github.com/Azure/typespec-azure/pull/5210

## Promotion sync policy

If review identifies a semantic gap, this PR should remain blocked until the user explicitly reopens the lintdiff repair workflow. The immutable source rule must be repaired and revalidated first; semantic changes should not be made only in this promotion PR.


msyyc:
Finding from the `ParametersInPointGet` migration check: we do **not** need a separate TypeSpec linter for `ParametersInPointGet`.

`ParametersInPointGet` is the older production Swagger rule for point **GET** operations only. The migrated TypeSpec rule in this PR, `valid-query-parameters-for-point-operations`, is the right shared implementation because it covers that GET behavior and also covers the broader staging-only Swagger rule `ValidQueryParametersForPointOperations` for point GET/PUT/PATCH/DELETE.

So the expected follow-up is documentation/evidence only: map `ParametersInPointGet` to this broader rule and explain the apparent TypeSpec-only projects as PUT/PATCH/DELETE coverage from the broader rule, not as false positives or a need for another rule.

catalinaperalta:
> while a clean TypeSpec @sharedRoute equivalent was not included because its emitted ?_overload=... disambiguator does not reproduce the upstream rule.

Is this saying that we skipped the check on the sharedRoute entries?

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.md
```diff
@@ -0,0 +1,69 @@
...
+}
+```
+
+#### ✅ Correct
+
+```tsp
+@route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Contoso/widgets/{widgetName}")
+interface Widgets {
+  @get
+  get(
```
catalinaperalta: I'm wondering if these examples should use an ARM operation template instead of boiler plate tsp operations.

packages/typespec-azure-rulesets/src/rulesets/resource-manager.ts
```diff
@@ -102,6 +102,7 @@ export default {
     "@azure-tools/typespec-azure-resource-manager/retry-after": false, // Disable https://github.com/Azure/typespec-azure/issues/3351
     "@azure-tools/typespec-azure-resource-manager/secret-prop": true,
     "@azure-tools/typespec-azure-resource-manager/unsupported-type": true,
+    "@azure-tools/typespec-azure-resource-manager/point-operation-invalid-query-parameter": true,
```
catalinaperalta: Set to false, unless you're ready to migrate this rule to the specs repo

catalinaperalta (approved):
overall lgtm and seems to match the original rule, we should double check though if we should include the sharedRoute instances

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.md
msyyc: Updated the examples to use ARM operation templates instead of hand-written TypeSpec operations.

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.md
```diff
@@ -0,0 +1,72 @@
+---
```
timotheeguerin, line 1: same comment on this pr and seems like all of them around the doc as https://github.com/Azure/typespec-azure/pull/5265/changes#r3871865499

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.ts
```diff
@@ -0,0 +1,54 @@
...
+              verb: httpOperation.verb,
+            },
+          });
+        }
+      },
+    };
+  },
+});
+
+function isPointOperationPath(path: string): boolean {
```
timotheeguerin, line 51: can you add tsdoc explaining what is a point operation.

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.md
```diff
@@ -0,0 +1,72 @@
...
+title: "point-operation-invalid-query-parameter"
+---
+
+```text title="Full name"
+@azure-tools/typespec-azure-resource-manager/point-operation-invalid-query-parameter
+```
+
+ARM point operations using GET, PUT, PATCH, or DELETE must not define query parameters other than `api-version`.
```
timotheeguerin, line 9: Not knowing much about arm terminology i would appreciate a short explanation in the doc of what is a point operation

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.ts
```diff
@@ -0,0 +1,54 @@
...
+const providerAndNamespace = "/providers/[^/]+";
+const resourceTypeAndResourceName = "(?:/\\w+/default|/\\w+/{[^/]+})";
+const queryParameter = "(?:\\?\\w+)";
+const pointOperationPathRegExp = new RegExp(
+  `${providerAndNamespace}${resourceTypeAndResourceName}+${queryParameter}?$`,
+  "i",
+);
+
+export const pointOperationInvalidQueryParameterRule = createRule({
+  name: "point-operation-invalid-query-parameter",
```
timotheeguerin, line 14: `no-query-param-point-op`?

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.ts
```diff
@@ -0,0 +1,54 @@
...
+      operation: (operation) => {
+        const [httpOperation] = getHttpOperation(context.program, operation);
+        if (
+          !pointOperationVerbs.has(httpOperation.verb) ||
+          !isPointOperationPath(httpOperation.path)
+        ) {
+          return;
+        }
+
+        for (const parameter of httpOperation.parameters.parameters) {
```
timotheeguerin, line 33: use httpOperation.parameters.properties over parameters

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.md
msyyc: Added a short definition and example of an ARM point operation, with contrast to collection and action operations. Fixed in 333a06618.

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.ts
msyyc: Renamed the rule to `no-query-param-point-op` consistently across the implementation, docs, tests, linter registration, ruleset, and generated references. Fixed in 4bd1340f3.

packages/typespec-azure-resource-manager/src/rules/no-query-in-point-op.ts
```diff
@@ -0,0 +1,54 @@
...
+              verb: httpOperation.verb,
+            },
+          });
+        }
+      },
+    };
+  },
+});
+
+function isPointOperationPath(path: string): boolean {
```
msyyc: Added TSDoc explaining that point paths target specific ARM resource instances and documenting the expected path shape. Fixed in f0b2967fa.

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.ts
msyyc: Switched query-parameter traversal to `httpOperation.parameters.properties`, using the query property's options and source property for diagnostics. Fixed in 004812831.

packages/typespec-azure-resource-manager/src/rules/point-operation-invalid-query-parameter.md
msyyc: Removed the title frontmatter and full-name block because `tspd` generates them. Fixed in d0c3a320a.

msyyc:
@timotheeguerin Thanks for your review and I updated the PR with your suggestions just now. Pls help take a review again when spare.

packages/typespec-azure-resource-manager/src/rules/no-query-param-point-op.ts
```diff
@@ -10,12 +10,12 @@ const pointOperationPathRegExp = new RegExp(
   "i",
 );
 
-export const pointOperationInvalidQueryParameterRule = createRule({
-  name: "point-operation-invalid-query-parameter",
-  docs: fileRef.fromPackageRoot("src/rules/point-operation-invalid-query-parameter.md"),
+export const noQueryParamPointOpRule = createRule({
+  name: "no-query-param-point-op",
```
timotheeguerin: to be inline with `no-query-in-post`, would also omit `param` here (or replace with `in` or omit `in` in both cases)

msyyc: Renamed the rule to `no-query-in-point-op` to align with `no-query-in-post`, including files, implementation symbol, tests, ruleset registration, and documentation references. Fixed in 23d4df74e.

msyyc:
[Timothee Guerin (@timotheeguerin)](https://github.com/timotheeguerin) Thanks for your review and I updated the PR with your suggestions just now.

timotheeguerin (approved):

PR #5276, opened 2026-08-20
[client-generator-core] Allow @clientOption value to reference a TypeSpec model

### LINKED ISSUE #5256: [client-generator-core] Allow clientOptions values to reference models
## Description

Extend the existing `@clientOptions` decorator so its `value` option can reference a TypeSpec model in addition to the currently supported `string`, `boolean`, and `number` values.

Some emitter-specific options need to preserve a reference to another TypeSpec model rather than encode its identity as a string. For example:

```tsp
@clientOptions({
  name: "composes",
  value: OpenAICreateResponseOptions,
  scope: "csharp"
})
model FoundryCreateResponseOptions {}
```

## Expected behavior

`clientOptions.value` should accept a model reference and preserve enough type information for the scoped emitter to resolve the referenced model and its customizations, including decorators such as `@alternateType`.

iscai-msft:
Fixes #5256

## Summary
Extends the `@clientOption` decorator so its `value` parameter can reference a TypeSpec model, in addition to the previously supported `string`, `boolean`, and `number` literal values.

```tsp
@clientOption("composes", OpenAICreateResponseOptions, "csharp")
model FoundryCreateResponseOptions {}
```

## Changes
- `lib/decorators.tsp`: updated `@clientOption`'s `value` parameter type to `unknown | valueof unknown` and added an example/description covering model reference values.
- `src/internal-utils.ts`: `getDecoratorArgValue` now routes `Model` kind decorator arguments through `getClientTypeWithDiagnostics` (same path already used for `EnumMember`), so a referenced model is resolved into its `SdkType`, preserving any customizations such as `@alternateType`.
- `generated-defs/Azure.ClientGenerator.Core.ts`: regenerated via `pnpm gen-extern-signature`.
- Added tests in `test/decorators/client-option.test.ts` covering:
  - A model reference as the `value`, verified via `getClientOptions`.
  - That `@alternateType` customizations on the referenced model are preserved when resolving the value.
- Added a changeset entry.

## Testing
- `pnpm vitest run` in `packages/typespec-client-generator-core` — 1405 passed, 2 skipped.
- `pnpm build` and `pnpm lint` pass.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


packages/typespec-client-generator-core/test/decorators/client-option.test.ts
```diff
@@ -193,6 +193,63 @@ describe("@clientOption with getClientOptions getter", () => {
...
   });
 
+  it("should support a model reference as the value", async () => {
+    const { program } = await SimpleTesterWithService.compile(`
+      model ComposedOptions {
+        a: string;
+      }
+
+      #suppress "@azure-tools/typespec-client-generator-core/client-option"
+      @clientOption("composes", ComposedOptions, "csharp")
```
iscai-msft, line 203: @tadelesh what do you think about usgaes for a client options model? i think we shouldn't add any by default, but woudd like your thoughts

tadelesh: Agree. We should not add usage for such models. Spec author could use `@usage` to override the usage if needed.

tadelesh (approved):

PR #5280, opened 2026-08-21
Fix double encoding of URI template parameter names for list/associative-array values

### LINKED ISSUE #5278: [Bug]: Double encoding of URL Template parameters in JS
### Describe the bug

Azure App Configuration has a `/labels` endpoint that uses a `$Select` parameter. The `$` is initially getting encoded by the emmiter into our auto generated code, which is incorect, then a second time when making the request resulting in `%2524Select=name` instead of `%24Select=name`.

When the parameter value is an array, expandUrlTemplate follows the getNonExpandedValue path. That function encodes the template variable name again:

```js
if (named && varName) {
  vals.push(encodeComponent(varName, reserved, op));
}
```

Since varName is already %24Select, the second encoding produces %2524Select.

The scalar-value path does not have this problem and explicitly preserves the already encoded variable name:

### Reproduction

The app config auto gen shows it already as encoded here https://github.com/Azure/azure-sdk-for-js/blob/e5610b2d19c65e4fe266a39f4276f39ee2a6acbc/sdk/appconfiguration/app-configuration/src/generated/api/operations.ts#L333

### Checklist

iscai-msft:
## Summary

Fixes Azure/typespec-azure#5278

Query parameter names are pre-encoded by the emitter (e.g. `$Select` → `%24Select`) before being passed as URI template variable names. The scalar-value expansion path in `expandUrlTemplate` already accounted for this and skipped re-encoding the variable name, but the array-expansion path (`getExpandedValue`) and the associative-array/list path (`getNonExpandedValue`) still called `encodeURIComponent`/`encodeComponent` on `varName`, causing it to be encoded twice (e.g. `%24Select` becoming `%2524Select`).

## Fix

Removed the redundant re-encoding of `varName` in both `getExpandedValue` and `getNonExpandedValue` in `packages/typespec-ts/static/static-helpers/urlTemplate.ts`, matching the existing (correct) behavior of the scalar-value path.

## Testing

Added repro tests in `packages/typespec-ts/test/modular-unit/static/url-template.test.ts` covering scalar, list, and associative-array query parameter values with a pre-encoded parameter name (`%24Select`). Verified the new list/associative-array tests fail on `main` with the exact reported symptom (`%2524Select`) and pass with this fix. All 67 tests in the file pass.

Also ran `pnpm lint` on the touched package (clean).


packages/typespec-ts/test/modular-unit/static/url-template.test.ts
```diff
@@ -510,5 +510,33 @@ describe("url-template", () => {
...
+      assert("{?%24Select}", "?%24Select=name,value");
+    });
+
+    it("should not double encode a pre-encoded query parameter name with an associative array value", () => {
+      const assert = createAssertion({
+        "%24Select": { name: "value" },
+      });
+      assert("{?%24Select*}", "?name=value");
+      assert("{?%24Select}", "?%24Select=name,value");
+    });
```
maorleger, line 540: Is it worth adding a test that combines encoded name with values requiring encoding?

Example (I think):

```ts
const assert = createAssertion({
  "%24Select": ["a b", "x=y", "*"],
});
assert(
  "{?%24Select*}",
  "?%24Select=a%20b&%24Select=x%3Dy&%24Select=%2A",
);
```

Or something similar? just a test that shows that we continue to encode the values even when preserving the already-encoded name.

maorleger (approved):
The fix LGTM but I'd love to get confirmation from the emitter folks 

Thanks for putting that together!

JialinHuang803 (approved):

JialinHuang803:
The fix LGTM, thank you!

packages/typespec-ts/test/modular-unit/static/url-template.test.ts
iscai-msft: Good call — added this test (using your exact example) in ac74e476. Confirms values are still encoded even when the pre-encoded parameter name is preserved.

PR #5281, opened 2026-08-24
[Swagger Linter Migration] ParametersInPost

msyyc:
## Original Swagger linter

- linter code: [ParametersInPost](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/parameters-in-post.ts)
- linter doc: [parameters-in-post.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/parameters-in-post.md)

The original rule checks:

- [x] Applies to ARM OpenAPI specs (`RPC-POST-V1-05`).
- [x] For each POST operation parameter, find parameters where `in === "query"`.
- [x] Exempt only the exact query parameter name `api-version`.
- [x] Report every remaining query parameter on the POST operation.
- [x] Message that POST operation query parameters must be moved into the payload.

The Spectral function receives the Swagger POST operation `parameters` array. It returns no diagnostics when the value is not an array or is empty. Otherwise it filters the operation parameters with `param.in === "query" && param.name !== "api-version"`, then emits one diagnostic per matching parameter. The diagnostic path is the Spectral context path for the operation parameters rather than the individual TypeSpec source parameter.

The Swagger rule reports emitted OpenAPI occurrences, so its raw diagnostic cardinality can differ from a semantic TypeSpec rule. The migration evidence notes one known cardinality skew in HybridConnectivity where Swagger emits two POST parameter occurrences while TypeSpec reports four semantic source locations for `expiresin`; that is not a project coverage gap.

- Validator rule id: `ParametersInPost`
- Local lintdiff rule name: `parameters-in-post`
- Canonical validator rule slug: `parameters-in-post`
- Source branch: `feature/lintdiff-parameters-in-post`
- Source commit: `e56688697d812f0c46ac0a5c15dc00aa038d708e`
- Source worktree: `C:\dev\worktrees\typespec-azure-lintdiff-parameters-in-post`
- Source worktree uncommitted rule changes: none observed for the inspected source/evidence files
- Source rule file: https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-parameters-in-post/packages/typespec-lintdiff/src/rules/parameters-in-post.ts

The user-marked done source rule was not modified during promotion.

## Destination analysis

Selected package: `@azure-tools/typespec-azure-resource-manager`.

Evidence:

- The fixture front matter says `ParametersInPost` applies to Resource Manager (ARM).
- Validator metadata has `applicability: ARM`, `sources: ["arm"]`, severity `error`, and RPC code `RPC-POST-V1-05`.
- The lintdiff implementation depends on ARM semantics via `isArmProviderNamespace`.
- The rule checks ARM POST operations, including non-resource ARM POST operations, and does not apply to data-plane services.
- `docs/validate-report.md` inferred the fixture rows as `resource-manager`.
- No existing official ARM/core rule was found with equivalent coverage.

Plausible alternative: `@azure-tools/typespec-azure-core`, but that would either over-scope an ARM-only RPC rule to data-plane services or require a rewrite that removes the ARM provider namespace boundary. The ARM package keeps the dependency direction correct and matches the source semantics.

The promoted rule is `@azure-tools/typespec-azure-resource-manager/no-query-parameters-in-post`. It registers in the ARM linter and the resource-manager ruleset.

For each operation, it:

1. Resolves the operation namespace from the containing interface namespace or operation namespace.
2. Uses `isArmProviderNamespace` to keep the rule scoped to ARM provider namespaces.
3. Uses `getHttpOperation` to inspect the HTTP verb and parameters.
4. Skips non-POST operations.
5. Reports each HTTP query parameter whose serialized name is not exactly `api-version`.
6. Targets the authorable TypeSpec parameter (`parameter.param`) and includes the query name in the diagnostic message.

Intentional promotion adaptation: the official rule uses an exact `name === "api-version"` exemption to match the original Swagger validator. During final review, the lintdiff source was found to use a case-insensitive helper; the source branch remains unchanged, and this PR documents the sync note below.

- [`query-param-in-post`](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-parameters-in-post/packages/typespec-lintdiff/test/fixtures/ParametersInPost/query-param-in-post/main.tsp) -> emits one warning for `mode` on an ARM POST resource action.
- [`multiple-query-params`](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-parameters-in-post/packages/typespec-lintdiff/test/fixtures/ParametersInPost/multiple-query-params/main.tsp) -> emits one warning each for `mode` and `format` on the same ARM POST operation.
- [`non-resource-post-query`](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-parameters-in-post/packages/typespec-lintdiff/test/fixtures/ParametersInPost/non-resource-post-query/main.tsp) -> emits a warning for `$filter` on a non-resource ARM POST operation.
- [`api-version-only`](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-parameters-in-post/packages/typespec-lintdiff/test/fixtures/ParametersInPost/api-version-only/main.tsp) -> remains valid when the only query parameter is the standard `api-version` supplied by ARM templates.
- Review regression: `@query("API-Version")` emits a warning, matching the Swagger validator's exact-name `api-version` exemption.
- Additional compliant case: non-ARM POST operations are ignored even when they have query parameters.

## Migration evidence

Migration evidence is captured in the source branch's [`migration.md`](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-parameters-in-post/packages/typespec-lintdiff/test/fixtures/ParametersInPost/migration.md), including the focused tests, real-service project comparison, latest full-corpus counts, one-sided project explanations, compile failures, diagnostic cardinality notes, and final uncertainty statement.

## Validation

Targeted validation completed:

- `pnpm --filter @azure-tools/typespec-azure-resource-manager exec vitest run test/rules/no-query-parameters-in-post.test.ts` (6 tests)
- `pnpm --filter @azure-tools/typespec-azure-resource-manager build`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager lint`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager regen-docs`
- `pnpm run format:check`
- `pnpm --filter "@azure-tools/typespec-azure-website..." run build`
- `pnpm --filter @azure-tools/typespec-azure-rulesets build`
- `pnpm --filter @azure-tools/typespec-azure-rulesets test`
- `pnpm --filter @azure-tools/typespec-azure-resource-manager test` (passed after rerun; the first full-package run timed out once in existing `test/enforce-constraint.test.ts`, and that test file passed when rerun directly)
- `git diff --check`

## Validation blocker

`pnpm validate:pr` was run with a bounded wait. It completed step 1, `Branch is up to date`, in 3.7s, then produced no further progress during the 5-minute bound. Because the targeted package, docs, website, rulesets, formatting, and diff validations above passed, the stalled broad pre-PR run was stopped and treated as an environmental/pre-existing validation blocker.

## Promotion sync policy

The lintdiff source branch remains the source of truth for the migration PR and was not changed during this promotion. If reviewers decide the source lintdiff rule must also use exact `api-version` matching, that should be handled by explicitly reopening lintdiff repair for `ParametersInPost`; this promotion PR should stay synchronized with the approved source or be blocked until that repair is completed.

This new ARM rule could affect existing Azure service specs, so the `int:azure-specs` label is appropriate.
NOTE: heavy


packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
```diff
@@ -0,0 +1,83 @@
...
+guidelines.
+
+The `api-version` query parameter is allowed.
+
+## Impact
+
+- **Area:** API
+
+Extra query parameters on ARM POST operations violate the ARM POST request
+contract and can lead to inconsistent generated SDK method signatures.
```
catalinaperalta: Given the description, I'd say this is still API, SDK impact

packages/typespec-azure-resource-manager/test/rules/no-query-parameters-in-post.test.ts
```diff
@@ -0,0 +1,218 @@
...
+        search(
+          @query("API-Version") apiVersion: string,
+        ): ArmResponse<SearchResponse> | ErrorResponse;
+      }
+      `,
+    )
+    .toEmitDiagnostics({
+      code: "@azure-tools/typespec-azure-resource-manager/no-query-parameters-in-post",
+      message:
+        "Query parameter 'API-Version' should be moved into the POST payload. POST operations must not contain query parameters other than api-version.",
```
catalinaperalta: Hmm so this test made me wonder about non-standard api-version parameters but that are marked with the `@apiVersion` decorator. If the swagger rule today flags on this sort of casing difference we're at parity at least, but maybe it's an improvement we could think on. Maybe in typespec world this rule shouldnt report on non-standard api version params that have the apiVersion decorator on them. cc @markcowl what do you think?

catalinaperalta (approved):
overall lgtm, left a question about the api version casing but we could improve it later on

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (!isArmProviderNamespace(context.program, namespace)) {
+          return;
+        }
+
+        const [httpOperation] = getHttpOperation(context.program, operation);
+        if (httpOperation.verb !== "post") {
+          return;
+        }
+
+        for (const parameter of httpOperation.parameters.parameters) {
```
timotheeguerin, line 27: use httpOperation.parameters.properties over parameters

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
```diff
@@ -0,0 +1,46 @@
...
+              name: parameter.name,
+            },
+          });
+        }
+      },
+    };
+  },
+});
+
+function isApiVersionParameter(name: string): boolean {
```
timotheeguerin, line 44: this is defined in use-api-version.ts rule can we share the helper(move it to common place)

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
```diff
@@ -0,0 +1,46 @@
+import { createRule, fileRef, paramMessage } from "@typespec/compiler";
+import { getHttpOperation } from "@typespec/http";
+import { isArmProviderNamespace } from "../namespace.js";
+
+export const noQueryParametersInPostRule = createRule({
+  name: "no-query-parameters-in-post",
```
timotheeguerin: `no-query-param-in-post` if changing to same terminology in other pr for shorter name

could also omit `param` tbh

packages/typespec-azure-resource-manager/test/rules/no-query-parameters-in-post.test.ts
markcowl: ARM apis should not use non-standard api-version parameters, as we account for here: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/use-api-version.ts

I think calling these out is duplicative, but fine.

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
```diff
@@ -0,0 +1,83 @@
...
+```text title="Full name"
+@azure-tools/typespec-azure-resource-manager/no-query-parameters-in-post
+```
+
+ARM POST operations should pass request-specific input in the request body, not
+as query parameters. Keeping POST inputs in the payload gives SDKs a stable
+request shape and keeps ARM APIs consistent with Azure Resource Manager RPC
+guidelines.
+
+The `api-version` query parameter is allowed.
```
markcowl:
```suggestion
Only the standard  `api-version` query parameter is allowed.
```

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
```diff
@@ -0,0 +1,83 @@
...
+
+model ActionResponse {
+  result: string;
+}
+
+@armResourceOperations
+interface Widgets {
+  @post
+  @armResourceAction(Widget)
+  doAction(...ResourceInstanceParameters<Widget>, @query mode?: string, @body body: ActionRequest):
```
markcowl: We should have at least one incorrect example using the 'Parameters' parameter of one of the ArmResourceAction templates to specify a query parameter.

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
```diff
@@ -0,0 +1,83 @@
...
+
+model ActionResponse {
+  result: string;
+}
+
+@armResourceOperations
+interface Widgets {
+  @post
+  @armResourceAction(Widget)
+  doAction(...ResourceInstanceParameters<Widget>, @body body: ActionRequest):
```
markcowl: Correct examples should *always* use the standard templates

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
```diff
@@ -0,0 +1,46 @@
...
+  docs: fileRef.fromPackageRoot("src/rules/no-query-parameters-in-post.md"),
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-query-parameters-in-post",
+  description: "ARM POST operations must not declare query parameters other than api-version.",
+  messages: {
+    default: paramMessage`Query parameter '${"name"}' should be moved into the POST payload. POST operations must not contain query parameters other than api-version.`,
```
markcowl:
```suggestion
    default: paramMessage`Query parameter '${"name"}' should be moved into the POST request body. POST operations must not contain query parameters other than api-version.`,
```

markcowl (commented):
Looks good, a few comments

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
markcowl: `no-query-in-post` has a nice ring to it

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
```diff
@@ -0,0 +1,83 @@
+---
+title: "no-query-parameters-in-post"
+---
+
+```text title="Full name"
+@azure-tools/typespec-azure-resource-manager/no-query-parameters-in-post
+```
```
msyyc: similar comment as https://github.com/Azure/typespec-azure/pull/5265#discussion_r3871865499

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
msyyc: Updated in cd2b41182 to iterate over `httpOperation.parameters.properties`, filter on `kind`, use the serialized `options.name`, and report against the source property.

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.ts
msyyc: Renamed the rule to `no-query-in-post` across its implementation, files, registration, ruleset, tests, generated reference links, and change descriptions in 58afe9ffd.

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
msyyc: Updated the incorrect example in 3547fe5f7 to pass an `ActionParameters` model containing the query parameter through `ArmResourceActionSync`'s `Parameters` argument. Follow-up 64004dad5 qualifies `Foundations.DefaultBaseParameters` so the documented example compiles.

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
msyyc: Updated the correct example in eb65a0c7c to use `ArmResourceActionSync<Widget, ActionRequest, ActionResponse>` directly.

packages/typespec-azure-resource-manager/src/rules/no-query-parameters-in-post.md
msyyc: Removed the manually authored frontmatter and `Full name` block in 94e4696ab so `tspd` owns the generated title and rule ID.

packages/typespec-azure-resource-manager/src/rules/no-query-in-post.ts
```diff
@@ -0,0 +1,46 @@
...
+              name: parameter.name,
+            },
+          });
+        }
+      },
+    };
+  },
+});
+
+function isApiVersionParameter(name: string): boolean {
```
msyyc: I kept these helpers separate because they answer different questions. This rule checks the exact serialized HTTP query name (`api-version`) for Swagger parity, while `use-api-version.ts` checks whether a model property originates from the common-types `ApiVersionParameter`. Sharing that helper would change this rule's semantics.

msyyc:
@timotheeguerin / @markcowl Thanks for your review. I updated PR with your suggestions and pls help take a review again.

packages/typespec-azure-resource-manager/src/rules/no-query-in-post.ts
timotheeguerin: still should at least share it across the new rules you add

msyyc: I see. Since other PRs are not merged yet, I plan to create another PR to extract common helper functions for the migrated rules after those PRs merged. @timotheeguerin

packages/typespec-azure-resource-manager/src/rules/no-query-in-post.md
```diff
@@ -0,0 +1,77 @@
...
+  @query mode?: string;
+}
+
+@armResourceOperations
+interface Widgets {
+  doAction is ArmResourceActionSync<
+    Widget,
+    ActionRequest,
+    ActionResponse,
+    Foundations.DefaultBaseParameters<Widget>,
```
markcowl: No reason to include this parameter - this is the default

packages/typespec-azure-resource-manager/src/rules/no-query-in-post.md
```diff
@@ -0,0 +1,77 @@
...
+    ActionParameters
```
markcowl:
```suggestion
    Parameters = ActionParameters
```

packages/typespec-azure-resource-manager/test/rules/no-query-in-post.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      }
+
+      model ActionResponse {
+        result: string;
+      }
+
+      @armResourceOperations
+      interface Widgets {
+        @post
+        @armResourceAction(Widget)
```
markcowl: let's use standard templates by default in tests

packages/typespec-azure-resource-manager/test/rules/no-query-in-post.test.ts
```diff
@@ -0,0 +1,218 @@
...
```
markcowl: same

packages/typespec-azure-resource-manager/test/rules/no-query-in-post.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      `,
+    )
+    .toEmitDiagnostics({
+      code: "@azure-tools/typespec-azure-resource-manager/no-query-in-post",
+      message:
+        "Query parameter '$filter' should be moved into the POST request body. POST operations must not contain query parameters other than api-version.",
+    });
+});
+
+it("does not emit a warning for an ARM POST operation with only api-version in query", async () => {
```
markcowl: This test is unnecessary if you use standard templates

markcowl (commented):
Some minor changes

packages/typespec-azure-resource-manager/test/rules/no-query-in-post.test.ts
msyyc: Updated this test to use `ArmResourceActionSync` with its `Parameters` argument in 08ead44b4.

packages/typespec-azure-resource-manager/test/rules/no-query-in-post.test.ts
msyyc: Removed the redundant api-version-only test in 32742d972. The standard-template diagnostic tests now include api-version implicitly while asserting that only the custom query parameters are reported.

packages/typespec-azure-resource-manager/src/rules/no-query-in-post.md
msyyc: Removed the explicit default `Foundations.DefaultBaseParameters<Widget>` argument in 0aae25e1f.

packages/typespec-azure-resource-manager/test/rules/no-query-in-post.test.ts
msyyc: Updated the multiple-query test to use `ArmResourceActionSync` with its `Parameters` argument in 92e9aed87.

packages/typespec-azure-resource-manager/src/rules/no-query-in-post.md
msyyc: Changed this to the named `Parameters = ActionParameters` argument in 03072b916.

msyyc:
Thanks @markcowl 's second review and I updated PR with your comments. Pls help review for my new change.

.chronus/changes/enable-parameters-in-post-ruleset-2026-08-21.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
```
timotheeguerin: internal and content here was incorrect anyway

msyyc: Updated in b5b9f7cb4: changed this to an `internal` change and corrected the description to say the rule is registered as disabled in the resource manager ruleset.

timotheeguerin (approved):

PR #5291, opened 2026-08-25
Label management C# emitter in coverage dashboard

live1206:
## Summary

- map `@azure-typespec/http-client-csharp-mgmt` to the `C#` dashboard display name
- ensure the Azure Management Plane table uses the same language label as the data-plane C# emitter
- allow the display-name grouping from microsoft/typespec#11709 to aggregate both C# emitters into one overview card

## Validation

- Verified the combined changes against live coverage data: the management-plane table header displays `C#`, the raw management emitter package name is absent, and the overview contains one aggregated C# card.


live1206:
typespec-dashboard-local-fixed-mapping

JoshLove-msft (approved):

PR #5294, opened 2026-08-25
[Swagger Linter Migration] PatchBodyParametersSchema

msyyc:
## Original Swagger linter

- Validator rule: [PatchBodyParametersSchema](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/patch-body-parameters.ts)
- [Validator documentation](https://github.com/Azure/azure-openapi-validator/blob/main/docs/patch-body-parameters-schema.md)
- ARM RPC guideline: `RPC-Patch-V1-10`

The Swagger validator traverses ARM PATCH body schemas and referenced definitions, reporting required, default-valued and create-only (`x-ms-mutability: ["create"]`) properties, including emitted discriminator properties. Top-level emitted `identity` is exempt. Its truthiness-based default check misses `false`, `0` and `""`; the TypeSpec rule intentionally rejects every authored default, consistent with the stated guidance.

## Immutable source and repair provenance

- Source rule ID/name: `PatchBodyParametersSchema` / `patch-body-parameters-schema`.
- Official rule: `@azure-tools/typespec-azure-resource-manager/no-unsafe-patch-body-properties`.
- Current source ref: `Azure/typespec-azure:feature/lintdiff-migration-new`.
- **Current immutable source SHA: `bbd50fc21b5c7fe3b7bcf57084e410ba17d6ea6b`**, the merged repair from #5480.
- Original validated repair head: `d85f9060d2096b9833644b01268f78fde22976fe`. Its complete tree equals the merged source tree, `0c609f2487f926e9fe797e1cec6c910843fdda22`.
- Previously pinned source: `9106cc7d74909f10b246cb1138ac58065b6960a2` (original migration, #5209).
- [Pinned source rule](https://github.com/Azure/typespec-azure/blob/bbd50fc21b5c7fe3b7bcf57084e410ba17d6ea6b/packages/typespec-lintdiff/src/rules/patch-body-parameters-schema.ts)
- [Pinned migration evidence](https://github.com/Azure/typespec-azure/blob/bbd50fc21b5c7fe3b7bcf57084e410ba17d6ea6b/packages/typespec-lintdiff/test/fixtures/PatchBodyParametersSchema/migration.md)
- [Pinned fixtures](https://github.com/Azure/typespec-azure/tree/bbd50fc21b5c7fe3b7bcf57084e410ba17d6ea6b/packages/typespec-lintdiff/test/fixtures/PatchBodyParametersSchema)

The source repair limits property traversal to HTTP bodies whose structured `bodyKind` is `single`. Multipart wrappers become form-data parameters rather than body-schema properties. File payloads have primitive binary schemas, not the properties of the native File model. The promotion imports this repair in `fab9afec1bcea08e71d1d9d2af1793330236361f`; it does not independently redesign source semantics.

The user merged #5480 and explicitly authorized this refresh. The source PR's automated review invocation stopped on a local timestamp-wrapper/collector recovery condition, with one Balanced request and no completed round. **Merged status is not being represented as a clean source review.** Independent source code reviews and the validation below predate that operational failure. This changed promotion head requires a new promotion-mode review invocation.

Both existing source and promotion worktrees are preserved. No source files, fixtures, dependency files or migration artifacts were changed during this promotion refresh.

## Destination analysis and applicability

Selected package: `@azure-tools/typespec-azure-resource-manager`, not core. Source fixture/catalog applicability is ARM; this rule enforces ARM PATCH semantics and the ARM identity-envelope exception. No equivalent official rule provided this coverage.

Explicit rule enablement is the destination applicability boundary. The promoted copy deliberately removes the lintdiff-only `resolveProviderNamespace` isolation guard: that helper searches a namespace and its descendants, not its ancestors, so it could skip operations nested under a decorated provider. The payload checks do not require provider metadata.

When enabled, the rule checks PATCH operations in ordinary/nested namespaces and interfaces, with or without `@armProviderNamespace`. It does not distinguish ARM from data-plane services within a mixed compilation. This approved destination adaptation is preserved by the repair. The shared `@azure-tools/typespec-azure-rulesets/resource-manager` entry remains **false**; explicit enablement is required. The previous main merge and both linter registrations are preserved.

## Native implementation

The rule visits HTTP PATCH operations, obtains the effective body through `getHttpOperation`, and traverses only single bodies. It uses `resolveRequestVisibility`, `createMetadataInfo`, `isTransformed`, `isPayloadProperty` and `isOptional`, not an emitter or generated OpenAPI references.

Model and single-non-null-variant traversal, required discriminator checks, falsy default detection, create-only visibility checks, `never` omission, top-level encoded `identity` exemption and diagnostic retargeting for imported declarations are unchanged. Multipart/file exclusion does not exempt operations from other ARM rules, such as JSON content-type guidance.

Fixture names below refer to the immutable pinned fixture directory linked above. Test titles are the actual native titles.

| Source fixture | Native test title |
| --- | --- |
| `required-patch-property` | `emits diagnostics for required PATCH body properties` |
| `nullable-body-required-property` | `emits diagnostics for required properties in nullable top-level PATCH bodies` |
| `nullable-model-required-property` | `emits diagnostics for required properties in nullable nested PATCH models` |
| `discriminator-required-patch-property` | `emits diagnostics for discriminator properties that Autorest requires` |
| `default-patch-property` | `emits diagnostics for PATCH body properties with defaults` |
| `create-only-patch-property` | `emits diagnostics for PATCH body properties that are only visible on create` |
| `encoded-non-identity-violating` | `emits diagnostics for authored identity properties encoded away from top-level identity` |
| `multi-model-union-compliant` | `allows unsupported unions with multiple model variants` |
| `never-property-compliant` | `allows required PATCH body properties omitted because their type is never` |
| `top-level-identity-compliant` | `allows top-level identity PATCH body properties` |
| `synthesized-identity-discriminator-compliant` | `allows top-level identity discriminator properties synthesized by Autorest` |
| `encoded-identity-compliant` | `allows PATCH body properties encoded as top-level identity` |
| `implicit-optional-patch-compliant` | `allows required and create-only source properties removed from the emitted PATCH schema` |
| `multipart-patch-body-compliant` | `does not inspect %s payload properties` with `multipart model` and `multipart tuple` |
| `file-patch-body-compliant` | `does not inspect %s payload properties` with `file` |

Additional native coverage: `allows optional PATCH body properties`; `checks PATCH bodies in a %s` and `ignores %s request bodies` under `applicability %s` cover service/nested/interface namespaces with and without provider metadata and POST/PUT controls. New controls use `does not inspect %s payload properties` with `single binary`, plus `preserves all checks for a single %s payload` for `PatchBody` and `PatchBody | null`.

### Current promotion refresh

Executed in the existing promotion worktree with the current promotion dependency context:

- `mise exec -- pnpm --filter @azure-tools/typespec-azure-resource-manager exec vitest run test\rules\no-unsafe-patch-body-properties.test.ts --hookTimeout 30000` — **31 passed**.
- `mise exec -- pnpm --filter @azure-tools/typespec-azure-resource-manager build` — passed, including library-linter compilation.
- `mise exec -- pnpm --filter @azure-tools/typespec-azure-resource-manager lint` — passed.
- `mise exec -- pnpm --filter @azure-tools/typespec-azure-resource-manager regen-docs` — passed; existing missing-documentation warnings concern unrelated declarations.
- Explicit `prettier --ignore-path NUL --write` and `--check` on the ARM README, authored rule Markdown, generated linter reference and generated rule page — passed; official names, paths, links, table entry and payload/applicability text inspected.
- `mise exec -- pnpm --filter @azure-tools/typespec-azure-rulesets build` and `test` — passed, **4 tests**.
- `mise exec -- pnpm --filter @azure-tools/typespec-azure-resource-manager test --hookTimeout 30000` — **495 passed across 66 files**.
- Repository `mise exec -- pnpm format` and `mise exec -- pnpm lint` — passed before commit.
- `git diff --check` — passed. Only the four intended maintained files changed in this refresh.

The first focused run passed 30/31 tests, including every new payload case, but the first pre-existing applicability case hit the default 10-second cold-start `beforeEach` timeout. A bounded 30-second hook timeout allowed both focused and full ARM suites to pass, without changing tests or production code to mask it.

No lintdiff corpus or website build ran as part of promotion. The target dependency closure had already been built; affected packages were rebuilt for this refresh.

### Reused source evidence (not rerun during promotion)

The exact-tree source repair passed build, scoped formatting/lint, six native tests, all fifteen fixture comparisons and two independent code reviews. Full corpus: 462/468 projects compiled, with the same six known compile failures; 93 validator projects, 105 TypeSpec projects, overlap 93, zero validator-only projects; 703 versus 872 diagnostics. Full details and limitations are in the pinned migration evidence.

### Historical broad-validation limitation

Original promotion `validate:pr` stopped after five minutes without progress beyond its successful branch-up-to-date check. The narrower required validation completed. This is historical evidence, not a fresh failure or a claim that broad validation passed. The incremental refresh uses the current targeted promotion checks above; no new PR was created.

## Promotion sync policy

Any new source-semantic defect must stop the promotion review and return to an explicitly authorized source-repair workflow. Do not silently repair the official copy alone, change the immutable source, inspect suppressed comments as a finding queue, or treat earlier reviews as verification of this changed head.


packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,296 @@
...
+  },
+});
+
+type Violation = {
+  target: DiagnosticTarget;
+  propertyName: string;
+  messageId: "required" | "default" | "createOnly";
+};
+
+function findViolations(program: Program, patchBody: Type, operation: Operation): Violation[] {
```
timotheeguerin, line 77: this looks to be quite a complex rule, can you incluide some benchmark of how fast/slow it runs on large spec(network) compared to others

msyyc: @timotheeguerin  I benchmarked this against the large Network spec from `azure-rest-api-specs` using the existing external benchmark runner (2 warmups + 10 measured iterations).

The rule averaged **10.24 ms** per compile (median **10.21 ms**, range **9.72–10.95 ms**, CV **4.02%**). That is **0.066%** of the total compile time and **0.465%** of total lint time. It ranked 17th of 83 enabled rules; for comparison, the most expensive rules in this run took 186–407 ms each.

The disabled/enabled end-to-end averages were 15.63s and 15.54s respectively, so there is no measurable total compile-time regression—the difference is within run noise.

packages/typespec-azure-resource-manager/src/rules/patch-body-invalid-property.ts
```diff
@@ -0,0 +1,296 @@
...
+  createMetadataInfo,
+  getHttpOperation,
+  resolveRequestVisibility,
+  Visibility,
+  type MetadataInfo,
+} from "@typespec/http";
+import { resolveProviderNamespace } from "../namespace.js";
+
+export const patchBodyInvalidPropertyRule = createRule({
+  name: "patch-body-invalid-property",
```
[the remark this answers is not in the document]
msyyc: Renamed the public rule to `no-unsafe-patch-body-properties` in `64a91f682` and updated the implementation, tests, ruleset registration, documentation, generated indexes, and changeset.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.md
```diff
@@ -0,0 +1,52 @@
...
+  displayName?: string;
+  enabled?: boolean;
+}
+
+@route("/widgets/{name}")
+@patch
+op update(@path name: string, @body body: WidgetPatchBody): void;
+```
+
+## LintDiff Equivalent
```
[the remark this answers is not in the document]
msyyc: Added the alphabetized ARM reviewer-table entry in `526713f68`, including the rule URL, `PatchBodyParametersSchema` equivalent, API impact, and the applicable `RPC-Patch-V1-10` guideline. (`R2016` is not the identifier for this legacy rule.)

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,296 @@
...
+      continue;
+    }
+    if (!metadataInfo.isPayloadProperty(property, schemaVisibility)) {
+      continue;
+    }
+    if (isNeverType(property.type)) {
+      continue;
+    }
+    const propertyTarget =
+      getLocationContext(program, property).type === "project" ? property : diagnosticTarget;
```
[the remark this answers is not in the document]
msyyc: Added coverage in `20ca081df` using both `Azure.Core.Page` and `Azure.ResourceManager.CommonTypes.ManagedServiceIdentity`. The test verifies library-originated violations are retargeted to local project properties and that the real top-level ARM `identity` exception remains intact.

msyyc:
Audit of `70527cbc3ca50d0c83e6bd68c2d90b3a58a25100` against the current migration skills found a concrete namespace-guard issue.

**The rule can skip PATCH operations nested beneath an ARM provider namespace.** In [no-unsafe-patch-body-properties.ts:42-44](https://github.com/Azure/typespec-azure/blob/70527cbc3ca50d0c83e6bd68c2d90b3a58a25100/packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts#L42-L44), applicability is tested with `resolveProviderNamespace(program, operationNamespace)`. That [helper searches the supplied namespace and its descendants](https://github.com/Azure/typespec-azure/blob/70527cbc3ca50d0c83e6bd68c2d90b3a58a25100/packages/typespec-azure-resource-manager/src/namespace.ts#L293-L308), not its ancestors. When only the enclosing service has provider metadata, the child operation namespace can return `undefined`, bypassing the required/default/create-only property checks.

Both skills explicitly flag this traversal mistake: [development applicability guidance](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/develop-lintdiff-rule/SKILL.md#arm-applicability-without-redundant-namespace-guards) and [promotion adaptation](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#4-move-and-adapt-source-code).

Please establish the intended destination applicability boundary, then remove the guard if it is only lintdiff isolation, or use an ancestor-aware check if per-service membership is genuinely required. Add a native regression with a PATCH endpoint in a nested namespace. If the selected official ruleset supplies applicability, also cover ordinary and nested namespaces without an unnecessary provider decorator. Do not mechanically swap helpers without deciding whether this guard belongs here.

Separately, the PR description still maps the official name to `patch-body-invalid-property`; the current code/tests/docs use `no-unsafe-patch-body-properties`. Please refresh that explanation.

Removing lintdiff-only isolation can be a promotion adaptation. Any verified defect in the source's intended semantics should instead follow source repair; this comment does not authorize changing the immutable source as part of promotion. The namespace finding is code-backed, not a claim that an affected production service was reproduced.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+      operation: (operation) => {
+        const [httpOperation] = getHttpOperation(context.program, operation);
+        if (httpOperation.verb !== "patch") {
+          return;
+        }
+
+        const patchBody = httpOperation.parameters.body?.type;
+        if (patchBody === undefined) {
+          return;
+        }
```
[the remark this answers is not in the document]
msyyc: Fixed in fab9afec1bcea08e71d1d9d2af1793330236361f by importing the merged source repair from #5480 (immutable source bbd50fc21b5c7fe3b7bcf57084e410ba17d6ea6b). The rule now traverses properties only when the structured HTTP body has `bodyKind === "single"`; multipart wrappers and file payloads are excluded. One clarification: file payloads can still emit an `in: body` parameter, but its schema is primitive binary rather than the File model's properties. Added multipart-model, multipart-tuple, file and single-binary regressions plus ordinary/nullable-model controls; all 31 focused tests and 495 ARM package tests pass. Existing enablement-based applicability and the disabled shared-ruleset entry are preserved.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.md
```diff
@@ -0,0 +1,60 @@
+ARM PATCH request body properties must be safe for partial updates. A property emitted in an ARM PATCH body must not be required, must not define a default value, and must not be create-only.
+
+## Impact
+
+- **Area:** API
+
+PATCH describes partial updates. Required PATCH body properties, default-valued properties, and create-only properties can make partial updates ambiguous for service authors and SDKs, and can produce ARM OpenAPI that violates PATCH request-body guidance.
```
timotheeguerin, line 11: can you tweak this doc to have a bullet list of the pattern that are not allowed for easy parsing

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+    !isTopLevelIdentityProperty([...path, discriminator.propertyName], discriminator.propertyName)
+  ) {
+    violations.push({
+      target: getLocationContext(program, model).type === "project" ? model : diagnosticTarget,
+      propertyName: [...path, discriminator.propertyName].join("."),
+      messageId: "required",
+    });
+  }
+
+  for (const property of getModelProperties(model)) {
```
timotheeguerin: can you use walkPropertiesInherited to simplify this, seems it might be a pattern across prs too check the others

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+    return true;
+  }
+
+  return (
+    visibility.size > 0 &&
+    [...visibility].every((member) => ["Read", "Create", "Update"].includes(member.name))
+  );
+}
+
+function getModelProperty(model: Model, name: string): ModelProperty | undefined {
```
timotheeguerin: we already expose that as `getProperty` I believe, same as other comment check other prs, seems a lot of helper either already exists or should be shared across rules here

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+  violations: Violation[],
+  path: string[] = [],
+  visited: Map<Model, Set<Visibility>> = new Map(),
+  diagnosticTarget: DiagnosticTarget,
+  metadataInfo: MetadataInfo,
+  visibility: Visibility,
+) {
+  const schemaVisibility = metadataInfo.isTransformed(model, visibility)
+    ? visibility
+    : Visibility.Read;
```
timotheeguerin: Switching to `Visibility.Read` based on `isTransformed`, together with `canSharePropertyUsingReadonlyOrXmsMutability`, makes emitter schema reuse affect which properties this linter treats as PATCH input. For example:

```tsp
model PatchBody {
  @visibility(Lifecycle.Read)
  id: string;

  @visibility(Lifecycle.Create)
  createdBy?: string;

  name?: string;
}

@patch op update(@body body: PatchBody): void;
```

TypeSpec's effective PATCH input contains only `name`, but the rule reports both `id` and `createdBy`. It also reports defaults on read-only properties. Adding an unrelated optional property with `@visibility(Lifecycle.Read, Lifecycle.Query)` makes the warnings disappear because that forces a separate schema representation.

This should use the resolved request visibility to determine payload membership and optionality, without AutoRest's canonical-schema sharing policy. A response-only or create-only property does not become part of the PATCH input just because an emitter can retain it in a shared schema.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+
+type Violation = {
+  target: DiagnosticTarget;
+  propertyName: string;
+  messageId: "required" | "default" | "createOnly";
+};
+
+function findViolations(program: Program, patchBody: Type, operation: Operation): Violation[] {
+  const violations: Violation[] = [];
+  const metadataInfo = createMetadataInfo(program, {
```
timotheeguerin: Create `metadataInfo` once in `create(context)` and keep per-operation traversal state in a closure. That preserves its caches and avoids threading eight arguments through both recursive helpers.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+export const noUnsafePatchBodyPropertiesRule = createRule({
+  name: "no-unsafe-patch-body-properties",
+  docs: fileRef.fromPackageRoot("src/rules/no-unsafe-patch-body-properties.md"),
+  severity: "warning",
+  description: "ARM PATCH body properties must not be required, have defaults, or be create-only.",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-unsafe-patch-body-properties",
+  messages: {
+    required: paramMessage`Properties of a PATCH request body must not be required, property:${"propertyName"}.`,
+    default: paramMessage`Properties of a PATCH request body must not have default value, property:${"propertyName"}.`,
+    createOnly: paramMessage`Properties of a PATCH request body must not be x-ms-mutability: ["create"], property:${"propertyName"}.`,
```
timotheeguerin: Describe this in TypeSpec terms: the property is only visible during `Lifecycle.Create`. Users should not need to understand `x-ms-mutability` to act on a TypeSpec diagnostic.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,290 @@
...
+  if (visitedVisibilities === undefined) {
+    visited.set(model, new Set([schemaVisibility]));
+  } else {
+    visitedVisibilities.add(schemaVisibility);
+  }
+
+  const discriminator = getInheritedDiscriminator(program, model);
+  if (
+    discriminator !== undefined &&
+    getModelProperty(model, discriminator.propertyName) === undefined &&
```
timotheeguerin: Do we need to synthesize a required property that is absent from the TypeSpec model? This should follow ARM/TypeSpec discriminator semantics, not reproduce what AutoRest happens to emit.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.md
msyyc: Addressed in ef0d397b6. The rule documentation now lists the three prohibited patterns as bullets: required properties, default values, and properties visible only during `Lifecycle.Create`.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
msyyc: Addressed in ef0d397b6. `metadataInfo` is now created once in `create(context)`, while each operation gets its own traversal closure, violations list, and visited map. The recursive helpers now take four arguments instead of eight. Added regressions for a shared recursive body across operations and different PATCH optionality settings in both operation orders. The existing schema-sharing policy is unchanged; the separate visibility comment remains open.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
msyyc: Addressed in ef0d397b6. Replaced the local `getModelProperty` helper with the compiler's public `getProperty`. Added a regression covering an inherited discriminator property so the lookup does not produce a duplicate synthesized-property diagnostic. Discriminator semantics are unchanged in this cleanup; the separate semantic comment remains open.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
msyyc: Addressed in ef0d397b6. Replaced the local `getModelProperties` helper with the compiler's `walkPropertiesInherited`. Added coverage for inherited properties and overridden properties to preserve traversal order, diagnostic targets, and deduplication. This change is scoped to this PR.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
msyyc: Addressed in ef0d397b6. The diagnostic now says that PATCH body properties must not be visible only during `Lifecycle.Create`, rather than referring to `x-ms-mutability`. Updated the diagnostic expectations and documentation accordingly; detection behavior is unchanged.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.md
```diff
@@ -0,0 +1,64 @@
+ARM PATCH request body properties must be safe for partial updates. A property emitted in an ARM PATCH body must not:
```
markcowl, line 1: I think we need to normalize these patch rules - we should only need to examine the PATCH request body once to determine:

- It is a subset of the properties of the resource type
- Known immutable properties ["id", "name", "type", "location", "properties.ProvisioningState"] are excluded or marked as readOnly
- All properties recursively meet the following criteria(modulo itemType models from array-typed properties):
  - optional 
  - include LifeCycle.Update visibility or ar readOnly
  - Have no default

I think this give us one rule that covers 3 lintdiff rules, but will be more efficient, and easier to understand

markcowl (changes requested):
I think we need to consolidate the ruls that examine PATCH bodies to avoid inefficiencies

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
msyyc: Addressed in 26f3c015e2d5d54fb3b31a3704c4befc4344b367, refreshed from reviewed source a02f54f67ecf82b934cfc0323ccc6fe788b147b2 in #5516.

PATCH payload membership and optionality now use the operation's resolved request visibility directly. The AutoRest canonical-schema sharing callback, `isTransformed` decision, and fallback to `Visibility.Read` have been removed.

Native regressions cover your exact example, adding an unrelated Read/Query-only property without changing existing diagnostics, defaults on excluded read-only properties, and shared models under different operation visibility/optionality settings. Ordinary PATCH excludes create-only properties; the create-only check is covered through an explicit Create request-visibility override instead of schema sharing.

This intentionally follows native TypeSpec semantics rather than emitted-Swagger parity; the source migration evidence and PR description record the differences. Validation passed: 49 focused tests, 513 ARM tests, 4 ruleset tests, builds and lint. The published refresh preserves the earlier promotion cleanup and keeps the ruleset disabled.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
msyyc: Addressed in 26f3c015e2d5d54fb3b31a3704c4befc4344b367, refreshed from reviewed source a02f54f67ecf82b934cfc0323ccc6fe788b147b2 in #5516.

Agreed: the rule no longer synthesizes an absent discriminator property or forces an authored optional discriminator to be required because of AutoRest emission. It traverses authored/inherited properties and applies effective PATCH membership and optionality, just as for other properties. Genuinely required discriminator properties are still diagnosed when they belong to effective PATCH input.

Regressions cover compiler-valid absent, optional, required, inherited and visibility-excluded discriminator properties, including legacy implicit PATCH optionality. Compiler-invalid discriminator shapes are not turned into duplicate linter checks. All 49 focused tests and the full 513-test ARM suite passed. This native-semantic divergence is documented in both source migration evidence and the promotion documentation.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      function collectViolations(
+        model: Model,
+        path: string[],
+        diagnosticTarget: DiagnosticTarget,
+        visibility: Visibility,
+      ) {
+        const visitedVisibilities = visited.get(model);
+        if (visitedVisibilities?.has(visibility)) {
+          return;
```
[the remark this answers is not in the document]
msyyc: Confirmed by independent source/adaptation assessment: the operation-wide visited model/visibility map also exists in immutable lintdiff source a02f54f67ecf82b934cfc0323ccc6fe788b147b2. Static tracing shows that two sibling properties using the same `Azure.Core.Page<Item>` instantiation traverse only the first path, including imported-property diagnostic retargeting. This is an inherited source-semantic issue, not an effect of the promotion's provider-guard adaptation.

Leaving this thread unresolved. A source-first repair should add a regression asserting both sibling paths and their local targets, use active-path recursion protection, and preserve cycle termination, effective PATCH visibility, and per-operation state. The published promotion cannot silently acquire different traversal semantics from its reviewed source.

The current source repair PR #5516 has already been merged by its author. The workflow stops at that publication boundary rather than recreating its deleted branch or opening a replacement PR automatically. Continuing needs separate authorization for a new source-repair PR, then source validation/review and a promotion refresh. #5294 remains open and unmerged; the two original visibility/discriminator findings are fixed and resolved.

packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.md
msyyc: @markcowl Agreed on consolidating this into one rule with a shared PATCH-body traversal and distinct diagnostic messages. Before implementing, could you confirm this scope and the semantic boundaries?

- **Scope:** combine `PatchBodyParametersSchema` (#5294), `UnSupportedPatchProperties` (#5384), and `ConsistentPatchProperties` (#5456). Should we also fold in the existing `arm-resource-patch` rule, preserving its coverage? I would leave PATCH-to-PUT correspondence (#5481) and required envelope-property presence separate, since they check different contracts.
- **Visibility:** check effective PATCH input using resolved request visibility. Read-only/Create-only properties may remain in shared models without warnings when excluded from PATCH input; optionality/default checks apply only to input properties. For example, `location` with Read + Create visibility remains valid for ordinary PATCH.
- **Exceptions:** retain the current top-level `identity` exemption unless we intentionally revise it. For array item models, I would interpret the exception as allowing required element properties, rather than exempting elements from every other applicable check. Is that your intent?

These are proposed defaults, not implemented consolidation changes. Keeping this thread open pending agreement.

CC @timotheeguerin

catalinaperalta: I think it makes sense to consolidate into 1 rule that validates patch properties, but we'll have to link to all of the old lintdiff rules in the documentation to help arm reviewers. In this case where we're consolidating multiple rules into 1 it might be helpful when we're linking to the old swagger lintdiff rules to have a table with the message folks would see and the corresponding lintdiff rule (only if it makes sense), otherwise a simple list with the links will be fine

msyyc: Consolidated into **`no-unsafe-patch-body-properties`**, using one PATCH-body traversal with distinct diagnostics for:

- Properties missing from the resource or at the wrong nesting level, compared by JSON name.
- Writable immutable fields: `id`, `name`, `type`, `location`, and `properties.provisioningState`.
- Effective PATCH input that is required, has a default, or lacks Update visibility.

Read/Create-only properties excluded from the effective request are not reported. The existing top-level `identity` exemption is retained for optionality/default/visibility checks, but not resource-shape checks.

I interpreted the array exception as allowing required properties in replacement elements, while retaining default, visibility, and resource-shape checks.

The [documentation maps each diagnostic to the corresponding legacy rule](https://github.com/msyyc/typespec-azure/blob/e9b88f465304f99ded9ded246331d3ae71215376/packages/typespec-azure-resource-manager/src/rules/no-unsafe-patch-body-properties.md#lintdiff-equivalent). #5384 and #5456 are closed as superseded by this PR.

`arm-resource-patch`, `patch-envelope`, and PATCH-to-PUT correspondence remain separate; default enablement is unchanged.

Issue #5254, opened 2026-08-18
Make TCGC decorator scope arguments evolvable with typed options


JoshLove-msft:
## Summary

Make TCGC decorators evolvable by introducing a shared application-options model and allowing scoped decorators to accept their final argument as `scope | optionsBag`, while preserving the existing string scope syntax.

This issue tracks the first workstream from the [spec-driven client generation project scope](https://gist.github.com/JoshLove-msft/ce64734dbc1034edca5e98949f0617f3): **Make decorators evolvable first**.

## Proposed direction

- Define a common typed options model for decorator application scope.
- Give each scoped decorator a typed options model that extends the common scope options.
- Centralize options normalization and validation so decorators do not independently implement compatibility behavior.
- Continue accepting positional string scope syntax during migration.
- For `@client`, add scope to the existing `ClientOptions`, temporarily retain the legacy third argument, and report a diagnostic when both forms specify conflicting scopes.

## Deliverables

- Update all scoped TCGC decorators to accept `scope | optionsBag` without breaking existing specifications.
- Add diagnostics for invalid options and conflicting scope declarations.
- Add code fixes and migration guidance for moving from positional scope syntax to options bags.
- Generate reference documentation for the typed decorator options.
- Define and document a deprecation policy for legacy positional scope syntax.
- Instrument or otherwise make it possible to measure legacy positional-scope and options-bag usage.

## Acceptance criteria

- [ ] A shared application-options contract is defined and used consistently by scoped decorators.
- [ ] Existing string scope syntax remains compatible.
- [ ] `@client` supports scope in `ClientOptions` and diagnoses conflicts with its legacy third argument.
- [ ] Compatibility tests cover positive, negative, grouped, and mixed scopes for both legacy and options-bag forms.
- [ ] Invalid and conflicting inputs produce actionable diagnostics.
- [ ] Migration code fixes and user-facing guidance are available.
- [ ] Generated decorator reference documentation reflects the new options models.
- [ ] The deprecation timeline and policy for positional scope syntax are documented.

## Outcome

Decorator signatures can evolve with additional typed settings without introducing more positional parameters or requiring another broad breaking change.

PR #5305, opened 2026-08-25
[tcgc] allow last param in decorator to accept scope or other options 

iscai-msft:
## Summary

First small step toward #5254 (Make TCGC decorator scope arguments evolvable with typed options).

This PR centralizes the type used for the `scope` argument across all scoped TCGC decorators into a single `Azure.ClientGenerator.Core.Scope` alias, defined once in `decorators.tsp` and reused from `legacy.tsp`. Having one shared alias means the scope type can evolve (e.g. toward a typed options bag as described in the issue) in one place instead of updating every decorator signature individually.

## Changes

- Added `alias Scope = string;` in `decorators.tsp` with doc comments describing supported language identifiers and valid patterns (moved from the previous per-decorator `@param scope` docs).
- Replaced every `scope?: valueof string` parameter in `decorators.tsp` with `scope?: valueof Scope`.
- Replaced every `scope?: valueof string` parameter in `legacy.tsp` with `scope?: valueof Azure.ClientGenerator.Core.Scope` (different namespace).
- Added tests in `test/decorators/scope.test.ts` verifying the alias can be referenced from user TypeSpec and that scoped decorators continue to behave identically.
- Added a changeset.

## Compatibility

No behavior or public API change — decorators still accept the existing string scope syntax (`"python"`, `"python, java"`, `"!csharp"`, etc.). Verified `generated-defs/*.ts` is unchanged after rebuilding, and the full TCGC test suite (1405 tests) passes.

## Testing

- `pnpm --filter @azure-tools/typespec-client-generator-core build`
- `pnpm --filter @azure-tools/typespec-client-generator-core test` (1405 passed, 2 skipped)
- `pnpm --filter @azure-tools/typespec-client-generator-core lint`
- `pnpm format`

Closes part of the first workstream in #5254; the remaining deliverables (typed options bag, diagnostics for conflicting scopes, codefixes, migration docs) are left for follow-up PRs.


packages/typespec-client-generator-core/design-docs/decorator-options-migration.md
```diff
@@ -0,0 +1,126 @@
...
+## Deprecation timeline
+
+There is currently **no deprecation timeline** for the legacy positional string scope syntax.
+Both forms (string and options bag) are supported side by side indefinitely. A future deprecation
+of the positional string form would only be considered once:
+
+1. Migration guidance has been available for at least one full TCGC release cycle.
+2. The majority of first-party TCGC-based emitters (C#, Java, Python, JavaScript/TypeScript, Go)
+   have migrated their own decorator usage to the options-bag form internally.
+3. Usage telemetry (see below) shows options-bag adoption is broadly viable for spec authors.
```
tadelesh: How can we telemetry this?

packages/typespec-client-generator-core/design-docs/decorator-options-migration.md
```diff
@@ -0,0 +1,126 @@
...
+
+If a deprecation is proposed in the future, it will follow the standard TypeSpec deprecation
+policy: a `@deprecated` warning period of at least two minor releases before any removal, tracked
+in a dedicated changeset and issue.
+
+## Migration guidance
```
tadelesh, line 61: My concern is when TypeSpec finish the conditional compile support, TCGC needs to remove all scope parameter, thus shall we flatten all options at that time from the option bag we introduce in current solution?

iscai-msft: no we should keep the options bag as is, just deprecate `scope` from the options bag. The options bag is so that decorators can gain new optional typed settings over time without more positional params.

when we drop `scope`, we will depreacte `scope` from the `DecoratorOptions` base options bag. Decorators with options bag that are only scope could keep the empty options bag as their last parameter for grow up, or we can also look into removal (i would say this is my preferred route since I don't love the vestigial parameter). Decorators with other values in the options bag will still keep those options bags

cc @timotheeguerin

packages/typespec-client-generator-core/design-docs/decorator-options-migration.md
iscai-msft: idk copilot just added that lmao, let me remove

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -155,7 +158,7 @@ export const $client: ClientDecorator = (
   context: DecoratorContext,
   target: Namespace | Interface,
   options?: Type,
-  scope?: LanguageScopes,
+  scope?: LanguageScopes | DecoratorOptions,
```
tadelesh, line 161: Is this correct? For legacy case, scope is a string param. For new case, scope should be in options. then options should be `DecoratorOptions & ClientOption` type?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -474,7 +498,7 @@ export const $protocolAPI: ProtocolAPIDecorator = (
   context: DecoratorContext,
   entity: Operation | Namespace | Interface,
   value?: boolean,
-  scope?: LanguageScopes,
+  scope?: LanguageScopes | DecoratorOptions,
 ) => {
   setScopedDecoratorData(context, $protocolAPI, protocolAPIKey, entity, value, scope);
```
tadelesh, line 503: I did not see you changed the `setScopedDecoratorData` with the new `DecoratorOptions` type. Then how could scope be set correctly?

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -172,6 +175,26 @@ export const $client: ClientDecorator = (
     options?.kind === "Model" ? options?.properties.get("service")?.type : undefined;
   const autoMergeServiceConfig =
     options?.kind === "Model" ? options?.properties.get("autoMergeService")?.type : undefined;
+  const optionsScopeConfig =
+    options?.kind === "Model" ? options?.properties.get("scope")?.type : undefined;
+  const optionsScope: string | undefined =
+    optionsScopeConfig?.kind === "String" ? optionsScopeConfig.value : undefined;
+  const legacyScope = normalizeScope(scope);
+
+  if (optionsScope !== undefined && legacyScope !== undefined && optionsScope !== legacyScope) {
```
JoshLove-msft: **P2: Compare scopes semantically before warning**

This compares the raw strings rather than the scopes they represent. For example, `"csharp, python"` and `"python,csharp"` select the same emitters but still produce `conflicting-scope`, which can fail consumers that treat warnings as errors. Please compare normalized parsed scope sets, including equivalent negation forms, before reporting a conflict.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1955,7 +1981,7 @@ export const $clientOption: ClientOptionDecorator = (
   target: Type,
   name: string,
   value: unknown,
-  scope?: LanguageScopes,
+  scope?: LanguageScopes | DecoratorOptions,
```
JoshLove-msft, line 2162: **P2: Normalize the options bag before requiring a scope**

With this expanded type, the `scope === undefined` validation below now checks the options object rather than its normalized scope. As a result, `@clientOption(..., #{})` bypasses `decorator-requires-scope`; `setScopedDecoratorData` then normalizes the empty bag to `undefined` and stores the option under `AllScopes`, causing it to silently apply to every emitter. Please validate `normalizeScope(scope)` and add an empty-options-bag regression test.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: yes it should be correct. Right now, the parameter `options` is `ClinetOptions`, and i've updated `ClientOptions` to extend from `DecoratorOptions`, so there's no need to type `options` as `DecoratorOptions & ClientOptions`

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: `setScopedDecoratorData` has been updated [here](https://github.com/Azure/typespec-azure/pull/5305/changes/BASE..7f04bd7f066ca2ccb075344cd936fdb9f7563020#diff-f947618675b2f7451341faa16ffbeed916f4a3150e9a62a6ec18725dc97e1f00R111), it's a little hard to see bc the function name isn't in the diff. It was updated to accept both and normalizes the two so there's really no difference in the function code

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -100,14 +102,37 @@ import { getSdkEnum, getSdkModel, getSdkUnion } from "./types.js";
 
 export const namespace = "Azure.ClientGenerator.Core";
 
+function isSemanticallyEqualScope(left: string, right: string): boolean {
+  const [leftNegationScopes, leftScopes] = parseScopes(left);
+  const [rightNegationScopes, rightScopes] = parseScopes(right);
+  return (
+    isEqualScopeList(leftNegationScopes, rightNegationScopes) &&
```
JoshLove-msft: **P2: Compare the effective mixed-scope selection**

Comparing the positive and negative arrays independently still produces false conflicts for valid mixed scopes. For example, `@client({ scope: "!java" }, "csharp, !java")` warns even though both values include C# and Python and exclude Java, matching the existing `isScopeApplicable` behavior. This can fail consumers that treat warnings as errors. Please canonicalize the effective selection (with negations, positives only remove overlapping exclusions) and add a mixed-scope conflict test.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
tadelesh: But why `scope` could be `DecoratorOptions`? There will be two option bag then.

JoshLove-msft: Can we make ClientOptions extend DecoratorOptions and just leave one options bag?

JoshLove-msft: Ah I see you have already done this. I agree we can leave scope parameter as is in this case. Otherwise we have two options bag parameters.

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: Fixed in 7ec17b9. Instead of comparing the positive/negative arrays independently, each scope string is now canonicalized via `getEffectiveScope` into the set of emitters it actually selects — either `all-except: [...]` (any negation present; positive scopes only cancel an overlapping exclusion, mirroring `setScopedDecoratorData`/`getScopedDecoratorData`) or `only: [...]`. The conflict is reported only when those canonical selections differ.

So `@client({ scope: "!java" }, "csharp, !java")` no longer warns, and neither does `"!(java, python)"` vs `"!python, !java"`, while genuinely different selections (e.g. `"!java"` vs `"csharp, !python"`, or `"!java"` vs `"csharp"`) still do. Added four regression tests in `test/decorators/client.test.ts` covering both directions.

packages/typespec-client-generator-core/src/decorators.ts
JoshLove-msft: Looks like we still have 2 options bags?

iscai-msft:
Good catch! I've updated the approach:

- Reverted `@client` and `@operationGroup` scope parameters back to `string` only (not `DecoratorOptions`)
- Made `ClientInitializationOptions extends DecoratorOptions` so it has access to the typed scope property
- Similarly, all decorators that have their own options model (`ClientOptions`, `ClientInitializationOptions`) now extend `DecoratorOptions` and read the scope from the options bag

This eliminates the dual-options-bags confusion. Decorators with existing options models handle scope through the model; decorators without an options model can continue using the trailing scope parameter.

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: oh i see what you mean. I ended up

- reverting `@client` and `@operationGroup` scope parameters back to `string` only (not `DecoratorOptions`)
- made `ClientInitializationOptions extends DecoratorOptions` so it has access to the typed scope property
- Similarly, all decorators that have their own options model (`ClientOptions`, `ClientInitializationOptions`) now extend `DecoratorOptions` and read the scope from the options bag

This eliminates the dual-options-bags confusion. Decorators with existing options models handle scope through the model; decorators without an options model can continue using the trailing scope parameter.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -185,6 +187,22 @@ export function getScopedDecoratorData(
...
+ * Centralizes normalization of the `scope` argument accepted by scoped TCGC decorators. Decorator
+ * implementations should call this instead of independently handling both the legacy plain-string
+ * form and the typed `DecoratorOptions` bag form, so any future evolution of the accepted shapes only
+ * needs to be implemented in one place.
+ *
+ * @param scope Either the legacy plain-string scope value, or a typed options bag (whose own
+ * decorator-specific model may extend `DecoratorOptions`) containing a `scope` property.
+ * @returns The plain-string scope value, or `undefined` if none was specified.
+ */
+export function normalizeScope(scope?: LanguageScopes | DecoratorOptions): string | undefined {
```
JoshLove-msft, line 200: **P1: Normalize the raw scope before materializing decorator arguments**

The options-bag form reaches `getTypeDecorators`, but that function first converts the `scope` model with `getDecoratorArgValue` and then passes the resulting SDK model to string-only `isScopeApplicable`. For example:

```typespec
@clientOption("flag", true, #{ scope: "python" })
model Payload {}
```

Once `Payload` is referenced by an operation, SDK package generation aborts with `TypeError: scope.match is not a function`. `@clientOption` is in the default decorator allowlist, so this affects normal generation. Please extract and normalize the raw scope before converting decorator arguments, and add an SDK-package materialization test for this form.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -100,14 +102,103 @@ import { getSdkEnum, getSdkModel, getSdkUnion } from "./types.js";
...
+ * @param legacyScope The legacy positional scope argument, if any.
+ * @returns The effective scope string, or `undefined` when no scope was specified.
+ */
+function resolveScopeFromOptions(
+  context: DecoratorContext,
+  decoratorName: string,
+  options: Type | undefined,
+  legacyScope?: string,
+): string | undefined {
+  const optionsScopeConfig =
```
JoshLove-msft, line 169: **P1: Preserve legacy parameter models that contain a `scope` property**

This helper reads `scope` from every model before determining whether the model is the new options wrapper or the supported legacy raw client-parameters model. As a result:

```typespec
model LegacyClientParameters {
  scope: "https://management.azure.com/.default";
}

@@clientInitialization(MyService, LegacyClientParameters);
```

interprets the actual client parameter as an emitter selector. The customization then silently disappears for normal emitters, with no diagnostic. Please identify the options-wrapper shape before consuming `scope` as decorator metadata, and add a compatibility test for a legacy parameter named `scope`.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1155,7 +1248,7 @@ export const $clientInitialization: ClientInitializationDecorator = (
       clientInitializationKey,
       target,
       options,
-      scope,
+      resolveScopeFromOptions(context, "clientInitialization", options, scope),
```
JoshLove-msft: **P2: Do not treat a scope-only options bag as client parameters**

After this stores `{ scope: "csharp" }` for the selected emitter, `getClientInitializationOptions` sees no `parameters` or `initializedBy` property and classifies the whole options bag as the legacy raw parameters model. The generated C# client then exposes an unintended public constructor parameter named `scope`.

Please distinguish a real `ClientInitializationOptions` bag from the legacy parameters form before returning the whole model as `parameters`, and cover the scope-only bag in an SDK-package materialization test.

--generated by Copilot

.chronus/changes/tcgc-scope-alias-2026-8-25-13-40-0.md
```diff
@@ -0,0 +1,9 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Make the `scope` argument accepted by scoped TCGC decorators evolvable via a shared, typed `Azure.ClientGenerator.Core.DecoratorOptions` model. Every scoped decorator now accepts either the legacy plain-string scope (e.g. `"csharp"`) or a typed options bag (e.g. `#{ scope: "csharp" }`), and individual decorators can later grow their own options model that extends `DecoratorOptions` without breaking others.
```
JoshLove-msft, line 7: **P2: Keep `@operationGroup` in the options-bag contract**

This says every scoped decorator accepts the typed options-bag form, but `@operationGroup` still declares only `scope?: valueof string` and has no separate options model. Consequently:

```typespec
@operationGroup(#{ scope: "python" })
interface Operations {}
```

fails with `invalid-argument`. Since `@operationGroup` does not already have an options bag, please update its extern/generated/runtime scope type to `DecoratorOptions | string` and normalize it before delegating to `@client`.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: Fixed in 90adbf13a. `getClientInitializationOptions` now returns `{ parameters: options }` only when the model is NOT a `ClientInitializationOptions` bag (via the same `isClientInitializationOptionsBag` predicate), so a scope-only bag like `{ scope: "csharp" }` no longer falls through to the legacy-parameters path and no longer surfaces a `scope` client-initialization parameter. Added a regression test asserting no `scope` parameter appears on the generated client for either the in-scope or out-of-scope emitter.

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: Fixed in 90adbf13a. Added an `isClientInitializationOptionsBag` predicate that identifies the options wrapper by its structural properties (`parameters`/`initializedBy`), and only treats an anonymous inline bag whose members are all known option keys (e.g. `{ scope: "csharp" }`) as the wrapper. A named legacy client-parameters model with a `scope` property is no longer read as an emitter selector — its `scope` stays a real client parameter. `$clientInitialization` only passes `options` into the scope resolver when it's a genuine bag. Added a regression test for a legacy `LegacyClientParameters { scope: "...default" }` passed directly.

.chronus/changes/tcgc-scope-alias-2026-8-25-13-40-0.md
iscai-msft: Fixed in 90adbf13a. `@operationGroup` now declares `scope?: valueof DecoratorOptions | string` (extern + regenerated generated-defs), and `$operationGroup` normalizes the scope via `normalizeScope` before delegating to `$client`, so `@operationGroup(#{ scope: "python" })` works instead of failing with `invalid-argument`. Added a regression test verifying the options-bag scope selects only the python emitter. The changeset's "every scoped decorator accepts the options-bag form" statement now holds.

packages/typespec-client-generator-core/src/internal-utils.ts
iscai-msft: Fixed in 90adbf13a. `getTypeDecorators` now special-cases the `scope` argument and runs it through `normalizeScope` (handling both the legacy string and the `DecoratorOptions` bag) before any decorator-argument materialization, so it's never sent through `getDecoratorArgValue`/`isScopeApplicable` as a model. Added a regression test that materializes an SDK package for a model carrying `@clientOption("flag", true, #{ scope: "python" })` and referenced by an operation — it previously threw `scope.match is not a function`.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1091,11 +1186,40 @@ export function getAlternateType(
...
+ */
+function isClientInitializationOptionsBag(options: Type): boolean {
+  if (options.kind !== "Model") {
+    return false;
+  }
+  if (options.properties.has("parameters") || options.properties.has("initializedBy")) {
+    return true;
+  }
+  return (
+    options.name === "" &&
```
JoshLove-msft: **P2: Use model provenance instead of anonymity to identify initialization options**

The `options.name === ""` heuristic reverses the intended behavior for two valid shapes:

```typespec
model CSharpOptions extends ClientInitializationOptions {
  scope: "csharp";
}

@@clientInitialization(MyService, CSharpOptions);
```

Because this options model is named, it is treated as the legacy raw parameters model; `scope` becomes a public constructor parameter for every emitter instead of selecting C#. Conversely, the previously supported anonymous legacy form:

```typespec
@@clientInitialization(
  MyService,
  { scope: "https://management.azure.com/.default" }
);
```

is now treated as the options bag, so the real client parameter disappears. Please use explicit `ClientInitializationOptions` ancestry/source-model information where available, and define a backward-compatible disambiguation for the anonymous scope-only shape rather than using the model name.

--generated by Copilot

iscai-msft: Fixed in bc0758c46. `isClientInitializationOptionsBag` now keys off model provenance rather than anonymity:

- `parameters`/`initializedBy` are declared only on `ClientInitializationOptions`, so their presence still unambiguously identifies the bag (covers the common inline `{ parameters, scope }` literal).
- A **named** model is an options bag when its `baseModel` chain reaches `ClientInitializationOptions` in `Azure.ClientGenerator.Core` — so your counterexample 1 (`model Foo extends ClientInitializationOptions { scope: "csharp"; }`) is now correctly treated as a bag, and `scope` selects the emitter instead of leaking as a client parameter.
- Anonymous literals are intentionally excluded from the ancestry check. The compiler links an inline literal to the expected `ClientInitializationOptions` parameter type, so an anonymous legacy model like `{ scope: "https://management.azure.com/.default" }` would otherwise look like a bag. Keeping the anonymous scope-only shape as legacy raw client parameters preserves backward compatibility — its `scope` stays a real client parameter (counterexample 2).

Added regression tests for all three shapes (named-extends bag with `parameters`, scope-only named-extends via ancestry, and the anonymous scope-only legacy form) in `client-initialization.test.ts`.

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -1091,11 +1186,56 @@ export function getAlternateType(
...
+    return false;
+  }
+  if (options.properties.has("parameters") || options.properties.has("initializedBy")) {
+    return true;
+  }
+  if (options.name === "") {
+    return false;
+  }
+  for (
+    let current: Model | undefined = options.baseModel;
```
JoshLove-msft: **P2: Traverse source-model provenance when identifying initialization options**

This follows only `baseModel`, so standard model transformations lose the options-bag identity. For example:

```typespec
model Base extends ClientInitializationOptions {
  scope: "csharp";
}

model Copy is PickProperties<Base, "scope">;
@@clientInitialization(MyService, Copy);
```

`Copy` has source-model provenance but no `baseModel`, so it is treated as legacy client parameters. The generated clients expose a public `scope` parameter for both C# and Python instead of using it to select C#. Please traverse explicit `sourceModel`/`sourceModels` provenance for named models as well as `baseModel`, and add coverage for `PickProperties`/`OmitProperties` or `model is`.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
```diff
@@ -100,14 +102,103 @@ import { getSdkEnum, getSdkModel, getSdkUnion } from "./types.js";
...
+ * @returns The effective scope string, or `undefined` when no scope was specified.
+ */
+function resolveScopeFromOptions(
+  context: DecoratorContext,
+  decoratorName: string,
+  options: Type | undefined,
+  legacyScope?: string,
+): string | undefined {
+  const optionsScopeConfig =
+    options?.kind === "Model" ? options.properties.get("scope")?.type : undefined;
```
JoshLove-msft: **P2: Resolve scope through the effective options inheritance chain**

This reads only the leaf model's `properties`, which excludes properties inherited with `extends`. For example:

```typespec
model Base extends ClientInitializationOptions {
  scope: "csharp";
}

model Final extends Base {
  parameters: Params;
}
```

Using `Final` with `@clientInitialization` applies the customization to Python as well as C#, and an inherited scope conflicting with the positional argument produces no warning. The same issue affects an inherited `ClientOptions.scope` on `@client`. Please resolve the nearest effective `scope` through the model inheritance chain and add multi-level inheritance tests for both decorators.

--generated by Copilot

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -524,7 +542,20 @@ export function getTypeDecorators(
...
         for (let i = 0; i < decorator.args.length; i++) {
-          decoratorInfo.arguments[decorator.definition.parameters[i].name] = diagnostics.pipe(
+          const parameterName = decorator.definition.parameters[i].name;
+          // The `scope` argument is emitter-selection metadata, not a client type. It can be
+          // provided either as the legacy plain string or as a typed options bag
+          // (`DecoratorOptions`). Normalize it to its plain-string form before materializing the
+          // remaining decorator arguments, otherwise the options-bag form is run through
+          // `getDecoratorArgValue` (which converts it to an SDK model) and later handed to the
+          // string-only `isScopeApplicable`, crashing with `scope.match is not a function`.
+          if (parameterName === "scope") {
```
JoshLove-msft: **P2: Preserve type-level string scopes on allowlisted decorators**

This special-cases every parameter named `scope`, but existing custom decorators can declare a type-level parameter:

```typespec
extern dec scopedMarker(target: unknown, scope: string);

@scopedMarker("python")
model Payload {}
```

For this signature, `jsValue` is a compiler string-literal type rather than a JavaScript string. Passing it directly to `normalizeScope` produces `undefined`, so emitted decorator metadata contains `scope: undefined` and the decorator is included for every emitter instead of only Python. Please preserve the normal `getDecoratorArgValue` conversion for compiler string literals, or restrict this normalization path to primitive strings and actual `DecoratorOptions` values.

--generated by Copilot

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: Good catch. `isClientInitializationOptionsBag` now delegates to a `derivesFromClientInitializationOptions` helper that walks `baseModel`, `sourceModel`, and `sourceModels[].model` with a visited set, so a scope-only copy declared via `model Copy is PickProperties<Base, "scope">` (or a spread) is recognized as deriving from `ClientInitializationOptions` and is no longer misclassified as a legacy positional argument. Added a regression test ("identifies a scope-only model via sourceModel provenance") in `client-initialization.test.ts`.

packages/typespec-client-generator-core/src/decorators.ts
iscai-msft: Fixed. `resolveScopeFromOptions` now resolves scope via a `getEffectiveScopePropertyType` helper that walks the `baseModel` chain to the nearest own `scope` property, instead of reading only the leaf `properties.get("scope")`. This applies to both `@clientInitialization` and inherited `ClientOptions.scope` on `@client`, since both route through `resolveScopeFromOptions`. Added multi-level inheritance regression tests in `client-initialization.test.ts` and `client.test.ts` (resolution plus a conflict-with-legacy-positional case).

packages/typespec-client-generator-core/src/internal-utils.ts
iscai-msft: Fixed. The `scope`-argument special-casing in `getTypeDecorators` is now guarded by an `isMarshalledScopeArg` check, so it only fires for the built-in marshalled forms (a legacy plain string or a `DecoratorOptions` bag — plain JS values with no `kind`/`entityKind`). A custom allowlisted decorator that declares a type-level `scope: string` parameter surfaces as a compiler string-literal type and now flows through the normal `getDecoratorArgValue` conversion, so its value is preserved (`scope: "python"` instead of `undefined`) and emitter-scope filtering still applies. Added a regression test ("preserves a custom decorator's type-level string scope argument") in `scope.test.ts`.

packages/typespec-client-generator-core/src/internal-utils.ts
```diff
@@ -524,7 +562,25 @@ export function getTypeDecorators(
...
+          // remaining decorator arguments, otherwise the options-bag form is run through
+          // `getDecoratorArgValue` (which converts it to an SDK model) and later handed to the
+          // string-only `isScopeApplicable`, crashing with `scope.match is not a function`.
+          //
+          // A custom decorator may instead declare a *type-level* `scope` parameter (e.g.
+          // `scope: string`), which surfaces here as a compiler string-literal type rather than a
+          // marshalled value. That must go through the normal `getDecoratorArgValue` conversion so
+          // its string value is preserved instead of being lost as `scope: undefined`.
+          if (parameterName === "scope" && isMarshalledScopeArg(decorator.args[i].jsValue)) {
+            decoratorInfo.arguments[parameterName] = normalizeScope(
```
timotheeguerin: does this not just drop the rest of the info?

iscai-msft: Good catch — it was collapsing the bag down to just the scope string. Fixed in 18148f205: we now keep the full marshalled value and only normalize to the string for the emitter applicability filter, so any extra `DecoratorOptions` fields survive instead of getting dropped. Added a regression test that fails without the change.

timotheeguerin (approved):

PR #5306, opened 2026-08-25
Fix default file upload Content-Type

### LINKED ISSUE #5288: [Bug]: Content-Type as */* is incorrect
### Describe the bug

When using `TypeSpec.Http.File`, if `ContentType` is not specified, TCGC models it as a `constant` with value `*/*`.  Note that `*/*` is not a valid MIME type and probably got confused with the `Accept` header.

Since `ContentType` is optional, if the use case doesn't specify any specific content types, then the `ContentType` param should be an optional string with a client-side default value of `application/octet-stream`.

### Reproduction

Build operation defined in https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/type/file/main.tsp#L96

### Checklist

iscai-msft:
Fixes #5288

## Summary
- Treat `*/*` from `@typespec/http` as the unconstrained `File` content-type marker, not a concrete request `Content-Type` value.
- For generated file upload clients without an explicit content type, expose `Content-Type` as an optional string with client default `application/octet-stream`.
- Keep download `Accept: */*` behavior unchanged.
- Update the file upload regression test and add a Chronus change entry.

## Validation
- `mise exec -- pnpm exec oxlint packages/typespec-client-generator-core/src/http.ts packages/typespec-client-generator-core/test/methods/file.test.ts --deny-warnings`
- `mise exec -- pnpm --filter @azure-tools/typespec-client-generator-core test -- methods/file.test.ts` *(blocked locally: this worktree has an unavailable/broken `core` submodule, so package imports from `@typespec/*` and `@azure-tools/typespec-azure-core` could not resolve)*


packages/typespec-client-generator-core/src/http.ts
```diff
@@ -385,9 +387,17 @@ function getSdkHttpParameters(
   }
   if (retval.bodyParam && !headerParams.some((h) => isContentTypeHeader(h))) {
     // if we have a body param and no content type header, we add one
+    // `*/*` is `@typespec/http`'s marker for an unconstrained (e.g. default File) content type,
+    // not a real value, so we exclude it from the doc. Besides being misleading, embedding a
+    // literal `*/*` in the doc string can prematurely close a generated `/** ... */` doc comment
+    // (e.g. in Java), so it must never end up there.
+    const knownContentTypes = retval.bodyParam.contentTypes.filter((ct) => ct !== "*/*");
```
tadelesh: Shall we leave it for languages? I believe every language has escape logic for special characters.

tadelesh (approved):

.chronus/changes/fix-file-content-type-2026-08-25.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@azure-tools/typespec-client-generator-core"
+---
+
+Use an optional string with an `application/octet-stream` client default for file uploads without an explicit content type.
```
tadelesh, line 7: I believe it is a behavior breaking change that we need to communicate with all languages.

packages/typespec-client-generator-core/src/http.ts
iscai-msft: we could, i was thinking that languages won't need to do anything with the catchall string and java was failing, but might as well pass it through and fix java

jorgerangel-msft (approved):

weidongxu-microsoft (approved):
LGTM

Java will need to do a release and check any GA lib need customize for this behavior change.

PR #5282, opened 2026-08-24
Add skill for TypeScript Spector tests

JialinHuang803:
## Summary

- add a project skill for implementing typespec-ts Spector integration coverage
- document focused client generation, mock-server validation, and tracked declaration regeneration
- include current package configuration and test naming templates

## Validation

- skill structure validation
- pnpm format
- pnpm lint

Example PR implemented by this skill https://github.com/Azure/typespec-azure/pull/5316.

PR #5312, opened 2026-08-26
[typespec-ts] Add Spector coverage report workflow

JialinHuang803:
## Summary

Adds a scheduled Spector coverage workflow for the TypeScript emitter. The workflow:

- Runs every Monday at 09:00 UTC+8 and supports manual dispatch.
- Builds `@azure-tools/typespec-ts`, runs its Spector suite, and reads `spector-coverage-typescript-azure.json`.
- Groups not-implemented scenarios, links each group to its source Spector case, and highlights newly discovered groups and scenarios.
- Preserves manually selected **Skip Implement** checkboxes and comments across report updates.
- Reads the dashboard tier configuration and automatically marks full Backlog groups, annotates mixed groups, and excludes Backlog scenarios from implementation tasks.
- Limits implementation tasks to tests for behavior already supported by the emitter; unsupported scenarios are reported rather than fixed in the task.
- Reuses an existing inactive implementation task instead of creating duplicate issues.
- Does not create or update a task while any generated task is assigned to Copilot or linked to an open implementation PR.
- Assigns created or refreshed tasks to `JialinHuang803` and `kazrael2119`.

## Workflow outputs

The workflow writes its results to GitHub issues; it does not commit generated files or publish a build artifact.

1. **Coverage report issue** — Creates or updates the **single** [[typespec-ts] Spector Coverage Report](https://github.com/Azure/typespec-azure/issues/5313) issue. It contains per-package pass/fail/not-implemented totals and coverage percentages, followed by grouped not-implemented scenarios with source links, preserved **Skip Implement** checkboxes, comments, and Backlog annotations. See the [report produced in the fork](https://github.com/JialinHuang803/typespec-azure/issues/2).
2. **Implementation task issue** — When actionable scenarios remain and no task is active, updates the newest inactive generated task or creates one if none exists. The task lists non-Backlog, non-skipped scenarios, requires a per-group success/failure report, and permits only passing test additions. It is assigned to `JialinHuang803` and `kazrael2119`. Manual runs can suppress this output with `report_only`. See the [task produced in the fork](https://github.com/JialinHuang803/typespec-azure/issues/4).

A task is considered active when it is assigned to Copilot or linked to an open implementation PR. In that case, the workflow leaves all existing task issues unchanged and creates no replacement. This behavior was exercised in the fork with active task #4 and implementation PR #7.

The implementation task uses the `typespec-ts-add-spector-test` skill from #5282.


.github/workflows/spector-coverage-report-typescript.yml
```diff
@@ -0,0 +1,601 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          submodules: recursive
+
+      - uses: ./.github/actions/setup
+
+      - name: Install dependencies
+        run: pnpm install
```
timotheeguerin:
```suggestion
        run: pnpm install --filter "@azure-tools/typespec-ts..."
```

you can do that too for slight speed bonus

.github/workflows/spector-coverage-report-typescript.yml
```diff
@@ -0,0 +1,601 @@
...
+        run: pnpm spector-test
+        timeout-minutes: 60
+
+      - name: Create or update coverage report issue
+        id: update-issue
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const fs = require("fs");
```
timotheeguerin: can this script be moved out of this into a ts file (and use modern js not common js)

.github/workflows/spector-coverage-report-typescript.yml
```diff
@@ -0,0 +1,601 @@
...
+                  ),
+                  comment: comment.trim(),
+                });
+              }
+            }
+
+            const validStatuses = new Set(["pass", "fail", "not-implemented"]);
+            const notImplementedByPackage = new Map();
+            const summaries = [];
+            const scenarioSources = new Map([
```
timotheeguerin: does the report job also need to check github api? feels a little of duplication/no good split between the job here

.github/workflows/spector-coverage-report-typescript.yml
```diff
@@ -0,0 +1,601 @@
...
+
+      - name: Build typespec-ts package
+        run: pnpm turbo run --filter "@azure-tools/typespec-ts..." build
+        timeout-minutes: 20
+
+      - name: Copy TypeSpec files
+        working-directory: packages/typespec-ts
+        run: pnpm copy:typespec
+
+      - name: Run Spector tests
```
timotheeguerin, line 53: Do you actually need to run the test, can you not just simply check the config and compare with the tests from spector?

JialinHuang803: The tests are run because the coverage JSON contains the most correct and direct status of the spector tests, while it's more complicated to check the spec config and search for tests that are skipped or partially implemented to get the final implementation results.

.github/workflows/spector-coverage-report-typescript.yml
JialinHuang803: The report job needs the GitHub api to locate the existing report and update the content. The task job uses the api for task detection and deduplication.
There is also a 'report-only' mode which can run only the report job when triggering the workflow manually.

timotheeguerin: can't it just find the issue in the next step and replace the content with the latest value, this all feels very complex here to reparse all the markdown expecting it to stay in the same format

.github/workflows/spector-coverage-report-typescript.yml
timotheeguerin: I see

.github/workflows/spector-coverage-report-typescript.yml
JialinHuang803: Seems @typespec/monorepo is required, so `"@azure-tools/typespec-ts..." --filter "@typespec/monorepo"` is added

.github/workflows/spector-coverage-report-typescript.yml
JialinHuang803: Moved the script to .github/scripts in 0dab7f7f00d592ee0054c655eff10f50bdaa139b.

.github/workflows/spector-coverage-report-typescript.yml
JialinHuang803: The idea of reparsing the existing markdown is to preserve manually edited Skip Implement checkboxes and Comments in the report and identify new scenarios. The report job doesn't fully replace the report content.

timotheeguerin: I mean is that really going to be reliable, this feels like if the comment is edited in the wrong way the parsing will fail and erase everything anyway.
There is explicit config you can have with spector to mark a test as unsupported we probably should rely on that instead no?

.github/scripts/spector-coverage-report-typescript.ts
```diff
@@ -0,0 +1,803 @@
...
+
+type Issue = {
+  number: number;
+  title: string;
+  body?: string | null;
+  pull_request?: object;
+  assignees?: Array<{ login: string }>;
+};
+
+type GitHub = {
```
timotheeguerin: can you not get those types from octokit/action lib?

.github/workflows/spector-coverage-report-typescript.yml
JialinHuang803: Yes, that makes sense. Let me investigate if we can leverage the spector.config.yaml to make things easier. 

Regarding the config to mark a test as unsupported, do you mean setting the scenario group to false as below?
```
# Disabled: multipart not yet supported
payload/multipart: false
```
I am thinking about marking the scenario group as `skipped` in the report if the scenario group is set to false in the config. However, this may introduce other problems. Seems we couldn't set a specific scenario to false in spector.config. If we want to implement some scenarios while leaving other scenarios in the same scenario group not-implemented, the scenario group in the config should still be set to `true`. 

And the reason why I added a Comment column to the report is that we can quickly scan or edit why we need to skip the tests. We had been using the report workflow for a while in the previous autorest.typescript repo, and most comments we added are just some texts or links which are less likely to cause the parsing to fail. Maybe we can try this and see what problems we can have then fix it.

.github/scripts/spector-coverage-report-typescript.ts
JialinHuang803: Updated to use types from Octokit.

timotheeguerin (approved):

PR #5318, opened 2026-08-26
[Swagger Linter Migration] QueryParametersInCollectionGet

msyyc:
## Original Swagger linter

- linter code: [QueryParametersInCollectionGet](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/query-parameters-in-collection-get.ts)
- linter doc: [query-parameters-in-collection-get.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/query-parameters-in-collection-get.md)

The original Swagger rule enforces the ARM collection-list query parameter contract:

- [x] Inspect each OpenAPI path item.
- [x] Select paths with a `get` operation that `isListOperationPath` classifies as a collection/list path.
- [x] Inspect the GET operation's `parameters` array.
- [x] Ignore non-query parameters.
- [x] Exempt exactly `api-version` and `$filter`, using case-sensitive names.
- [x] Report every remaining query parameter independently.

The Spectral function receives an object whose keys are OpenAPI path strings. For each path, it checks for a GET operation and delegates collection-shape detection to `isListOperationPath`. It filters the operation-level parameter array to query parameters whose names are neither `api-version` nor `$filter`, then emits one diagnostic per disallowed parameter.

Swagger diagnostics target the GET operation's `parameters` array rather than each parameter object. The validator operates on emitted OpenAPI occurrences, so the same semantic TypeSpec declaration can appear in multiple emitted files or versions. The promoted rule intentionally preserves the behavior, exemptions, and effective diagnostic cardinality without copying occurrence-based identities or emitted JSON locations.

- Validator rule ID: `QueryParametersInCollectionGet`
- Local lintdiff rule name: `query-parameters-in-collection-get`
- Canonical validator slug: `query-parameters-in-collection-get`
- Source branch: `feature/lintdiff-query-parameters-in-collection-get`, merged into `feature/lintdiff-migration-new`
- Source commit: `e7064db5a5574d08e28381cea4b9cfb45cf7521b`
- Source worktree: `C:\dev\worktrees\lintdiff-query-parameters-in-collection-get`
- Uncommitted source-rule changes: none
- Source: [`query-parameters-in-collection-get.ts`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/src/rules/query-parameters-in-collection-get.ts)

The user-marked done lintdiff source was treated as immutable and was not modified during promotion.

## Destination analysis

The rule belongs in `@azure-tools/typespec-azure-resource-manager`:

- its implementation depends on `getArmProviderNamespace`;
- it applies only to operations inside ARM provider namespaces, including child namespaces;
- fixture metadata and catalog applicability are `ARM`;
- the validation report resolves its fixtures to the resource-manager ruleset; and
- the rule checks ARM collection-path semantics rather than shared data-plane API style.

`@azure-tools/typespec-azure-core` was considered but rejected because removing ARM provider detection would broaden the done semantics, while adding an ARM dependency to Azure Core would violate package dependency direction. No equivalent official rule exists.

The promoted rule is named `collection-get-invalid-query-parameter` to follow the official subject-oriented naming convention. It listens to semantic operations, resolves each operation with `getHttpOperation`, and limits evaluation to GET operations in a resolved ARM provider namespace whose emitted path has collection shape. It permits only exact `api-version` and `$filter` query names.

Project-authored parameters receive diagnostics on their model properties. Parameters inherited from library models are retargeted to the local operation so authors can suppress or fix the diagnostic in their own source. Repeated semantic visits are deduplicated by resolved ARM provider identity, emitted HTTP path, and parameter name; identical paths in different providers remain distinct.

The rule evaluates the compiler program or projection supplied to the linter and therefore respects version projections without maintaining a separate version map. Its severity remains `warning`, matching the done TypeSpec source. Promotion-only adaptations are the concise official rule name, destination-relative ARM helper import, native docs metadata, and package/ruleset registration. The resource-manager ruleset lists the new rule as plain `false`, so promotion does not enable new diagnostics for existing services.

| --- | --- | --- |
| [`extra-query-param`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/extra-query-param/main.tsp) | `it("reports one extra query parameter on an ARM collection list operation", async () => {` | Registered ARM list operation with one disallowed parameter. |
| [`multiple-query-params`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/multiple-query-params/main.tsp) | `it("reports every extra query parameter on a collection GET", async () => {` | Preserves one diagnostic per disallowed parameter. |
| [`api-version-and-filter`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/api-version-and-filter/main.tsp) | `it("allows api-version and $filter on a collection GET", async () => {` | Exact standard exemptions are compliant. |
| [`raw-collection-get`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/raw-collection-get/main.tsp) | `it("reports a raw collection-shaped GET without ARM list registration", async () => {` | Path-based selection does not require ARM list registration. |
| [`mis-cased-filter`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/mis-cased-filter/main.tsp) | `it("reports a mis-cased $FILTER query parameter", async () => {` | Allowed names remain case-sensitive. |
| [`library-query-parameters`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/library-query-parameters/main.tsp) | `it("reports library-provided query parameters on the local operation", async () => {` | Uses `Azure.Core.StandardListQueryParameters` and asserts local operation targets. |
| [`nested-provider-namespace`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/nested-provider-namespace/main.tsp) | `it("checks operations in a child namespace of an ARM provider", async () => {` | Child namespace inherits ARM provider identity. |
| [`point-get-extra-query`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/point-get-extra-query/main.tsp) | `it("allows extra query parameters on a point GET", async () => {` | Point GET remains outside the rule. |
| [`non-get-collection`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/non-get-collection/main.tsp) | `it("allows extra query parameters on a non-GET collection operation", async () => {` | Non-GET operation remains outside the rule. |

## Migration evidence

The checked-in [`migration.md`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/QueryParametersInCollectionGet/migration.md) contains the focused fixture results, selected-version projection analysis, real-service comparison, full-corpus counts, compile failures, diagnostic-cardinality explanation, and remaining uncertainty.

## Validation

- Focused rule tests: 12 passed.
- ARM dependency-closure build passed.
- ARM package build and lint passed.
- ARM package tests: 383 passed.
- ARM docs regeneration passed; generated package and website indexes updated.
- Azure rulesets build passed; rulesets tests: 4 passed.
- Website dependency-closure build and website build passed.
- `pnpm chronus status`, `pnpm format`, `pnpm lint`, `pnpm run format:check`, and `git diff --check` passed.
- Focused promotion review found no source semantic issues; its generated-index finding was resolved by docs regeneration.

## Validation blocker

`pnpm validate:pr` completed its branch-up-to-date check, then produced no further output for five minutes. It was stopped at the workflow's bounded timeout after the narrower required validations above had passed.

## Promotion sync policy

If review discovers a semantic gap, this promotion should be blocked until the user explicitly reopens lintdiff repair. Source-rule changes must be completed and revalidated in the lintdiff workflow before being synchronized into this PR.


.chronus/changes/promote-query-parameters-in-collection-get-2026-08-26.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Add the `collection-get-invalid-query-parameter` ARM lint rule, disabled by default in the resource-manager ruleset.
```
timotheeguerin: `no-query-in-collection` to be inline with the other rules

packages/typespec-azure-resource-manager/src/rules/collection-get-invalid-query-parameter.ts
```diff
@@ -0,0 +1,103 @@
...
+    const reportedParameters = new Set<string>();
+
+    return {
+      operation: (operation) => {
+        const namespace = operation.interface?.namespace ?? operation.namespace;
+        if (namespace === undefined) {
+          return;
+        }
+
+        const providerNamespace = getArmProviderNamespace(context.program, namespace);
```
timotheeguerin: why do this rule need to be in there but not others? Unless really needed we should keep that simple and leave the filtering to the ruleset

packages/typespec-azure-resource-manager/src/rules/collection-get-invalid-query-parameter.ts
```diff
@@ -0,0 +1,103 @@
...
+        if (providerNamespace === undefined) {
+          return;
+        }
+
+        const [httpOperation] = getHttpOperation(context.program, operation);
+        if (httpOperation.verb !== "get" || !isCollectionPath(httpOperation.path)) {
+          return;
+        }
+
+        for (const parameter of httpOperation.parameters.parameters) {
```
timotheeguerin: use parameters.properties instead

.chronus/changes/promote-query-parameters-in-collection-get-2026-08-26.md
msyyc: Renamed the rule to `no-query-in-collection` for consistency with `no-query-in-point-op`, including its files, docs URL, tests, registrations, and changelog entry.

packages/typespec-azure-resource-manager/src/rules/collection-get-invalid-query-parameter.ts
msyyc: Removed the rule-level ARM provider namespace filtering and left applicability to the resource-manager ruleset. I also removed the provider-based deduplication and related tests.

packages/typespec-azure-resource-manager/src/rules/collection-get-invalid-query-parameter.ts
msyyc: Updated the rule to iterate over `httpOperation.parameters.properties` and use `kind`, `options.name`, and `property`. It now follows the same HTTP-property pattern as the related rules.

msyyc:
@timotheeguerin I updated PR with your comments and pls help review again. Thanks!

timotheeguerin (approved):

PR #5328, opened 2026-08-27
Fix TypeScript generation for multiple services

### LINKED ISSUE #5301: [typespec-ts] Multiple services bind ServiceA operation groups to ServiceB
## Description

When `@azure-tools/typespec-ts` generates a package containing multiple implicit root clients, the operation groups for the first service are incorrectly bound to the second service.

This is reproduced by the [`service/multiple-services`](https://github.com/Azure/typespec-azure/tree/main/packages/azure-http-specs/specs/service/multiple-services) Spector spec. The spec defines `ServiceAClient` and `ServiceBClient`, each with an `operations` group and a `subNamespace` group.

## Affected scenarios

- `Service_MultipleServices_ServiceA_Operations_opA`
- `Service_MultipleServices_ServiceA_SubNamespace_subOpA`

The corresponding ServiceB scenarios generate and run successfully:

- `Service_MultipleServices_ServiceB_Operations_opB`
- `Service_MultipleServices_ServiceB_SubNamespace_subOpB`

## Reproduction

Enable `service/multiple-services` in `packages/typespec-ts/spector.config.yaml`, generate the client, and run the integration test:

```powershell
pnpm --dir packages/typespec-ts generate-tsp-only:client
pnpm --dir packages/typespec-ts integration-test:alone -- service-multiple-services.test.ts
```

The failing calls are:

```ts
await serviceAClient.operations.opA({ apiVersion: "av2" });
await serviceAClient.subNamespace.subOpA({ apiVersion: "av2" });
```

## Actual behavior

Both ServiceA calls fail at runtime:

```text
TypeError: serviceAClient.operations.opA is not a function
TypeError: serviceAClient.subNamespace.subOpA is not a function
```

The generated declaration assigns ServiceB operation-group types to `ServiceAClient`:

```ts
export declare class ServiceAClient {
  readonly operations: ServiceBClientOperationsOperations;
  readonly subNamespace: ServiceBClientSubNamespaceOperations;
}
```

The generated ServiceA implementation also imports its operation-group factories from ServiceB:

```ts
import {
  OperationsOperations,
  _getOperationsOperations,
} from "../serviceB/classic/operations/index.js";
import {
  SubNamespaceOperations,
  _getSubNamespaceOperations,
} from "../serviceB/classic/subNamespace/index.js";
```

As a result, `serviceAClient.operations` exposes `opB` instead of `opA`, and `serviceAClient.subNamespace` exposes `subOpB` instead of `subOpA`.

## Expected behavior

Each root client should use the operation groups generated for its own service:

```ts
serviceAClient.operations.opA(...);
serviceAClient.subNamespace.subOpA(...);
serviceBClient.operations.opB(...);
serviceBClient.subNamespace.subOpB(...);
```

`ServiceAClient.operations` and `ServiceAClient.subNamespace` should use the ServiceA operation-group types and factories rather than the ServiceB variants.

## Versions

- `@azure-tools/typespec-ts`: `0.56.0`
- TypeSpec compiler: `1.14.0`

Related epic: #4738

JialinHuang803:
## Summary
- scope classic operation-group declarations and references to their owning client
- use each generated service context directly instead of invalid `Client.<context>` qualification
- trim the model namespace prefix shared by all services while preserving service-specific folders
- enable the `service/multiple-services` Spector case and add modular coverage for client and model hierarchy

## Validation
- `pnpm format`
- `pnpm lint`
- `pnpm -C packages/typespec-ts run build`
- `pnpm -C packages/typespec-ts run unit-test` (666 passed)
- `pnpm -C packages/typespec-ts run test-next` (254 passed)
- regenerated `service/multiple-services` successfully
- `service-multiple-services.test.ts` (4 passed)
- existing `service-multi-service.test.ts` (2 passed)


packages/typespec-ts/src/modular/emit-models.ts
```diff
@@ -325,22 +354,11 @@ export function getModelNamespaces(context: SdkContext, model: SdkType): string[
...
-    if (segments.length > rootNamespace.length) {
-      while (segments[0] === rootNamespace[0]) {
-        segments.shift();
-        rootNamespace.shift();
-      }
-      return segments;
-    }
-    return [];
+    const commonNamespace = getCommonServiceNamespace(context);
+    const modelUsesCommonNamespace = commonNamespace.every(
```
qiaozha, line 358: Not sure why we still need to leverage every in this case? can commonNamespace.length > 0 help?

packages/typespec-ts/src/modular/build-classical-client.ts
```diff
@@ -268,19 +260,14 @@ function buildClientOperationGroups(
...
-    clientType = `Client.${clientClass.getName()}`;
-  }
   const methodMap = getMethodHierarchiesMap(dpgContext, client);
   for (const [prefixKey, operations] of methodMap) {
     const prefixes = prefixKey.split("/");
     const layer = 0;
     if (prefixKey === "") {
       operations.forEach((op) => {
-        const methods = generateMethod(dpgContext, clientType, [prefixes, op]);
+        const methods = generateMethod(dpgContext, "Client", [prefixes, op]);
```
qiaozha: wonder why we still need this Client as parameter? Maybe we can move this within the generateMethod?

qiaozha (approved):

packages/typespec-ts/src/modular/build-classical-client.ts
JialinHuang803: Updated. `generateMethod` now uses the fixed `Client` context type internally, so callers no longer pass it as a parameter.

packages/typespec-ts/src/modular/emit-models.ts
JialinHuang803: `commonNamespace.length > 0` is not sufficient because models can be defined outside the common service namespace. For example, `Contoso.ServiceA` and `Contoso.ServiceB` can both use `Shared.CommonModel`. The common service prefix is `Contoso`, but the model namespace starts with `Shared`. Checking only the length would strip `Shared` and incorrectly emit the model under `models/models.ts`; the `every()` check preserves `models/shared/models.ts`.

```tsp
namespace Shared {
  model CommonModel {
    value: string;
  }
}

namespace Contoso {
  @service namespace ServiceA {
    @get @route("/a") op getA(): Shared.CommonModel;
  }

  @service namespace ServiceB {
    @get @route("/b") op getB(): Shared.CommonModel;
  }
}
```

Added a modular unit scenario covering this case.

PR #5331, opened 2026-08-27
Add TCGC exact-name support to TypeScript emitter

qiaozha:
## Summary

- honor TCGC `exact()` names across TypeScript clients, operations, parameters, models, properties, and enum members
- keep exact names consistent in serializers, classic clients, generated samples/tests, and client hierarchy paths
- report deduplicated warnings at emission sites when an exact name cannot form a valid TypeScript identifier
- add focused unit coverage and opt the emitter into the shared exact-name Spector scenario with a generated API baseline

## Validation

- `mise exec -- pnpm format`
- `mise exec -- pnpm lint`
- `mise exec -- pnpm --filter @azure-tools/typespec-ts build`
- `mise exec -- pnpm --dir packages/typespec-ts unit-test` (669 tests)
- `mise exec -- pnpm --dir packages/typespec-ts test-next` (254 tests)
- exact-name Spector integration (5 tests)


packages/typespec-ts/src/utils/name-utils.ts
```diff
@@ -208,6 +217,99 @@ export function normalizeName(
...
+
+export function isValidTypeScriptIdentifierName(name: string): boolean {
+  return /^[$_\p{ID_Start}][$\u200c\u200d\p{ID_Continue}]*$/u.test(name);
+}
+
+export function isValidTypeScriptIdentifier(name: string): boolean {
+  if (!isValidTypeScriptIdentifierName(name)) {
+    return false;
+  }
+  const result = ts.transpileModule(`export const ${name} = 0;`, {
```
JialinHuang803: Would this `transpileModule` more time-consuming than a normal regex? Since the `normalizeModelName` function that calls this function is called from many places, do we need to add a cache?

packages/typespec-ts/src/utils/name-utils.ts
```diff
@@ -208,6 +217,99 @@ export function normalizeName(
...
+  options: NormalizeNameOption = {},
+): string {
+  if (!sdkName.isExactName) {
+    return normalizeName(sdkName.name, nameType, options);
+  }
+  return sdkName.name;
+}
+
+export function normalizeSdkPropertyName(sdkName: SdkName): string {
+  return sdkName.isExactName ? sdkName.name : normalizeName(sdkName.name, NameType.Property);
```
JialinHuang803: Can this be `return normalizeSdkName(sdkName, NameType.Property);` to avoid duplicate implementation?

packages/typespec-ts/test/modular-unit/exact-name.test.ts
```diff
@@ -0,0 +1,285 @@
+import { afterAll, describe, expect, it } from "vitest";
```
JialinHuang803: I suggest we move the tests to scenarios/ and use the Markdown pattern.

packages/typespec-ts/src/modular/emit-models.ts
```diff
@@ -840,7 +853,16 @@ export function normalizeModelName(
     (isPagedResultModel(context, type) && !pagedModelsKeptPublic.has(type)) || type.isGeneratedName
       ? "_"
       : "";
-  return `${internalModelPrefix}${normalizeName(namespacePrefix + type.name, nameType, true)}${unionSuffix}`;
+  const normalizedNamespacePrefix =
```
JialinHuang803: I am not sure what the expected behavior should be when `enableModelNamespace` is true. Do we need to add the namespace prefix for the models with exact name? But I believe this is a rare case, just want to confirm the behavior. Maybe we can add e unit test for it as well.

packages/typespec-ts/src/modular/emit-models.ts
```diff
@@ -840,7 +853,16 @@ export function normalizeModelName(
...
+    type.isExactName && namespacePrefix
+      ? normalizeName(namespacePrefix, nameType)
+      : namespacePrefix;
+  const modelName = normalizeSdkName(
+    { name: normalizedNamespacePrefix + type.name, isExactName: type.isExactName },
+    nameType,
+    { shouldGuard: true },
+  );
+  return `${internalModelPrefix}${modelName}${unionSuffix}`;
```
JialinHuang803: Would it be clearer if we separate namespace prefix from model name as now they have different normalize logic:
```
const namespacePrefix = context.emitterOptions?.enableModelNamespace
  ? normalizeName(segments.join(""), nameType)
  : "";

const modelName = normalizeSdkName(type, nameType, { shouldGuard: true }});

return `${internalModelPrefix}${namespacePrefix}${modelName}${unionSuffix}`;
```

packages/typespec-ts/src/modular/build-classical-client.ts
```diff
@@ -289,7 +298,7 @@ function buildClientOperationGroups(
       const operationName = `_get${rawGroupName}Operations`;
       const propertyType = `${rawGroupName}Operations`;
       // The `groupName` is used to any places where we don't need normalized name again
-      const groupName = normalizeName(rawGroupName, NameType.Property);
+      const groupName = normalizeName(prefixes[0] ?? "", NameType.Property);
       const existProperty = clientClass.getProperties().filter((p) => {
         return p.getName() === normalizeName(groupName, NameType.Property);
```
JialinHuang803: Maybe it's an issue with the existing code, not introduced by this PR - seems like we don't need to call `normalizedName` again because the `groupName` is already normalized.

I just tried with nested operation groups:
```
namespace Operations {
  @route("/direct")
  @get
  op directOperation(): void;

  namespace ChildOne {
    @route("/child-one")
    @get
    op childOneOperation(): void;
  }

  namespace ChildTwo {
    @route("/child-two")
    @get
    op childTwoOperation(): void;
  }
}

#suppress "experimental-feature" "exact name test"
@@clientName(Operations, exact("my_group"));
```
And the generated code has three `my_group` properties in the classic client.

JialinHuang803 (approved):

PR #5334, opened 2026-08-27
[python] reenable streaming test

### LINKED ISSUE #5332: Re-enable streaming/sse spec for the Python emitter (duplicate model name across namespaces)
The `streaming/sse` spec is currently skipped for `@azure-tools/typespec-python` in [`packages/typespec-python/spector.config.yaml`](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-python/spector.config.yaml) and needs to be re-enabled.

## Why it is skipped

microsoft/typespec#11613 added a `Protocol` namespace to `streaming/sse/main.tsp` containing a `model Info`, alongside the pre-existing `Unnamed.Info`:

```tsp
namespace Streaming.Sse;

@route("unnamed")
namespace Unnamed {
  model Info {
    desc: string;
  }
  ...
}

@route("protocol")
namespace Protocol {
  model Info {
    message: string;
  }
  ...
}
```

The Python generator emits every model flat into a single `_models.py` per package, so the two namespaced models collide and generation fails hard:

```
File "generator/pygen/codegen/models/code_model.py", line 404, in _sort_model_types_helper
    raise ValueError(f"We have already generated a schema with name {current.name}")
ValueError: We have already generated a schema with name Info
```

This blocked the `core` submodule bump (build & regenerate job), so the spec was disabled to unblock it.

## Repro

```bash
cd packages/typespec-python
node ./eng/scripts/ci/regenerate.ts --name streaming/sse --flavor unbranded
```

## Possible resolutions

- Handle duplicate model names across TypeSpec namespaces in `@typespec/http-client-python` (disambiguate or nest by namespace), or
- Rename `Streaming.Sse.Protocol.Info` upstream in `microsoft/typespec` so spec model names stay unique.

Once resolved, flip `"streaming/sse"` back to `true` in `packages/typespec-python/spector.config.yaml` and add mock-api coverage for the new protocol scenarios.

iscai-msft:
fixes #5332


packages/typespec-python/spector.config.yaml
```diff
@@ -141,4 +141,4 @@ specs:
   type/file: false
   # `Unnamed.Info` and `Protocol.Info` collide once the Python emitter flattens
   # models into a single `_models.py`: https://github.com/Azure/typespec-azure/issues/5332
-  "streaming/sse": false
+  "streaming/sse": true
```
timotheeguerin: remove comment abovce?

timotheeguerin (approved):

PR #5336, opened 2026-08-28
[Swagger Linter Migration] GuidUsage

msyyc:
## Original Swagger linter

- linter code: [GuidUsage](https://github.com/Azure/azure-openapi-validator/blob/6243cb01c16c7535cd3b8df6f45fbeb3c095ed7f/packages/rulesets/src/spectral/az-arm.ts#L1227-L1237)
- linter doc: [guid-usage.md](https://github.com/Azure/azure-openapi-validator/blob/6243cb01c16c7535cd3b8df6f45fbeb3c095ed7f/docs/guid-usage.md)

The original rule performs these checks:

- [x] Traverses the unresolved ARM OpenAPI document recursively.
- [x] Selects every node whose property name is `format` and value is `uuid`.
- [x] Covers named definitions, parameters, request and response bodies, response headers, and any other emitted UUID format occurrence.
- [x] Reports an error at each emitted `format: uuid` node with Azure API review guidance.

The Spectral rule runs the unresolved JSONPath `$..[?(@property === 'format' && @ === 'uuid')]` and applies `falsy`. It has no authored-shape exemptions and does not resolve references before traversal. Its diagnostic location is the emitted format node, so one TypeSpec declaration may be reported repeatedly when it is expanded across paths, operations, schemas, visibility variants, or API versions.

That representation also produces findings which should not be copied into a native TypeSpec rule: imported `Azure.Core.RequestIdResponseHeader` expansions are library-owned, and retained Swagger can contain stale `Azure.Core.uuid` definitions absent from the current authored TypeSpec surface. The native rule therefore diagnoses actionable authored targets and deduplicates them rather than matching emitted occurrence counts.

- Validator rule ID: `GuidUsage`
- Local lintdiff rule: `tsp-lintdiff-local-linter/guid-usage`
- Canonical validator slug: `guid-usage`
- Source rule: [`guid-usage.ts`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/src/rules/guid-usage.ts)
- Source branch: `feature/lintdiff-guid-usage`
- Source commit used for promotion: `cf3f5db38bf3a540c929a1d2ba9b7fea1638d7f7`
- Source worktree: `C:\dev\worktrees\lintdiff-guid-usage`
- Uncommitted source-rule changes: none

The user marked this source rule done for promotion. The lintdiff source, fixtures, snapshots, manifests, and documentation were not modified during promotion.

## Destination analysis

The promoted rule belongs in `@azure-tools/typespec-azure-resource-manager`:

- validator metadata declares `applicability: ARM` and `sources: ["arm"]`
- every focused fixture is inferred as `resource-manager`
- the implementation needs ARM provider namespace discovery, resolved ARM resources, resource lifecycle/list/action operations, and generated resource-key parameters
- the migration evidence specifically evaluates ARM services and ARM emitter behavior

`@azure-tools/typespec-azure-core` was considered because its `no-format` rule recommends `Azure.Core.uuid` instead of `@format("uuid")`. It is not equivalent: `no-format` governs decorator usage across Azure APIs, while this rule prohibits semantic UUID wire shapes throughout ARM APIs, including UUID scalar aliases and template-generated parameters. Placing the rule in core would also require removing its ARM-only resource and provider semantics.

The official rule is named `no-uuid`, following the TypeSpec convention for a banned construct rather than carrying the validator slug into the user-facing rule name.

`@azure-tools/typespec-azure-resource-manager/no-uuid`:

- limits named-model traversal to ARM service namespaces
- resolves ARM HTTP services with `getAllHttpServices`
- checks parameters, request bodies, response bodies, and response headers
- recognizes scalar-level and property-level UUID formats
- follows custom scalar bases, arrays, records, tuples, unions, inherited properties, and cloned `sourceProperty` chains
- inspects unreferenced named models because the OpenAPI emitter can retain those definitions
- excludes imported library declarations and client-only namespaces
- deduplicates diagnostics by authored `ModelProperty` or `Operation`
- maps generated resource-name parameters back to authored ARM operations

Current `main` retains a library-owned source property for generated resource-key parameters. The official-package adaptation therefore prioritizes the resource-key operation fallback before the generic source-property lookup. This preserves the lintdiff rule's intended authored-operation diagnostic target for both typed and formatted `ResourceNameParameter` instantiations.

The rule is available but registered as `false` in the resource-manager ruleset so promotion does not introduce diagnostics into existing Azure service specs.

| --- | --- | --- |
| [`uuid-property`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-property/main.tsp) | `it("reports a UUID-typed model property")` | Direct UUID model property |
| [`uuid-query-parameter`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-query-parameter/main.tsp) | `it("reports a UUID-typed query parameter")` | Resolved HTTP query parameter |
| [`uuid-template-parameter`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-template-parameter/main.tsp) | `it("reports a UUID-typed resource name template parameter")` | Typed generated resource key; asserts four authored operation diagnostics |
| [`uuid-template-format-parameter`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-template-format-parameter/main.tsp) | `it("reports a UUID-formatted resource name template parameter")` | Formatted generated resource key; asserts four authored operation diagnostics |
| [`uuid-body`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-body/main.tsp) | `it("reports a direct UUID request body")` | Direct request payload |
| [`uuid-response`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-response/main.tsp) | `it("reports a direct UUID response body")` | Direct response payload |
| [`uuid-custom-scalar`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-custom-scalar/main.tsp) | `it("reports a custom scalar derived from UUID")` | Custom scalar base traversal |
| [`uuid-format-property`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-format-property/main.tsp) | `it("reports a property-level UUID format")` | Property format after model spread |
| [`uuid-array-property`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/uuid-array-property/main.tsp) | `it("reports an array property containing UUID values")` | Array container traversal |
| [`non-uuid-shapes`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/non-uuid-shapes/main.tsp) | `it("allows non-UUID shapes and UUIDs in client-only namespaces")` | Compliant strings and excluded client-only model |
| [`unreachable-uuid-property`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/unreachable-uuid-property/main.tsp) | `it("reports a UUID property on an unreferenced named model")` | Unreferenced emitted named model |
| N/A (review regression) | `it("reports UUID values nested in records, tuples, and unions")` | Container and union traversal |
| N/A (review regression) | `it("reports a UUID-typed response header")` | Authored response-header target |
| N/A (review regression) | `it("deduplicates a property reached through declarations and HTTP payloads")` | Authored-target deduplication |

## Migration evidence

See [`GuidUsage/migration.md`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-guid-usage/packages/typespec-lintdiff/test/fixtures/GuidUsage/migration.md) for the declared focused tests, real-service comparison, latest full-corpus counts, one-sided project explanations, compile failures, and remaining uncertainty.

## Validation

- ARM dependency closure build
- focused `no-uuid` test: 14 passed
- ARM package build and lint
- ARM documentation regeneration
- ARM package tests: 385 passed
- Azure rulesets build and tests: 4 passed
- repository format check and required pre-commit `pnpm format && pnpm lint`
- website dependency closure and website build
- Chronus status and LF line-ending validation
- two focused promotion code reviews with no high-confidence findings
- `git diff --check`

## Validation blocker

`pnpm validate:pr` reported that the already-tested branch was behind a newly advanced `origin/main`, then produced no further progress for five minutes. It was stopped under the bounded validation policy. The narrower native validations listed above completed successfully.

## Promotion sync policy

Any newly discovered semantic gap should block this promotion until the user explicitly reopens lintdiff repair. The done lintdiff source is not modified as part of promotion.

NOTE: heavy


packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
```diff
@@ -0,0 +1,280 @@
...
+            property.type,
+            target,
+            reportedTargets,
+            new Set(),
+            true,
+            property,
+          );
+        }
+      },
+      root: () => {
```
timotheeguerin, line 66: why do you need to make your own traverse here? can't you just use the linter engine, this seems like this should be a pretty simple rule

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
```diff
@@ -0,0 +1,280 @@
...
+  target: ModelProperty | Operation,
+  reportedTargets: Set<ModelProperty | Operation>,
+  seen = new Set<Type>(),
+  canReportTarget = true,
+  formatSource?: ModelProperty,
+): void {
+  const formattedProperty = formatSource ?? (target.kind === "ModelProperty" ? target : undefined);
+  if (
+    formattedProperty !== undefined &&
+    getFormat(context.program, formattedProperty) === "uuid" &&
```
timotheeguerin: do not rely on `getFormat` check for the actual scalar

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: The extra traversal is intentional because this rule is trying to cover the resolved HTTP projection of the API while reporting on actionable authored TypeSpec targets. The regular linter traversal handles ordinary declarations, but it does not fully cover generated ARM HTTP shapes with a useful diagnostic location.

For example:

```tsp
model Widget is TrackedResource<WidgetProperties> {
  ...ResourceNameParameter<
    Resource = Widget,
    KeyName = "widgetName",
    SegmentName = "widgets",
    Type = Azure.Core.uuid
  >;
}

@armResourceOperations
interface Widgets {
  get is ArmResourceRead<Widget>;
}
```

This emits an HTTP path parameter with `format: uuid`, which the Swagger `GuidUsage` rule reports. The corresponding `widgetName` model property is generated from an ARM library template and is library-owned, so reporting from a basic `modelProperty` listener would point into code the service author cannot meaningfully fix or suppress. The HTTP/resource traversal recognizes it as the resource key and maps the finding back to the authored `Widgets.get` operation.

The same projection traversal covers direct `ArmResponse<uuid>` bodies and formats applied to generated HTTP parameters. It also lets us exclude imported library-owned findings and deduplicate repeated emitted occurrences. These cases came from the lintdiff corpus investigation and have focused tests in this PR.

Some of the recursive type inspection may still be refactorable, but replacing the resolved HTTP traversal with only semantic listeners would narrow the migrated rule's established coverage or change its diagnostic targets.

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: Updated this to resolve `Azure.Core.uuid` once with `program.resolveTypeReference("Azure.Core.uuid")` and compare scalar identity while walking base scalars, following the existing `no-offsetdatetime` pattern.

I retained `getFormat` only as a separate detection path for explicitly formatted shapes, such as:

```tsp
@format("uuid")
scalar WidgetId extends string;
```

and property-level `@format("uuid")`, because those do not have `Azure.Core.uuid` identity but still emit `format: uuid`. Added focused regression coverage for the explicitly formatted custom scalar as well.

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
```diff
@@ -0,0 +1,312 @@
...
+    "ARM APIs should avoid UUID-typed schemas unless they have explicit Azure API review approval.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-uuid",
+  messages: {
+    default:
+      "UUID usage is not recommended. If UUIDs are required in your service, get sign-off from the Azure API review board.",
+  },
+  create(context) {
+    const reportedTargets = new Set<ModelProperty | Operation>();
+    const [uuidType] = context.program.resolveTypeReference("Azure.Core.uuid");
```
timotheeguerin: use typekit `$.type.resolve()`

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
timotheeguerin: hhm really don't like that by that reasoning every single rule should do their own traversal due to the same limitation. I think if this is critical and can't be hacked in the diagnostic target resolution function right now we need to figure out a better way built in for this.

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: Updated to use `$(context.program).type.resolve("Azure.Core.uuid", "Scalar")` in 44bab683a.

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: Can't find a better solution. Let us hold on this PR for now and see if other PRs have similar request for traversal or we could find other solution. CC @catalinaperalta

timotheeguerin: can you not just check modelProperty type(and any other types you want to check) and report on the model property? The typespec engine should report template instantiation trace if it happens in a template

msyyc: @timotheeguerin  Yes, this works for the ordinary cases, and I refactored the rule in afd2b72bb to use `modelProperty` and `operation` listeners for authored properties, parameters, headers, direct return types, containers, unions, and custom scalars. The resolved HTTP traversal for ordinary parameters, request/response bodies, and headers has been removed.

There is one ARM-specific exception. An instantiated `ResourceNameParameter.name` property does carry a template mapper, so the compiler can produce an instantiation trace. However, its primary source location remains the library declaration in `Azure.ResourceManager.ResourceNameParameter`. The linter filters library-targeted diagnostics before presenting related instantiation locations, so reporting that generated property directly produces no user-visible diagnostic.

This is not just theoretical in the migration corpus: there are 9 UUID resource-name declarations across 6 projects (8 across 5 successfully compiled projects; Quota was excluded by its compile failure), producing 35 Swagger operation-parameter findings. Dropping the case would therefore create a known gap in 5 of the 46 projects detected by the final TypeSpec corpus run. (detailed report is here: https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GuidUsage/migration.md)

**As a compromise, the only remaining HTTP projection logic now finds resolved ARM resource-key parameters and maps those library-owned generated parameters to their authored operations. Everything else uses the standard linter traversal. I also added coverage for `ArmResponse<Azure.Core.uuid>` to ensure a UUID hidden behind a library response wrapper remains detected without recursively reporting project response models. The implementation is reduced by 184 lines, and all 16 focused tests still pass.**

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
```diff
@@ -0,0 +1,190 @@
...
+    const resourceKeyByOperation = getResourceKeyByOperation(context.program);
+
+    return {
+      modelProperty: (property) => {
+        if (!isInArmService(property.model?.namespace, armServices)) {
+          return;
+        }
+
+        if (
+          getFormat(context.program, property) === "uuid" ||
```
timotheeguerin: remove `getFormat` this doesn't count as a uuid and we have a separate linter to prevent usage of `@format`

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
```diff
@@ -0,0 +1,190 @@
...
+            if (resourceKey === undefined) {
+              continue;
+            }
+
+            const parameter = httpOperation.parameters.parameters.find(
+              (parameter) => parameter.param.name === resourceKey,
+            );
+            if (
+              parameter !== undefined &&
+              (getFormat(context.program, parameter.param) === "uuid" ||
```
timotheeguerin: format usage here too

msyyc:
The migration-skill audit found a **source-provenance verification gap**, rather than a confirmed production-rule violation.

The initial code review used `5dded4ae52306f000067a91e2b62370655f952ed`. The head subsequently advanced to `ab6238cb21a30922948e0ca8f533d1bed4544160`; before posting, I rechecked the current PR description and its source references. This comment is limited to that still-current provenance issue, not a claim to have re-audited the new code head.

The description identifies:

- source branch `feature/lintdiff-guid-usage`;
- source commit `cf3f5db38bf3a540c929a1d2ba9b7fea1638d7f7`;
- source-rule, fixture, and `GuidUsage/migration.md` links through that branch.

The GitHub contents API returns 404 for the source rule at that branch, and the commit API returns 422 / "No commit found" for the recorded SHA in `Azure/typespec-azure`. Consequently, readers cannot currently reproduce the source-to-native mapping or inspect the linked migration evidence from the declared provenance.

Please restore the intended source ref or correct the branch/commit and evidence links to a resolvable source. Preserve the actual immutable source identity; if the historical commit exists only locally, explain that limitation and provide verifiable provenance rather than silently substituting a newer implementation.

This follows the [promotion skill's source identification and PR evidence requirements](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#10-review-commit-push-and-create-a-draft-pr). An inaccessible reference does not prove the original validation or source work did not happen.

.chronus/changes/promote-guid-usage-2026-08-28.md
timotheeguerin, line 1: so @markcowl mentioned that the ARM team was looking to actually get rid of this rule

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: Removed the format-based generated resource-parameter check and its format-specific regression coverage in 6e749fc50. The ARM fallback now checks only whether the resolved parameter type is `Azure.Core.uuid` or derives from it.

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: Removed all `getFormat` handling in 6e749fc50. `no-uuid` now detects only `Azure.Core.uuid` scalar identity (including derived scalars); explicit `@format("uuid")` usage remains the responsibility of the enabled Azure Core `no-format` rule.

markcowl (changes requested):
Correct, the ARM team wants to make this rule a 'warning' in lintdiff, which means we should not cover it.  There are too many valid usages of GUID.

Suhas asked me to create a pr changing the status of this lintdiff rule last week.

.chronus/changes/promote-guid-usage-2026-08-28.md
msyyc: Confirmed by Mark's changes-requested review: the ARM team intends to classify `GuidUsage` as a lintdiff warning because there are too many valid GUID usages, so it should not be promoted. I’ll close this PR without merging.

packages/typespec-azure-resource-manager/src/rules/no-uuid.ts
msyyc: Closing this as moot based on the ARM policy decision confirmed by Mark: `GuidUsage` will be treated as a lintdiff warning and should not be promoted to a native TypeSpec rule. We will close this PR rather than continue refining the traversal.

msyyc:
I understand that we don't need this PR anymore? @markcowl

catalinaperalta:
@msyyc why was this PR reopened?

msyyc:
> Yuchao Yan (@msyyc) why was this PR reopened?

Before receiving explicit confirmation from MarkC, it is better to keep the PR open to track for now.

PR #5358, opened 2026-09-01
[Swagger Linter Migration] GetCollectionOnlyHasValueAndNextLink

msyyc:
## Original Swagger linter

- linter code: [GetCollectionOnlyHasValueAndNextLink](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/get-collection-only-has-value-and-next-link.ts)
- linter doc: [get-collection-only-has-value-and-next-link.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/get-collection-only-has-value-and-next-link.md)

The original rule performs these checks:

- Selects resolved `properties` objects under `200` responses for ARM GET operations in both `paths` and `x-ms-paths`.
- Excludes raw path keys ending in `}`, `operations`, or `default`.
- Uses the path portion after the provider namespace and treats an even provider-tail segment count as a collection path.
- Requires exactly two response-envelope properties.
- Requires those properties to be named `value` and `nextLink`.

The Spectral selector traverses resolved Swagger `paths` and `x-ms-paths`, selects each eligible GET operation's `responses.200.schema.properties`, and passes that object plus its JSON path to the custom function. The function derives the provider tail by splitting a path component containing `.` and then `/`; odd tail lengths are treated as point operations and skipped. For collection-shaped tails, it reports when the property map does not contain exactly `value` and `nextLink`.

The selector's suffix exclusions run against the raw Swagger path key. Therefore, `.../operations` and `.../default` are excluded, while the same paths followed by a query suffix are not. Direct arrays, property-less objects, file responses, and multipart responses have no selected `schema.properties` node and are skipped. Because Swagger diagnostics are attached to emitted OpenAPI occurrences, one authored TypeSpec declaration can produce several validator diagnostics; this migration intentionally reports semantic TypeSpec targets instead of reproducing emitted-occurrence duplication.

- Validator rule ID: `GetCollectionOnlyHasValueAndNextLink`
- Local lintdiff rule: `get-collection-only-has-value-and-next-link`
- Canonical validator slug: `get-collection-only-has-value-and-next-link`
- Source PR: [#5310](https://github.com/Azure/typespec-azure/pull/5310)
- Source branch: `feature/lintdiff-get-collection-only-value-next-link` at `efee405b7e91ad3774d6ec0b9e77fffa53c96e51`
- Inspected source worktree: `C:\dev\worktrees\lintdiff-get-collection-only-value-next-link`
- [Source lintdiff rule](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/src/rules/get-collection-only-has-value-and-next-link.ts)

The inspected source had no uncommitted rule or fixture changes and matched the merged source branch. The user-marked-done lintdiff source was not modified during promotion.

## Destination analysis

The rule belongs in `@azure-tools/typespec-azure-resource-manager`. Its metadata and fixtures are ARM-only, its semantics depend on provider namespace and ARM collection-path conventions, and its implementation needs `resolveProviderNamespace`. The validation report also infers the resource-manager ruleset.

`@azure-tools/typespec-azure-core` was considered but rejected: this is not a shared data-plane response-envelope rule, and moving it to core would either introduce an invalid dependency on the ARM library or weaken the ARM-specific provider and path semantics.

The official rule is named `collection-response-only-value-and-next-link` and listens to semantic operations. It:

1. Uses `resolveProviderNamespace` to limit evaluation to ARM provider namespaces.
2. Uses `getHttpOperation` to inspect the projected HTTP GET operation and its raw route.
3. Reproduces the validator's provider-tail parity and raw `operations`/`default` suffix behavior.
4. Finds a single-model `200` response body while skipping direct arrays, property-less models, file responses, and multipart responses that the Swagger selector cannot reach.
5. Requires exactly the `value` and `nextLink` properties.
6. Targets the first authored extra property, otherwise the response model; when the response model comes from library code, it falls back to the authored operation or interface.

The compiler invokes the rule over the active projection, so no separate version-state traversal is needed. It emits at most one semantic diagnostic per operation rather than duplicating diagnostics for every emitted Swagger occurrence. The rule is registered as available but disabled by default in the ARM ruleset.

| --- | --- | --- |
| [`extra-collection-props`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/extra-collection-props/main.tsp) | `it("reports an extra property on a collection response")` | Reports an authored extra envelope property. |
| [`extension-scope-value-only`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/extension-scope-value-only/main.tsp) | `it("reports a value-only response on an extension-scope collection path")` | Covers provider-tail collection detection and a missing `nextLink`. |
| [`only-value-and-nextlink`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/only-value-and-nextlink/main.tsp) | `it("accepts a response containing only value and nextLink")` | Accepts the required two-property envelope. |
| [`array-response-body`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/array-response-body/main.tsp) | `it("accepts a named array response body")` | Preserves the missing `schema.properties` exemption for a named array. |
| [`direct-array-response-body`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/direct-array-response-body/main.tsp) | `it("accepts a direct array response body")` | Preserves the direct-array exemption. |
| [`record-response-body`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/record-response-body/main.tsp) | `it("accepts a record response body")` | Preserves the property-less object exemption. |
| [`file-response-body`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/file-response-body/main.tsp) | `it("accepts a file response body")` | Preserves the Swagger file-schema exemption. |
| [`multipart-response-body`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/multipart-response-body/main.tsp) | `it("accepts a multipart response body")` | Preserves the Swagger multipart/string-schema exemption. |
| [`terminal-resource-invalid-response`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/terminal-resource-invalid-response/main.tsp) | `it("accepts an invalid collection shape on a point path with a query suffix")` | Confirms query stripping for provider-tail point classification. |
| [`operations-suffix-invalid-response`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/operations-suffix-invalid-response/main.tsp) | `it("accepts an invalid collection shape when the raw path ends with operations")` | Preserves the raw suffix exclusion. |
| [`operations-query-suffix-invalid-response`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/operations-query-suffix-invalid-response/main.tsp) | `it("reports an invalid collection shape when operations is followed by a query suffix")` | Preserves the validator's raw-path query behavior. |
| [`default-suffix-invalid-response`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/default-suffix-invalid-response/main.tsp) | `it("accepts an invalid collection shape when the raw path ends with default")` | Preserves the raw suffix exclusion. |
| [`default-query-suffix-invalid-response`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/default-query-suffix-invalid-response/main.tsp) | `it("reports an invalid collection shape when default is followed by a query suffix")` | Preserves the validator's raw-path query behavior. |

## Migration evidence

The checked-in [migration evidence](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/GetCollectionOnlyHasValueAndNextLink/migration.md) records the focused fixture results, investigated real-service projects, final full-corpus comparison, compile-failure exclusions, emitted-occurrence count differences, and remaining uncertainty.

## Validation

- ARM dependency closure and target package build
- Native rule test: 13 passed
- ARM package lint
- ARM documentation regeneration
- Azure rulesets build and test: 4 passed
- Full ARM suite: 382 passed with two unrelated timeout cases; both timeout files passed on focused rerun (29 tests)
- Focused promotion code review: no findings
- `pnpm validate:pr --skip-build --skip-test`: branch, lint, format, spelling, changeset, and diff checks passed
- Local website build intentionally skipped; the dedicated CI Website job owns full documentation regeneration, Astro checking, and website building

## Promotion sync policy

If review finds a semantic gap in the user-marked-done lintdiff source, this promotion must pause until the user explicitly reopens lintdiff repair. Promotion-only adaptations can be fixed here, but the source rule will not be changed implicitly.


packages/typespec-azure-resource-manager/src/rules/collection-response-only-value-and-next-link.ts
```diff
@@ -0,0 +1,124 @@
...
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/collection-response-only-value-and-next-link",
+  messages: {
+    default:
+      "Collection GET response models must declare exactly the `value` and `nextLink` properties.",
+  },
+  create(context) {
+    return {
+      operation: (operation) => {
+        const namespace = operation.interface?.namespace ?? operation.namespace;
+        if (resolveProviderNamespace(context.program, namespace) === undefined) {
```
timotheeguerin: necessary?

packages/typespec-azure-resource-manager/src/rules/list-response-envelope.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        context.reportDiagnostic({
+          target: getDiagnosticTarget(context.program, operation, invalidTarget),
+        });
+      },
+    };
+  },
+});
+
+function isCollectionGetPath(path: string): boolean {
```
timotheeguerin, line 56: This also seems like something that could be moved to a common helper file/something to figure out what kind of arm operation is an endpoint

.chronus/changes/promote-collection-response-rule-arm.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add the `collection-response-only-value-and-next-link` rule for ARM collection GET response envelopes.
```
timotheeguerin: `list-response-envelope`? inline with https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/patch-envelope/

packages/typespec-azure-resource-manager/src/rules/collection-response-only-value-and-next-link.ts
```diff
@@ -0,0 +1,124 @@
...
+      }
+
+      return bodyModel;
+    }
+  }
+
+  return undefined;
+}
+
+function getInvalidTarget(responseModel: Model): Model | ModelProperty | undefined {
```
timotheeguerin: should that not report for every single extra property?

packages/typespec-azure-resource-manager/src/rules/list-response-envelope.ts
msyyc: You left similar comment in other PRs and I will do the optimization after merging these PRs. Here is track issue https://github.com/Azure/typespec-azure/issues/5374

packages/typespec-azure-resource-manager/src/rules/collection-response-only-value-and-next-link.ts
msyyc: The ARM scope check is necessary because linter visitors also see operations outside ARM provider namespaces. I replaced `resolveProviderNamespace` with `getArmProviderNamespace`, which checks the operation's namespace ancestry and correctly handles nested namespaces under an ARM provider.

.chronus/changes/promote-collection-response-rule-arm.md
msyyc: Renamed the rule to `list-response-envelope` for consistency with `patch-envelope`, including the implementation, tests, ruleset registration, changeset, and generated documentation.

packages/typespec-azure-resource-manager/src/rules/collection-response-only-value-and-next-link.ts
msyyc: The original Swagger rule reports once for the response property map, so the initial implementation preserved that diagnostic-count parity. Reporting each authored extra property is better TypeSpec-native UX, though, so I adopted the improvement. The rule now reports every extra property and additionally targets the response model when `value` or `nextLink` is missing.

packages/typespec-azure-resource-manager/src/rules/list-response-envelope.ts
```diff
@@ -0,0 +1,124 @@
...
+
+function getInvalidTargets(responseModel: Model): (Model | ModelProperty)[] {
+  const properties = [...responseModel.properties.values()];
+  const invalidTargets: (Model | ModelProperty)[] = properties.filter(
+    (property) => property.name !== "value" && property.name !== "nextLink",
+  );
+
+  if (
+    !properties.some((property) => property.name === "value") ||
+    !properties.some((property) => property.name === "nextLink")
```
timotheeguerin: combine in the same `.some` to not iterate over it twice?

timotheeguerin: also isn't that going to create an extra diagnostic?

msyyc: Good points. I changed this to classify the properties in one pass. It still reports every extra property, but now reports the response model for missing `value` or `nextLink` only when there are no extras, avoiding the redundant diagnostic.

msyyc:
@timotheeguerin updated with your latest comments and pls help take a review again. Thanks!

timotheeguerin (approved):

PR #5361, opened 2026-09-01
[Swagger Linter Migration] ParametersSchemaAsTypeObject

msyyc:
## Summary

Promotes the ARM request-body guideline represented by Swagger validator rule [`ParametersSchemaAsTypeObject`](https://github.com/Azure/azure-openapi-validator/blob/main/docs/parameters-schema-as-type-object.md) as the idiomatic TypeSpec rule `use-model-request-body`.

## TypeSpec-first behavior

The native rule validates authored TypeSpec semantics rather than reproducing AutoRest's emitted Swagger schema details:

- request bodies must use a plain `Model` without an indexer
- named and inline models are allowed
- model-property references are checked using their underlying type
- arrays, records, primitives, unions, files, and other non-model body shapes are rejected
- absent bodies, synthetic `void` bodies, and multipart bodies are exempt
- encoding, format, schema-reference, and emitter-specific schema replacement behavior is intentionally not simulated

This follows the review decision to prefer idiomatic TypeSpec validation and preserve Swagger parity only where it maps naturally to TypeSpec. Constructs already rejected by Azure rules or emitter diagnostics do not justify additional AutoRest simulation in this rule.

## Relationship to lintdiff evidence

The original lintdiff [`migration.md`](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-migration-new/packages/typespec-lintdiff/test/fixtures/ParametersSchemaAsTypeObject/migration.md) records the earlier exact-emission parity investigation and remains useful historical evidence. Its exact diagnostic-parity conclusion is not the acceptance criterion for this promoted rule; differences caused by invalid or non-idiomatic TypeSpec are intentional. The source document lives only on `feature/lintdiff-migration-new`, so revising it requires a separate change on that branch rather than adding the lintdiff package to this `main`-targeted PR.

## Validation

- 10 focused `use-model-request-body` tests pass
- ARM package build and lint pass
- generated ARM rule references updated
- Azure rulesets build and 4 ruleset-definition tests pass


packages/typespec-azure-resource-manager/src/rules/request-body-must-be-object.md
```diff
@@ -0,0 +1,54 @@
...
+@post
+op submitItems(@body body: SubmitItemsRequest): void;
+
+@post
+op submitNullable(@body body: SubmitRequest | null): void;
+```
+
+## LintDiff Equivalent
+
+This rule corresponds to the Swagger validator rule [ParametersSchemaAsTypeObject](https://github.com/Azure/azure-openapi-validator/blob/main/docs/parameters-schema-as-type-object.md).
```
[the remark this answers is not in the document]
msyyc: Fixed in 3e0e8b92627b87f05ca5b3cea06476cbb3aca89a by adding suppression guidance limited to preserving compatibility for an existing API; new APIs should use an object model.

packages/typespec-azure-resource-manager/src/linter.ts
```diff
@@ -83,6 +84,7 @@ const rules = [
   operationsInterfaceMissingRule,
   patchEnvelopePropertiesRules,
   patchOperationsRule,
+  requestBodyMustBeObjectRule,
```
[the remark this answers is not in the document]
msyyc: Fixed in 3e0e8b92627b87f05ca5b3cea06476cbb3aca89a by adding the alphabetized `request-body-must-be-object` entry to the ARM rules inventory with its LintDiff mapping and impact.

packages/typespec-azure-resource-manager/src/rules/request-body-must-be-object.ts
```diff
@@ -0,0 +1,364 @@
...
+      schema = {
+        hasType: encodedSchema.hasType,
+        format: isSupportedAutorestFormat(mergedFormat) ? mergedFormat : schema.format,
+      };
+    }
+  }
+  return schema;
+}
+
+const allowedAutorestFormats = new Set([
```
timotheeguerin: why do we have that, format shouldn't be used in typespec azure specs

.chronus/changes/promote-parameters-schema-as-type-object-2026-08-31.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Add the `request-body-must-be-object` ARM lint rule, migrated from the Swagger `ParametersSchemaAsTypeObject` validator rule and disabled by default in the resource-manager ruleset.
```
timotheeguerin: object is a swagger/json schema concept.

I propose `use-model-request-body`
is there no rule for response?

msyyc: Renamed the rule to `use-model-request-body` across its implementation, diagnostics, tests, ruleset, docs, and change description in 89df075a7. The migrated `ParametersSchemaAsTypeObject` validator is request-only, and there is no general rule requiring every response body to be a model; existing response rules cover body presence, status codes, and ARM resource-operation response shapes, so response validation remains out of scope here.

packages/typespec-azure-resource-manager/src/rules/request-body-must-be-object.ts
msyyc: Removed explicit `@format` handling and the duplicated AutoRest format allowlist in 89df075a7. The remaining scalar metadata is only used to resolve standard TypeSpec scalar and `@encode` behavior; Azure specs continue to rely on the existing `no-format` rule.

msyyc:
@timotheeguerin updated with your comments and pls take a review again.

packages/typespec-azure-resource-manager/src/rules/use-model-request-body.ts
```diff
@@ -0,0 +1,318 @@
...
+      },
+    };
+  },
+});
+
+function isModelSchemaType(type: Type): boolean {
+  return type.kind === "Model" && !isArraySchemaType(type);
+}
+
+function getEmittedSchemaType(
```
timotheeguerin: can you explain + add some docs on why this need all this logic, would that already have been resolved by the http operation body, we check if that type is a model otherwise its an error?

packages/typespec-azure-resource-manager/src/rules/use-model-request-body.ts
```diff
@@ -0,0 +1,318 @@
...
+  program: Program,
+  target: Scalar | ModelProperty,
+): "none" | "typed" | "untyped" {
+  const encoding = getEncode(program, target);
+  if (!encoding || encoding.type === target) {
+    return "none";
+  }
+  const targetFormat = isSecret(program, target) ? "password" : undefined;
+  const encodedSchema = getEmittedScalarSchema(program, encoding.type);
+  const mergedFormat = mergeFormatAndEncoding(
```
timotheeguerin: why do we need this whole format thing

msyyc: This is not handling authored `@format`; that support was removed. It mirrors AutoRest's internal `@encode` behavior because encoding only replaces the schema type when `mergeFormatAndEncoding()` produces a format.

For example:

```typespec
model Payload {
  value: string;
}

model Request {
  @encode("custom", int32)
  payload: Payload;
}

@post op create(@body body: Request.payload): void;
```

Although the underlying TypeSpec property is model-shaped, AutoRest applies the encoding and emits an integer schema such as `{ type: "integer", format: "int32" }`; the migrated Swagger rule must therefore report it. Other encodings can produce no usable format and leave an untyped schema, which the original rule does not report.

So the format value is only an intermediate signal used to determine whether AutoRest applies the encoding and changes `schema.type`. I agree the reason is subtle and will add documentation around this logic.

packages/typespec-azure-resource-manager/src/rules/use-model-request-body.ts
msyyc: @timotheeguerin 
`getHttpOperation()` gives us the semantic TypeSpec body type, but this migrated rule checks the final Swagger `schema.type`, and those do not always match.

For example, an array body is represented as a TypeSpec `Model`:

```typespec
@post op create(@body items: string[]): void;
```

but AutoRest emits `schema: { type: "array", items: { type: "string" } }`, so the Swagger rule reports it. Checking only `body.type.kind === "Model"` would incorrectly allow it.

The reverse also occurs:

```typespec
scalar OpaqueBody;
@post op create(@body body: OpaqueBody): void;
```

This can emit `schema: {}` with no `type`. The original `ParametersSchemaAsTypeObject` predicate only reports an explicitly emitted non-`object` type, so this remains valid; checking simply for a non-model TypeSpec type would incorrectly report it.

Nullable/singleton unions and file models add similar differences. That is why this code resolves the effective AutoRest schema type rather than checking the raw HTTP body type. I agree this is non-obvious and will document that intent in the implementation.

packages/typespec-azure-resource-manager/src/rules/use-model-request-body.ts
timotheeguerin: but this doesn't make sense, the goal of this rule is to check the body is a model, either it is or it;s not. If it has any encode to anything else it is not(If that is even worth checking), that's it don't need to check further. The rules should NOT try to mimic exactly what autorest output did they should work wiht what you can do in typespec.

msyyc: # Context
A TypeSpec `Model` is a compiler-level construct, not necessarily a Swagger `type: object`.

Examples:

```tsp
model Request {
  value: string;
}
```

emits an object schema.

```tsp
model Request is Array<string>;
```

is also a TypeSpec `Model`, but emits:

```json
{ "type": "array", "items": { "type": "string" } }
```

Visibility transformations, model properties, templates, files, and encodings can further change the effective emitted schema.

The reverse also matters: a non-`Model` TypeSpec type can emit no Swagger `type`, such as `unknown`; the Swagger rule ignores it even though it is not a TypeSpec model.

# Need discussion
Now the gap is clear and we should answer the question: whether real behavior equivalence is strict standard for the rule migration. And I need discussion with @catalinaperalta

timotheeguerin: yeah by model I mean a plain model (one without an indexer) which is easily checkable, what does this have to do with encode.

msyyc: This resolves my response to the array example: although arrays have `kind === "Model"`, they have an indexer, so the simple plain-model check still rejects them. **The remaining decision is product-level: whether this promotion prioritizes Swagger diagnostic parity or translates the Swagger guideline into idiomatic TypeSpec policy.** CC @catalinaperalta 

> yeah by model I mean a plain model (one without an indexer) which is easily checkable, what does this have to do with encode.

timotheeguerin: it is still feature parity, you cannot do that in TypeSpec, the check that swagger was doing with format has no equivalence in TypeSpec, this is just dead code. Can you show me a test that would pass here without all that extra noise.

catalinaperalta: We should focus on being idiomatic to TypeSpec and not bringing in swagger logic into the linter rules. Wherever relevant we should maintain parity with the swagger checks but only if it's not violating the previous statement of not being idiomatic to typespec

msyyc: Updated in a29fa6537 following this guidance and Catalina's TypeSpec-first decision. The rule now checks the authored body semantically: a valid single body must resolve to a `Model` without an indexer. It retains only the HTTP-level `void` and multipart exemptions and project-owned diagnostic targeting. All AutoRest encoding, format, union-emission, scalar-emission, and reference-resolution simulation was removed. The focused suite was rewritten around plain models, model-property references, arrays/records, transformed ARM action arrays, file bodies, primitives/non-models, synthetic `void`, and multipart bodies; all 10 tests pass.

msyyc:
Audit of `50b037ec880e52a0a69bc2806e6001046eaad1bd` against the current `develop-lintdiff-rule` and `lintdiff-rule-promote` skills found **no confirmed production native-boundary violation** in the inspected changes. The remaining finding is incomplete promotion handoff evidence.

Please refresh the PR description to include the substantive information required by the [promotion PR handoff](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#10-review-commit-push-and-create-a-draft-pr):

- Direct original validator source and documentation links, with its specific checks and exemptions.
- The lintdiff source-rule link, source branch, pinned source commit, source worktree/uncommitted-state provenance, and an accurate explanation of any promotion adaptation.
- Destination analysis and a description of the native semantic contract, including intentional differences from emitted Swagger behavior.
- An explicit `Original lintdiff fixture | Native vitest case | Coverage note` table using exact `it("...")` titles, including compliant and regression cases.
- A usable direct link to the source rule's `migration.md`, distinguishing source corpus evidence from validation of the promoted implementation.

The existing summary/validation note is not enough to verify fixture conversion and the source-to-promotion handoff. This is an evidence deliverable gap, not an assertion that the missing documentation means tests or reviews were never run. The title's `(devbox-heavy)` suffix also differs from the skill's stable `[Swagger Linter Migration] ParametersSchemaAsTypeObject` pattern.

.chronus/changes/promote-use-model-request-body-2026-08-31.md
```diff
@@ -0,0 +1,7 @@
+---
```
timotheeguerin, line 1: just cleanup the pr title

timotheeguerin (approved):

msyyc:
(devbox-heavy)

.chronus/changes/promote-use-model-request-body-2026-08-31.md
msyyc: done

PR #5364, opened 2026-09-01
[Swagger Linter Migration] GetInOperationName

msyyc:
## Original Swagger linter

- Linter code: [GetInOperationName](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/get-in-operation-name.ts)
- Guideline: [GetInOperationName (R1005)](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/openapi-authoring-automated-guidelines.md#r1005)
- Source TypeSpec lintdiff PR: #5317

The Swagger rule checks GET operation IDs for a `Get` or `List` verb, either directly or after an operation-group prefix. Its underlying purpose is to ensure generated SDK methods clearly communicate whether they retrieve one resource or list several resources.

## Promoted TypeSpec rule

The promoted `@azure-tools/typespec-client-generator-core/get-operation-name` rule validates the SDK method name modeled by TCGC rather than reconstructing an AutoRest-specific Swagger `operationId`.

The rule:

- visits concrete GET operations and ignores template declarations and non-GET operations;
- resolves the unscoped/default TCGC method name, including an unscoped `@clientName` override;
- ignores `@operationId`, client/operation-group location, and emitter-scoped name overrides because those do not define the common TCGC SDK method name;
- accepts `get` or `list` case-insensitively so the language-neutral TCGC name can use its normal lower-camel casing;
- reports one diagnostic on the TypeSpec operation.

This keeps the rule in TCGC, where the generated SDK API name is modeled, without introducing an AutoRest dependency or duplicating AutoRest's operation-ID resolver. The rule is registered but disabled by default in both Azure data-plane and resource-manager rulesets.

## Migration note

This is an intent-preserving migration rather than a byte-for-byte check of the Swagger `operationId`. Explicit `@operationId` values and operation-group prefixes can therefore differ from the value checked by the native rule. Those differences are intentional: the native rule enforces the generated SDK method-name guideline at the TCGC abstraction layer.

## Validation

- TCGC dependency-closure and package builds
- Focused native rule tests: 11 passed
- TCGC lint
- TCGC documentation regeneration
- Repository formatting and linting
- `git diff --check`


packages/typespec-client-generator-core/src/rules/get-operation-name.md
```diff
@@ -0,0 +1,37 @@
...
+@route("/widgets")
+@get
+op listWidgets(): Widget[];
+```
+
+## LintDiff Equivalent
+
+This rule is equivalent to the Azure OpenAPI Validator
+[`GetInOperationName`](https://github.com/Azure/azure-openapi-validator/blob/main/docs/get-in-operation-name.md)
+rule.
```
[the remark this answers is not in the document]
msyyc: Fixed in 488ca37faf1345852da80dc4dd9aba5bccfab322. Added Impact and Suppression guidance and updated the LintDiff reference to the canonical R1005 guideline.

.chronus/changes/promote-get-in-operation-name-rulesets.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Register the Azure Core `get-operation-name` rule as disabled by default.
```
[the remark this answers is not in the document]
msyyc: Fixed in 488ca37faf1345852da80dc4dd9aba5bccfab322. The user-facing changelog now attributes the rule to TypeSpec Client Generator Core.

packages/typespec-client-generator-core/test/rules/get-operation-name.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        op ResourceRead<T>(): T;
+
+        @route("/widgets/details")
+        op getWidget is ResourceRead<string>;
+      `,
+      )
+      .toBeValid();
+  });
+
+  it("accepts the AutoRest operation ID resolved from client name and location", async () => {
```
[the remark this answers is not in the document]
msyyc: Fixed in 488ca37faf1345852da80dc4dd9aba5bccfab322. Added focused coverage for interface, namespace/service, direct interface/namespace, AutoRest versus other scopes, and Azure.Core/ResourceManager library-type exclusion behavior.

packages/typespec-client-generator-core/src/rules/get-operation-name.rule.ts
```diff
@@ -0,0 +1,115 @@
...
+const validGetOperationId = /^(?:\w+_(?:Get|List)|Get|List)/;
+
+export const getOperationNameRule = createRule({
+  name: "get-operation-name",
+  docs: fileRef.fromPackageRoot("src/rules/get-operation-name.md"),
+  description: "GET operation IDs should use 'Get' or 'List' as the verb prefix.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/rules/get-operation-name",
+  messages: {
+    default: paramMessage`GET operation ID '${"operationId"}' should use 'Get' or 'List' as the verb prefix. Changing an operation ID after an SDK has shipped may be a breaking change.`,
```
timotheeguerin: should that really be about autorest operationId, we don't want to add any dependency on autorest
cc @catalinaperalta

msyyc: @timotheeguerin  Good question. The original rule checks Swagger `operationId`, but its intent is to ensure the generated SDK API name uses `Get` or `List`. Since TCGC now directly models the generated SDK method name, I think the rule can remain in TCGC while validating TCGC’s operation/method name instead of reproducing AutoRest’s `operationId` resolution. That would remove the AutoRest-specific context and duplicated resolver logic. We should treat emitter-scoped `@clientName` separately and validate the default TCGC method name here.

Updated in https://github.com/Azure/typespec-azure/pull/5364/commits/4983184db8f3a8c7cecf2cd1b4818624bccb6895

.chronus/changes/promote-get-in-operation-name-rulesets.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-rulesets"
+---
+
+Register the `@azure-tools/typespec-client-generator-core` `get-operation-name` rule as disabled by
+default.
```
timotheeguerin: might have missed that in the other PR but I think this should just be an internal change, it shouldn't have any effect on users

packages/typespec-azure-rulesets/src/rulesets/data-plane.ts
```diff
@@ -49,6 +49,7 @@ export default {
 
     // TCGC rules
     "@azure-tools/typespec-client-generator-core/require-client-suffix": true,
+    "@azure-tools/typespec-client-generator-core/get-operation-name": false,
```
timotheeguerin: should that be in client-sdk ruleset instead?

packages/typespec-client-generator-core/src/rules/get-operation-name.rule.ts
```diff
@@ -0,0 +1,53 @@
...
+const validGetOperationName = /^(?:get|list)/i;
+
+export const getOperationNameRule = createRule({
+  name: "get-operation-name",
+  docs: fileRef.fromPackageRoot("src/rules/get-operation-name.md"),
+  description: "GET SDK method names should use 'Get' or 'List' as the verb prefix.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/rules/get-operation-name",
+  messages: {
+    default: paramMessage`GET SDK method name '${"operationName"}' should use 'Get' or 'List' as the verb prefix. Changing a method name after an SDK has shipped may be a breaking change.`,
```
timotheeguerin:
```suggestion
    default: paramMessage`GET SDK method name '${"operationName"}' should start with 'get' or 'list'. Changing a method name after an SDK has shipped may be a breaking change.`,
```
and do we really need to include that extended info about the consequence to every messages?

packages/typespec-client-generator-core/src/rules/get-operation-name.rule.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (isTemplateDeclarationOrInstance(operation)) {
+          return;
+        }
+
+        const [httpOperation] = getHttpOperation(context.program, operation);
+        if (httpOperation.verb !== "get") {
+          return;
+        }
+
+        const operationName = getLibraryName(tcgcContext, operation, AllScopes);
```
timotheeguerin, line 42: does this actually need tcgc at all, is that not just `operation.name`?

msyyc: Kept the TCGC resolution intentionally and clarified it in 7118644c5. An unscoped `@clientName` changes the generated `SdkServiceMethod.name`, while `operation.name` remains the TypeSpec source name. `getLibraryName(tcgcContext, operation, AllScopes)` validates that common SDK method name and excludes emitter-specific overrides.

packages/typespec-client-generator-core/src/rules/get-operation-name.rule.ts
msyyc: Fixed in 7118644c5. The diagnostic now says the method name should start with `get` or `list`; the breaking-change guidance remains in the rule documentation instead of every diagnostic.

.chronus/changes/promote-get-in-operation-name-rulesets.md
msyyc: Fixed in 7118644c5. The rulesets change is now marked `internal` because the rule remains disabled and has no user-visible effect.

packages/typespec-azure-rulesets/src/rulesets/data-plane.ts
msyyc: Fixed in 7118644c5. The disabled registration now lives in the `client-sdk` ruleset instead of the data-plane and resource-manager rulesets.

packages/typespec-azure-rulesets/src/rulesets/client-sdk.ts
```diff
@@ -8,6 +8,8 @@ export default {
     "@azure-tools/typespec-client-generator-core/csharp-model-suffix": true,
   },
   disable: {
```
timotheeguerin, line 11: I think its better to have in enable as false. Makes upgrading after just toggling it. Disabled should more be used in tspconfig

timotheeguerin: also just thinking about this being a tcgc rule was that supposed to apply to data plane too or just ARM?

msyyc: Updated in 963d971c9. The rule is now listed under `enable` with value `false`, so activating it later only requires toggling the value.

msyyc: https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/openapi-authoring-automated-guidelines.md#r1005-getinoperationname shows the rule shall apply for both data-plane and ARM. TCGC rule is also for data-plane and ARM, so it is proper to put in TCGC.

msyyc:
Audit of `963d971c9592049dac8db08ac5c98703b3ba26e1` against the current migration skills found **stale/incomplete promotion evidence**, not a confirmed production dependency violation.

The TCGC destination and common SDK-name semantics are supported by the review discussion. This is not an ARM-package dependency on TCGC, and I am not recommending moving the rule back merely because the generic promotion skill primarily describes core/ARM destinations.

Please refresh the PR body:

- It says the rule is registered as disabled in both data-plane and resource-manager, but the [current registration](https://github.com/Azure/typespec-azure/blob/963d971c9592049dac8db08ac5c98703b3ba26e1/packages/typespec-azure-rulesets/src/rulesets/client-sdk.ts#L5-L10) is `false` in client-SDK, following [the review decision](https://github.com/Azure/typespec-azure/pull/5364#discussion_r3947959497).
- Add the original validator documentation link alongside the source-code link, pinned lintdiff source provenance, a usable `migration.md` link, and the required exact fixture-to-`it("...")` mapping table.
- Describe the reviewed TCGC adaptation separately from the immutable source's historical parity evidence; do not imply a fresh corpus result for a changed naming contract.

These are the [promotion PR evidence requirements](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#10-review-commit-push-and-create-a-draft-pr).

There is also a **skill-wording ambiguity**, not a demonstrated rule bug: [the OpenAPI operation-ID tests](https://github.com/Azure/typespec-azure/blob/963d971c9592049dac8db08ac5c98703b3ba26e1/packages/typespec-client-generator-core/test/rules/get-operation-name.test.ts#L57-L88) prove that `@OpenAPI.operationId` is ignored. They do not make the production decision depend on OpenAPI. The updated skill's blanket native-test wording should explicitly permit this narrowly scoped negative independence regression, or place it in comparison coverage. Do not delete useful regression coverage merely to satisfy an import-only interpretation.

timotheeguerin (approved):

PR #5371, opened 2026-09-01
Code model improvements

jhendrixMSFT:
Added generic type params to Constant, Map, Scalar, and Slice for their underlying types along with type guards to simplify narrowing. Replaced some duplicate helpers with the new ones.
Constrained slice element/map value types to applicable types instead of the wider WireType.
Added type guard for additional properties model fields which simplified checks at the call site.

No functional changes.


jhendrixMSFT:
@microsoft-github-policy-service rerun

heaths (approved):

jhendrixMSFT:
@microsoft-github-policy-service rerun

PR #5383, opened 2026-09-03
Fix empty AutoRest feature file emission

### LINKED ISSUE #5322: [typespec-autorest]:  Update @featureFiles support to not emit OpenAPI with no contents and omit featurs enum
If OpenAPI contains no paths, parameters, or definitions, we should not emit it - this allows seamlessly adding new feature values in new versions.

Also should key from the setting for omitting versions enum to omit the features enum as well.

markcowl:
## Summary
- skip legacy feature-file OpenAPI documents that contain no paths, x-ms-paths, parameters, or definitions
- apply version-enum-strategy to the feature enum as well as the API version enum
- add regression coverage for versioned empty feature files and enum include/omit behavior

## Testing
- full repository build
- repository lint and formatting
- @azure-tools/typespec-autorest coverage suite (512 tests)
- focused regression tests

Fixes #5322

### DIFF
--- a/packages/typespec-autorest/src/openapi.ts
+++ b/packages/typespec-autorest/src/openapi.ts
@@ -10,8 +10,6 @@ import {
   hasUniqueItems,
 } from "@azure-tools/typespec-azure-core";
 import {
-  $featureFiles,
-  $features,
   type ArmFeatureOptions,
   getArmCommonTypeOpenAPIRef,
   getArmIdentifiers,
@@ -19,6 +17,8 @@ import {
   getCustomResourceOptions,
   getExternalTypeRef,
   getFeature,
+  getFeatureFileSet,
+  getFeatureOptions,
   getInlineAzureType,
   getResourceFeatureSet,
   isArmCommonType,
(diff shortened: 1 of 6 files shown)


packages/typespec-autorest/src/openapi.ts
```diff
@@ -10,6 +10,8 @@ import {
   hasUniqueItems,
 } from "@azure-tools/typespec-azure-core";
 import {
+  $featureFiles,
```
markcowl: You should not import the decorator implementation from typespec-azure-resource-manager.  Instead, you should import and use the associated accessor

packages/typespec-autorest/src/openapi.ts
```diff
@@ -3302,6 +3312,9 @@ function createFeatureDocumentProxy(
         for (const [defName, [_, defSchema]] of definitionsForFeature) {
           featureItem.document.definitions![defName] = defSchema;
         }
+        if (!hasOpenApiContent(featureItem.document)) {
```
markcowl, line 3311: I think this needs to come after finalizeOpenAPI2Document, in case that call adds any parameters, paths, or definitions

packages/typespec-autorest/test/arm/arm.test.ts
```diff
@@ -572,6 +579,98 @@ it("can split resources and operations by feature", async () => {
...
+        Student: "student",
+        Teacher: "teacher",
+      }
+
+      enum Versions {
+        v1,
+        v2,
+      }
+
+      @Azure.ResourceManager.featureFile(Features.Student)
```
markcowl, line 601: You should have some type or operation belonging to the teacher feature (without versioning decoration) and validate that teacher.json is emitted in both versions

markcowl (commented):
Please fix the remaining issues

timotheeguerin (approved):

PR #5398, opened 2026-09-04
[Swagger Linter Migration] XmsExamplesRequired (origin)

msyyc:
CONTEXT:
TCGC has related logic to map operation and its example file so I plan to reuse the logic then put the rule into TCGC. But TCGC owner said there will be refactor for tsp about example then I should be able to use the native compiler function to check example mapping for each operation. So I will block this PR for now.

## Original Swagger linter

- linter code: [XmsExamplesRequired](https://github.com/Azure/azure-openapi-validator/blob/1198225afecbb818c3050d4d2a91da92e14e56ce/packages/rulesets/src/spectral/functions/xms-examples-required.ts)
- linter doc: [xms-examples-required.md](https://github.com/Azure/azure-openapi-validator/blob/1198225afecbb818c3050d4d2a91da92e14e56ce/docs/xms-examples-required.md)

Checks performed by the original rule:

- Visits Swagger path operations for `get`, `put`, `post`, `patch`, `delete`, `options`, and `head` through the shared az-common Spectral ruleset.
- Reports when the operation object does not define `x-ms-examples`.
- Treats any defined `x-ms-examples` value as passing, including an empty object, because of the shipped presence-check bug.

The Spectral function receives an emitted OpenAPI operation object and reports at that operation's JSON path when `x-ms-examples` is absent. It does not validate the examples' contents, referenced files, request/response coverage, or minimum/maximum payload quality.

The TypeSpec rule visits HTTP operations with the same verbs. It already accepted `@extension("x-ms-examples", ...)`; this change also accepts `@Autorest.example(...)`, because the autorest emitter writes those decorator entries into the emitted operation's `x-ms-examples` object. The diagnostic target remains the source operation so authors can fix the operation-level examples evidence directly.

External `examples-dir` files are intentionally not used to suppress diagnostics in this PR: modeling emitter option interpolation and version-specific file loading exactly requires separate emitter-aware work, and an unsafe broad filesystem scan was rejected during review.

## Migration evidence

See [migration.md](packages/typespec-lintdiff/test/fixtures/XmsExamplesRequired/migration.md) for focused fixtures, final full-corpus counts, one-sided project lists, compile failures, and remaining uncertainty.


packages/typespec-lintdiff/src/rules/xms-examples-required.ts
```diff
@@ -1,16 +1,9 @@
+import { getExamples } from "@azure-tools/typespec-autorest";
```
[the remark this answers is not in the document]
msyyc: Fixed in e0c3f29 by declaring `@azure-tools/typespec-autorest` as a peer dependency for `tsp-lintdiff-local-linter` while keeping the dev dependency for local tests and fixture validation.

markcowl, line 1: I wonder if we shouldn't implement this rule in typespec-autorest, and not export any new APIs

packages/typespec-lintdiff/test/fixtures/XmsExamplesRequired/autorest-example/main.tsp
```diff
@@ -0,0 +1,37 @@
...
+
+@doc("A widget")
+model Widget {
+  @doc("Widget name")
+  name: string;
+}
+
+@route("/widgets/{name}")
+@doc("Get a widget")
+@Autorest.example("./examples/getWidget.json", "getWidget")
```
markcowl, line 29: I don't think we should include these - examples should coem from standard example matching only

markcowl (commented):
I wonder if we shouldn't just implement this linting rule in typespec-autorest?

packages/typespec-lintdiff/src/rules/xms-examples-required.ts
catalinaperalta: I suppose with the new ruleset we're going to add in the specs repo we can still include this rule there, even if it will ultimately be added in typespec-autorest

PR #5376, opened 2026-09-02
Backmerge release/august-2026 to main

weidongxu-microsoft:
## Summary

Backmerge `release/august-2026` into `main`, including the `@azure-tools/typespec-java` 0.46.2 hotfix.

This PR preserves the release core pin, Java sync output, package version, and changelog while retaining the newer repository structure from `main`.

PR #5400, opened 2026-09-04
Backmerge release/august-2026 into main

weidongxu-microsoft:
Backmerge the `release/august-2026` TypeSpec Java hotfix into `main`.


weidongxu-microsoft:
Chronus reports the backmerged Java diff because the release change entry was already consumed when versioning `0.46.3`. The published changelog entry is included in this PR, so adding a new pending change entry would duplicate the same release note. This matches the prior August backmerge handling in Azure/typespec-azure#5376 and Azure/typespec-azure#5351.

XiaofeiCao (approved):

PR #5412, opened 2026-09-07
Run Spector coverage only on main

weidongxu-microsoft:
## Summary

- run the Spector Coverage stage only on `main`
- prevent `release/*` pipeline runs from overwriting the coverage index with an older package version

## Validation

- `pnpm format`
- `pnpm lint`

### DIFF
--- a/eng/pipelines/publish.yml
+++ b/eng/pipelines/publish.yml
@@ -34,6 +34,7 @@ extends:
       - stage: spector_coverage
         displayName: Spector Coverage
         dependsOn: build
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft:
Jiaqi found that when we release from release branch, the coverage report regress on Golang. I think this is the reason.

Since we would anyway backmerge to main, it seems safe just publish coverage from main branch.

JiaqiZhang-Dev (approved):

XiaofeiCao (approved):

msyyc (approved):

tadelesh (approved):

timotheeguerin (approved):

PR #5420, opened 2026-09-08
[Swagger Linter Migration] TenantLevelAPIsNotAllowed

msyyc:
Superseded by #5420, which uses the same branch and commit from the Azure/typespec-azure repository.


packages/typespec-azure-resource-manager/src/rules/no-tenant-level-apis.ts
```diff
@@ -0,0 +1,40 @@
...
+              });
+            }
+          }
+        }
+      },
+    };
+  },
+});
+
+function isTenantLevelPutPath(path: string): boolean {
```
timotheeguerin: nit I think the function name doesn't actually describe what this is checking, it doesn't check put at all

timotheeguerin (approved):

packages/typespec-azure-resource-manager/src/rules/no-tenant-level-apis.ts
msyyc: Renamed the helper to `isTenantLevelPath` so its name reflects that it only checks the path; the PUT verb check remains in the caller.

PR #5423, opened 2026-09-08
[Swagger Linter Migration] PutRequestResponseSchemeArm + ConsistentResponseSchemaForPut

msyyc:
## Summary

Introduce one ARM linter rule, **`put-resource-schema-consistency`**, that encourages reuse of a single native resource model across a PUT request and its `200`/`201` response bodies.

This supersedes the separate request/response proposal in #5493 and replaces this PR's earlier response-only `put-response-schema-consistency` rule. It implements the consolidation discussed in [review](https://github.com/Azure/typespec-azure/pull/5423#discussion_r4037099949).

The rule remains explicitly **disabled** in the resource-manager ruleset. This change makes the rule available without enabling new diagnostics for existing service specs.

## Guideline-driven design, not Swagger implementation parity

The design combines the intent of two ARM guidelines:

- [PutRequestResponseSchemeArm / RPC-Put-V1-25](https://github.com/Azure/azure-openapi-validator/blob/main/docs/put-request-response-scheme-arm.md): encourage reuse of the resource model between PUT input and its response.
- [ConsistentResponseSchemaForPut / RPC-Put-V1-29](https://github.com/Azure/azure-openapi-validator/blob/main/docs/consistent-response-schema-for-put.md): creation and replacement responses should represent the same resource.

The user explicitly approved redesigning the native rule from these guidelines rather than mechanically combining the existing implementations. The selected contract is **resource-model reuse**, not general structural schema equivalence, TypeSpec assignability, emitted OpenAPI equality, or prediction of generated SDK types.

The historical Swagger implementations use different comparison mechanisms: the [request/response implementation](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/put-request-response-scheme.ts) uses a schema comparator, while the [response-consistency implementation](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/consistent-response-schema-for-put.ts) compares resolved JavaScript object identity. Those implementations are historical context, not the specification for this native rule.

## Native contract

The rule visits authored PUT operations, resolves HTTP bodies through `getHttpOperation`, and compares the available identifiable resource models in a stable order: request, exact `200`, exact `201`.

| Situation | Behavior |
| --- | --- |
| Request, 200 and 201 have identifiable resource models | Require the same native model identity across all three |
| Request and only one response body are available | Compare the available pair |
| Request is absent but both response bodies are available | Compare 200 against 201 |
| 200 is absent or bodyless, with a request and 201 body | Compare the request against 201 |
| Fewer than two identifiable participants | No consistency warning; this rule does not require bodies or status codes |
| Other statuses, including 202 and ordinary error responses | Outside this rule's scope |
| One status has distinct body types or body kinds | Skip that ambiguous status in either variant order; still compare other identifiable participants |

Only ordinary bodies resolving to **named models without an own or inherited indexer** participate. `createMetadataInfo().getEffectivePayloadType` can recover the original named model from some anonymous HTTP bodies reusing its properties, including implicit bodies formed through spread. Request visibility is obtained from `resolveRequestVisibility`; responses use `Visibility.Read`.

Type aliases preserve identity. Separately declared models remain distinct, including declarations using `is`, `extends`, or spread, even when they have the same properties. This is an intentional authoring policy: share the resource model and express legitimate differences using lifecycle visibility rather than duplicating the resource declaration.

The rule compares the shared model, not serialized property sets. Read-only/write-only properties, defaults and constraints on the same model therefore do not create mismatches. Content types alone do not change model identity.

Anonymous bodies without a recoverable named model, scalars, unions, tuples, arrays, dictionaries (including inherited indexers), multipart bodies and file bodies are not compared. Skipping unsupported or ambiguous input is not a claim that the input is otherwise valid. Bodyless response variants do not introduce ambiguity.

When participating models differ, emit one warning on the authored operation, listing their body roles and model names. For example:

> PUT bodies must reuse the same resource model. Found request: WidgetInput; 200 response: Widget; 201 response: Widget.

Library declarations and source templates are excluded; concrete aliases and inherited concrete operations are covered. Explicitly enabling the ARM rule checks ordinary and nested namespaces without requiring a provider annotation. The rule inspects the source program without emitter-specific version projections.

## Relationship to other rules and PRs

- **#5493 is superseded by this combined rule.** Its separate `put-request-response-schema` implementation should not also be shipped. This change does not transplant its broad structural-equivalence engine.
- `arm-resource-operation-response` checks operation responses against a registered canonical ARM resource across lifecycle operations. This rule checks consistency among the bodies within one PUT, including a request-only mismatch and operations without resource registration.
- Azure Core's `response-schema-problem` examines non-error response schemas across verbs/statuses; it does not compare the PUT request with its responses.
- Historical lintdiff repairs, including #5392, #5424, #5435 and #5491, remain intact. The new design intentionally supersedes the earlier immutable-source promotion contract; it no longer claims to be an exact copy of source SHA `72d8f52cc766daf2da5537bee14ba1ce10660d10`.
- Historical [migration evidence](https://github.com/msyyc/typespec-azure/blob/72d8f52cc766daf2da5537bee14ba1ce10660d10/packages/typespec-lintdiff/test/fixtures/ConsistentResponseSchemaForPut/migration.md) describes the earlier source rule, not measured behavior of this new model-reuse policy. Earlier failure/recovery records are preserved; no corpus-parity claim is made for this redesign.

## Implementation and documentation

- Replace the response-only source, documentation and tests with `put-resource-schema-consistency` in `@azure-tools/typespec-azure-resource-manager`.
- Remove the recursive anonymous-model/tuple comparator; compare native model identity after supported HTTP normalization.
- Preserve order-independent response-variant handling and produce one diagnostic per operation.
- Update linter registration, the disabled resource-manager ruleset entry, generated README/reference listings, authored ARM rule table, and feature/internal Chronus entries.
- Document the strict model-reuse policy, unsupported inputs, lifecycle behavior, suppression guidance and both guideline links.

## Validation and review evidence

- TDD coverage includes request-only and response-only mismatches, both exact statuses, missing/bodyless participants, implicit bodies, aliases, separately declared models, namespace identity, visibility/defaults/constraints, ambiguous variant ordering, content variants, inherited indexers, standard ARM templates, library/template exclusions, versioned operations and real-pipeline suppression.
- All **46 focused native cases** pass. Before the upstream integration, focused coverage reported **100% branch coverage** for the rule.
- Independent implementation review identified an inherited-indexer exclusion gap; it was reproduced, corrected and covered by regressions before publication.
- After integrating the existing remote branch, the affected dependency closure built successfully with TypeSpec 1.16.0. The full ARM run passed 509/510 cases, with one unrelated existing `enforce-constraint` test hitting its 10-second cold-start limit. An unchanged targeted rerun of that entire file plus the new rule passed **56/56**. Rulesets build and **5/5** cases passed.
- Repository formatting/lint and required ARM documentation regeneration completed; existing unrelated missing-documentation warnings remain. Final changes relative to the previous remote head are confined to the combined-rule redesign.
- Dependency refresh initially encountered stale built exports and missing compiler dependencies. Node dependencies were restored; the workspace install's unrelated Python prepare step failed on a PyPI TLS handshake. The affected Node dependency closure and the checks above subsequently completed without changing Python setup or relaxing assertions/timeouts.
- The existing `int:azure-specs` label is retained. No fresh GitHub Copilot review is claimed for this new design. Earlier review results apply only to their historical commits.

## Broader validation limitation

The historical full-repository `validate:pr` attempt stopped at its five-minute no-progress bound. Broad local validation remains incomplete; this update used the affected dependency/package checks above. CI's dedicated Website job remains authoritative. No generated lintdiff corpus output is included. The merge preserves existing upstream lockfile/submodule changes; the redesign itself does not alter dependencies.

## Publication

Redesign commit: `76f6a53a0`.

Published head: `94d4b87fba0ca0f00d03f9c7066115cd902af6c4`, incorporating the remote branch's existing `db39bde` main integration without a force-push. No PR has been merged.

NOTE: heavy


packages/typespec-azure-resource-manager/src/rules/put-response-schema-consistency.ts
```diff
@@ -0,0 +1,221 @@
...
+  const array = program.checker.getStdType("Array");
+  const indexer = array.decorators.find(
+    ({ node }) =>
+      node?.parent === array.node &&
+      node?.kind === SyntaxKind.DecoratorExpression &&
+      node.target.kind === SyntaxKind.Identifier &&
+      node.target.sv === "indexer",
+  )?.decorator;
+  return model.decorators.every(({ decorator }) => decorator === indexer);
+}
```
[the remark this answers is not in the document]
msyyc: Keeping this unchanged after checking the compiler metadata and native regressions. The standard `Array` supplies the intrinsic indexer decorator identity. An empty decorator list contains no additional schema metadata, so `every` being true there is not evidence of an incorrect normalization; the caller also checks array shape, unknown element type, naming/template identity, and friendly-name metadata. The constrained/named/friendly-named/typed-array and custom same-named decorator regressions pass. No reachable case demonstrating a missing intrinsic identity or false equivalence was established, so this speculative change is not adopted.

packages/typespec-azure-resource-manager/src/rules/put-response-schema-consistency.md
```diff
@@ -0,0 +1,96 @@
...
+consistent result type for the same PUT operation.
+
+## Suppression
+
+Suppress only when an existing API contract cannot be corrected without a
+breaking change and the inconsistency has been reviewed. Prefer reusing the same
+response model. Place the directive above the affected operation:
+
+```tsp
+#suppress "put-response-schema-consistency" "Existing API contract returns different create and replace bodies."
```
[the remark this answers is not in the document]
msyyc: Fixed in ed7d1bf6eb94d29ce37a05570d99a37ce2ff172f: the documentation now uses `@azure-tools/typespec-azure-resource-manager/put-response-schema-consistency`. Added a regression through the compiler's real linter pipeline, with an unsuppressed control, to confirm the documented directive suppresses the warning.

packages/typespec-azure-resource-manager/src/rules/put-response-schema-consistency.ts
```diff
@@ -0,0 +1,221 @@
...
+  const contentTypes: string[] = [];
+  for (const content of response.responses) {
+    if (content.body !== undefined) {
+      body = content.body;
+      contentTypes.push(...content.body.contentTypes);
+    }
+  }
+
+  return body === undefined ? undefined : { body, contentTypes };
+}
```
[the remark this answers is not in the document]
msyyc: Synced the authorized source repair from merged #5435 in ed7d1bf6eb94d29ce37a05570d99a37ce2ff172f. Either status now skips this comparison when its variants have different body-type identities; native regressions cover both statuses and reversed ordering. AutoRest already rejects these groups with `duplicate-body-types`, so this prevents secondary order-dependent warnings rather than fixing a demonstrated mismatch on valid emitted Swagger. Shared body types still aggregate all content types. No lintdiff source was changed during promotion.

msyyc:
Audit of `ed7d1bf6eb94d29ce37a05570d99a37ce2ff172f` found a **production native-boundary violation** under the updated migration skills.

[put-response-schema-consistency.ts:97-146](https://github.com/Azure/typespec-azure/blob/ed7d1bf6eb94d29ce37a05570d99a37ce2ff172f/packages/typespec-azure-resource-manager/src/rules/put-response-schema-consistency.ts#L97-L146) decides equivalence through `haveEquivalentEmittedSchemas` / `getConstantSchemaCategory`. These collapse distinct native bodies into `"file"`, `"string"`, and `"array-any"` categories based on their emitted representation. In particular, multipart bodies and standard strings share a category; tuples and selected unknown arrays share another. The following helper also inspects standard Array decorator syntax to recover the intrinsic indexer identity.

This goes beyond a direct native response identity/shape check. The [native implementation boundary](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/develop-lintdiff-rule/SKILL.md#native-typespec-implementation-boundary) explicitly rejects reproducing AutoRest/Swagger formatting or emission decisions merely to achieve parity, even when implemented using compiler APIs with no emitter import. [Promotion now inherits that boundary](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#native-semantic-boundary).

Please define the intended supported TypeSpec response-consistency contract directly, then remove emitter-equivalence categories/special cases that are justified only by the emitted schema. For each retained branch, provide evidence that it affects valid supported native inputs under that contract. Keep emitter-only equivalences in comparison evidence and document intentional parity gaps instead of encoding them as production exemptions. Update the native tests, rule documentation, and migration explanation accordingly.

This is an architectural finding, not a claim that a current production service has failed. If resolving it changes the immutable lintdiff source's semantics, stop promotion and use the authorized source-repair workflow rather than silently implementing different behavior only in the official copy.

.chronus/changes/put-response-schema-consistency-arm.md
timotheeguerin, line 1: cleanup pr title

.chronus/changes/put-response-schema-consistency-arm.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add the `put-response-schema-consistency` linter rule to require matching ARM PUT 200 and 201 response schemas.
```
timotheeguerin: wondering here I was very confused with the different with https://github.com/Azure/typespec-azure/pull/5493

could we combine both those rules to make sure request and all response match the same schema this feels way to close

.chronus/changes/put-response-schema-consistency-arm.md
msyyc: done

.chronus/changes/put-response-schema-consistency-arm.md
msyyc: Agreed—these belong together. Updated this PR in 94d4b87fba0ca0f00d03f9c7066115cd902af6c4 to provide one native rule, `put-resource-schema-consistency`, and closed #5493 as superseded.

The rule now checks the PUT request and all available **200/201 resource response bodies** together, producing one diagnostic identifying the differing models. It still compares responses when the request is absent, and compares the request with 201 when 200 is absent or bodyless. Other statuses, including 202 and error responses, are outside this rule's scope.

We also reconciled the two rules' different equality policies: this is now explicitly a **resource-model reuse** rule, based on the intent of both ARM guidelines rather than the Swagger implementations. It expects the same identifiable native model, permits lifecycle visibility differences on that shared model, and does not attempt general structural or emitted-schema equality. Separate named declarations remain distinct. Ambiguous or unsupported bodies are skipped rather than selecting an arbitrary variant.

The tests, documentation, registration and PR description have been updated for the combined contract. The rule remains disabled by default.

.chronus/changes/put-response-schema-consistency-arm.md
```diff
@@ -0,0 +1,7 @@
+---
```
timotheeguerin, line 1: Rule looks good but just pending another rule that might be combined here #5503

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
```diff
@@ -0,0 +1,101 @@
...
+```typespec
+model Widget {
+  @visibility(Lifecycle.Read)
+  id: string;
+
+  value: string;
+}
+
+@put
+op createOrReplace(@body body: Widget): ArmResponse<Widget> | ArmCreatedResponse<Widget>;
```
markcowl, line 62: Always have the furst correct response use a standard template

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
```diff
@@ -0,0 +1,101 @@
+Use the same resource model for an ARM PUT request and its `200` and `201`
+response bodies. This keeps creation and replacement consistent and encourages
+reuse of the resource model across operations. Prefer the standard ARM
+create-or-replace operation templates.
+
+This is a **model-reuse rule**, not a structural schema-equivalence check.
```
markcowl, line 6: This is unnecessary except for the first paragraph - these are regurgitated prompts and not rule documentation

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
```diff
@@ -0,0 +1,101 @@
...
+the resource contract and make model reuse harder for API consumers. This rule
+encourages a single authored resource model; it does not predict generated SDK
+types or require identical serialized payloads across lifecycle visibilities.
+
+## Suppression
+
+For an existing API that intentionally uses separate models and cannot change
+compatibly, suppress the warning with a justification. For new APIs, prefer
+sharing one resource model or using the standard ARM operation templates.
+The rule is registered as disabled in the resource-manager ruleset.
```
markcowl, line 79:
```suggestion
```

this should never be in any documentation

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
```diff
@@ -0,0 +1,101 @@
...
+
+```typespec
+#suppress "@azure-tools/typespec-azure-resource-manager/put-resource-schema-consistency" "Existing API uses separate request and response models."
+@put
+op createOrReplace(@body body: WidgetInput): ArmResponse<Widget>;
```
markcowl, line 84: Use a standard template with the request parameter override

packages/typespec-azure-resource-manager/test/rules/put-resource-schema-consistency.test.ts
```diff
@@ -0,0 +1,381 @@
...
+  `,
+    )
+    .toEmitDiagnostics([
+      diagnostic(`request: ${request}; 200 response: ${ok}; 201 response: ${created}`),
+    ]);
+});
+
+it("compares responses even without a request", async () => {
+  await tester
+    .expect(`${models} @put op put(): Ok<Widget> | Created<Other>;`)
```
markcowl, line 72: SHouldn't it also throw a diagnostic for no request body?

markcowl (commented):
A few nits.

We should also likely deprecate/remove this rule: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-put-response-codes.ts  When this one rolls out.  and think about how to normalize with this rule: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-resource-operation-response.ts

.chronus/changes/put-response-schema-consistency-arm.md
msyyc: Combined the native ARM-resource registration check from #5503 into this rule in 8d87839e9. In addition to requiring matching model identities, the identifiable bodies must use a model registered with `getArmResource`; a shared plain model now warns even without `name`/`type` properties. Both 200 and 201 are checked rather than preferring the first 200. Imported library operations remain excluded, while project aliases are checked. The rule stays disabled pending rollout; #5503 is superseded by this implementation.

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
msyyc: Agreed. Removed the implementation/process exposition in 8d87839e9. The page now focuses on the required resource model, how to fix violations using standard templates, impact, and suppression. Detailed scope and implementation tradeoffs are kept in the PR description.

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
msyyc: Changed in 8d87839e9. The first example is now the correct pattern: a tracked `Widget` resource with `createOrReplace is ArmResourceCreateOrReplaceSync<Widget>`. A test reads and compiles the example directly from the Markdown.

packages/typespec-azure-resource-manager/test/rules/put-resource-schema-consistency.test.ts
msyyc: Yes—added `ARM PUT operations must have a request body.` in 8d87839e9. It diagnoses absent and explicit `void` request bodies, including operations with only header/query parameters or no response body. This diagnostic takes priority over response mismatches, so the operation gets one actionable warning. Unsupported but present bodies are not treated as missing. Added regressions for each case.

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
msyyc: Removed the disabled-registration sentence from the user-facing documentation in 8d87839e9. Rollout status remains in the PR description; the actual ruleset entry is unchanged.

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.md
msyyc: Updated in 8d87839e9 to use `Azure.ResourceManager.Foundations.ArmCreateOperation` with a `WidgetInput` override for its `BodyParameter` and `Widget` responses. `ArmResourceCreateOrReplaceSync` does not expose a request-body parameter, so the suppression example uses the supported foundational template instead. Tests exercise the actual Markdown examples and verify the suppression through the compiler pipeline.

msyyc:
Addressed the inline feedback in 8d87839e9 and updated the PR description. Regarding the review overview: I kept arm-put-operation-response-codes because the new rule neither requires the complete PUT status-code set nor is enabled yet. I also kept arm-resource-operation-response: it checks the canonical resource across lifecycle operations, which intra-PUT consistency cannot replace. A regression now demonstrates the distinction: all PUT bodies can consistently use resource B while the existing lifecycle rule correctly rejects them for an operation registered against resource A. Retiring or consolidating those checks should wait until their coverage and rollout are explicitly preserved. #5503 is now closed as superseded by the combined native registration check.

.chronus/changes/put-response-schema-consistency-arm.md
timotheeguerin: from @markcowl comment on one of the superseeded prs, isn't all of this already part of https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-resource-operation-response.ts

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.ts
```diff
@@ -0,0 +1,129 @@
...
+    default: paramMessage`PUT bodies must reuse the same resource model. Found ${"bodies"}.`,
+    missingRequest: "ARM PUT operations must have a request body.",
+    unregistered: paramMessage`PUT bodies must use registered ARM resource models. Found ${"bodies"}.`,
+  },
+  create(context) {
+    const metadata = createMetadataInfo(context.program);
+
+    function getResourceModel(body: HttpPayloadBody | undefined, visibility: Visibility) {
+      if (body?.bodyKind !== "single") return undefined;
+      const type = metadata.getEffectivePayloadType(body.type, visibility);
```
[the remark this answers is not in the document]
msyyc: I checked the HTTP resolver and could not reproduce this distinction for single property-backed payloads. `resolveExplicitBodyProperty` sets `type: item.property.type` for both `@body` and `@bodyRoot`, so `body.type` already is the payload type, not the enclosing model. The fallback in `list-response-envelope.ts` does not establish a different resolver contract.

Kept normalization on `body.type` and added regression coverage in c6c79688c asserting that both the request and response resolve to their respective property types for `@body` and `@bodyRoot`. Existing mismatch and registration regressions remain passing. If there is a concrete input where these identities differ, we can revisit it with that repro.

.chronus/changes/put-response-schema-consistency-arm.md
msyyc: There is overlap, but the existing rule does not cover the request or require registration of response candidates. It traverses responses of registered lifecycle operations and reports a mismatch only when the candidate is itself a registered ARM resource.

Added executable comparisons in c6c79688c showing three gaps, even on a registered create/update operation:
- No request body: the existing rule is silent; this rule warns.
- Request uses resource B while the response correctly uses canonical A: the existing rule is silent; this rule warns.
- Request and response share an unregistered model: the existing rule is silent; this rule warns.

The reverse distinction is also covered: when all PUT bodies use registered B but the operation belongs to A, this rule sees consistent bodies while the existing lifecycle rule correctly warns. That is why I retained the existing rule rather than replacing it. The new rule remains disabled pending rollout. Sharing implementation could still make sense, but dropping either set of checks would lose coverage.

packages/typespec-azure-resource-manager/src/rules/put-resource-schema-consistency.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+      },
+    };
+  },
+});
+
+function getUnambiguousBody(response: HttpOperationResponse): HttpPayloadBody | undefined {
+  let body: HttpPayloadBody | undefined;
+  for (const content of response.responses) {
+    if (content.body === undefined) continue;
```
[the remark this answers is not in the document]
msyyc: Confirmed and fixed in c6c79688c. Explicitly status-coded responses can retain a `void` payload, which was incorrectly making the status ambiguous and hiding the model variant. `getUnambiguousBody` now skips both undefined and void bodies before comparing variants.

Added regressions for 200 and 201 in both variant orders, verifying that mismatches and unregistered response models still warn, and that matching resources or void-only responses remain valid. Updated the documentation and change description. All 69 focused cases pass after incorporating the latest PR branch updates.

PR #5429, opened 2026-09-08
fix(tcgc): address review feedback on override response support

iscai-msft:
Follow-up to #5173 addressing the unresolved review comments from @tadelesh, plus a regression fix.

## Changes

Removed the explicit model name/namespace fallback in the return-type comparison.

The diagnostic docs were single-line stubs and now follow the standard pattern (Impact, Incorrect Usage, Diagnostic Message, How to Fix, Suppression).

### 3. Fix regression: plain `@override` incorrectly flagged as a response replacement
The `Azure REST API Specs` validation started failing on many existing specs (SignalR, RecoveryServicesBackup, DataBoxEdge, KeyVault, etc.) with `override-response-replacement`/`override-response-mismatch` errors.

**Root cause:** `@override` is used heavily for *parameter customization*, where the customization operation commonly declares `void` (or a different type) just to satisfy the signature — the override operation's return type was historically ignored. The new response check inferred "intentional replacement" from the declared return type, so every ordinary parameter-only override that didn't repeat the exact response tripped the diagnostic. The `@azure-tools/typespec-java` `method-override.tsp` emitter test reproduces this (original ops return `{}`, customization ops return `void`).

**Fix:** The response check now keys off the `responseOverrideKey` marker set by `replaceResponseWithVoid` / `replaceResponseWithBytes` instead of the declared return type, so only genuine response replacements emit the `override-response-replacement` warning. Plain overrides ignore the return type as before. Removed the `override-response-mismatch` diagnostic and its doc: it could only fire for plain overrides declaring a different return type, which is long-standing allowed behavior, so reporting it was a regression.

Added regression tests asserting that a plain parameter-only override declaring `void` or an unrelated return type produces no response diagnostic, and that the warning still fires for `replaceResponseWithVoid` / `replaceResponseWithBytes`.

## Validation
- `pnpm build` (tcgc) — passes, including doc/lint validation
- `pnpm lint` (tcgc) — passes
- Full TCGC vitest suite — 1466 passed, 2 skipped


packages/typespec-client-generator-core/src/decorators.ts
timotheeguerin, line 1: should there be a fix for the new warnings added to spec repo(or are those expected and should be suppressed?)

iscai-msft: this functionality didn't exist before, so there aren't any existing cases. i can run the integration check to make 100% sure though

iscai-msft: oh you already did, thanks!

timotheeguerin (approved):

PR #5431, opened 2026-09-08
[ARM] Prepare resolveArmResources for versioned views

markcowl:
## Summary

- add an RFC for versioned and customizable `resolveArmResources` views
- add a maintenance skill documenting resolver invariants and implementation workflow
- complete Phase 1 by refactoring resolution around an explicit provider context without changing public behavior
- add a characterization test covering cached results, resource operations, parent/scope relationships, and provider operations

## Validation

- all 54 `resource-resolution.test.ts` tests pass
- ARM package suite: 393 tests passed; 2 tests exceeded the default 10-second timeout and all 29 tests in those files passed when rerun with a 30-second timeout
- `pnpm format`
- `pnpm lint`
- `pnpm --filter "@azure-tools/typespec-azure-resource-manager" regen-docs`

### DIFF
--- a/packages/typespec-azure-resource-manager/src/resource.ts
+++ b/packages/typespec-azure-resource-manager/src/resource.ts
@@ -498,10 +498,20 @@ function mapResourceKind(
   }
 }
 
+interface ArmResourceResolutionContext {
+  program: Program;
+  providerNamespace: Namespace;
+}
+
 export function resolveArmResources(program: Program): Provider {
-  const provider = resolveProviderNamespace(program);
-  if (provider === undefined) return {};
-  const resolvedResources = getResolvedResources(program, provider);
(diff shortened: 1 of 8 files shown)


packages/typespec-azure-resource-manager/src/resource.ts
```diff
@@ -498,18 +556,140 @@ function mapResourceKind(
...
+ * This overload preserves the original resolver behavior.
+ */
+export function resolveArmResources(program: Program): Provider;
+/**
+ * Resolves ARM resources and operations for an exact API version or with customized logical names.
+ *
+ * Selected versions are projected with `@typespec/versioning`, so returned TypeSpec references
+ * reflect availability, renames, and type changes in that version.
+ */
+export function resolveArmResources(
```
markcowl, line 588: Why are there two exported functions here?

PR #5438, opened 2026-09-09
Clarify native ARM lint dependency and API boundaries

msyyc:
## Request and evidence

During review of the `LroErrorContent` migration, the user identified three production-rule constraints that the development skill did not state: an ARM rule must not depend on TCGC, must not use `@typespec/openapi`, and must not mutate compiler graphs through unsafe APIs. The existing source used all three, making its intended ARM promotion unsuitable.

## Changes

Document destination-compatible dependency direction, prohibit those APIs in rule logic and its implementation helpers, and forbid private-state or copied-helper workarounds. Distinguish research/comparison fixtures from native tests and existing transitive library dependencies. Require explicit contract, regression, and migration-evidence updates when removing these dependencies changes SDK scope, overrides, inlining, or historical-version coverage. Add the same checks to independent review.

This is the requested skill-only update, separate from the follow-up LroErrorContent source repair. No rule code, corpus outputs, dependency manifests, or validation gates change here.

## Validation

Prettier check with the prepared source checkout's configuration and `git diff --check` passed. The complete diff contains only `.github/skills/develop-lintdiff-rule/SKILL.md`.

### DIFF
--- a/.github/skills/develop-lintdiff-rule/SKILL.md
+++ b/.github/skills/develop-lintdiff-rule/SKILL.md
@@ -446,6 +446,39 @@ An emitter helper that only reads metadata is still an emitter dependency.
 Do not conceal it behind a wrapper, dynamic import, private state-map key,
 decorator-name scraping, or copied emitter implementation.
 
+Design production rules for their intended official destination before choosing
+dependencies. ARM rules must not import or call
+`@azure-tools/typespec-client-generator-core` (TCGC): ARM is a dependency of
+TCGC, so promoting such a rule would introduce the wrong dependency direction.
+Do not add a TCGC dependency to ARM to preserve SDK scope or legacy LRO markers;
+use supported Azure Core and ARM semantic APIs instead.
+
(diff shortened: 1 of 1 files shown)


msyyc:
Resolved the conflict in commit 3138f5ca7 by merging the current target branch, preserving the shared native-TypeSpec-first intent of this PR and #5446.

- Retained #5446's supported-TypeSpec completeness, evidence, and review requirements; did not restore the obsolete exhaustive-emitter-branch requirement.
- Retained this PR's explicit prohibition of TCGC in ARM rule logic, @typespec/openapi usage, and unsafe compiler mutation.
- Clarified that supported Azure semantic APIs and required transitive test-host registrations remain permitted, without allowing wrappers around prohibited functionality.
- Aligned behavior-change evidence with #5446: native regression tests cover supported inputs; comparison fixtures or code-backed rejection evidence explain intentional parity gaps without requiring emitter simulation.

The resulting PR delta remains limited to .github/skills/develop-lintdiff-rule/SKILL.md.

PR #5450, opened 2026-09-10
Prototype: TypeSpec breaking change analysis tool

catalinaperalta:
## Purpose

This draft presents the complete TypeSpec-native breaking-change tool as one pull request so the
overall implementation, size, architecture, and review surface can be evaluated before it is
reconstructed as a sequence of smaller subsystem pull requests.

This is an exploratory draft.

## Included

- Same-version (Phase A) and cross-version (Phase B) analysis
- HTTP operation canonicalization and structural diffing
- Direction-aware breaking-change policy
- Source-origin tracing, finding consolidation, and source locations
- Approval decorators, matching, lifecycle classification, ambiguity detection, diagnostics, and
  codefixes
- Console, GitHub, JSON, and Markdown reporting
- CLI execution and git-revision baseline checkout
- Unit, scenario, and optional real-spec tests
- Workspace, lockfile, and Chronus package registration

## Intentionally excluded

- Breaking-change RFCs
- Package documentation and prototype evaluation documents
- Prototype planning, demo, presentation, and deployment material
- Generated output
- Unrelated repository and submodule changes

## Validation

- `pnpm -r --filter "@azure-tools/typespec-breaking-change..." build` passes against current
  `main`.
- The package suite completed with 450 passing tests and 11 configured real-spec skips. One
  import-mocking CLI test exceeded its 30-second timeout during the full parallel suite; rerunning
  `test/cli-main.test.ts` alone passed all 7 tests in 7.5 seconds.

## Review focus

- Whether the package and subsystem boundaries are appropriate
- The scale and coupling visible in a monolithic review
- Which contracts should be stabilized before splitting the implementation
- Whether the proposed full-file subsystem PR sequence provides suitable review units


packages/typespec-breaking-change/lib/decorators.tsp
```diff
@@ -0,0 +1,86 @@
+using TypeSpec.Reflection;
```
markcowl, line 1: We need to move the decorators into typespec-azure-core and add a dependency

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+  buildPhaseBPairs,
+  createVersionedView,
+  enumerateVersions,
+} from "../dist/src/pipeline/versions.js";
+import { applySuppressions } from "../dist/src/suppression/suppression.js";
+
+const SPEC_ROOTS = [
+  {
+    name: "Network",
+    root: "C:\\Users\\markcowl\\session2\\azure-rest-api-specs\\specification\\network\\resource-manager\\Microsoft.Network\\Network\\Network",
```
markcowl, line 21: Likely need a point-in-time uri for this, or a relative reference

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+  },
+  {
+    name: "AppConfiguration",
+    root: "C:\\Users\\markcowl\\session2\\azure-rest-api-specs\\specification\\appconfiguration\\resource-manager\\Microsoft.AppConfiguration\\AppConfiguration",
```
markcowl, line 25: same

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+  },
+  {
+    name: "ContainerService/fleet",
+    root: "C:\\Users\\markcowl\\session2\\azure-rest-api-specs\\specification\\containerservice\\resource-manager\\Microsoft.ContainerService\\fleet",
```
markcowl, line 29: same

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+    note: "The task text listed Contoso.Management, but the stated 13-version / 8-pair profile matches fleet.",
```
markcowl, line 30:
```suggestion
```

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+    const resolvedOutPath = path.resolve(outPath);
+    fs.mkdirSync(path.dirname(resolvedOutPath), { recursive: true });
+    fs.writeFileSync(resolvedOutPath, `${JSON.stringify(result, null, 2)}\n`, "utf8");
+  }
+
+  process.stdout.write(`${JSON.stringify(result, null, 2)}\n`);
+}
+
+async function evaluateSpec(spec) {
+  const entrypoint = path.join(spec.root, "main.tsp");
```
markcowl, line 82: Will probably need a "download spec" functionality before this

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+      stableVersions: service.versions.filter((version) => !version.endsWith("-preview")),
+      comparisonPairs: buildPhaseBPairs(service.versions, service.versions).map((pair) => ({
+        baseVersion: pair.baseVersion,
+        headVersion: pair.headVersion,
+      })),
+      operationCounts,
+    };
+  });
+
+  const phaseA = analyzeBaseAndHead(program, program, { phase: "same-version" });
```
markcowl, line 111: seems like the phaseA for this is not particularly useful, except for perf measurements

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+    finalCount: finalFindings.length,
+    originCount,
+    resolvedCount,
+    traceCounts,
+    lowTraceFindings,
+    fallbackCategories,
+  };
+}
+
+function deduplicateBySourceType(findings) {
```
markcowl, line 218: This seems like it should be a common function - if this is here for additional tracing, we probably want to introduce tracing into the common function that does this, with a standard tracing context or sink configured at startup.

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+
+  return result;
+}
+
+function mergeRequestResponseToResource(findings) {
+  const result = [];
+  const consumed = new Set();
+  const responseIndex = new Map();
+
+  for (const finding of findings) {
```
markcowl, line 258: Seems like this can be optimized into a single loop through the findings

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+    if (!consumed.has(finding)) {
+      result.push(finding);
+    }
+  }
+
+  return result;
+}
+
+function collapsePhaseADuplicates(findings) {
```
markcowl, line 306: same comment - thsi should be a common library function

packages/typespec-breaking-change/scripts/evaluate-source-tracing.mjs
```diff
@@ -0,0 +1,424 @@
...
+    if (!finding.diff.baseType && !finding.diff.headType) {
+      return {
+        category: "operation-lifecycle",
+        why: "No base/head type or origin was available, so the resolver used operationSourceLocation.",
+      };
+    }
+
+    if (finding.diff.identity.element.startsWith("query.") || finding.diff.identity.element.startsWith("path.") || finding.diff.identity.element.startsWith("headers.")) {
+      return {
+        category: "parameter-without-declaration-anchor",
```
markcowl, line 365: Hmm, want to examine this carefully.  We should have an anchor in many cases here

packages/typespec-breaking-change/src/cli/cli.ts
```diff
@@ -0,0 +1,399 @@
...
+          options.entry = arg;
+        }
+    }
+  }
+
+  return options;
+}
+
+function printUsage(): void {
+  console.log(`
```
markcowl, line 152: We probably want a general log sink that we can map th the console at execution time.

markcowl: Note that, we still need to use console in the standard case, and we often just add this to the top of cli files that need console: /* eslint-disable no-console */

packages/typespec-breaking-change/src/cli/cli.ts
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    // Markdown file output
+    if (options.markdownOutput) {
+      const mdPath = resolve(options.markdownOutput);
+      await ensureParentDir(mdPath);
+      const mdOptions: MarkdownReportOptions = {
+        specPaths: [options.entry],
+        showTiming: true,
+        githubServerUrl: process.env.GITHUB_SERVER_URL,
```
markcowl, line 339: If this is standard for specs tools, maybe this should be in some library code there (eventually)  this is a future comment, not something to take action on now

packages/typespec-breaking-change/src/cli/git-checkout.ts
```diff
@@ -0,0 +1,223 @@
...
+    // surfacing the error, so a failed checkout doesn't also leak a
+    // worktree registration.
+    await execFileAsync("git", ["worktree", "remove", "--force", worktreeRoot], {
+      cwd: repoRoot,
+    }).catch(() => undefined);
+    await rm(worktreeRoot, { recursive: true, force: true }).catch(() => undefined);
+
+    throw new Error(
+      `Failed to check out revision "${commitish}" (resolved to ${sha}) as a git worktree: ` +
+        `${error instanceof Error ? error.message : String(error)}`,
```
markcowl, line 145: add a cause option pointing to the caught error

packages/typespec-breaking-change/src/diff/operation-identity.ts
```diff
@@ -0,0 +1,79 @@
...
+  /** The original TypeSpec Operation node (for source location tracing). */
+  operation: Operation;
+}
+
+/**
+ * Normalize a path by replacing parameter names with {}.
+ * "/subscriptions/{subscriptionId}/providers/{rpName}" → "/subscriptions/{}/providers/{}"
+ */
+export function normalizePath(path: string): string {
+  return path.replace(/\{[^}]+\}/g, "{}");
```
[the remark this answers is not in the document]
markcowl: I think we can maybe do a split on `/`, replace any segments starting with `{` with {} and re-join ratehr than doing a transform

packages/typespec-breaking-change/src/pipeline/orchestrator.ts
```diff
@@ -0,0 +1,605 @@
...
+ * and kinds are the Request/Response pair (e.g., RequestPropertyRemoved +
+ * ResponsePropertyRemoved → ResourcePropertyRemoved).
+ */
+function mergeRequestResponseToResource(findings: Finding[]): Finding[] {
+  const result: Finding[] = [];
+  const consumed = new Set<Finding>();
+
+  // Index Response findings by (element + version + suffix + suppressed)
+  const responseIndex = new Map<string, Finding>();
+  for (const f of findings) {
```
markcowl, line 504: same comment as before, not sure we can't do this in one pass.

packages/typespec-breaking-change/src/pipeline/versions.ts
```diff
@@ -0,0 +1,185 @@
+import type { Namespace, Program } from "@typespec/compiler";
+import { unsafe_mutateSubgraphWithNamespace } from "@typespec/compiler/experimental";
+import { getVersioningMutators } from "@typespec/versioning";
+import { listServices } from "@typespec/compiler";
+import type { VersionedView, ComparisonPhase, VersionPair, ComparisonPair } from "../types.js";
```
markcowl, line 5:
```suggestion
import type { VersionedView, ComparisonPhase, VersionPair } from "../types.js";
```

packages/typespec-breaking-change/src/pipeline/versions.ts
```diff
@@ -0,0 +1,185 @@
...
+  headVersions: string[],
+  classifier: VersionClassifier = defaultVersionClassifier,
+): VersionPair[] {
+  const phaseA = buildPhaseAPairs(baseVersions, headVersions);
+
+  // Phase B candidates (statically known): new versions in head
+  const baseSet = new Set(baseVersions);
+  const newVersions = headVersions.filter((v) => !baseSet.has(v));
+
+  const phaseB = buildPhaseBPairs(headVersions, newVersions, classifier);
```
markcowl, line 182: I wonder if we shouldn't build this after phaseA, we should have changed versions from phase A which *should* also be added to this comparison set.

packages/typespec-breaking-change/src/suppression/ambiguity.ts
```diff
@@ -0,0 +1,98 @@
...
+ * the same logical finding across 2+ distinct stable version transitions.
+ *
+ * When detected:
+ * - The earliest occurrence remains suppressed
+ * - Later occurrences are marked unsuppressed with `ambiguousSuppressionDetected`
+ *
+ * @param findings - All findings after suppression has been applied
+ * @returns Findings with ambiguous later occurrences un-suppressed
+ */
+export function detectAmbiguousSuppressions(findings: Finding[]): Finding[] {
```
markcowl, line 28: We likely need soem additional testing of this functionality

markcowl: This is meant to detect cases where  type evolves multiple times:

`a: string -> a: int32 -> a:float64`  - this should result in separate suppressions for each break (2 in this case)

packages/typespec-breaking-change/src/suppression/codefixes.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Create a codefix that adds @approvedBreakingChange decorator to suppress a finding.
+ *
+ * Targets the origin declaration type when available (suppresses all uses of that declaration),
+ * falling back to the head type (wire type where the change was detected).
+ *
+ * @param finding - The finding to create a codefix for
+ * @returns A CodeFix, or undefined if no suitable target exists
+ */
+export function createApproveBreakingChangeCodeFix(finding: Finding): CodeFix | undefined {
```
markcowl, line 13: We likely need to augment this for ambiguous breaking changes

markcowl: This is not necessary for initial rollout

packages/typespec-breaking-change/src/suppression/decorators.ts
```diff
@@ -0,0 +1,200 @@
...
+  /** Optional identity path from suppression site to violation target. */
+  path?: string;
+}
+
+export interface ResolvedSuppression {
+  suppression: SuppressionMetadata;
+  target: Type;
+}
+
+export function $approvedBreakingChange(
```
markcowl, line 24: Belonmgs in Azure.Core

packages/typespec-breaking-change/src/lib.ts
```diff
@@ -0,0 +1,25 @@
+import { createTypeSpecLibrary, paramMessage } from "@typespec/compiler";
+
+export const $lib = createTypeSpecLibrary({
+  name: "@azure-tools/typespec-breaking-change",
+  diagnostics: {
+    "invalid-suppression-kind": {
+      severity: "warning",
+      messages: {
+        default: "Invalid DiffKind specified in suppression decorator.",
```
markcowl, line 9: SHould include the specified diffkind

markcowl (commented):
Review 1:  code

packages/typespec-breaking-change/test/cli-main.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    }));
+    vi.doMock("../src/pipeline/orchestrator.js", () => ({
+      analyzeBaseAndHead: vi.fn(),
+      analyzeProgram: vi.fn(),
+    }));
+
+    const errorSpy = vi.spyOn(console, "error").mockImplementation(() => undefined);
+    const { main } = await import("../src/cli/cli.js");
+
+    const code = await main(["head.tsp"]);
```
markcowl, line 26: seems like we need to mock this file someplace

packages/typespec-breaking-change/test/codefixes.test.ts
```diff
@@ -0,0 +1,364 @@
...
+          phase: "cross-version" as const,
+          suppressed: false,
+          versionPair: { baseVersion: "v1", headVersion: "v2", phase: "cross-version" as const },
+        },
+      ],
+    };
+
+    const diagnostics = emitFindingDiagnostics(program, modifiedResult as any);
+    expect(diagnostics).toHaveLength(0);
+  });
```
markcowl, line 363: Seems like we should have some tests over the contents of the codefix

markcowl (commented):
A few more comments

packages/typespec-breaking-change/src/cli/git-checkout.ts
```diff
@@ -0,0 +1,223 @@
...
+    // The checked-out revision is git-tracked source only — it has no
+    // `node_modules`, since dependencies are never committed to the repo.
+    // Without this, compiling the checked-out revision fails to resolve
+    // every npm import (e.g. `@typespec/http`, `@typespec/versioning`) and
+    // silently produces an empty, diagnostics-only Program instead of
+    // throwing — which callers can easily mistake for "nothing to compare"
+    // rather than "compilation was completely broken". Link the caller's
+    // own `node_modules` into the worktree root so the checked-out revision
+    // resolves dependencies exactly like the caller's own working tree does.
+    await linkNodeModulesIfPresent(repoRoot, worktreeRoot);
```
haolingdong-msft, line 133: I love this tool, but just would like to understnad more regarding to git checkout and dependency management.

It looks like the base worktree uses the head checkout's `node_modules` (I may be missing some context, so please correct me if I understand this incorrectly). Could a TypeSpec dependency upgrade (e.g., `@typespec/compiler`, `@typespec/http`, or `@typespec/versioning`) cause the base source to fail or produce a different result? Or is TypeSpec assumed to always be backward compatible across upgrades?

packages/typespec-breaking-change/src/cli/cli.ts
```diff
@@ -0,0 +1,399 @@
...
+  /** Custom report title for markdown output. */
+  reportTitle?: string;
+  /** Omit the H2 title line from markdown output. */
+  omitTitle?: boolean;
+}
+
+/**
+ * Parse CLI arguments into CliOptions.
+ */
+export function parseArgs(args: string[]): CliOptions {
```
timotheeguerin: Use node `parseArgs` or `yargs`

PR #5384, opened 2026-09-03
[Swagger Linter Migration] UnSupportedPatchProperties (devbox-heavy)

msyyc:
## Original Swagger linter

- linter code: [UnSupportedPatchProperties](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts)
- linter doc: [un-supported-patch-properties.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/un-supported-patch-properties.md)

The original rule performs these checks on every ARM Swagger PATCH operation:

- [x] Rejects writable top-level `id`.
- [x] Rejects writable top-level `name`.
- [x] Rejects writable top-level `type`.
- [x] Rejects writable top-level `location`.
- [x] Rejects writable `properties.provisioningState`.
- [x] Exempts `readOnly: true` properties.
- [x] Exempts properties whose `x-ms-mutability` excludes `update`.

## How the Swagger linter works

The ARM Spectral ruleset selects PATCH operations under both `paths` and `x-ms-paths` with resolved references enabled. `validatePatchBodyParamProperties` finds the first body parameter, resolves its schema through `getProperties`, checks the configured top-level denylist, then separately resolves the `properties` bag and checks `provisioningState`. Every failure targets the PATCH body parameter rather than the individual schema property.

For each selected property, the validator treats `readOnly: true` or an `x-ms-mutability` array without `update` as exempt. Spectral's resolved traversal can drop annotations that are siblings of `$ref`; the migration corpus confirmed false positives for referenced read-only `provisioningState` and immutable `location` properties. The TypeSpec rule intentionally preserves the authored lifecycle semantics instead of reproducing that validator defect. Swagger also reports emitted occurrences, while one authored TypeSpec property can be reused across operations and versions.

## Source TypeSpec lintdiff rule

- Validator rule ID: `UnSupportedPatchProperties`
- Local lintdiff rule: `unsupported-patch-properties`
- Canonical validator slug: `un-supported-patch-properties`
- Source branch: `feature/lintdiff-un-supported-patch-properties`
- Merged source revision: `de8d12e0244eb128604d2c4208f19325837fa852`
- Source repair PR: #5386
- Source worktree: `C:\dev\worktrees\lintdiff-un-supported-patch-properties`
- Source state: clean; there were no uncommitted rule changes
[...]

PR #5451, opened 2026-09-11
[Swagger Linter Migration] LroErrorContent

msyyc:
## Original Swagger linter

- Linter code: [LroErrorContent](https://github.com/Azure/azure-openapi-validator/blob/a970d991d2785184d2786b85e0a345dc3f37bc25/packages/rulesets/src/spectral/az-arm.ts#L233-L249)
- Linter documentation: [lro-error-content.md](https://github.com/Azure/azure-openapi-validator/blob/a970d991d2785184d2786b85e0a345dc3f37bc25/docs/lro-error-content.md)

The original validator:

- [x] Traverses operations under both `paths` and `x-ms-paths`.
- [x] Selects operations whose `x-ms-long-running-operation` value is exactly `true`.
- [x] Examines existing `schema.$ref` values on default, 4xx, and 5xx responses.
- [x] Requires those references to match ARM common-types v2-or-later `ErrorResponse`.
- [x] Rejects local references, common-types v1, and references to a different definition.
- [x] Accepts later versions, including multi-digit versions such as v10.

The Spectral rule walks emitted Swagger operations, filters on the raw LRO extension, and inspects only error response objects that already contain a top-level `schema.$ref`. It does not resolve references or require a response body, and it skips inline schemas because they have no `$ref`. Its regex is intentionally unanchored and the dot in `types.json` is unescaped. These are historical Swagger-validator behaviors, not behaviors implemented by the current native TypeSpec identity check. Generated reference occurrences and emitter schema-selection choices are not the diagnostic unit of the native rule.

- Validator rule ID: `LroErrorContent`
- Local lintdiff rule: `lro-error-content`
- Canonical validator slug: `lro-error-content`
- Historical source branch: `feature/repair-lintdiff-lro-error-content` (no longer available on GitHub)
- Recorded immutable promotion source commit: `e00fa5661d34fdf404ea83e155efadff6e5c4468`
- Recorded reviewed source head: `97d02dd81d170ce36ce122aa89cdb0905454cf01`
- Reachable, immutable historical source: [lro-error-content.ts at `274c4c327a4eef84ba6416281707dc9c2735781d`](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/src/rules/lro-error-content.ts)
- Historical migration evidence: [migration.md at the same reachable commit][migration-evidence]

The source rule was assumed done for the original promotion run and was not modified by this PR. The original promotion recorded that its pinned rule/test/fixture content matched the reviewed source head. The reachable links above replace the deleted-branch links; they do not represent a new source-repair run or promotion refresh. During this description update, the source-rule Git blob and every immediate fixture-directory tree/file blob, including `migration.md`, were verified identical between the recorded source commit in the local Git object database and the reachable commit `274c4c327a4eef84ba6416281707dc9c2735781d`.

The historical lintdiff source accepts qualifying ARM common-type references. The current promoted implementation intentionally differs following native-identity review feedback, as described below. Historical provenance must not be read as a claim of current behavioral equivalence.

## Destination analysis

The selected package is `@azure-tools/typespec-azure-resource-manager`. This rule enforces the ARM-specific standard error contract: error payloads of native long-running operations should use `Azure.ResourceManager.CommonTypes.ErrorResponse`. Its ownership follows those ARM semantics and the source fixture/migration classification, not a dependency on OpenAPI reference helpers. `@azure-tools/typespec-azure-core` was considered, but this is an ARM modeling requirement rather than a general Azure Core error-shape rule. Existing ARM LRO rules cover response status, final-result, and header behavior, not this error-payload identity requirement.

The implementation uses Azure Core's `getLroMetadata`, compiler model/namespace/template information, and HTTP operation/payload metadata. It does not call `isArmCommonType`, `getArmCommonTypeOpenAPIRef`, `getExternalTypeRef`, or `findArmCommonTypeRecord`, and it does not infer identity or versions from an OpenAPI reference string. A version-record lookup is not required merely to perform the identity check implemented here; the rule does not separately enforce a common-types version floor.

The official TypeSpec name is `use-standard-lro-error`, following the preferred-pattern naming convention. It is registered as `false` in the shared resource-manager ruleset so promotion does not enable a new diagnostic in that ruleset by default.

Current implementation: [use-standard-lro-error.ts](https://github.com/Azure/typespec-azure/blob/c6851a91baf51498557dd855b44a0d0b201982ff/packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts).

The rule uses an `operation` visitor, filters template declarations/instances with `isTemplateDeclarationOrInstance` and `isTemplatedInterfaceOperation`, and obtains each operation's HTTP metadata through `getHttpOperation`. It selects non-GET operations with native `getLroMetadata` and examines every existing default, 4xx, and 5xx response body, including overlapping HTTP status-code ranges. It obtains effective payload types with `createMetadataInfo` at `Visibility.Read`.

A compliant payload is a model whose `sourceModel` chain ends at the model named `ErrorResponse` in the namespace `Azure.ResourceManager.CommonTypes`. Aliases, `model is` copies (including chained copies), and nullable unions with exactly one non-null standard member are accepted. Custom models, derived models, spreads, inline models, primitives, literals, collections, binary bodies, and multipart bodies do not substitute for the standard type. A custom model is rejected even if its external reference points to the standard Swagger `ErrorResponse`; an external-reference override likewise does not disqualify a native standard-model copy.

Missing bodies, success responses, synchronous operations, and GET operations are ignored. Diagnostics are deduplicated by authored operation node, including shared declarations used across template instantiations. The rule checks the unprojected authored program, does not reconstruct historical return types, and does not consult raw OpenAPI extensions, AutoRest, TCGC, SDK scope, or emitter overrides to make its decisions.

Operations do not need `@service` or `@armProviderNamespace`. The rule no longer traverses `listServices`/`getHttpService` or applies the lintdiff-only ARM provider guard. ARM package/ruleset ownership expresses the intended audience; it is not an implicit compiler namespace filter. Native tests cover ordinary namespaces, nested services, operations without `@service`, unused templates, and exclusion of imported Azure library operations.

### Intentional changes from the historical source

[Human review feedback](https://github.com/Azure/typespec-azure/pull/5451#discussion_r3989118056) requested checking the TypeSpec type rather than accepting an OpenAPI reference as evidence of identity. Commit [`f1d9927f1`](https://github.com/Azure/typespec-azure/commit/f1d9927f1677aa8999ae9e7b4c3abb1a0c386692) implemented that direction; the subsequent instruction to ignore [the suggestion to restore reference acceptance](https://github.com/Azure/typespec-azure/pull/5451#discussion_r4003600568) retained the native identity contract.

This is a deliberate semantic change, not a behavior-preserving helper substitution:

- Historical source: qualifying v2+ native/legacy common-type references could establish compliance, including references on custom models.
- Current promotion: the native TypeSpec model identity establishes compliance; external-reference lookalikes are rejected, including v2/v5/v10 examples.
- The current predicate does not reproduce the historical regex, its matching quirks, or an explicit v2-or-later version check. Native tests cover the standard model under supported v3/v4/v5/v6 selections without claiming equivalent reference/version validation.
- Traversal was also adapted from provider-guarded service enumeration to the native operation visitor described above.

These changes are documented as intentional differences from the immutable source. The historical source and its corpus evidence have not been rewritten or rerun to imply equivalence.

The following maps historical fixture intent to the [current native test file](https://github.com/Azure/typespec-azure/blob/c6851a91baf51498557dd855b44a0d0b201982ff/packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts). Titles are exact strings from `it(...)` or the title template supplied to `it.each(...)`. Mapping a fixture does not imply unchanged historical expected results; the external-reference acceptance change is explicit.

| Original lintdiff fixture | Current native test title | Coverage / intentional difference |
| --- | --- | --- |
| [external-references][fixture-external] | `does not accept a custom model based on the external reference %s` (`it.each`) | Rejects custom-model references to v1/v2/v5/v10 ErrorResponse, a wrong definition, and a local definition. Valid-looking v2+ references are intentionally rejected rather than accepted as in the historical source. |
| [external-references][fixture-external] | `does not let an external reference override the standard TypeSpec model` | A `model is` copy remains standard even with a nonstandard external reference. |
| [inline-and-standard][fixture-inline] | `accepts common-type errors, model-is copies, and nullable standard errors` | Native identity, aliases, chained copies, and nullable standard payloads. |
| [inline-and-standard][fixture-inline] | `rejects inline, primitive, literal, tuple, unknown, and bytes error payloads` | Inline and non-model payload rejection. |
| [inline-and-standard][fixture-inline] | `rejects binary and multipart error payloads` | Body-specialization rejection. |
| [inline-and-standard][fixture-inline] | `ignores success responses, absent error bodies, synchronous operations, and GET operations` | Selector exclusions and no-body behavior. |
| [inline-and-standard][fixture-inline] | `does not let a raw OpenAPI extension disable native LRO metadata` | Raw `false` does not override native metadata. |
| [non-standard-error][fixture-nonstandard] | `rejects a custom error supplied to an ARM template` | Native ARM async template with a custom error. |
| [non-standard-error][fixture-nonstandard] | `does not treat a raw OpenAPI extension as native LRO metadata` | Raw `true` alone does not establish a native LRO. |
| [reference-shapes][fixture-shapes] | `rejects custom model reference shapes` | Named, derived, spread, and effective inline custom models. |
| [reference-shapes][fixture-shapes] | `rejects named scalar, enum, union, collection, generic, and nullable custom errors` | Named non-model/custom references and nullable custom errors. |
| [reference-shapes][fixture-shapes] | `checks ordinary and nested namespaces without an ARM provider decorator` | Native traversal without the lintdiff-only provider guard. |
| [standard-error][fixture-standard] | `accepts the standard error from ARM templates` | Compliant native ARM template control. |
| [template-and-versions][fixture-templates] | `rejects a custom error supplied to an ARM template` | Custom template errors remain invalid. |
| [template-and-versions][fixture-templates] | `does not let a raw OpenAPI extension disable native LRO metadata` | Explicit raw extension does not alter native checking. |
| [template-and-versions][fixture-templates] | `checks authored return types without reconstructing historical versions` | Removed authored operation remains visible; historical return type is not reconstructed. |
| N/A (review regression) | `does not import TCGC, OpenAPI, or experimental compiler APIs` | Import-boundary regression coverage. |
| N/A (review regression) | `accepts the standard TypeSpec model with common-types %s` (`it.each`) | Standard native model under v3/v4/v5/v6 common-type selections; not an explicit version-floor check. |
| N/A (review regression) | `rejects a same-named model with the standard shape in a different namespace` | Name/shape alone outside the standard namespace does not establish identity. |
| N/A (review regression) | `checks default, explicit, and ranged error status codes` | Default, numeric 5xx, and overlapping 4xx range selectors. |
| N/A (review regression) | `reports once across multiple error responses` | Per-operation diagnostic deduplication. |
| N/A (review regression) | `checks operations without a service decorator` | Operation visitor does not require `@service`. |
| N/A (review regression) | `does not diagnose imported Azure.Core or Azure.ResourceManager library operations` | Imported library operations do not produce diagnostics. |
| N/A (review regression) | `reports a nested service operation without an ARM provider decorator` | Nested service behavior without provider metadata. |
| N/A (review regression) | `ignores unused operation templates` | Unused-template filtering. |
| N/A (review regression) | `reports a shared operation declaration once across template instantiations` | Authored-node deduplication across instantiations. |

## Migration evidence

The linked [historical migration report][migration-evidence] records focused fixture results, real-service comparisons, corpus counts, one-sided project explanations, compile failures, and uncertainty for the historical lintdiff implementation. It is not a corpus evaluation of the identity-only implementation introduced by `f1d9927f1`. The changed external-reference expectations and traversal contract mean historical diagnostic counts must not be presented as evidence that the current implementation has identical behavior or corpus results.

No lintdiff corpus harness was rerun for this promotion's native-identity review change or for this description correction. Current native regression cases are mapped above; historical comparison data remains historical.

### Historical original-promotion validation

The original promotion description recorded the following before the subsequent native-identity/traversal change. These are preserved as historical results, not fresh results for the current head:

- ARM dependency-closure build, ARM package build, and ARM package lint passed.
- Focused native rule suite passed: 19/19.
- ARM documentation regeneration and formatting checks passed.
- Azure rulesets build and tests passed: 4/4.
- Full ARM package tests passed: 478/478.
- Repository `pnpm format` and `pnpm lint` passed.
- Chronus status passed with destination `feature` and disabled-ruleset `internal` entries.
- Original focused reviews and follow-up test-target assertions were reported clean for the implementation reviewed at that time. They do not establish semantic equivalence or a clean review of the later identity-only implementation.

### Subsequent documentation-only fix

Commit [`c6851a91b`](https://github.com/Azure/typespec-azure/commit/c6851a91baf51498557dd855b44a0d0b201982ff) added suppression guidance and the manual ARM rules-table entry without changing the rule. Affected documentation regeneration, explicit source/generated-document Prettier checks, repository `pnpm format`, repository `pnpm lint`, and `git diff --check` passed. Source/package tests and corpus validation were not rerun for those documentation-only edits. The review/fix loop stopped after publication because of a publication-precheck retry outside its approval gate; no subsequent clean verification review is claimed.

This description correction changes no implementation, tests, or historical evidence, and does not claim a new build, unit-test run, or corpus run.

## Validation blocker

The original broad `TYPESPEC_SKIP_WEBSITE_BUILD=true pnpm validate:pr` run completed 8/9 checks. Branch freshness, full repository build, lint, format, spelling, docs regeneration, changeset validation, and diff hygiene passed. Its broad test step had three 10-second timeouts while 3,380 tests passed: two `@azure-tools/typespec-autorest` setup hooks (`options.test.ts` and `x-ms-examples.test.ts`) and the playground `compile Widget Manager` sample. This is the historical broad-validation limitation reported during original promotion; it has not been rerun here and is not a claim about current CI status.

## Promotion sync policy

The immutable lintdiff source remains the provenance and historical comparison baseline. The native-identity and operation-visitor changes above are explicitly recorded exceptions to behavioral equivalence, following human review; this PR must not simultaneously claim those changes and unchanged source semantics.

Future semantic changes require explicit review and the applicable authorized source-repair/promotion-refresh workflow. No source repair, source modification, or refresh is implied by correcting this description. If current behavior is later propagated back to lintdiff, that repair needs its own validation and updated provenance; historical evidence must not be relabeled as validation of the changed implementation.

[migration-evidence]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/migration.md
[fixture-external]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/external-references/main.tsp
[fixture-inline]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/inline-and-standard/main.tsp
[fixture-nonstandard]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/non-standard-error/main.tsp
[fixture-shapes]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/reference-shapes/main.tsp
[fixture-standard]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/standard-error/main.tsp
[fixture-templates]: https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/packages/typespec-lintdiff/test/fixtures/LroErrorContent/template-and-versions/main.tsp


packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
```diff
@@ -0,0 +1,114 @@
...
+        return members.length === 1 && isStandardError(members[0], service);
+      }
+      return false;
+    }
+
+    return {
+      root() {
+        for (const service of listServices(program)) {
+          const [httpService] = getHttpService(program, service.type);
+          for (const httpOperation of httpService.operations) {
```
timotheeguerin, line 72: why do we iterate of the service here an not the operation like the other rules

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
```diff
@@ -0,0 +1,114 @@
...
+} from "../common-types.js";
+
+const standardErrorReference =
+  /.*\/common-types\/resource-management\/v(([1-9]\d+)|[2-9])\/types.json#\/definitions\/ErrorResponse/;
+
+export const useStandardLroErrorRule = createRule({
+  name: "use-standard-lro-error",
+  docs: fileRef.fromPackageRoot("src/rules/use-standard-lro-error.md"),
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/use-standard-lro-error",
+  description:
```
timotheeguerin, line 30: we don't support less than v3 in arm lib so I would just change the docs + message to say you have to use the ErrorResponse from common types

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
```diff
@@ -0,0 +1,114 @@
...
+      let reference = getExternalTypeRef(program, type);
+      if (
+        !reference &&
+        isArmCommonType(type) &&
+        (type.kind === "Model" ||
+          type.kind === "ModelProperty" ||
+          type.kind === "Enum" ||
+          type.kind === "Union")
+      ) {
+        reference = getArmCommonTypeOpenAPIRef(program, type, { service });
```
timotheeguerin, line 52: why are we referencing openapi, we should check we are referencing the TypeSpec type

timotheeguerin: if that is already a common type added in another rule this seems like something missed in the review and the other rule should be fixed too

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
msyyc: Agreed. Updated the description, diagnostic, documentation, and change entry in f1d9927f1 to require `Azure.ResourceManager.CommonTypes.ErrorResponse`, without the Swagger-specific “v2 or later” wording.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
msyyc: Agreed. Removed the OpenAPI reference helpers and regex in f1d9927f1. The rule now checks the TypeSpec model name and namespace, following `sourceModel` for `model is` copies. Aliases and nullable standard errors remain supported; custom models no longer pass solely because an external reference points at ErrorResponse. Added regression coverage for both cases.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
msyyc: Fixed in f1d9927f1. Replaced the service traversal with an `operation` listener and `getHttpOperation()`. Kept template filtering and authored-node deduplication, and added regression coverage for operations without `@service`.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
msyyc: I checked the merged linter rules on the synced main branch (`5f4088cb9`) and the current heads of #5271, #5294, #5336, #5361, #5364, #5384, #5423, #5451, #5456, and #5457. I found no other occurrences of the same issue: accepting a TypeSpec type as the standard common type solely because of its OpenAPI reference. The implementation in this PR is now fixed in f1d9927f1.

There is a narrower related cleanup in #5271: it already recognizes common types through native metadata, but converts common-type records into OpenAPI references and then parses them back to obtain version information. I left the details and a suggestion to use the native records directly here: https://github.com/Azure/typespec-azure/pull/5271#issuecomment-5661318416. No equivalent issue was found in the other checked rules.

msyyc:
Audit of `1f60b21d5688d5803b722f6ff876635db476a504` against the updated migration skills found **stale promotion documentation**, not a confirmed remaining instance of the earlier external-reference acceptance bug.

The current [identity check](https://github.com/Azure/typespec-azure/blob/1f60b21d5688d5803b722f6ff876635db476a504/packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts#L70-L87) examines native model name/namespace and the `sourceModel` chain. The tests reject custom models merely pointing at standard ErrorResponse external references. Checking native identity is not itself prohibited OpenAPI-reference inference, and the absence of `findArmCommonTypeRecord` alone does not establish a defect.

However, the PR body's destination analysis still describes dependencies on `isArmCommonType`, `getArmCommonTypeOpenAPIRef`, and legacy external-reference handling that the current implementation no longer uses. Please update the implementation/destination explanation to describe the actual native identity contract and rejection of external-reference lookalikes. Reconcile the fixture mapping and source-versus-promotion evidence with that change, and refresh any source-branch evidence links that no longer resolve.

The [promotion skill](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#10-review-commit-push-and-create-a-draft-pr) requires the PR explanation to match the current APIs, semantics, and intentional adaptations. Historical migration evidence should remain labeled as historical rather than implying it was rerun for the updated implementation.

This comment does not request reintroducing reference parsing or replacing every native identity check with version-record resolution.

msyyc:
> Audit of `1f60b21d5688d5803b722f6ff876635db476a504` against the updated migration skills found **stale promotion documentation**, not a confirmed remaining instance of the earlier external-reference acceptance bug.
> 
> The current [identity check](https://github.com/Azure/typespec-azure/blob/1f60b21d5688d5803b722f6ff876635db476a504/packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts#L70-L87) examines native model name/namespace and the `sourceModel` chain. The tests reject custom models merely pointing at standard ErrorResponse external references. Checking native identity is not itself prohibited OpenAPI-reference inference, and the absence of `findArmCommonTypeRecord` alone does not establish a defect.
> 
> However, the PR body's destination analysis still describes dependencies on `isArmCommonType`, `getArmCommonTypeOpenAPIRef`, and legacy external-reference handling that the current implementation no longer uses. Please update the implementation/destination explanation to describe the actual native identity contract and rejection of external-reference lookalikes. Reconcile the fixture mapping and source-versus-promotion evidence with that change, and refresh any source-branch evidence links that no longer resolve.
> 
> The [promotion skill](https://github.com/Azure/typespec-azure/blob/274c4c327a4eef84ba6416281707dc9c2735781d/.github/skills/lintdiff-rule-promote/SKILL.md#10-review-commit-push-and-create-a-draft-pr) requires the PR explanation to match the current APIs, semantics, and intentional adaptations. Historical migration evidence should remain labeled as historical rather than implying it was rerun for the updated implementation.
> 
> This comment does not request reintroducing reference parsing or replacing every native identity check with version-record resolution.

Fixed

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
```diff
@@ -0,0 +1,97 @@
...
+        ) {
+          return;
+        }
+
+        const [httpOperation] = getHttpOperation(program, operation);
+        if (httpOperation.verb === "get" || getLroMetadata(program, operation) === undefined) {
+          return;
+        }
+
+        const invalidBody = httpOperation.responses.some(
```
markcowl, line 51: I think this doesn't ctach cases in which there is no error response at all, which should also be a violation

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.md
```diff
@@ -0,0 +1,55 @@
...
+  restart is ArmResourceActionAsync<Widget, void, void, Error = CustomError>;
+}
+```
+
+## ✅ Correct
+
+```tsp
+@armResourceOperations
+interface Widgets {
+  restart is ArmResourceActionAsync<Widget, void, void, Error = CommonTypes.ErrorResponse>;
```
markcowl, line 48:
```suggestion
  restart is ArmResourceActionAsync<Widget, void, void>;
```

markcowl: No reason to override the error type for the default error type

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
```diff
@@ -0,0 +1,429 @@
...
+        op success(): Accepted | { @statusCode statusCode: 200; @body body: string; };
+        @route("/sync") @post op sync(): Accepted | Failure<string>;
+        @Azure.Core.pollingOperation(poll) @route("/get") @get
+        op getOperation(): Accepted | Failure<string>;
+      `,
+      )
+      .toBeValid();
+  });
+
+  it("does not treat a raw OpenAPI extension as native LRO metadata", async () => {
```
markcowl, line 321: This test is unnecessary

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
```diff
@@ -0,0 +1,429 @@
...
+        `${header}
+        @TypeSpec.OpenAPI.extension("x-ms-long-running-operation", true)
+        @route("/raw") @post
+        op rawExtension(): Accepted | Failure<string>;
+      `,
+      )
+      .toBeValid();
+  });
+
+  it("does not let a raw OpenAPI extension disable native LRO metadata", async () => {
```
markcowl, line 333: This test is unnecessary

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
```diff
@@ -0,0 +1,429 @@
...
+      `,
+      )
+      .toBeValid();
+  });
+
+  it("reports a shared operation declaration once across template instantiations", async () => {
+    await tester
+      .expect(
+        `${header}
+        interface Actions<T> { @route("/run") op run is Lro<T>; }
```
markcowl, line 408: This should not be necessary

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
```diff
@@ -0,0 +1,429 @@
...
+        @route("/one") interface One extends Actions<string> {}
+        @route("/two") interface Two extends Actions<int32> {}
+      `,
+      )
+      .toEmitDiagnostics([diagnostic]);
+  });
+
+  it("checks authored return types without reconstructing historical versions", async () => {
```
markcowl, line 416: This is a feature of how linter rules work,  we should not test it explicitly

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
```diff
@@ -0,0 +1,429 @@
...
+      .expect(
+        `${header.replace("@service", "@service @TypeSpec.Versioning.versioned(Versions)")}
+        enum Versions { v1: "2024-01-01", v2: "2025-01-01" }
+        @TypeSpec.Versioning.returnTypeChangedFrom(Versions.v2, Accepted | Failure<string>)
+        @route("/changed") op changed is Lro<CommonTypes.ErrorResponse>;
+        @TypeSpec.Versioning.removed(Versions.v2)
+        @route("/removed") op removed is Lro<string>;
+      `,
+      )
+      .toEmitDiagnostics([diagnostic]);
```
markcowl, line 427: Need a test of an lro operation without an error

markcowl (commented):
A few minor issues, and one extra arear of coverage I think we should have

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
msyyc: Fixed in 71397aa59. Non-GET native LROs now require at least one default, 4xx, or 5xx response, with a dedicated diagnostic when none exists. Existing error payloads still require the native standard ErrorResponse type. A declared error response without a body remains allowed; that is distinct from having no error response. The docs and change entry now describe this intentional expansion beyond the historical rule.

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
msyyc: Removed in 71397aa59. The rule uses native LRO metadata and does not read raw OpenAPI extensions, so this migration-specific test is unnecessary here.

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
msyyc: I kept this test because it covers deduplication implemented by this rule's `reported` set, keyed by authored operation node. The compiler walker deduplicates semantic Type objects, but separate template instantiations can share an authored node; the linter reporting path does not provide that same deduplication. This regression checks that those instantiations produce only one diagnostic.

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
msyyc: Removed in 71397aa59, along with the other raw OpenAPI-extension test. The native ARM template and polling-operation cases remain.

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
msyyc: Added `rejects a long-running operation without an error response` in 71397aa59, asserting the dedicated missing-response diagnostic. I also adjusted the success-payload exclusion case to include a standard error response, and retained coverage that bodyless errors, synchronous operations, and GET operations remain allowed. The focused suite passes all 30 cases.

packages/typespec-azure-resource-manager/test/rules/use-standard-lro-error.test.ts
msyyc: Agreed; removed the historical-version reconstruction test in 71397aa59. That behavior belongs to the linter framework rather than this rule.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.md
msyyc: Applied in 71397aa59. The correct example now uses `ArmResourceActionAsync<Widget, void, void>` and notes that the template supplies the standard ErrorResponse by default.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.md
```diff
@@ -0,0 +1,61 @@
+Non-GET long-running ARM operations must define at least one default, 4xx, or 5xx
+error response and use `Azure.ResourceManager.CommonTypes.ErrorResponse` for every
+such response body.
+
+The rule checks the TypeSpec model, not an emitted OpenAPI reference. Aliases,
```
timotheeguerin: still some openapi references

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
```diff
@@ -0,0 +1,106 @@
...
+          reported.add(source);
+          return;
+        }
+
+        const invalidBody = errorResponses.some((response) =>
+          response.responses.some(
+            ({ body }) =>
+              body !== undefined &&
+              (body.bodyKind !== "single" ||
+                !isStandardError(metadata.getEffectivePayloadType(body.type, Visibility.Read))),
```
timotheeguerin: do we really need to use the visibility here? shouldn't it just be using the plain common types model?

simlar for the logic in isStandardError

msyyc: I checked this against concrete HTTP payload cases. The normalization is unnecessary for a direct `CommonTypes.ErrorResponse` or explicit `@body body: CommonTypes.ErrorResponse`, but it matters for this wrapper:

```tsp
model Failure {
  ...CommonTypes.ErrorResponse;
  @statusCode status: 400;
  @header requestId: string;
}
```

HTTP processing removes the status/header properties and leaves an anonymous body. `getEffectivePayloadType(..., Visibility.Read)` recovers `CommonTypes.ErrorResponse` from its property origins; a direct identity check on `body.type` would reject it. `Read` selects the response payload. This normalization was inherited from the original lintdiff rule, and 12d438c66 adds regressions for anonymous, named-spread, and model-is wrappers plus a short explanatory comment.

There was a separate issue in `isStandardError`: following `sourceModel` unconditionally also accepted copies with extra payload fields. The same commit now checks each copy against its source before following the chain, preserving unchanged copies and the existing nullable handling while rejecting modified copies. The focused suite passes 40 cases.

My recommendation is to preserve standard-payload wrappers. Do you agree these should be supported, or do you want the rule to require the standard model directly? I am keeping this discussion open for that contract decision rather than treating the retained normalization as agreed.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.md
msyyc: Updated in 12d438c66. Removed the OpenAPI-reference/Swagger-override discussion from the user-facing guidance. The docs now describe the native TypeSpec contract, distinguish standard-payload wrappers from independently defined lookalikes, and show a status/header wrapper. The historical Swagger link remains only in the LintDiff Equivalent provenance section.

packages/typespec-azure-resource-manager/src/rules/use-standard-lro-error.ts
timotheeguerin: yeah but that example doesn't look like something we should allow in arm @markcowl. Feels like we make the rule more complex here to allow a pattern we shouldn't anyway

A few other rules also use visibility would be good to go through all and make sure we do need it, it adds cost to the rule to transform so we should be sure it does allow pattern we want.

PR #5456, opened 2026-09-11
[Swagger Linter Migration] ConsistentPatchProperties

msyyc:
## Original Swagger linter

linter code: [ConsistentPatchProperties](https://github.com/Azure/azure-openapi-validator/blob/6243cb01c16c7535cd3b8df6f45fbeb3c095ed7f/packages/rulesets/src/spectral/functions/consistent-patch-properties.ts)

linter doc: [consistent-patch-properties.md](https://github.com/Azure/azure-openapi-validator/blob/6243cb01c16c7535cd3b8df6f45fbeb3c095ed7f/docs/consistent-patch-properties.md)

- [x] Inspect ARM PATCH operations selected by `$.paths.*.patch`.
- [x] Select the first body parameter's schema; skip absent request schemas.
- [x] Select PATCH `200`, then PATCH `201`, then same-path GET `200` or `201`; skip absent comparison schemas.
- [x] Include inherited `allOf` properties and direct properties.
- [x] Require every PATCH property at the same nesting level in the resource schema.
- [x] Recurse into corresponding named property schemas while allowing a subset.
- [x] Report missing or wrongly nested properties; do not compare requiredness, scalar types, array items, or additional-property values.

The Spectral function reads the resolved PATCH object, first `in: body` parameter, and selected response schema. GET fallback looks up the same path in the resolved document. `diffSchema` flattens `allOf` property maps (direct properties override inherited ones), then recursively compares named `properties`. Missing properties produce dotted names, but diagnostics target the PATCH body's `parameters[index].schema`, not the property declaration. Non-object PATCH values and absent request/comparison schemas produce no findings. There is no registered lifecycle-operation requirement, name whitelist, or type-equality check.

This supports custom operations such as Application Insights `WorkItemConfigurationsOperationGroup.updateItem`, beyond the existing official `arm-resource-patch` coverage. Known differences are not presented as complete Swagger parity: AutoRest can omit scoped declarations; native status ranges can be eligible when emitted `2XX` is missed; native missing-object findings expand to authored leaves instead of a parent at the operation body path. Reused declarations and unprojected versions also explain count differences. Source evidence records external-common-types inherited-envelope noise and an Informatica omission whose internal validator cause remains unisolated.

Native inheritance overrides deliberately differ from emission: `never` can remove a property that emitted `allOf` retains; an emitted derived scalar discriminator can override an inherited object that the native comparison retains. The validator correctly evaluates those emitted schemas. These are contract differences, not validator defects or reasons to recreate emitter ordering.

Source rule: [consistent-patch-properties.ts](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/src/rules/consistent-patch-properties.ts).

Validator ID `ConsistentPatchProperties`; local name `tsp-lintdiff-local-linter/consistent-patch-properties`; canonical validator slug `consistent-patch-properties`. Source branch `feature/lintdiff-consistent-patch-properties-native`, immutable commit **`52a52ac4d60a9a208f40d103ff295e573fd34672`**, worktree `C:\dev\worktrees\lintdiff-consistent-patch-properties`. Source was clean before copying and remained clean; no uncommitted source-rule changes were included. Original PR #5399 and native-boundary repair #5439 are merged. Reviewed source repair PR #5453 supplies the approved inherited-override/discriminator fixes and remains the semantic source of truth while this official PR is prepared.

The named source rule was **assumed done for this run**, not marked done in persistent metadata or claimed as explicitly user-confirmed. It was not modified during promotion. Promotion uses the separate `C:\dev\worktrees\promote-lintdiff-consistent-patch-properties` worktree, branch `promote-consistent-patch-properties-to-arm`, originally based on canonical main `0b96c781da85bdbf2f86115fad158803dba2a46b`.

## Destination analysis

Selected **`@azure-tools/typespec-azure-resource-manager`** under per-run do-not-ask authorization: an evidence-backed agent decision, not a claim that the user explicitly selected the destination. Fixture metadata and both catalog records classify the contract as ARM (`RPC-Patch-V1-01`). Its recursive resource/PATCH layout requirements are ARM-specific even though compiler/HTTP APIs implement most traversal.

Azure Core is technically plausible for generic HTTP traversal but inappropriate for this ARM contract. Existing `arm-resource-patch` only partially overlaps: registered resource/top-level envelope checks do not provide this recursive same-level comparison and custom-operation response fallback. It remains unchanged. The concise official name stays `consistent-patch-properties`.

No new dependencies, emitter calls, TCGC context, or private-state access are introduced. The new rule is explicitly **false** in the shared resource-manager ruleset. Separate Chronus entries describe ARM availability as a `feature` and disabled-only rulesets registration as `internal`.

`getAllHttpServices` supplies native HTTP operations. Eligible PATCH model bodies compare against PATCH `200`, PATCH `201`, same-path GET `200`, then GET `201`. An exact status precedes a containing range; the selected response's first available body is used. A scalar PATCH `200` is not replaced by a later model response: named top-level PATCH properties are then inconsistent.

The approved source helpers are preserved. `getProperty` selects effective declarations by authored name before JSON encoding and `never` filtering. `resolveEncodedName(..., "application/json")`, inheritance, `isNullType`, and `getDiscriminator` provide the native shape. The complete effective authored inheritance chain is collected before adding missing discriminator metadata; inherited encoded objects keep their nested type. An authored discriminator source name encoded differently prevents a second synthetic property. Nullable single-model unions recurse. Arrays/records/empty models contribute named properties only, not elements/indexer values.

Warning severity and diagnostic intent are preserved. Authored findings target properties with JSON paths; implicit discriminators target their model. Active model-pair tracking terminates cycles without globally deduplicating sibling paths or repeated operations sharing a source target. There is no additional API-version projection or emitter/client scope filtering: the current compiler program is checked.

**Intentional promotion adaptations:** remove the documented mixed-lintdiff provider-namespace guard; add existing ARM `isInternalTypeSpec` filtering; add official docs/URL/registration. Neighboring `arm-no-path-casing-conflicts` likewise traverses HTTP services without provider metadata and excludes internal declarations. Ordinary and nested no-provider namespaces are tested. Ruleset selection defines the intended audience, not automatic compiler namespace isolation: explicitly selecting this rule checks all non-internal discovered HTTP PATCH operations. Compiler/HTTP operation enumeration separately excludes uninstantiated templates.

All **22 fixtures** map below to literal cases in `packages/typespec-azure-resource-manager/test/rules/consistent-patch-properties.test.ts`; **55 additional regression rows** complete **77 distinct tests**. No snapshots or harness outputs were copied.

Scope cases retain equivalent native declarations without loading TCGC or a scope adapter; they assert native outcomes, not adapter integration. The two never fixtures and two inherited-discriminator fixtures overlap with corresponding source regressions, adding intermediate inheritance; these tests are counted once. Named `Payload` preserves the fixture's object shape while allowing exact authored-target assertions. Target-sensitive cases use supported `file`/`pos`/`end` locations, including reused targets and synthesized models.

| --- | --- | --- |
| [inconsistent-patch](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/inconsistent-patch/main.tsp) | `it("reports displayName moved out of properties in an ARM resource PATCH")` | Wrong-level `displayName`. |
| [nested-extra-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/nested-extra-property/main.tsp) | `it("reports the nested extraPatchOnly property in an ARM resource PATCH")` | `properties.extraPatchOnly`. |
| [custom-patch-operation](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/custom-patch-operation/main.tsp) | `it("reports displayName at the wrong level in an unregistered custom PATCH")` | No lifecycle registration needed. |
| [patch-201-response](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/patch-201-response/main.tsp) | `it("uses a PATCH 201 resource response to report a moved displayName")` | PATCH 201 response selection. |
| [get-201-fallback](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/get-201-fallback/main.tsp) | `it("uses a same-path GET 201 fallback to report a moved displayName")` | GET 201 fallback. |
| [response-precedence](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/response-precedence/main.tsp) | `it("selects a scalar PATCH 200 before a matching PATCH 201 model")` | Top-level finding, not leaf expansion. |
| [payload-property-shape](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/payload-property-shape/main.tsp) | `it("matches different authored property names by their JSON wireName")` | Matching JSON names are valid. |
| [nullable-object-mismatch](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/nullable-object-mismatch/main.tsp) | `it("reports properties.child.displayName inside mismatched nullable objects")` | Nullable nested mismatch. |
| [nullable-object-match](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/nullable-object-match/main.tsp) | `it("accepts matching nullable child objects")` | Nullable matching control. |
| [non-model-property-shape](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/non-model-property-shape/main.tsp) | `it("accepts a same-named array PATCH property against a scalar resource property")` | No nested named properties. |
| [scoped-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/scoped-property/main.tsp) | `it("checks the native clientOnly property from the scoped-property fixture")` | Native property remains visible. |
| [scoped-get-fallback](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/scoped-get-fallback/main.tsp) | `it("uses the native GET fallback from the scoped-get-fallback fixture")` | Native GET remains eligible. |
| [scoped-patch-operation](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/scoped-patch-operation/main.tsp) | `it("checks the native PATCH endpoint from the scoped-patch-operation fixture")` | Native endpoint remains eligible. |
| [scoped-response-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/scoped-response-property/main.tsp) | `it("accepts the native description property from the scoped-response-property fixture")` | Native response property matches. |
| [synthesized-discriminator](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/synthesized-discriminator/main.tsp) | `it("reports a synthesized properties.details.kind discriminator on its model")` | Model target verified. |
| [encoded-discriminator-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/encoded-discriminator-property/main.tsp) | `it("compares an encoded authored discriminator instead of a synthesized kind")` | Authored leaf target verified. |
| [same-level-subset](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/same-level-subset/main.tsp) | `it("accepts a same-level properties.description subset")` | Valid nested subset. |
| [async-get-fallback](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/async-get-fallback/main.tsp) | `it("accepts an async PATCH 202 body using the same-path ARM GET resource")` | Valid asynchronous fallback. |
| [never-patch-override](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/never-patch-override/main.tsp) | `it("ignores a never override in a top-level PATCH model")` | Native-valid; also intermediate inheritance regression. |
| [never-resource-override](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/never-resource-override/main.tsp) | `it("reports a property removed by never in a top-level resource model")` | Native-invalid; also intermediate inheritance regression. |
| [inherited-encoded-discriminator-patch](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/inherited-encoded-discriminator-patch/main.tsp) | `it("checks the nested shape of an inherited encoded PATCH property")` | `details.kind.extra`; authored target, intermediate inheritance. |
| [inherited-encoded-discriminator-response](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/inherited-encoded-discriminator-response/main.tsp) | `it("uses the nested shape of an inherited encoded response property")` | Native matching object; intermediate inheritance. |
| N/A (native regression) | `it("uses a PATCH response range containing 200")` | PATCH 200-299 range. |
| N/A (native regression) | `it("uses a PATCH response range starting at 201")` | PATCH 201-299 range. |
| N/A (native regression) | `it("uses a GET fallback response range containing 200")` | GET 200 range. |
| N/A (native regression) | `it("uses a GET fallback response range starting at 201")` | GET 201 range. |
| N/A (native regression) | `it("prefers an exact PATCH 200 over an overlapping range")` | Exact selection. |
| N/A (native regression) | `it("prefers an exact PATCH 201 over an overlapping range")` | Exact selection. |
| N/A (native regression) | `it("prefers an exact GET 200 over an overlapping fallback range")` | Exact fallback selection. |
| N/A (native regression) | `it("prefers an exact GET 201 over an overlapping fallback range")` | Exact fallback selection. |
| N/A (native regression) | `it("prefers PATCH 200 over PATCH 201 and both GET success responses")` | Response priority. |
| N/A (native regression) | `it("prefers PATCH 201 over GET 200 and GET 201")` | Response priority. |
| N/A (native regression) | `it("prefers GET 200 over GET 201 when PATCH has no comparison response")` | Fallback priority. |
| N/A (native regression) | `it("falls through a bodyless PATCH 200 to PATCH 201")` | Missing body fallback. |
| N/A (native regression) | `it("does not replace a bodyless exact 200 with its overlapping range")` | Exact-response semantics. |
| N/A (native regression) | `it("skips PATCH operations with no request body")` | Request guard. |
| N/A (native regression) | `it("skips PATCH operations with no comparison response body")` | Response guard. |
| N/A (native regression) | `it("does not use a GET resource from a different path")` | Exact path boundary. |
| N/A (native regression) | `it("reports JSON-encoded paths while targeting the authored nested property")` | Encoded path and source location. |
| N/A (native regression) | `it("accepts matching inherited JSON properties and ignores never properties")` | Inheritance and never control. |
| N/A (native regression) | `it("reports a missing inherited property on its base declaration")` | Base property target. |
| N/A (native regression) | `it("accepts matching synthesized discriminators on inherited model shapes")` | Matching discriminator control. |
| N/A (native regression) | `it("accepts same-named records without comparing indexer values")` | Indexers excluded. |
| N/A (native regression) | `it("accepts same-named empty objects against scalars")` | No nested names. |
| N/A (native regression) | `it("accepts same-named scalars without comparing scalar types")` | Not type equality. |
| N/A (native regression) | `it("accepts an array request body with no named properties")` | Array control. |
| N/A (native regression) | `it("accepts a record request body with no named properties")` | Record control. |
| N/A (native regression) | `it("accepts an empty-object request body with no named properties")` | Empty-object control. |
| N/A (native regression) | `it("skips a scalar request body")` | Non-model body. |
| N/A (native regression) | `it("reports missing array record scalar and empty-object properties as leaves")` | Exactly four leaves. |
| N/A (native regression) | `it("reports only top-level PATCH properties against a scalar response")` | Exactly two top-level findings. |
| N/A (native regression) | `it("reports nested object leaves against a same-named scalar property")` | Exactly two nested findings. |
| N/A (native regression) | `it("terminates matching recursive model pairs without diagnostics")` | Recursive control. |
| N/A (native regression) | `it("terminates recursive model pairs and reports exactly one extra path")` | Recursive mismatch. |
| N/A (native regression) | `it("terminates a missing recursive subtree with exactly two leaf paths")` | Missing recursive subtree. |
| N/A (native regression) | `it("terminates an object-to-scalar recursive subtree with exactly two paths")` | Recursive object/scalar comparison. |
| N/A (native regression) | `it("does not deduplicate the same property target across sibling model pairs")` | Two paths, one source location. |
| N/A (native regression) | `it("does not deduplicate the same property target across PATCH operations")` | Two operations, one source location. |
| N/A (native regression) | `it("collects every missing subtree sibling path for a shared leaf model")` | Shared leaf model paths. |
| N/A (native regression) | `it("checks PATCH in an ordinary service namespace without provider metadata")` | Provider guard adaptation. |
| N/A (native regression) | `it("checks PATCH in a nested service namespace without provider metadata")` | Nested applicability. |
| N/A (native regression) | `it("ignores non-PATCH operations with mismatched request bodies")` | PUT/POST excluded. |
| N/A (native regression) | `it("ignores internal TypeSpec library PATCH operations")` | Library filtering. |
| N/A (native regression) | `it("ignores internal Azure.Core library PATCH operations")` | Library filtering. |
| N/A (native regression) | `it("ignores internal Azure.ResourceManager library PATCH operations")` | Library filtering. |
| N/A (native regression) | `it("ignores uninstantiated operation and interface templates")` | Template filtering. |
| N/A (native regression) | `it("checks an instantiated PATCH operation template")` | Instantiated operation remains eligible. |
| N/A (native regression) | `it("ignores a never override in a nested PATCH model")` | Nested override through intermediate inheritance. |
| N/A (native regression) | `it("reports a property removed by never in a nested resource model")` | Nested response override; authored target. |
| N/A (native regression) | `it("shadows an inherited property by source name before resolving its encoded name")` | Never override changing JSON name. |
| N/A (native regression) | `it("does not let a never property hide a differently named inherited payload property")` | Unrelated source names sharing JSON name. |
| N/A (native regression) | `it("uses a redeclared property's encoded name rather than its base declaration")` | Non-never encoded redeclaration. |
| N/A (native regression) | `it("checks the nested shape of an inline encoded PATCH property")` | Inline authored leaf control. |
| N/A (native regression) | `it("uses the nested shape of an inline encoded response property")` | Inline matching response control. |
| N/A (native regression) | `it("does not synthesize a second discriminator for an inherited authored name encoded differently")` | Source kind encoded as wireKind. |
| N/A (native regression) | `it("still reports a synthesized discriminator with no authored property")` | Synthetic model target. |
| N/A (native regression) | `it("preserves a derived never override when synthesizing a discriminator")` | No authored-shape resurrection. |

## Migration evidence

The authoritative [migration.md](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-consistent-patch-properties-native/packages/typespec-lintdiff/test/fixtures/ConsistentPatchProperties/migration.md) records focused fixtures, real-service comparisons, the latest retained corpus counts, one-sided projects, compile exclusions, and remaining uncertainty. It distinguishes native correctness from partial Swagger equivalence. This promotion did not rerun the corpus or copy harness/snapshot artifacts, and does not duplicate that corpus declaration here.

## Validation

Passed: pinned source TypeScript build with output redirected outside its checkout; ARM dependency closure/final package build; 77 focused tests; all 536 ARM tests in 65 files; rulesets build and four tests; ARM lint; repository `pnpm format` then `pnpm lint`; Chronus status; whitespace checks. Both required independent focused reviews of the final adaptation/publication diff found no significant issues.

Required ARM docs regeneration completed. README/reference entries and generated rule page title/ID were inspected and explicitly Prettier-checked, including ignored generated docs. Only expected tracked index additions resulted; documentation warnings concern existing unrelated entries.

Website-skipped `pnpm validate:pr --verbose` completed without reaching the silence cap. All 64 build tasks, engineering checks, lint, formatting, spelling and the 92-task full docs-regeneration graph completed. The broader suite had 3,441 passing tests, four skipped tests and four timeouts below.

## Validation blocker

The aggregate pre-PR command was not green. Four playground sample cases (`Widget Manager`, `Tracked Resource`, `Proxy Resource`, `Tenant Resource`) exceeded their existing 10-second limit in the concurrent broad suite. All four passed when rerun together in isolation with the same timeout and no changes (17.86 seconds total). This broad-run environmental evidence is retained, not hidden by changing unrelated tests or timeouts.

The precommit docs check flagged the expected uncommitted new row in `website/.../azure-resource-manager/reference/linter.md`; it compares working-tree differences rather than regeneration-only drift. Full regeneration completed and the only diff was that intended row, now committed. Before committing, its `origin/main...HEAD` changeset/diff steps also saw no committed source changes. Postcommit inspection confirms the nine promotion files, both change entries, and clean docs/worktree.

An earlier additional test-inclusive `tsc -p tsconfig.json --noEmit` reported 16 existing errors outside the new test: six unsupported diagnostic `target` fields, eight non-type-only imports and two shared-vitest-config rootDir/file-list errors. No errors remained in the new test; package source builds and type-aware repository lint pass. No unrelated repair was made.

Canonical main advanced to `791829b71e049fb9c38720a0de1c706d332fbbf8` during this run. The pre-PR check warned that this branch is behind, but its merge-tree check was conflict-free. This PR keeps the requested original base commit and targets current canonical `main`. CI remains authoritative for a clean broad run and the dedicated website build.

## Promotion sync policy

The pinned lintdiff source remains the semantic source of truth. Future source-semantic gaps block this promotion until the user explicitly reopens lintdiff repair; promotion authorization does not authorize source edits. Destination-only adaptations belong in this promotion worktree. The requested subsequent promotion-mode review loop is coordinated by the parent after this draft is published.

heavy


packages/typespec-azure-resource-manager/src/rules/consistent-patch-properties.ts
```diff
@@ -0,0 +1,250 @@
...
+        for (const service of services) {
+          for (const httpOperation of service.operations) {
+            if (
+              httpOperation.verb !== "patch" ||
+              isInternalTypeSpec(context.program, httpOperation.operation)
+            ) {
+              continue;
+            }
+
+            const patchBody = getObjectModel(httpOperation.parameters.body?.type);
```
timotheeguerin, line 43: have you tried just using the TypeSpec built in type system to see if the patch model can be assigned to the parent model and not have to reimplement all of this?

packages/typespec-azure-resource-manager/src/rules/consistent-patch-properties.ts
```diff
@@ -0,0 +1,250 @@
...
+
+interface PayloadProperty {
+  target: Model | ModelProperty;
+  type?: Type;
+}
+
+function getPayloadProperties(program: Program, model: Model): Map<string, PayloadProperty> {
+  const properties = new Map<string, PayloadProperty>();
+
+  for (let current: Model | undefined = model; current !== undefined; current = current.baseModel) {
```
timotheeguerin: can you use walkPropertiesInherited to simplify this, seems it might be a pattern across prs too check the others

packages/typespec-azure-resource-manager/src/rules/consistent-patch-properties.ts
timotheeguerin, line 43: Can we use `getResourceOperation(program, operation)?.resourceType` to identify the resource first, and only fall back to responses for custom operations without a resource association?

For example, `@armResourceUpdate(Widget)` already tells us the resource is `Widget`, but this currently skips validation if PATCH only returns `202` and there is no GET operation. Inferring the resource from `200`/`201` responses was necessary in Swagger, but TypeSpec already has that relationship.

msyyc: Fixed in b07f5e9cf. Resource selection now prefers `getResourceOperation(program, operation)?.resourceType` from `@typespec/rest`, with response-based fallback only when no association exists. Added coverage for 202-only PATCH without GET and for association precedence over PATCH/GET responses, and documented the difference from Swagger behavior.

packages/typespec-azure-resource-manager/src/rules/consistent-patch-properties.ts
msyyc: Updated in b07f5e9cf to use `walkPropertiesInherited`. Kept the separate base-model traversal for discriminator metadata, which the property iterator does not expose.

packages/typespec-azure-resource-manager/src/rules/consistent-patch-properties.ts
msyyc: I looked into assignability, but it has different semantics: PATCH -> resource rejects omitted required fields and differing value types, while allowing extra properties on named models. It also compares authored names rather than JSON encoded names. Reversing the direction still misses optional PATCH-only properties. This rule needs to allow partial updates but reject extra JSON properties, so I kept the layout comparison and documented the rationale in b07f5e9cf.

.chronus/changes/arm-consistent-patch-properties.md
```diff
@@ -0,0 +1,9 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add the `consistent-patch-properties` lint rule to check that PATCH body properties exist at the same nesting level in the resource model.
```
markcowl, line 7: This rule is redundant with the PatchPropertiesCorrespondToPutProperties rule - if the properties in the patch body are, recursively, a proper subset, then they must have the same nesting level and no additional proeprties

markcowl (changes requested):
This is another case where we should drop the 1:1 correspondence - we should have one rule that validates that OPATCH request bodies are a proper subset of the resource (and return the resource type).

msyyc:
waiting for https://github.com/Azure/typespec-azure/pull/5294#discussion_r4056139098

msyyc:
Closing as superseded by #5294, which consolidates UnSupportedPatchProperties, ConsistentPatchProperties, and PatchBodyParametersSchema into `no-unsafe-patch-body-properties`.

The consolidated PR documents the legacy-rule mapping, regression coverage, and intentional native semantic differences. Source consolidation is in merged #5540. Please continue review on #5294.

PR #5465, opened 2026-09-11
feat(autorest): support unified examples.yaml via examples-format option

timotheeguerin:
Part of the [Unified Examples Format epic (Azure/typespec-azure#4831)](https://github.com/Azure/typespec-azure/issues/4831).

Today `typespec-autorest` can only read the legacy per-version `x-ms-examples` JSON files. That blocks any service from adopting the unified `examples.yaml` format, because switching would drop their examples from the emitted OpenAPI.

This teaches the emitter to read `examples.yaml` and re-materialize the legacy `x-ms-examples` files for the version being emitted, so a service can migrate to the new format with **no change** to the produced OpenAPI or to downstream consumers.

New `examples-format` emitter option:

- `"auto"` (default) — use `examples.yaml` when the service has one, otherwise fall back to the legacy per-version JSON files.
- `"legacy"` — always use the legacy files.
- `"unified"` — always use `examples.yaml`.

To keep the specs-repo rollout diff minimal, the materialized files reproduce the original layout:

- File name defaults to the Azure convention `<OperationId>.json` (`<OperationId>_<slug(title)>.json` for titled variants), and honors the original name when `examples.yaml` preserves it via `legacyFilename`.
- The `x-ms-examples` key defaults to the file name without extension, matching the dominant Azure convention, and honors an explicit `title` when present.

The naming convention lives in a single shared source of truth in `@azure-tools/typespec-azure-examples` that both the migrate tool and this emitter import, so the round-trip can't drift.


timotheeguerin:
Addressed review feedback:
- **Silent data loss**: the `x-ms-examples` key and the file-emission record were both keyed by `doc.title` with no disambiguation, so two variants of one operation that derive the same title (e.g. an untitled lineage next to a lineage explicitly titled the same as the `operationId`) silently overwrote each other — dropping a file whose distinct name had already been reserved. Keys are now disambiguated the same way file names are (`uniqueExampleKey`). Added a collision regression test.

PR #5466, opened 2026-09-11
Refactor XML support

### LINKED ISSUE #5179: [typespec-go] Nested XML model serialization ignores the property element name

jhendrixMSFT:
Simplified the implementation now that it's no longer shared with autorest which had different semantics for some things (arrays). Fixed marshalling arrays so that the field's XML name takes precedence over the model type's XML name.

Fixes https://github.com/Azure/typespec-azure/issues/5179


packages/typespec-go/src/codegen/core/models.ts
```diff
@@ -1238,6 +1272,12 @@ function recursivePopulateDiscriminator(
   return text;
 }
 
+/** returns the alias type name to use when aliasing a model type */
+function getXMLNestedModelAliasTypeName(type: go.Model | go.Ptr<go.Model>): string {
+  const fieldType = go.unwrapPtr(type);
+  return naming.uncapitalize(fieldType.name);
```
heaths, line 1278: What is this used for? Seems we shouldn't be changing case on XML elements, especially when, IIRC, Storage elements are often PascalCase. Or is that not what's going on here?

heaths (approved):
Only had one question. I'm just not understanding how it's used without more context not here, but assuming it's fine.

packages/typespec-go/src/codegen/core/models.ts
jhendrixMSFT: This is used as the alias type name for the model in `fieldType`.  So if the model name is `Foo` we create an alias type with the name `foo`.  Here's a complete snippet of where this is emitted.
```go
func (m ModelWithRenamedNestedModel) MarshalXML(enc *xml.Encoder, start xml.StartElement) error {
	type author Author // <-- the alias type for Author
	type alias ModelWithRenamedNestedModel
	aux := &struct {
		*alias
		Author *author `xml:"author"`
	}{
		alias: (*alias)(&m),
	}
	if m.Author != nil {
		aux.Author = (*author)(m.Author)
	}
	return enc.EncodeElement(aux, start)
}
```

tadelesh:
This refactor still could not cover two cases. Opened an issue [here](https://github.com/Azure/typespec-azure/issues/5470).

PR #5477, opened 2026-09-14
Fix required eTag conditional headers in Python

l0lawrence:
## Summary

- add an Azure HTTP scenario for a required `If-Match` header typed as `Azure.Core.eTag`
- verify the Python emitter exposes it as direct required `if_match: str` input while optional conditional headers retain `etag` + `match_condition`
- add sync and async runtime coverage and regenerate the scenario summary
- update the core submodule to microsoft/typespec#11964

## Validation

- `pnpm -r --filter "@azure-tools/azure-http-specs..." build`
- TypeSpec Python emitter tests: 24 passed
- branded conditional-request sync/async tests: 10 passed
- core conditional-request sync/async tests: 10 passed

Depends on microsoft/typespec#11964.


l0lawrence:
Closing because the emitter fix will proceed without adding another Spector scenario in typespec-azure.

PR #5481, opened 2026-09-15
[Swagger Linter Migration] PatchPropertiesCorrespondToPutProperties (devbox-heavy)

msyyc:
## Original Swagger linter

- linter code: [PatchPropertiesCorrespondToPutProperties](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/patch-properties-correspond-to-put-properties.ts)
- linter doc: [patch-properties-correspond-to-put-properties.md](https://github.com/Azure/azure-openapi-validator/blob/main/docs/patch-properties-correspond-to-put-properties.md)

The Swagger rule pairs PUT and PATCH operations by path, requires request bodies, and recursively compares PATCH schema properties with PUT. This promotion preserves the property-presence intent without copying whole-schema metadata equality, stale emitted-schema behavior, or emitted-occurrence duplication. It reports on the PATCH operation for missing/empty bodies and on the source PATCH property for a missing corresponding JSON name.

The source is `PatchPropertiesCorrespondToPutProperties`, local rule `patch-properties-correspond-to-put-properties`, on branch `feature/lintdiff-patch-properties-correspond-to-put-properties`, pinned to reviewed commit `8d350eb5f85c259275516a69422325b6cadb63f3` in `C:\dev\worktrees\lintdiff-patch-properties-correspond-to-put-properties`.

Source rule: [patch-properties-correspond-to-put-properties.ts](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/src/rules/patch-properties-correspond-to-put-properties.ts)

The source was assumed done for this promotion and was not modified here. Cycle 1 repaired cross-service diagnostic deduplication in the source, completed validation, and received a fresh clean review before this promotion refresh.

## Destination analysis

Selected `@azure-tools/typespec-azure-resource-manager`: the rule pairs ARM lifecycle PUT/PATCH operations and is ARM-specific. Azure Core is not appropriate because this behavior is tied to ARM resource update semantics. The lintdiff-only provider-isolation guard is deliberately omitted: the official ARM package/ruleset supplies the applicability boundary, and native tests cover ordinary and nested namespaces without provider metadata. Registration is explicitly `false` to avoid enabling new diagnostics for existing services.

The rule uses native compiler, HTTP, and versioning metadata. It groups available PUT/PATCH operations by route for every resolved service/dependency version, ignores same-endpoint overload siblings, resolves request visibility, and compares JSON leaf names. Added/removed availability applies to operations, interfaces, explicit body parameters, and properties. Body diagnostics deduplicate by PATCH operation; missing properties deduplicate by source target plus JSON name across versions and services. Shared models therefore report once per missing property, while distinct targets remain separate. Historical renames, type changes, body-type changes, and route changes are not reconstructed.

|---|---|---|
| [patch-extra-property](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/patch-extra-property/main.tsp) | `it("reports a nested PATCH leaf absent from PUT without provider metadata")` | Nested missing leaf |
| [encoded-name-mismatch](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/encoded-name-mismatch/main.tsp) | `it("reports different JSON names for the same authored property")` | JSON names differ |
| [allof-wrapper-name-mismatch](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/allof-wrapper-name-mismatch/main.tsp) | `it("retains inherited-only nested wrappers as leaves")` | Inherited wrapper behavior |
| [missing-patch-body](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/missing-patch-body/main.tsp) | `it("requires a PATCH request body")` | Missing body |
| [missing-patch-body](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/missing-patch-body/main.tsp) | `it("treats a void PATCH body as missing")` | Void body |
| [empty-patch-model](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/empty-patch-model/main.tsp) | `it("rejects a PATCH body with no properties")` | Empty body model |
| [missing-put-body](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/missing-put-body/main.tsp) | `it("requires a PUT body when PATCH has a body")` | Missing PUT body |
| [synthesized-discriminator-mismatch](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/synthesized-discriminator-mismatch/main.tsp) | `it("reports a synthesized PATCH discriminator absent from PUT")` | Synthesized discriminator |
| [inherited-synthesized-discriminator](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/inherited-synthesized-discriminator/main.tsp) | `it("reports an inherited synthesized PATCH discriminator")` | Inherited discriminator |
| [compliant-subset](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/compliant-subset/main.tsp) | `it("accepts a PATCH subset of PUT")` | Compliant subset |
| [different-nesting-compliant](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/different-nesting-compliant/main.tsp) | `it("accepts matching leaf names at different nesting levels")` | Path-insensitive match |
| [encoded-name-compliant](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/encoded-name-compliant/main.tsp) | `it("accepts different authored names with the same JSON name")` | Encoded-name match |
| [schema-value-validator-discrepancy](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/schema-value-validator-discrepancy/main.tsp) | `it("ignores documentation differences on matching properties")` | Avoids Swagger metadata equality |
| [synthesized-discriminator-compliant](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/synthesized-discriminator-compliant/main.tsp) | `it("accepts matching synthesized discriminators")` | Compliant discriminator |
| [type-family-compliant](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/type-family-compliant/main.tsp) | `it("accepts matching scalar and array leaves")` | Supported leaf families |
| [type-family-compliant](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/type-family-compliant/main.tsp) | `it("preserves fallback behavior for non-ARM payload shapes")` | Source fallback behavior |
| N/A (source-repair regression) | `it.each(["shared", "distinct"] as const)("deduplicates cross-service findings by property target with %s PATCH models")` | Shared targets deduplicate; distinct targets remain separate |

Additional native cases cover namespaces without provider metadata, unmatched routes, repeated operation pairs, overload selection, service/dependency versions, inheritance/spreads, nonconcurrent declarations, jointly added properties, and removed explicit body parameters.

## Migration evidence

See [migration.md](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-patch-properties-correspond-to-put-properties/packages/typespec-lintdiff/test/fixtures/PatchPropertiesCorrespondToPutProperties/migration.md) for focused fixtures, real-service comparison, full-corpus counts, compiler failures, and remaining uncertainty.

## Validation

Passed focused native tests (31), ARM build and lint, ARM docs regeneration and explicit four-file formatting/check, rulesets build and tests (4), full ARM tests (490), Chronus status, and diff checks. Required focused prepublication review found no significant issues.

### Validation blocker

The final best-effort website-skipped `pnpm validate:pr` completed 7/9 steps. Build, lint, format, spelling, regenerated-doc cleanliness, changesets, and diff hygiene passed. Its broad test step had one unrelated timeout in `@azure-tools/typespec-azure-playground-website` while compiling the pre-existing `Tracked Resource` sample at 10,009 ms; 3,398 tests passed and 4 skipped. This PR changes no playground files or samples, and the affected ARM focused/full suites passed. The branch-up-to-date step was a warning because the app-owned promotion branch remains based on its verified `origin/main` comparison base rather than being rebased during the preserved queue workflow.

## Promotion sync policy

If review finds a source-semantic gap, this PR stops and returns evidence to the queue. Only the queue may start a bounded lintdiff source-repair cycle; promotion refreshes only after that source head is cleanly reviewed.


markcowl (changes requested):
Fundamentally, I think we need to align this check and the arm-resource-patch check: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-resource-patch.ts

If we ensure that the PATCH model is a subset of the corresponding resource, we don't need to compare it to PUT, because we already have rules that require a PUT and require PUT to contain the resource.

This is a place where exact equivalence should give way to functional equivalence.

However, I think the existing rule arm-resource-patch is only checking the properties bag when it should also be checking the resource envelope.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.ts
```diff
@@ -0,0 +1,387 @@
...
+export const patchPropertiesCorrespondToPutRule = createRule({
+  name: "patch-properties-correspond-to-put",
+  docs: fileRef.fromPackageRoot("src/rules/patch-properties-correspond-to-put.md"),
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/patch-properties-correspond-to-put",
+  description: "ARM PATCH body properties must correspond to properties in the PUT request body.",
+  severity: "warning",
+  messages: {
+    missingPatchBody: "The PATCH operation must have a request body.",
+    emptyPatchBody: "The PATCH request body must contain at least one property.",
+    missingPutBody: "A PATCH request body requires the corresponding PUT operation to have a body.",
```
markcowl, line 45: We have a separate rule that checks for PUT request bodies, there is no reson to duplicate that check here.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.ts
```diff
@@ -0,0 +1,387 @@
...
+
+export const patchPropertiesCorrespondToPutRule = createRule({
+  name: "patch-properties-correspond-to-put",
+  docs: fileRef.fromPackageRoot("src/rules/patch-properties-correspond-to-put.md"),
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/patch-properties-correspond-to-put",
+  description: "ARM PATCH body properties must correspond to properties in the PUT request body.",
+  severity: "warning",
+  messages: {
+    missingPatchBody: "The PATCH operation must have a request body.",
+    emptyPatchBody: "The PATCH request body must contain at least one property.",
```
markcowl, line 44: The rules here are a bit stronger - for example, if there *is* a tags property (i.e. TrackedResource) the PATCH request must contain a tags property.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.ts
```diff
@@ -0,0 +1,387 @@
...
+    missingPutBody: "A PATCH request body requires the corresponding PUT operation to have a body.",
+    missingProperty: paramMessage`The property '${"propertyName"}' in the PATCH body does not correspond to a property in the PUT body.`,
+  },
+  create(context) {
+    return {
```
markcowl, line 49: I think this can be much simpler - we should be able to handle this on a per (Operation) basis, looking for PATCH operations, validating that they are over a resource, and that the request body is a non-empty subset of the resource properties.   I don't think we need to pair with PUT correlation in order to have the same scenario coverage

markcowl (commented):
I think this rule is covering too much ground already checked by other rules - we certainly should make use of transitive properties (i.e. if the PUT is required to have the resource as a request body amd the PATCH is required to be a proper non-empty subset of settable resource properties, and every resource with a PATCH must have a PUT, then we don't need to do this extra check of PUT request bodies.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.ts
msyyc: I checked these proposals together. In the standard-template cases I tested, the checks compose as proposed: #5294 owns recursive resource-layout validation; #5423 rejects a separate named PUT request model that differs from the resource responses; existing ARM rules cover missing PATCH bodies and required tags.

There is still an effective-input gap even with the **same resource model**. A standard PUT template with `@parameterVisibility(Lifecycle.Create)` excludes an Update-only resource property, while a custom PATCH can expose it. Both proposed rules and the enabled ARM/Core checks accept that case.

#5481 detects that mismatch only after an unrelated Read/Query-only property is added. The native PUT/PATCH input properties remain unchanged. So its current emission-oriented visibility logic also needs repair; keeping it unchanged would not give a reliable native check.

I recommend retaining only the **effective PATCH-to-PUT input correspondence** responsibility, without duplicating resource-layout or body/tags checks. If ARM policy instead excludes these visibility overrides, we can retire this PR as an intentional scope restriction, rather than claiming identical coverage. Scalar and multi-model PUT bodies also need an explicit boundary because #5423 skips their model comparison.

I'm leaving this PR and the design discussion open; #5294 and default enablement are unchanged.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.md
```diff
@@ -0,0 +1,148 @@
...
+effective PATCH leaf property has a matching JSON name in the effective PUT input. It does not
+require property descriptions, constraints, or types to be identical.
+Direct nested payload properties are compared by leaf name rather than their full nesting path;
+inherited-only nested wrappers remain leaves. Explicit JSON names, inherited properties, request
+visibility, and authored discriminator properties participate in the comparison.
+
+## Effective input and rule ownership
+
+Each operation's resolved request visibility determines its input. The rule does not reuse
+AutoRest canonical schemas, fall back to Read visibility, or synthesize discriminator properties.
```
msyyc: The AutoRest canonical-schema comparison is migration detail and distracts from the native authoring rule. Please describe the behavior directly here: each operation's resolved request visibility determines its input, and only authored discriminator properties participate. Keep the following visibility and `@body`/`@bodyRoot` guidance, and move any comparison with AutoRest schema reuse to `LintDiff Equivalent` or the PR description.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.md
```diff
@@ -0,0 +1,148 @@
...
+
+The rule does not diagnose a missing PUT body. PUT-body validity belongs to the separate
+`put-resource-schema-consistency` rule proposed in
+[#5423](https://github.com/Azure/typespec-azure/pull/5423); it must be available and enabled to provide
+that coverage. This rule does not require a PUT operation to exist either. An absent or `void`
+PUT body skips correspondence, including when a body parameter is unavailable in a service version.
+
+For paired operations, missing or `void` PATCH bodies and object bodies with no effective input
+properties are still diagnosed. These safeguards remain until equivalent coverage is provided
+by the PATCH-body rules; declared properties alone do not establish nonempty effective input.
```
msyyc: This section mixes current rule behavior with development coordination: "proposed in #5423" and "These safeguards remain until equivalent coverage is provided" describe the implementation plan, not how authors should use the rule. Please retain the current missing/empty-body behavior and the distinction between related rules, but remove the temporary handoff language and use stable rule documentation links where available. Apply the same cleanup to the #5294 reference below. Keep PR dependencies and future redistribution of checks in the PR description or migration evidence.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.ts
msyyc: Agreed that supported tags must be exposed in the PATCH schema. The existing `arm-resource-patch` and `patch-envelope` rules already own that requirement, so 631f5dbd1 preserves those checks rather than adding another tags diagnostic here. The documentation now distinguishes required envelope-property presence from PATCH-to-PUT input correspondence, and clarifies that exposing `tags` does not mean requiring it in every PATCH request. The neighboring PATCH/envelope tests passed; default enablement is unchanged.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.md
msyyc: Addressed in cc7f8e3c9. Removed the temporary PR references and future-handoff language from the authoring guidance. The section describes current missing/empty-body behavior and the responsibilities of related rules, with stable relative links to the existing `patch-envelope` and `arm-resource-patch` documentation. The separate PUT and consolidated PATCH rules are named without development-coordination language.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.ts
msyyc: Addressed in 631f5dbd1. Removed the `missingPutBody` diagnostic and its deduplication state. Missing, void, or version-unavailable PUT bodies now skip correspondence without a duplicate warning; PUT-body validity remains the responsibility of `put-resource-schema-consistency` when enabled. Regression coverage includes absent/void PUT bodies and removed body parameters. Missing/empty effective PATCH-body safeguards remain independent of PUT-body validity.

packages/typespec-azure-resource-manager/src/rules/patch-properties-correspond-to-put.md
msyyc: Addressed in cc7f8e3c9. The effective-input section now states the native behavior directly: resolved request visibility determines input, and only authored discriminator properties participate. The visibility and `@body`/`@bodyRoot` guidance remains. Emitted-schema sharing and synthesized-discriminator comparisons are confined to `LintDiff Equivalent`.

PR #5502, opened 2026-09-17
[Swagger Linter Migration] TagsAreNotAllowedForProxyResources

msyyc:
## Original Swagger linter

- linter code: [TagsAreNotAllowedForProxyResources](https://github.com/Azure/azure-openapi-validator/blob/1198225afecbb818c3050d4d2a91da92e14e56ce/packages/rulesets/src/spectral/functions/tags-are-not-allowed-for-proxy-resources.ts)
- linter doc: [tags-are-not-allowed-for-proxy-resources.md](https://github.com/Azure/azure-openapi-validator/blob/1198225afecbb818c3050d4d2a91da92e14e56ce/docs/tags-are-not-allowed-for-proxy-resources.md)

The staging-only Swagger function scans emitted definitions. It reports a JSON `tags` property when its containing object lacks `location`, and also searches a nested `properties` object. This catches proxy-resource envelope and property-bag tags, but the absence-of-`location` heuristic also produces known false positives for generated tracked-resource update envelopes and arbitrary non-resource models. Those emitted-shape defects are not copied into the native rule.

- Validator rule: `TagsAreNotAllowedForProxyResources`
- Lintdiff rule: `tags-are-not-allowed-for-proxy-resources`
- Canonical validator slug: `tags-are-not-allowed-for-proxy-resources`
- Source branch: `msyyc:feature/lintdiff-tags-are-not-allowed-for-proxy-resources`
- Pinned reviewed source commit: `396f80453ba8dc953e1a47fc2f969e8b56e7b5c2`
- Source worktree: `C:\dev\worktrees\lintdiff-tags-are-not-allowed-for-proxy-resources`
- Source implementation: [tags-are-not-allowed-for-proxy-resources.ts](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/src/rules/tags-are-not-allowed-for-proxy-resources.ts)

The source rule was assumed done for this promotion run, had a clean worktree at the reviewed commit, and was not modified during promotion.

## Destination analysis

The agent selected `@azure-tools/typespec-azure-resource-manager` and the official name `no-tags-on-proxy-resources`. The rule is ARM-only, classifies registered ARM resources through `getArmResources`, and distinguishes proxy resources from tracked resources. Azure Core is not a viable destination because it cannot depend on ARM resource metadata. No equivalent official rule exists.

The rule enumerates registered ARM resources and evaluates only `kind === "Proxy"`. It resolves JSON encoded names using the compiler's `resolveEncodedName` API, checks both the resource envelope and resource-specific properties hierarchy, and reports on the authored property whose JSON name is `tags`. This preserves the reviewed source semantics, including inherited properties and encoded-name overrides, while avoiding Swagger's emitted-definition heuristic and duplicate occurrences. The official ARM ruleset registers the new rule as disabled by default.

| --- | --- | --- |
| [proxy-without-tags](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/proxy-without-tags/main.tsp) | `it("is valid for a proxy resource without tags")` | Compliant proxy resource. |
| [tracked-with-tags](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/tracked-with-tags/main.tsp) | `it("is valid for a tracked resource with tags")` | Tags remain supported on tracked resources. |
| [proxy-tags-encoded-away](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/proxy-tags-encoded-away/main.tsp) | `it("is valid when a property named tags is encoded to another JSON name")` | Authored name is not treated as the emitted JSON name. |
| [proxy-with-envelope-tags](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/proxy-with-envelope-tags/main.tsp) | `it("emits a warning for tags on the proxy resource envelope")` | Envelope violation. |
| [proxy-with-tags](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/proxy-with-tags/main.tsp) | `it("emits a warning for inherited tags in the properties bag")` | Inherited property-bag violation. |
| [proxy-with-encoded-tags](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/proxy-with-encoded-tags/main.tsp) | `it("emits a warning when another property is encoded as tags")` | JSON encoded-name violation. |

## Migration evidence

See the source [migration.md](https://github.com/msyyc/typespec-azure/blob/feature/lintdiff-tags-are-not-allowed-for-proxy-resources/packages/typespec-lintdiff/test/fixtures/TagsAreNotAllowedForProxyResources/migration.md) for focused fixture evidence, the 462/468-project corpus, staging comparison, complete project sets, version-projection analysis, compile failures, and intentional Swagger-only differences.

## Promotion sync policy

The lintdiff source commit above is immutable provenance for this PR. If promotion review finds a source-semantic defect, this PR must stop while the owning queue performs a bounded return-to-development cycle. Promotion may refresh only from a newly reviewed source SHA; source semantics are never repaired only in the official copy.

## Validation blocker

The optional broad `pnpm validate:pr` check was stopped after producing no output for more than five minutes following successful branch and full-repository build checks. The last completed step was `Build succeeds` at 338.3 seconds. Required targeted validation passed: focused rule tests, ARM package build/lint/docs, generated-doc formatting, rulesets build/test, full ARM package tests, Chronus status, and diff hygiene.

heavy


packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
```diff
@@ -0,0 +1,78 @@
...
+    };
+  },
+});
+
+function getResourcePropertiesModel(program: Program, resourceModel: Model): Model | undefined {
+  const propertiesProperty = getPropertyInHierarchy(program, resourceModel, "properties");
+  return propertiesProperty?.type.kind === "Model" ? propertiesProperty.type : undefined;
+}
+
+function getPropertyInHierarchy(
```
timotheeguerin, line 100: is there really use case where encoded name was used we need to cover?

.chronus/changes/promote-no-tags-on-proxy-resources-arm.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add the `no-tags-on-proxy-resources` lint rule for detecting unsupported tags on ARM proxy resources.
```
timotheeguerin: I think singular name match more the existing convention, realize now a few rule rules were merged with pluaral in this migration

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
```diff
@@ -0,0 +1,78 @@
...
+          );
+          const propertiesTags =
+            propertiesModel && getPropertyInHierarchy(context.program, propertiesModel, "tags");
+
+          for (const tagsProperty of [resourceTags, propertiesTags]) {
+            if (tagsProperty) {
+              context.reportDiagnostic({
+                target: tagsProperty,
+                format: {
+                  resourceName: armResource.name,
```
timotheeguerin, line 54: i think a simple codefix to remove the property here could be added

.chronus/changes/promote-no-tags-on-proxy-resources-arm.md
msyyc: Renamed the rule to `no-tags-on-proxy-resource` in 6c997ce52, including the implementation/test filenames, exported symbol, ruleset registration, documentation links, and change descriptions. Kept `tags` plural as the literal property name.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
msyyc: I don't have a concrete production example for this; the encoded-name cases in the migration evidence are synthetic fixtures. I kept the handling because this rule targets the JSON field name: `labels` encoded as `tags` should be diagnosed, while `tags` encoded as `labels` should not. `resolveEncodedName` matches the emitter's naming behavior and avoids those false negatives/positives. Added an explicit note to the rule documentation in babcbede4; detection semantics are unchanged.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
msyyc: Added a `Remove tags property` quick fix in babcbede4 for properties declared directly on a non-template proxy resource envelope, including properties encoded as JSON `tags`. It removes the declaration, attached decorators/docs, and its separator. Inherited/copied declarations, reusable resource templates, and properties-bag declarations remain diagnostic-only to avoid deleting shared declarations. Added coverage for these boundaries and for preserving adjacent properties.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
```diff
@@ -0,0 +1,114 @@
...
+              });
+            }
+          }
+        }
+      },
+    };
+  },
+});
+
+function createRemoveTagsCodeFix(resource: Model, property: ModelProperty): CodeFix | undefined {
```
timotheeguerin, line 65: create this as a reusable `createDeletePropertyCodeFix` like we have for other compiler codefix(keep it here for now but then we can move it upstream when we need it)

I think some of the other rules could reuse it after too(some already merged too)

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
```diff
@@ -0,0 +1,114 @@
...
+    return undefined;
+  }
+
+  return defineCodeFix({
+    id: "remove-proxy-resource-tags",
+    label: "Remove tags property",
+    fix(context) {
+      const location = getSourceLocation(node);
+      // Include the separator even when whitespace or comments follow the property type.
+      const separator = /^(?:\s|\/\*[\s\S]*?\*\/|\/\/[^\r\n]*)*[;,]/.exec(
```
timotheeguerin, line 84: This can delete unrelated code when the last property has no separator:

```typespec
model Widget is ProxyResource<{}> {
  @key @segment("widgets") name: string;
  tags?: Record<string> /* first */
}
model Other { value: string /* second */; }
```

The regex backtracks past the first `*/` to the second one, removing `Other` too. We should scan comments without backtracking and add a regression test that preserves the other model.

timotheeguerin, line 84: This also treats a semicolon or comma inside a trailing line comment as the separator:

```typespec
tags?: Record<string> // keep; comment
}
```

The fix deletes through `// keep;`, leaving `comment` as code and causing a syntax error. We should add a regression test for this too.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.md
```diff
@@ -0,0 +1,49 @@
...
+shared declarations.
+
+## ❌ Incorrect
+
+```tsp
+@armProviderNamespace
+namespace MyService;
+
+model Widget is ProxyResource<WidgetProperties> {
+  @key @segment("widgets") name: string;
```
markcowl, line 22: Always use ...ResourceNameParameter in examples.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.md
```diff
@@ -0,0 +1,49 @@
+Proxy resources do not support ARM tags. Declaring a JSON property named `tags` on either the
+resource envelope or its resource-specific properties can give callers the false impression that the
+resource participates in Azure Resource Manager tag operations. Use a tracked resource when tags are
+required.
+
+The rule checks JSON property names, including names assigned with `@encodedName`.
+
```
markcowl, line 7: This needs Impact and Suppression sections

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
```diff
@@ -0,0 +1,114 @@
...
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-tags-on-proxy-resource",
+  messages: {
+    default: paramMessage`Proxy resource '${"resourceName"}' must not declare \`tags\` on its resource envelope or in its properties bag. Use a tracked resource if tags are required.`,
+  },
+  create(context) {
+    return {
+      root: () => {
+        for (const armResource of getArmResources(context.program)) {
+          if (armResource.kind !== "Proxy") {
```
markcowl, line 31: I think the better gate here is to skip when armResource.kind === "Tracked"

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
```diff
@@ -0,0 +1,114 @@
...
+      );
+      return context.replaceText(
+        { ...location, end: location.end + (separator?.[0].length ?? 0) },
+        "",
+      );
+    },
+  });
+}
+
+function getResourcePropertiesModel(program: Program, resourceModel: Model): Model | undefined {
```
markcowl, line 95: In this rule, you should be looking in the envelope, not in the 'properties' property

packages/typespec-azure-resource-manager/test/rules/no-tags-on-proxy-resource.test.ts
```diff
@@ -0,0 +1,268 @@
...
+        `
+        @armProviderNamespace namespace MyService;
+
+        model Widget is ProxyResource<WidgetProperties> {
+          @key @segment("widgets") name: string;
+        }
+
+        model WidgetProperties {
+          @encodedName("application/json", "labels")
+          tags?: Record<string>;
```
markcowl, line 188: This is not where we are looking for tags.  tags should be in the resource envwelope, not the resource properties.

markcowl (changes requested):
The biggest thing is that 'tags' is an envelope property, not an rp-specific properties property. Otherwise, a few nits.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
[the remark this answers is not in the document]
msyyc: Fixed in 74ce28973: removed the backtracking comment regex. The reusable deletion helper scans forward and ends each block comment at its first closing */. Added a regression case with 10,000 repeated block-comment fragments.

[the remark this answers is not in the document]
msyyc: Fixed in 74ce28973: replaced the repeated regex alternatives with a forward-only scan. Line comments are consumed through their line ending without reconsidering earlier characters. Added a long-line-comment regression case with 10,000 repeated // fragments.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
msyyc: Extracted createDeletePropertyCodeFix into src/codefixes/delete-property.ts in 74ce28973, with independent helper tests. It takes a ModelPropertyNode and handles only deletion mechanics; the rule keeps the ARM-specific eligibility checks. It remains local to this package for now so other rules can reuse it before a future upstream move.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
msyyc: Fixed in 74ce28973. The scanner uses the first closing */ and stops at the next non-trivia token, including the model closing brace. If there is no separator, only the property node is deleted and trailing comments are preserved. Added your following-model regression case at both helper and rule levels.

msyyc: Fixed in 74ce28973. Line comments are consumed completely through LF/CRLF, so a semicolon or comma inside the comment cannot become the separator. Added both trailing // keep; comment and // keep, comment regressions, plus comments followed by an actual separator on the next line.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.md
msyyc: Updated all documentation examples to use ...ResourceNameParameter<Widget> in 74ce28973, including the new suppression example.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.md
msyyc: Added Impact and Suppression sections in 74ce28973. Impact describes the ARM contract and misleading tagging support; suppression is limited to an existing contract with an ARM-approved exception and includes a #suppress directive example. Also added the rule to the ARM rule matrix.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
msyyc: Changed the gate to skip only armResource.kind === "Tracked" in 74ce28973. Added envelope-tag cases for extension, virtual, generic, and custom resources alongside proxy resources and the valid tracked-resource case. Diagnostics and docs now say non-tracked rather than calling all diagnosed resources proxy resources.

packages/typespec-azure-resource-manager/src/rules/no-tags-on-proxy-resource.ts
msyyc: Agreed; removed the resource-properties-model lookup and properties-bag scan in 74ce28973. The rule now checks only envelope tags. Direct, inherited, and encoded tags in the RP-specific properties bag are valid controls. Updated the diagnostics, docs, changeset, and PR migration rationale to make this correction explicit rather than claiming unchanged source semantics.

packages/typespec-azure-resource-manager/test/rules/no-tags-on-proxy-resource.test.ts
msyyc: Moved the encoded-name examples onto the resource envelope in 74ce28973. The tests now check that envelope labels encoded as tags are diagnosed and envelope tags encoded as labels are allowed. Added valid cases for direct, inherited, and encoded properties-bag tags, which the rule no longer checks.

msyyc:
@markcowl Addressed this [review](https://github.com/Azure/typespec-azure/pull/5502#pullrequestreview-5253016418) in 74ce28973. The rule now checks only resource-envelope tags and skips tracked resources; it no longer checks the RP-specific properties bag. The encoded-name cases now exercise the envelope, and direct/inherited/encoded properties-bag tags are valid controls. Also updated the documentation examples to use ResourceNameParameter and added Impact and Suppression sections. Replied to each inline comment separately and left the review threads open for your re-review.

Issue #5274, opened 2026-08-20
Migrate ARM swagger linter rule to typespec linter rule


msyyc:
| id | swagger linter | tsp lib | tsp linter PR | PR created time | PR merged time | approved by | new comment after latest commit | Status |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | [AdditionalPropertiesAndProperties](https://github.com/Azure/azure-openapi-validator/blob/main/docs/additional-properties-and-properties.md) | @azure-tools/typespec-azure-core/bad-record-type | - | - | - | - | - | Done |
| 2 | [AllProxyResourcesShouldHaveDelete](https://github.com/Azure/azure-openapi-validator/blob/main/docs/all-proxy-resources-should-have-delete.md) | @azure-tools/typespec-azure-resource-manager/no-resource-delete-operation | - | - | - | - | - | Done |
| 3 | [AllTrackedResourcesMustHaveDelete](https://github.com/Azure/azure-openapi-validator/blob/main/docs/all-tracked-resources-must-have-delete.md) | @azure-tools/typespec-azure-resource-manager/no-resource-delete-operation | - | - | - | - | - | Done |
| 4 | [ApiHost](https://github.com/Azure/azure-openapi-validator/blob/main/docs/api-host.md) | TypeSpec structural | - | - | - | - | - | Done |
| 5 | [ApiVersionParameterRequired](https://github.com/Azure/azure-openapi-validator/blob/main/docs/api-version-parameter-required.md) | @azure-tools/typespec-azure-core/operation-missing-api-version, @azure-tools/typespec-azure-resource-manager/arm-resource-operation | - | - | - | - | - | Done |
| 6 | [APIVersionPattern](https://github.com/Azure/azure-openapi-validator/blob/main/docs/api-version-pattern.md) | @azure-tools/typespec-azure-resource-manager/arm-resource-invalid-version-format | - | - | - | - | - | Done |
| 7 | [ArmResourcePropertiesBag](https://github.com/Azure/azure-openapi-validator/blob/main/docs/arm-resource-properties-bag.md) | @azure-tools/typespec-azure-resource-manager/arm-resource-duplicate-property | - | - | - | - | - | Done |
| 8 | [ArrayMustHaveType](https://github.com/Azure/azure-openapi-validator/blob/main/docs/array-must-have-type.md) | TypeSpec structural | - | - | - | - | - | Done |
[...]

PR #5503, opened 2026-09-17
[Swagger Linter Migration] XmsResourceInPutResponse 

msyyc:
## Original Swagger linter

- linter code: [XmsResourceInPutResponse](https://github.com/Azure/azure-openapi-validator/blob/6243cb01c16c7535cd3b8df6f45fbeb3c095ed7f/packages/rulesets/src/spectral/functions/with-xms-resource.ts)
- linter doc: [xms-resource-in-put-response.md](https://github.com/Azure/azure-openapi-validator/blob/6243cb01c16c7535cd3b8df6f45fbeb3c095ed7f/docs/xms-resource-in-put-response.md)

The original `RPC-Put-V1-12` rule:

- Selects PUT operations under `paths` and `x-ms-paths`.
- Selects the first 200 response schema, otherwise 201; skips operations without a selected schema.
- Accepts a truthy `x-ms-azure-resource` value, including metadata found recursively in inline `allOf` objects.
- Reports the PUT operation when the selected schema lacks that metadata.

The helper does not itself resolve `$ref`; surrounding resolver behavior affects external common-type ancestry. The native rule does not simulate that resolution or infer identity from generated reference strings.

- Validator ID: `XmsResourceInPutResponse`; canonical slug: `xms-resource-in-put-response`.
- Local rule: `tsp-lintdiff-local-linter/xms-resource-in-put-response`.
- [Original TypeSpec source rule](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/src/rules/xms-resource-in-put-response.ts).
- Source repository/branch: `Azure/typespec-azure:feature/lintdiff-xms-resource-in-put-response`, development PR #5407.
- Pinned reviewed source: `ffbb3d7037b93fbbe9a8dd0ea5f2c73df3cb56d9`.
- Source worktree: `C:\dev\worktrees\lintdiff-xms-resource-in-put-response`, clean with no uncommitted changes.

The source rule was assumed done for this promotion run and was not modified during promotion. Development review terminated with no new comments on the exact pinned head. The reviewed message correction requiring **registered ARM resource identity** is retained.

## Destination analysis

The agent selected `@azure-tools/typespec-azure-resource-manager`, with official rule name **`use-resource-model-for-put`**. Catalog applicability is ARM; fixtures use ARM resource operations, the validation report infers the resource-manager ruleset, and the implementation requires ARM `getArmResource` identity. Azure Core is not an appropriate alternative because it must not depend on ARM.

Existing `arm-resource-operation-response` checks consistency among registered resource models. It does not reject the decorated manual operation's unregistered response. Native tests demonstrate that the existing response-consistency, interface, and operation-decorator rules are silent for that case.

The new rule is registered but **disabled** in the resource-manager ruleset. There is no new data-plane registration and no immediate enablement.

For a PUT operation, select the first model body under 200, otherwise 201. If its direct or inherited properties include both `name` and `type`, require registered ARM identity through `getArmResource`. Report on the response model, once per visited operation. Separate operations sharing a response retain separate warnings; a source-backed regression also preserves the two visited nodes for a template instantiation and its concrete alias.

The implementation uses compiler, HTTP, and relative ARM semantic APIs only. It does not import an emitter, OpenAPI metadata helpers, TCGC, private state, or unsafe mutation APIs. It checks the supplied semantic program rather than reconstructing historical versions.

Destination adaptations:

- Rename the public rule to `use-resource-model-for-put` and add native documentation metadata.
- Use relative ARM imports and the existing `getProperties` hierarchy helper.
- Remove the provider-namespace guard that isolated ARM cases in lintdiff's mixed runner. The official rule is selected for ARM specifications; ordinary, nested, and global project operations do not need a provider decorator merely to be checked.
- Separately exclude library declarations through structured compiler location metadata and omit uninstantiated template declarations using compiler/existing package helpers.

The narrower native contract is unchanged: arbitrary non-resource-shaped schemas, raw OpenAPI overrides, and Legacy escape hatches are not reproduced for Swagger parity. Exact equivalence is partial. OpenAPI and Legacy authoring motivating special exemptions is already rejected by Azure prerequisite rules.

Snapshots were not copied. Native tests compile TypeSpec without an emitter and assert diagnostic messages, response-model targets, and occurrence counts.

| --- | --- | --- |
| [put-interface-missing-azure-resource](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/put-interface-missing-azure-resource/main.tsp) | `it("reports a decorated manual response not checked by existing operation rules")` | Supported manual operation; existing official checks remain silent. |
| [put-missing-azure-resource](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/put-missing-azure-resource/main.tsp) | `it("reports a resource-shaped response from a namespace-level PUT")` | Direct rule test; unrelated prerequisite suppression omitted. |
| [put-nested-namespace-missing-azure-resource](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/put-nested-namespace-missing-azure-resource/main.tsp) | `it("reports a resource-shaped response in a nested provider namespace")` | Nested traversal and response target. |
| [put-arm-resource](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/put-arm-resource/main.tsp) | `it("accepts a registered tracked resource")` | Standard tracked resource remains compliant. |
| [put-with-azure-resource](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/put-with-azure-resource/main.tsp) | `it("accepts a registered proxy resource")` | Standard proxy resource remains compliant; no extension override. |
| [patch-ignored](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/patch-ignored/main.tsp) | `it("ignores PATCH responses")` | Non-PUT selector boundary. |
| [global-put-ignored](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/global-put-ignored/main.tsp) | `it("checks global PUT responses when the ARM rule is selected")` | Deliberate adaptation: lintdiff-only isolation is not an exemption under an explicitly selected ARM rule. |
| N/A (promotion applicability regression) | `it("checks ordinary namespaces without a provider decorator")` | No unnecessary provider marker. |
| N/A (promotion applicability regression) | `it("checks nested namespaces without a provider decorator")` | Same applicability in nested namespaces. |
| N/A (source native regression) | `it("checks inherited resource envelope properties")` | Hierarchy helper preserves inherited-property detection. |
| N/A (source native regression) | `it("checks a 201 response when no 200 model exists")` | 201 fallback. |
| N/A (source native regression) | `it("prefers a 200 model over a 201 model")` | Status selection remains unchanged. |
| N/A (source native regression) | `it("ignores an unrelated response status")` | 202 does not enter the selected-response check. |
| N/A (native body boundary regression) | `it("ignores responses without a body")` | Empty response. |
| N/A (source native regression) | `it("ignores non-resource-shaped model bodies")` | Narrow resource-model contract. |
| N/A (source native regression) | `it("ignores scalar bodies")` | Non-model body exclusion. |
| N/A (diagnostic count regression) | `it("reports once per PUT when operations share a response model")` | Two warnings target the shared response model. |
| N/A (promotion template filtering) | `it("ignores uninstantiated operation and interface templates")` | Non-endpoint definitions. |
| N/A (source-backed cardinality regression) | `it("preserves source diagnostic counts for a template instantiation and alias")` | Two visited operation nodes; no silent deduplication change. |
| N/A (documentation regression) | `it("verifies the documented incorrect and correct examples")` | Both complete documentation snippets compile with the expected native lint outcomes. |

## Migration evidence

See the source [migration.md](https://github.com/Azure/typespec-azure/blob/feature/lintdiff-xms-resource-in-put-response/packages/typespec-lintdiff/test/fixtures/XmsResourceInPutResponse/migration.md) for focused fixtures, full-corpus analysis, one-sided project explanations, selected-version attribution, failure exclusions, and equivalence limits.

The final development-review message-only correction reran the full corpus; its post-fix shard counts were not captured correctly before cleanup. The earlier checked-in counts remain separately identified semantic evidence, not newly recaptured review-fix statistics. Production selection and identity semantics did not change in that correction.

## Validation

- Focused rule tests: **20 passed**, including both complete documentation examples.
- Full ARM package: **484 tests passed in 66 files**; package build and lint passed.
- Azure rulesets: build, lint, and **5 tests passed**.
- ARM `regen-docs` completed; explicit README, authored rule, generated linter reference, and ignored generated rule page passed formatting checks.
- Separate ARM feature and disabled-rulesets internal Chronus entries passed status validation with LF line endings.
- Independent local reviews found no significant issues in native/authored changes or the subsequent generated-documentation and exact fixture-mapping scope.

## Validation blocker — broad repository check

The best-effort `TYPESPEC_SKIP_WEBSITE_BUILD=true pnpm validate:pr` completed with exit 1.
Full-repository build, lint, formatting, and spelling passed. Its test run had **3,459 passed,
6 timed out, and 4 skipped**. Timeouts occurred in existing AutoRest `options.test.ts` and
`x-ms-examples.test.ts`, Azure Core `lro-status.test.ts`, and three playground sample cases
(`Widget Manager`, `Tracked Resource`, `Proxy Resource`). The new rule's 20 tests passed in
that broader run too. No unrelated tests, timeouts, or source files were changed.

The precommit docs-clean check also flagged the intended new linter-reference entry relative
to HEAD. All nine promotion-file hashes were identical before and after the broad run, and
scoped generated-doc formatting passed again; this was not regeneration drift. The
prepared base `c52f850d4c858f28b2d7a2f5c7fcf8badef080ae` was preserved rather than rebased
onto newer `main`. All required narrow promotion checks listed above passed.

## Promotion sync policy

The source migration remains authoritative. A verified source-semantic gap blocks promotion and returns to the owning queue for an authorized bounded source-repair cycle. This PR may be refreshed only after that source repair passes a new clean review. Promotion does not edit source semantics, merge either PR, or enable the rule automatically.

heavy


.chronus/changes/arm-use-resource-model-for-put.md
timotheeguerin, line 1: what does devbox heavy mean in the title, also update with the current rule name

packages/typespec-azure-resource-manager/src/rules/use-resource-model-for-put.ts
```diff
@@ -0,0 +1,67 @@
...
+import { getProperties, isTemplatedInterfaceOperation } from "./utils.js";
+
+export const useResourceModelForPutRule = createRule({
+  name: "use-resource-model-for-put",
+  docs: fileRef.fromPackageRoot("src/rules/use-resource-model-for-put.md"),
+  description: "Use registered ARM resource models for PUT resource responses.",
+  severity: "warning",
+  url: "https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/use-resource-model-for-put",
+  messages: {
+    default: "PUT 200/201 resource response models must be registered as ARM resources.",
```
timotheeguerin:
```suggestion
    default: "Create or update response must be an ARM resource.",
```
what does registered mean here? is my message more what is the rule about(I think the doc and description could be updatged a little accordingly too)

Feels to me this all "looks like a resource" is just a way to try to detect it no?

.chronus/changes/arm-use-resource-model-for-put.md
msyyc: I have 2 devbox to do the migration so I add a suffix to make it convenient for me to know quickly where this PR is from.

I use the swagger rule name to make it easy for me to record the status of overall migration: https://github.com/Azure/typespec-azure/issues/5274

packages/typespec-azure-resource-manager/src/rules/use-resource-model-for-put.ts
msyyc: Agreed. "Registered" referred to the ARM library's internal resource state, populated by the standard resource templates, not a separate step for spec authors. Updated the diagnostic in f51dd2278 to "The success response for a PUT create or update operation must be an ARM resource model." The description, docs, changelog, and reference pages now use author-facing wording too.

The `name`/`type` check is indeed a heuristic for selecting responses to inspect, not what makes the model an ARM resource. The docs now state that explicitly and retain the current 200/201 selection and other scope limits. This change does not alter enforcement behavior.

.chronus/changes/arm-use-resource-model-for-put.md
timotheeguerin: can you include the name of the new rule too(like `XmsResourceInPutResponse  -> use-resource-model-for-put``, from a git history POV this will make it easier to understand the commit if needed in the future

.chronus/changes/arm-use-resource-model-for-put.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@azure-tools/typespec-azure-resource-manager"
+---
+
+Add the `use-resource-model-for-put` lint rule to require ARM resource models for PUT create or update success responses.
```
timotheeguerin, line 7: is that rule also not really a subset of #5423
`PutRequestResponseSchemeArm + ConsistentResponseSchemaForPut`

From what I understand the final expectation is that for put request the request and all responses are the same arm resource.
If you think this should stay separate I guess maybe that is still ok but I do find the naming and documentation not plannign for the final result which will require not just hte response but the request to also have that be a resource and the same one

markcowl (changes requested):
I think this duplicates checks alread in: arm-resource-operation-response: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-resource-operation-response.ts

.chronus/changes/arm-use-resource-model-for-put.md
msyyc: make sense and done for all PRs including merged ones.

msyyc:
Superseded by #5423, published in 8d87839e902c5c37a6bd2072633d8c52e44df26e. The combined put-resource-schema-consistency rule now requires native ARM registration through getArmResource, alongside a PUT request body and shared model identity across identifiable request/200/201 bodies. It checks both success statuses, uses no name/type property heuristic, and skips imported library declarations while checking project aliases. Shared unregistered models now diagnose. Unsupported or ambiguous bodies retain the explicitly documented native scope; this is not literal Swagger parity. The rule remains disabled pending rollout. Existing status-code and canonical lifecycle-response rules remain intact. Closing this separate proposal to avoid shipping overlapping rules; its branch is preserved.

PR #5508, opened 2026-09-17
Fix TCGC model-valued decorator discovery

### LINKED ISSUE #5490: Model-valued decorator arguments create duplicate SDK models and numeric type names
## Summary

Upgrading `@azure-tools/typespec-client-generator-core` from `0.71.2` to `0.72.0` causes model-valued decorator arguments to participate in SDK type discovery. In Azure Resource Manager specifications, this can create two SDK model objects for the same logical TypeSpec model.

The duplicate objects then reach `@typespec/http-client-csharp`, whose generic naming-conflict fallback renames the second model by appending a number—for example, `BillingReservation` becomes `BillingReservation1`. This produces widespread, unexpected SDK changes and can break existing custom code.

This was found while validating the management-generator dependency update in Azure/azure-sdk-for-net#62963.

## Regression

The relevant TCGC behavior changed from processing only enum-member decorator arguments:

```ts
if (arg.kind === "EnumMember") {
  return getClientTypeWithDiagnostics(context, arg);
}
```

to processing model arguments as SDK client types as well:

```ts
if (arg.kind === "EnumMember" || arg.kind === "Model") {
  return getClientTypeWithDiagnostics(context, arg);
}
```

As a result, a model used as decorator metadata can be converted before normal SDK/resource traversal. In versioned ARM specifications, the later normal traversal may encounter a distinct TypeSpec model object representing the same logical definition.

## Observed result

For the Billing management library, the resulting C# code model contained two models with the same logical identity:

```text
name: BillingReservation
crossLanguageDefinitionId: Microsoft.Billing.Reservation

name: BillingReservation1
crossLanguageDefinitionId: Microsoft.Billing.Reservation
```

This propagated into generated names such as:

```text
BillingReservation1Resource
BillingReservation1Data
```

The two objects are not two intentional SDK types; they represent the same logical TypeSpec definition.

## Where the `1` is added

TCGC introduces the duplicate model discovery, but the specific `BillingReservation1` spelling is assigned downstream by `@typespec/http-client-csharp` in `fixNamingConflicts()`:

```ts
const key = `${model.namespace}.${model.name}`;
const count = modelNameMap.get(key);
if (count) {
  modelNameMap.set(key, count + 1);
  model.name = `${model.name}${count}`;
} else {
  modelNameMap.set(key, 1);
}
```

The Azure branded emitter and the management emitter do not contain equivalent numeric model-name deduplication.

TCGC itself also contains numeric conflict helpers (`resolveDuplicateGenearatedName` and `resolveConflictGeneratedName`), but the client-name override used for this Billing model bypasses that path. Both duplicate objects arrive at the base C# emitter with the same `BillingReservation` name, and the base emitter numbers the second one.

## Controlled verification

We performed dependency-by-dependency local regeneration using the management generator's `RegenSdkLocal.ps1`:

1. New C# generator with the complete former TypeSpec dependency stack: zero Billing diff.
2. Only TCGC upgraded from `0.71.2` to `0.72.0`: reproduced the complete Billing drift and `BillingReservation1` names.
3. Management decorator-argument sanitization enabled versus disabled: identical generated output, proving the management sanitizer does not cause the duplicate discovery.
4. Locally reverting only the new `arg.kind === "Model"` conversion in TCGC 0.72: restored Billing to zero diff.
5. The same one-line diagnostic revert restored Enclave, IoT Hub, Chaos, and Storage Mover to zero diff and clean builds.

The sanitizer cannot prevent this because it runs after `createSdkContext()` has already completed SDK model discovery and name allocation.

## Impact

The full management SDK regeneration produced unexpected changes in hundreds of files. Effects included:

- numeric public type names such as `BillingReservation1`;
- missing or duplicated generated models;
- generated types no longer matching existing customizations;
- invalid generated collection interfaces;
- package compilation failures.

Some packages compiled but still had generated drift despite no intended API changes.

## Suggested fix

Model-valued decorator arguments should remain available as decorator metadata without independently adding their complete model graph to normal SDK model discovery/name allocation.

Possible approaches:

1. Do not call the normal `getClientTypeWithDiagnostics()` SDK-discovery path for model decorator arguments.
2. Resolve model decorator arguments through a metadata-specific representation that does not mutate the SDK package/type caches.
3. Canonicalize models by logical identity so decorator traversal and normal traversal reuse the same SDK model, including across versioned TypeSpec projections.

It would also be useful to add a regression test asserting that:

- a model-valued decorator argument is preserved as metadata;
- only one SDK model exists for each nonempty `crossLanguageDefinitionId`;
- traversal order does not change the generated model name;
- no numeric suffix is introduced for the duplicated logical identity.

## Related issue

The same early decorator-model conversion also exposed a separate reference-preserving JSON problem: a model's `$id` can first be serialized inside decorator arguments that the C# reader treats as opaque data, followed later by a normal `$ref` that cannot be resolved. That deserialization issue is tracked in microsoft/typespec#11981.

tadelesh:
## Summary
- limit model-valued decorator SDK type conversion to `@clientOption`
- prevent unrelated decorator metadata from participating in SDK model discovery
- add regression coverage for `@finalLocation(Model)`

Fixes #5490


tadelesh:
Superseded by https://github.com/Azure/typespec-azure/pull/5509 after renaming the head branch to publish/hotfix/tcgc-september-2026.

